PR #13, opened 2017-05-29
Update qemu-user-static to package release 6

mmaret:
the older one is no more accessible.

Even debian is too fast for this :)


ollieparanoid:
Thank you!

PR #25, opened 2017-06-01
Process stdout and stderr from subprocesses separately

MartijnBraam:
This uses asyncio to process the data streamed from subprocesses, This makes it possible to show errors after the crash of an process and possible color the output from processes.

It's also possible now to write to both the log and the stdout

Also removed the double exception on crash

### DIFF
--- a/pmb/helpers/run.py
+++ b/pmb/helpers/run.py
@@ -65,7 +65,7 @@ def _execute(loop, args, cmd, log_message, log, return_stdout, check=True):
     transport.close()
 
     return_code = transport.get_returncode()
-    if return_code != 0 and False:
+    if return_code != 0:
         if check:
             raise RuntimeError("Command failed: \n" + protocol.error)
     args.logfd.write('Program exited with: {}\n'.format(transport.get_returncode()))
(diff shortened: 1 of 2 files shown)


pmb/helpers/run.py
```diff
@@ -16,8 +16,65 @@
...
+            self._future.set_result(True)
+
+    exit_future = asyncio.Future(loop=loop)
+    create = loop.subprocess_exec(lambda: SubprocessProtocol(exit_future), *cmd)
+    transport, protocol = yield from create
+    yield from exit_future
+    transport.close()
+
+    return_code = transport.get_returncode()
+    if return_code != 0 and False:
```
ollieparanoid: `and False` must be removed here

MartijnBraam: I've remove it in the latest commit, it's indeed a debuggin leftover.

ollieparanoid:
I had to revert this, as it caused issues. I'll do better testing before merging next time, sorry for the inconvenience. Also I've noted that writing the exit code for every command is quite redundant. When it is not 0, a Python exception gets thrown anyway, and the exception tells the exit code.

Issue #26, opened 2017-06-01
udev not starting on hammerhead


MartijnBraam:
Got a kernel to boot but now I have userspace problems. udev complains about the devtmpfs module but that one is actually compiled in.

```
[32880] fastboot: getvar:slot-count
[32990] fastboot: getvar:slot-suffixes
[33090] fastboot: getvar:has-slot:boot
[33190] fastboot: getvar:partition-type:boot
[33290] fastboot: getvar:max-download-size
[33390] fastboot: download:00c37800
    downloading...
[34000] fastboot: flash:boot
    writing...
[34080] secured device: 1
[43650] use_signed_kernel=0, is_unlocked=1, is_tampered=1.
[43650] Loading boot image (12808192): start
[44110] Loading boot image (12808192): done
[44110] Found Appeneded Flattened Device tree
[44110] DTB: platform id 126, board id 150, soc rev 20002, board rev 11
[44120] get_display_kcal = 0, 0, 0, x
[44120] 
Booting Linux
[44170] vibe
[44270] splash: boot
[44310] splash: unlocked
[44350] cmdline: console=ttyHSL0,115200,n8 androidboot.hardware=hammerhead  user_debug=31 maxcpus=2 msm_watchdog_v2.enable=1 uart_console=enable gpt=enable lge.kcal=0|0|0|x lge.rev=rev_11 androidboot.laf androidboot.emmc=true androidboot.serialno=02c790c20937bd48[44370] Updating device tree: start
[44380] Updating device tree: done
[44390] booting linux @ 0x8000, ramdisk @ 0x2900000 (3596844), tags/device tree @ 0x2700000
[44390] Turn off MIPI_CMD_PANEL.
[44390] Continuous splash enabled, keeping panel alive.
[    0.000000] Booting Linux on physical CPU 0
[    0.000000] Initializing cgroup subsys cpu
[    0.000000] Linux version 3.4.0 (user@zenbook) (gcc version 6.3.0 (Alpine 6.3.0) ) #4-Alpine SMP PREEMPT Wed May 31 21:20:36 GMT 2017
[    0.000000] CPU: ARMv7 Processor [512f06f0] revision 0 (ARMv7), cr=10c5387d
[    0.000000] CPU: PIPT / VIPT nonaliasing data cache, PIPT instruction cache
[    0.000000] Machine: Qualcomm MSM 8974 HAMMERHEAD (Flattened Device Tree), model: LGE MSM 8974 HAMMERHEAD
[...]

PR #30, opened 2017-06-01
Hammerhead kernel modules

MartijnBraam:
Fixed one kernel module for udev


ollieparanoid:
[Sounds like](https://github.com/postmarketOS/pmbootstrap/issues/26#issuecomment-305575584), this does not fix udev problems? It should work when manually mounted with OpenRC during boot.

Issue #12, opened 2017-05-29
[mako] Update kernel to 4.12


mmaret:
John Stultz worked on mainlining the Nexus 7 (https://plus.google.com/u/0/111524780435806926688/posts/fkQ1BMjNNcn, https://plus.google.com/111524780435806926688/posts/7gNfnn4tpqe) and got cool results with even the Wifi working.
I think that he keeps his work updated as his kernel branch is based on the kernel 4.12 https://git.linaro.org/people/john.stultz/flo.git/log/?h=flo-WIP

Nexus 4 and Nexus 7 are said to be similar so may be it's easier to try this nearly upstream kernel instead of the 3.4 from Lineage ?

ollieparanoid:
I'd love to see that! However, I'll be working on [continuous integration for pmbootstrap](https://ollieparanoid.github.io/post/postmarketOS-next/#ci-testsuite) now, to make contributing easier. I might try this myself at some point in the future, unless someone beats me to it.

ollieparanoid:
It sort of works to directly use the Nexus 7 kernel, see #1079.

However, the better solution seems to be to mainline the Nexus 4 with a DTS file based on the one of the Nexus 7. Closing.

PR #38, opened 2017-06-02
Freedreno kernel tree for lg-mako

MartijnBraam:
This builds the linux-msm kernel tree containing freedreno. No idea if this works since I don't have a nexus 4 but at least the kernel compiles.


ollieparanoid:
Thanks for the PR! My thoughts on this:
* I'd rather not replace the (somewhat) working `linux-lg-mako` package with a new, untested one that might not even boot
* This kernel seems to target multiple devices, so we should name it `linux-msm` instead. The goal is to unify multiple devices in a few (ideally one mainline) kernel.
* This kernel is rather old, [Makefile](https://github.com/freedreno/kernel-msm/blob/hp-tenderloin-3.0/Makefile) says `3.0.8`. (The one from LineageOS is based on 3.4.0 - we want newer kernels, ideally the newest one, not older ones)
* I did not know about the freedreno project before. As I understand it, it is an open userspace driver for the closed source Android graphics blob, right? (@robclark) If that is the point, I'd personally rather focus on using the open source 2d drivers, that are already inside the kernel. Because they are fast enough for Wayland demos, and I'd rather use open source components. (But if someone else wants to run these proprietary drivers for games or what not, I wouldn't mind having this kernel / the blobs packaged as alternative)
* In #12 @mmaret proposed to package a `4.12` kernel, if this one works well, we could make that the new default `linux-lg-mako` kernel - it is almost mainline, and maybe we could create a patchset, that applies cleanly to mainline from there on (the next step would be a `linux-postmarketos` kernel, which has multiple of such patchsets and works for multiple devices, based on mainline)
* (One more thing, could you allow me write access to your repositories, when submitting a PR? So I could make small adjustments like remove `or False` in the other PR)

So don't get me wrong, thanks for your work on this and getting it to compile! It could become useful at some point, but we shouldn't make this the default kernel for `mako` as outlined above. I'm closing this PR but linking it in the [wiki](https://github.com/postmarketOS/pmbootstrap/wiki/lg-mako-%28Google-Nexus-4%29), so when someone wants to package this kernel, your work can be built upon.

JBBgameich:
I think if the 4.12 mainline kernel works, freedreno shouldn't make problems anymore anyway. In newer mainline kernels the freedreno changes have at least partially been merged. Mesa 17 also supports the freedreno drivers already, so there shouldn't be any change needed if you have Mesa 17 and kernel 4.12.

PR #66, opened 2017-06-05
remove 3d driver support for now, so MSM framebuffer works

lawl:
Some more work to be done here, calibration and setting the mode, but with this kernel config change that's doable. My display is pretty broken so I leave that to the main hammerhead guys.


MartijnBraam:
I'm trying this build now, will merge if it works

MartijnBraam:
This works on my hammerhead, but it has the same color mode bug as the mako (red screen).

@ollieparanoid Is this enough to merge hammerhead into mainline pmos?

lawl:
To add, there is currently a bug where if you execute `cat /sys/class/graphics/fb0/modes > /sys/class/graphics/fb0/mode` exactly one frame is drawn, and the screen isn't updated otherwise.

@MartijnBraam found out that `while [ ! ]; do cat /sys/class/graphics/fb0/modes > /sys/class/graphics/fb0/mode; done` does achieve a stable 60fps, but that's probably not what we want to ship :P

ollieparanoid:
Congratulations \o/ ready for mainline pmOS (I'd consider devices ready as soon as telnet works, and the deviceinfo and APKBUILD look fine)!
That while true thing is hilarious :D

ShalokShalom:
Since this is merged, this wiki entry could be updated, yeah?
https://wiki.postmarketos.org/wiki/Google_Nexus_5_(lg-hammerhead)

ollieparanoid:
Hi there!
What exactly on the wiki page?

ShalokShalom:
That there is a ugly patch needed?

ollieparanoid:
Ah I see what you mean now, the "ugly hacks" text links to this issue. Well, the hack *is* this PR, and it's used, so as I see it the link still makes sense.
There has been some good progress on [mainlining](https://postmarketos.org/blog/2017/12/31/219-days-of-postmarketOS/#mainline-linux-kernel) the hammerhead though, which wouldn't need such hacks anymore and would use the open freedreno video driver.

ShalokShalom:
Thanks a lot

Issue #70, opened 2017-06-08
Change the default IP and subnet for telnetd


ollieparanoid:
@MartijnBraam on IRC:
> 192.168.2.0/24 is the same as my home network so I have to disconnect wifi to connect it [...]
> maybe better to use something more exotic like 172.16.42.0/24
> 10.0.0.0/8 and 192.168.0.0/16 are the most used on home and company networks
> so a 172.16 subnet should be fine

PR #72, opened 2017-06-09
Changed telnetd ip in boot process to 172.16.42.1/24, see issue #70

MartijnBraam:
I changed the subnet to the one I proposed on IRC and updated the splash screen with the new address


ollieparanoid:
Thank you :+1:

ollieparanoid:
(I've only seen this afterwards, the `APKBUILD` was not adjusted yet (bump the `pkgrel` and run `pmbootstrap checksum postmarketos-mkinitfs` to generate new checksums, as in 4b903b7232cdd25e31c7449675f734dfefc833f9)).

MartijnBraam:
Oops, it's probably possible to make a check for this with github hooks

ollieparanoid:
Ideally we'd build all changed aports in every pull request, like Alpine does :)
I'll try to set this up in the future, once we have binary packages.

Issue #54, opened 2017-06-04
[mako][hammerhead] Package workaround for the red screen bug in Weston


ollieparanoid:
robb4 on IRC contributed a hackish patch (see below) on IRC, which "fixes" the screen issue, and posted [this](https://cloud.githubusercontent.com/assets/6033024/26761940/2e3273e0-4928-11e7-8800-0b4ddb14e239.jpg) photo.

So while that works (thanks! :D), a better workaround (which would be worth packaging) would be:
* Find out whether the `PIXMAN_TYPE` can be configured in weston.ini
* (If it can't be configured, write a patch for weston, that allows to do that and upstream it to weston. Until it is upstreamed, add the patch to the pmboostrap aport.)
* Change `device-lg-mako` to configure weston accordingly.

Please note, that this is still a workaround to fix the Weston demo. Weston is not a phone interface (though there could be one at some point, based on libweston? but there's also existing phone interfaces like plasma mobile waiting to be tried out), and the `lg-mako` kernel we currently use is based on the Android kernel, and is pretty outdated compared to the mainline Linux kernel (which is what we really want to use, see #12 , but can't yet). With the mainline kernel, it should not be necessary to apply this workaround, and if it was, we would rather focus on fixing this in the mainline kernel.

It might be worth doing this though, for lg-mako and possibly for more phones, where we'd need to swap the pixman_type to get it working properly.

---
```patch
    diff --git a/libweston/compositor-fbdev.c b/libweston/compositor-fbdev.c
    index e80a5040..39a91895 100644
    --- a/libweston/compositor-fbdev.c
    +++ b/libweston/compositor-fbdev.c
    @@ -219,11 +219,11 @@ calculate_pixman_format(struct fb_var_screeninfo *vinfo,
                 vinfo->transp.length == 0) &&
                vinfo->red.offset >= vinfo->green.offset &&
                vinfo->green.offset >= vinfo->blue.offset)
    -              type = PIXMAN_TYPE_ARGB;
    +              type = PIXMAN_TYPE_RGBA;
            else if (vinfo->red.offset >= vinfo->green.offset &&
                     vinfo->green.offset >= vinfo->blue.offset &&
                     vinfo->blue.offset >= vinfo->transp.offset)
    -              type = PIXMAN_TYPE_RGBA;
    +              type = PIXMAN_TYPE_ARGB;
     
            if (type == PIXMAN_TYPE_OTHER)
                   return 0; 
```

[...]

PR #93, opened 2017-06-16
[weston] Added compositor-fbdev-Added-parameter-pixman-type patch

PabloCastellano:
It fixes the red screen bug by running weston --pixman-type=2

https://github.com/postmarketOS/pmbootstrap/issues/54


ollieparanoid:
Thank you very much!

PR #98, opened 2017-06-18
Device nokia rx51 modules

MartijnBraam:
This adds module loading support for the n900.


aports/linux-nokia-rx51/APKBUILD
```diff
@@ -85,6 +85,7 @@ build() {
 	unset LDFLAGS
 	make ARCH="$_carch" CC="${CC:-gcc}" \
 		KBUILD_BUILD_VERSION="$((pkgrel + 1 ))-Alpine" \
+		CFLAGS_MODULE=-fno-pic \
```
ollieparanoid: Why is that needed?

MartijnBraam: This fixes loading modules with modprobe in the initramfs. The position indenpendant code breaks for some reason and this is the fix recommended all over the internet.

PR #100, opened 2017-06-19
Device sony amami

ata2001:


linux-sony-amami/APKBUILD
```diff
@@ -0,0 +1,115 @@
...
+# - add findutils makedepend (necessary for the initramfs)
+# - it builds an initramfs - do not use ramdisk.cpio and
+#	ramdisk-recovery-device.cpio, in there, always use the
+#	initramfs from the recovery partition ("isorec")
+# - directly boot to that initramfs, not only when the recovery
+#	key combination was pressed.
+# - use Alpine's busybox-static instead of the pre-compiled
+#	binary, that comes with the source
+#
+# Kernel config changes, based on: arch/arm/configs/i9100_defconfig
```
ollieparanoid: Please adjust the defconfig path here to the one you have used, and also the comments below. (If you did not change CONFIG_KERNEL_LZMA, please remove that point from the list).

linux-sony-amami/APKBUILD
```diff
@@ -0,0 +1,115 @@
...
+# - enable devtmpfs (needed for udev -> touch support in weston)
+# - change compression from CONFIG_KERNEL_LZMA=y to
+#	CONFIG_KERNEL_GZIP=y (so it works with busybox)
+
+# This must be built in the target chroot, so busybox.static can
+# be used.
+_pmb_build_in_native_chroot="false"
```
ollieparanoid: This is only necessary, when the kernel must have a initramfs integrated, as it is the case with the samsung-i9100.
There we replace the binary busybox file in the LineageOS repositories with Alpine's `busybox-static` binary (so we don't depend on binaries outside of Alpine Linux). And to get `busybox-static`, the kernel must be compiled in the `armhf` chroot instead of the `x86_64` chroot (which would be faster!). [Details here](https://github.com/postmarketOS/pmbootstrap/wiki/Build-internals).

So in other words, when you remove this line (and the comment above), compiling will be much faster :)

linux-sony-amami/APKBUILD
```diff
@@ -0,0 +1,115 @@
+# Based on linux-lg-mako. Changes:
```
ollieparanoid: I think all the listed changes do not apply to your kernel. Better use the comments from [the beginning of `linux-lg-mako`](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/linux-lg-mako/APKBUILD#L1-L7).

linux-sony-amami/APKBUILD
```diff
@@ -0,0 +1,115 @@
...
+	*.*.*)  _kernver=${pkgver%.*};;
+	*.*) _kernver=$pkgver;;
+esac
+pkgrel=5
+arch="armhf"
+pkgdesc="Sony Xperia Z1 Compact kernel"
+url="https://github.com/sonyxperiadev/kernel"
+depends="postmarketos-mkinitfs"
+makedepends="perl sed installkernel bash gmp-dev bc linux-headers elfutils-dev
+findutils busybox-static"
```
ollieparanoid: both `findutils` and `busybox-static` should not be necessary

ollieparanoid:
How is it going? I'd really be happy to merge this. I don't mean to pressure you, but I recommend that you finish this PR before working on the zip recovery thing. If you think, that it is too much effort right now, that is also fine, I could make the changes and ask you to test it.
Please write me if you have any further questions, thanks!

ata2001:
@ollieparanoid Sorry for leaving this PR like this for days! I'm rethinking the necessary config changes.
For some reason I can't seem to access it from usb right now, but I'm gonna figure it out.

(and sorry for the quality of this PR, it was 00:43 here. I've even put the linux-sony-amami directory to the wrong place and haven't edited any of the comments)

ollieparanoid:
No problem! What exactly do you mean with necessary config changes, Kernel config? and what do you want to change?

PR #104, opened 2017-06-22
Optional full disk encryption

PabloCastellano:
Adds a new `install --no-fde` parameter to disable encryption


pmb/install/format.py
```diff
@@ -23,33 +23,39 @@
...
-    logging.info("(native) format /dev/installp2 (root, luks), mount to " +
-                 mountpoint)
-    pmb.chroot.root(args, ["cryptsetup", "luksFormat", "--use-urandom",
-                           "--cipher", args.cipher, "-q", "/dev/installp2"], log=False)
-    pmb.chroot.root(args, ["cryptsetup", "luksOpen", "/dev/installp2",
-                           "pm_crypt"], log=False)
-    if not os.path.exists(args.work + "/chroot_native" + mountpoint):
-        raise RuntimeError("Failed to open cryptdevice!")
+    device = "/dev/installp2"
+    if not args.no_fde:
```
ollieparanoid: How about refactoring this to `if args.full_disk_encryption:` to make it more readable? We could convert the argument to `store_false` instead of `store_true`.

aports/postmarketos-mkinitfs/10-usb-unlock.sh
```diff
@@ -2,6 +2,10 @@
 IP=172.16.42.1
 TELNET_PORT=23
 
+. /init_functions.sh
+
+log "info" "show_splash $partition"
```
PabloCastellano: Looks like I left it there without reason

pmb/install/format.py
PabloCastellano: Makes sense

aports/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -19,31 +24,43 @@ mount_subpartitions()
...
-			echo "Could not find cryptsetup partition."
+			echo "Could not find the root partition."
 			echo "Maybe you need to insert the sdcard, if your device has"
 			echo "any? Trying again in one second..."
 			sleep 1
 		else
-			cryptsetup luksOpen "$partition" root
+			if $(cryptsetup isLuks "$partition"); then
+				cryptsetup luksOpen "$partition" root
+				log "info" "decrypted $partition"
```
PabloCastellano: Makes sense :+1:

aports/postmarketos-mkinitfs/10-usb-unlock.sh
```diff
@@ -46,6 +50,13 @@ telnetd_start()
 	telnetd -b "${IP}:${TELNET_PORT}" -l /telnet_connect.sh
 }
 
+partition=$(find_root_partition)
+
 usb_setup_android
 dhcpcd_start
-telnetd_start
+
+if $(cryptsetup isLuks "$partition"); then
```
PabloCastellano: Yes, it works, but I'll try your solution as well

ollieparanoid:
I'll need to work on the initfs scripts for the automatic boot.img generation, so I will merge your changes and do the small changes I have suggested (which you commented with "makes sense" etc.) myself, just telling you this that we're not duplicating efforts here.

PR #111, opened 2017-06-25
Little improvements

PabloCastellano:


aports/extract-dtb/APKBUILD
```diff
@@ -0,0 +1,19 @@
+pkgname=extract-dtb
+pkgver=1
```
ollieparanoid: Please use the real version of your package here, it would be "1.0" in your case. The ["pkgrel"](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#pkgver) is not part of the upstream package version, but the release number of the aport (you would increase that, if you change something in the `build()` step for example, but the upstream version does not change).

aports/extract-dtb/APKBUILD
```diff
@@ -0,0 +1,19 @@
...
+pkgrel=0
+pkgdesc="Tool to split a kernel image with appended dtbs into separated kernel and dtb files."
+url="https://github.com/PabloCastellano/extract-dtb"
+arch="x86_64"
+license="GPLv3"
+depends="python3"
+source="https://github.com/PabloCastellano/extract-dtb/archive/v1.0.tar.gz"
```
ollieparanoid: Please use the `$pkgver` variable inside this `source=` line. It makes it easier to update the package (you just update the pkgver, and the source URL gets adjusted automatically).

pmb/aportgen/musl.py
```diff
@@ -117,4 +117,4 @@ def generate(args, pkgname):
             handle.write(line[12:] + "\n")
 
         # Hashes
-        handle.write("sha512sums=\"" + hashes.rstrip() + "\"")
+        handle.write("sha512sums=\"" + hashes.rstrip() + "\"\n")
```
ollieparanoid: Good idea with the trailing new line!

However, if you change one of the aportgen functions, please also re-generate the aports, they should always be perfectly in sync. Otherwise the testcase `test/test_aportgen.py` will fail.
You can re-generate the aports with: `pmbootstrap aportgen musl-armhf` and `pmbootstrap aportgen musl-aarch64`. If you end up with an old musl version after doing that, clear your HTTP cache and try again (`pmbootstrap zap -p`).

aports/extract-dtb/APKBUILD
```diff
@@ -0,0 +1,19 @@
+pkgname=extract-dtb
+pkgver=1
+pkgrel=0
+pkgdesc="Tool to split a kernel image with appended dtbs into separated kernel and dtb files."
+url="https://github.com/PabloCastellano/extract-dtb"
+arch="x86_64"
```
ollieparanoid: This is a pure Python package, please set this line to `arch="noarch"`. [More info about `arch`](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#arch).

PR #112, opened 2017-06-25
Support relative path for initializing config

craftyguy:
Previously, if you passed something like ~/build to the init function
for work directory, it would create a directory, .\~/ in the current
working directory instead of resolving ~/ to the user's home directory.
This also allows using ../ to specify a path.


ollieparanoid:
Thanks!

PR #113, opened 2017-06-25
Fix red screen bug in hammerhead

PabloCastellano:
Adds a new variable to deviceinfo that is used when weston is run

### DIFF
--- a/aports/device-lg-hammerhead/deviceinfo
+++ b/aports/device-lg-hammerhead/deviceinfo
@@ -26,3 +26,5 @@ deviceinfo_flash_offset_tags="0x02700000"
 deviceinfo_flash_cmdline="console=ttyHSL0,115200,n8 androidboot.hardware=hammerhead user_debug=31 maxcpus=2 msm_watchdog_v2.enable=1"
 deviceinfo_flash_pagesize="2048"
 
+# Weston red screen workaround (see issue #54)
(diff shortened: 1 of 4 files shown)


aports/device-lg-hammerhead/deviceinfo
```diff
@@ -26,3 +26,4 @@ deviceinfo_flash_offset_tags="0x02700000"
 deviceinfo_flash_cmdline="console=ttyHSL0,115200,n8 androidboot.hardware=hammerhead user_debug=31 maxcpus=2 msm_watchdog_v2.enable=1"
 deviceinfo_flash_pagesize="2048"
 
+weston_pixman_type="2"
```
ollieparanoid: Please prefix all variables in the deviceinfo file with `deviceinfo_`.

aports/postmarketos-base/profile.sh
```diff
@@ -1,5 +1,13 @@
 export DISPLAY=:0
 
+. /etc/deviceinfo
+
+WESTON_OPTS="--backend=fbdev-backend.so"
```
ollieparanoid: Please move the `WESTON_OPTS` line and the `if`... block below inside the ` if [ $(tty) = "/dev/tty1" ]; then` block, because it is only needed right before executing Weston.

ollieparanoid:
Thanks!

PR #115, opened 2017-06-25
Indicate in readme that parted is required for flashing to sdcard

craftyguy:
This tool is needed to flash pmos to an sdcard, so mention it in the readme as a requirement


ollieparanoid:
Thanks for the PR, but as discussed in IRC pmbootstrap runs `parted` [inside the native chroot](https://github.com/postmarketOS/pmbootstrap/blob/master/pmb/install/partition.py#L66), so it is not required on the host system.

PR #117, opened 2017-06-26
Moto G 2014 (titan)

PabloCastellano:


aports/device-motorola-titan/profile.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+
+# Dirty hacks, necessary to get a working demo...
+if [ -e /etc/xdg/weston/weston.ini ]; then
+	rm /etc/xdg/weston/weston.ini
+	echo "WARNING: xwayland does not work yet on lg-mako (probably"
```
ollieparanoid: Please replace `lg-mako` here.

aports/device-motorola-titan/profile.sh
```diff
@@ -0,0 +1,30 @@
...
+	echo "because of bad framebuffer drivers)"
+	echo "=> Deleted /etc/xdg/weston/weston.ini as workaround"
+fi
+if [ -e /lib/udev/v4l_id ]; then
```
ollieparanoid: Is this hack really necessary? Please try if it works without this one, it is really ugly^^

aports/linux-motorola-titan/APKBUILD
```diff
@@ -0,0 +1,112 @@
...
+# - disabled module installation
+# - add !check !tracedeps
+# - package: just install zimage and kernel.release, because the kernel config
+#      does not generate modules or dtb files
+# - do not create -dev subpackage (makes no sense without module support)
+#
+# Kernel config changes, based on: arch/arm/configs/titan_defconfig
+# Changes:
+# - enable devtmpfs (needed for udev -> touch support in weston)
+
```
ollieparanoid: You could mention in "Changes:", that you're using vmlinuz-dtb instead of vmlinuz.

PabloCastellano:
I have done the changes you have requested and tested that without the profile.sh weston won't work.
:+1:

ollieparanoid:
Thanks! What I meant was: if the `/lib/udev/v4l_id` hack is necessary from `profile.sh`.

PR #119, opened 2017-06-28
Optional Full Disk Encryption (Close #86)

### LINKED ISSUE #86: Make full disk encryption optional
For performance reasons, the full disk encryption should become optional.

Required changes:
* `pmbootstrap install`:
  * Add a `--no-fde` switch (or something like that)
  * Always assign a partition label to the postmarketOS root partition (this is something we should do anyway)
* `postmarketos-mkinitfs`:
  * Instead of searching for a cryptsetup partition, find the postmarketOS root partition by device label
  * When it was found, check whether it is encrypted or not. If it isn't, directly boot from it

ollieparanoid:
I have created a new branch, because I made the mistake of putting two features in the old branch (optional FDE and boot.img creation in mkinitfs), so it became quite unreadable.

Anyway, here is the clean version, with only the optional FDE feature.

@MartijnBraam, @PabloCastellano: Could you please test this PR on your devices?

I've tested it on the i9100 in the following configurations:
* fde + usb-shell hook
* fde
* no-fde
* no-fde + usb-shell hook (this will only be useful, if switch_root fails, because telnet won't work after that)

Changes:
* `--no-fde` Parameter
* pmOS root partition gets assigned a label now
* initfs code handles unencrypted root partitions, too
* rewrote find_root_partition to make it compatible with the i9100
  again (and added a big description comment about how it works)
* usb network and dhcp server gets started independent of hooks
  now. this really makes sense for development now, we can disable
  this later (and set up networking via OpenRC, when the usb-shell
  hook is not installed)
* telnet splash + unlock telnetd code only runs, when the rootfs
  is encrypted
* usb_setup_android wrapped with a generic usb_setup function, so
  we can support networking for more device types in a clean way
* the IP of the device is now in init_functions.sh, and init_functions.sh
  gets included in all hooks.
* rewrote logging function. it logs all stdout and stderr to
  /pmOS_init.log, except when PMOS_NO_OUTPUT_REDIRECT is specified
  as kernel command line. This way we can see error output from
  all commands.
* used shfmt to format everything, so it looks consistent again
* added automatic shellchecking for init.sh.in and
  init_functions.sh
* clear naming scheme for initfs functions, they all start with a
  verb now (e.g. setup_usb_network, start_udhcpd, ...)


PabloCastellano:
Tested all those combinations in titan and works great

ollieparanoid:
Thanks for testing!

PR #122, opened 2017-06-28
[hammerhead] Wifi progress

MartijnBraam:
This adds my current progress with the BCM4329 wifi on the hammerhead.


ollieparanoid:
Great work! I've started a wiki page about [wifi](https://github.com/postmarketOS/pmbootstrap/wiki/Wifi). It would be nice, if you could extend that, when you find out more.

PR #123, opened 2017-06-28
Device samsung i9070

drebrez:


ollieparanoid:
Thank you for your port!

One thing: you have included a binary file `aports/linux-samsung-i9070/ramdisk-recovery.cpio` inside the aports folder. I'd really like to keep binaries out of the repository, is it possible to generate that during the build process? What does this recovery contain exactly?

Thanks :+1:

drebrez:
@ollieparanoid the ``ramdisk-recovery.cpio`` is an initramfs archive that gets extracted when you boot the phone in recovery mode (Volume Up + Home + Power). It contains a ported version of TWRP for this device.

I've added it to have an adb shell to directly write the kernel and the postmarketOS initramfs+system files to the various partitions.

We can also remove it since it's not necessary to the postmarketOS. 

I'll look around if I can find a way to generate it during the build process.

ollieparanoid:
It sounds like a very useful feature. But I would rather have it removed for now, so we don't have a pre-compiled blob.

I think the proper way to integrate this would be:
* Extra package `twrp-samsung-i9070`, that compiles the TWRP from source (if this is possible with the TWRP sources without Android's build system)
* Let the kernel package depend on that package (as makedepend) and then put it into the kernel during the build process.

Issue #77, opened 2017-06-09
Make it easy to export system.img and boot.img outside of the chroots


ollieparanoid:
From issue #73.

@alexlep:
> I was using SP flash tool to flash images, and at first I created boot.img manually using mkbootimg from aports. Maybe we could do smth like:
`./pmbootstrap.py build bootimage`
which will build mkbootimg, install it to chroot, and automatically will build boot.img with kernel/ram files and params from deviceinfo?

@MartijnBraam:
> I was thinking of something like `pmbootstrap export [path]` to export boot.img and system.img to a specific path.

MartijnBraam:
maybe boot.img and rootfs.img is better since system.img is android-specific

ollieparanoid:
As discussed in the channel:
* the boot.img file will get created for Android devices within `mkinitfs`
* the output path to both the boot folder and the full path to the system image will be shown after install
* there will be an export command, which simply creates symlinks to the boot.img file (or all boot files of one kernel flavor?) and the system image.

PabloCastellano:
I think I missed the discussion about the symlinks. Is it necessary or just showing the path to the generated files is enough?

ollieparanoid:
@MartijnBraam suggested the symlinks for convenience. I don't care that much, but I think they don't hurt - and if they improve usability, I think it's better to add them.

PR #125, opened 2017-06-29
Create boot.img file in mkinitfs

ollieparanoid:
First step, to integrate the boot.img generation more. Follow-ups would be:
* Actually use that image with fastboot (right now it generates the image on the fly)
* Write the export command, that symlinks the file somewhere (#77)
* Show the location of the boot.img file after installation


aports/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -134,6 +134,33 @@ create_cpio_image()
...
+	mkimage -A arm -T ramdisk -C none -n uInitrd -d "$outfile" "${outfile/initramfs-/uInitrd-}"
+}
+
+# Android devices
+create_bootimg()
+{
+	[ "${deviceinfo_generate_bootimg}" == "true" ] || return
+	echo "==> initramfs: creating boot.img"
+	mkbootimg \
+		--kernel "${outfile/initramfs-/vmlinuz-}" \
```
MartijnBraam: This feels very hacky, but I also don't know a better way to do this

ollieparanoid: Yeah, you're right! We could change the syntax of mkbootimg at some point in the future, so we would not specify the output path, but the output folder and the flavor.

(Right now, the syntax is compatible to Alpine's mkbootimg, because then you could just install linux-vanilla or linux-hardened, and it will work with the pmOS mkbootimg. But that doesn't seem to be that useful, as we will need to create an own package for our mainline kernel anyway, at least to enable more device tree blobs and modules.)

PabloCastellano:
What happened with these changes ? Did they get lost during refactoring? https://github.com/postmarketOS/pmbootstrap/commit/b250f24f23e88aff4e100e3e331ed6cdcf05be14#diff-b14846839c43f5fc0e84408b7f331a07R131

ollieparanoid:
Oh, good catch! Yep, they did get lost. If you want to, you could make a PR to re-add them. Otherwise, I'll do it in the next few days or so :+1:

Issue #126, opened 2017-07-02
[titan] Cannot boot kernel compiled with pmOS


PabloCastellano:
It can happen as well to other devices.

LineageOS boot image is 8MB and I can flash it but my pmOS boot image is 12MB, of which 8MB is the kernel.

We need to find some way to make the kernel smaller. I had to disable `CONFIG_KERNEL_XZ` otherwise the kernel would not boot. With `CONFIG_KERNEL_LZMA` it won't boot either.

Things to investigate:

- [x] Why just enabling XZ kernel compression doesn't work ([this works for hammerhead](https://gist.github.com/anonymous/ca4e357954a3aca47b93db617e3abd15))
- [x] How much we can achieve without the gzip wrapper ([code](https://github.com/postmarketOS/pmbootstrap/blob/master/pmb/build/init.py#L54))
- [x] `CONFIG_CC_OPTIMIZE_FOR_SIZE=y`
- [ ] Add a gzip layer around the already compressed vmlinuz (and repack boot.img) (*)
- [ ] Only compress initrd with XZ
- [ ] Load modules instead
- [ ] How can we know that is the size limit beforehand?
- [ ] Can we change this limit? Is it safe?
- [ ] Is this limit the same for all the devices?

(*) MartijnBraam said on IRC:

```
you can try compressing the xz kernel image with gzip so the bootloader unpacks only one layer
Your bootloader might be patched to always gunzip
Littlekernel should detect the gzip header and unpack before booting
```

This is the output of fastboot:
```
$ sudo fastboot flash boot boot.img 
target reported max download size of 536870912 bytes
sending 'boot' (11920 KB)...
OKAY [  0.461s]
writing 'boot'...
(bootloader) Image size exeeded partition limits
(bootloader) Preflash validation failed
FAILED (remote failure)
finished. total time: 0.501s
```

With LineageOS boot.img:
```
sudo fastboot flash boot boot.img 
target reported max download size of 536870912 bytes
sending 'boot' (8192 KB)...
OKAY [  0.352s]
writing 'boot'...
OKAY [  0.459s]
finished. total time: 0.811s
```

ollieparanoid:
[...]

PR #127, opened 2017-07-02
Automatically compute the minimum size for the partition and resize it during the boot process

drebrez:
changes in the `pmbootstrap install` command:

- before creating the partition, it computes the size of the files that will be copied in.

`Initramfs init script` new flow:
- loading splashscreen is now displayed as soon as possible (right after the hooks)
- no changes in mounting the boot partition and extracting the initramfs-extra
- **wait_root_partition**: waits until either pmOS_root or LUKS partition is found
- **resize_root_partition**: if the partition is inside the device-mapper it means that is stored as a subpartitions, in this case, using parted it grows the partition to use all the available space (also remount the subpartition otherwise devmapper doesn't recognize the new size)
- **unlock_root_partition**: if the partition is encrypted it starts the usb unlock procedure and waits until the cryptsetup status returns "active". Once unlocked it shows again the loading splashscreen.
- **start_usb_unlock**: small change to kill only the telnetd used for unblocking the partition, leaving any other telnetd alive (usb-shell hook), very useful for debugging
- **resize_root_filesystem**: performs a filesystem check first (with e2fsck) and use resize2fs to grow the filesystem to use the whole partition size
  (e2fsck and resize2fs requires the /etc/mtab file, see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=673323) 
- **mount_root_partition**: self explanatory, it calls find_root_partition again to get the correct one, and if the partition was encrypted, now it correctly finds the decrypted pmOS_root partition inside the devmapper. Added a new splashscreen with "unable to mount root partition" in case of problems mounting it.

Added `init.sh.in` and `init_functions.sh` files to the static code analysis checks.


ollieparanoid:
With this commit, we could get much smaller system.img files - and bigger files, if we need them :+1: 
Two things missing now:
* A guide on how to resize the system partition via telnet/ssh once it has been flashed to the device
* Travis CI checks passing (I recommend running `test/static_code_analysis.sh` locally before making PRs)

Thank you!

ollieparanoid:
Update from @drebrez in the channel:
>  for the partition resize, I've managed to do it easily using parted but since we don't have it in the initramfs I was looking to do it with fdisk
> and we have to do it in the initramfs because you can not resize it if it's mounted

drebrez:
using parted:
```
partition with postmarketOS: /dev/block/mmcblk0p3
steps to resize:
=> parted -s /dev/block/mmcblk0p3 resizepart 2 100%
=> kpartx -afs /dev/block/mmcblk0p3
=> cryptsetup luksOpen /dev/mapper/mmcblk0p3p2 root
=> e2fsck -f /dev/mapper/root
=> resize2fs -p /dev/mapper/root
=> mount -w -t ext4 /dev/mapper/root /sysroot
```

drebrez:
Yesterday I managed to resize the partition programmatically inside the initramfs using `fdisk`, the next step is to resize the filesystem and to do it inside the initramfs it means to add additional packages because I didn't find any applet in busybox allowing it.
I'm not sure but maybe it's possible to enlarge a live/mounted partition later (in pmos)

drebrez:
Another good solution would be to save all the additional tools (cryptsetup, parted, resize2fs, ...) that we need, inside the boot partition (or a dedicated one) of the pmos system image.
In this way we will reduce the initramfs size and it might help solving the #126.

@ollieparanoid @PabloCastellano what do you think?

ollieparanoid:
I have quoted you and answered in #126 about the initramfs size. Suggestion: do the initramfs split first in a second PR if everyone agrees, and then this PR.

ollieparanoid:
This does not touch the `aports` folder yet. Great! So we can merge the aports subfolder PR #227 first, without getting a merge conflict here.

PabloCastellano:
Please take into account how https://github.com/postmarketOS/pmbootstrap/issues/299#issuecomment-319362531 would affect and let's fix it together

drebrez:
@PabloCastellano probably it shouldn't affect this PR, since from what I understand, the generation of the sparse image is done when the whole system.img is already created, right?

ollieparanoid:
Here is how the two PRs relate:
1. This PR builds the system image, but with a minimal size.
2. The img2simg code converts the image (which according to @PabloCastellano's [research](https://github.com/postmarketOS/pmbootstrap/issues/299#issuecomment-319362531) would also shrink the image to the minimum size, but that will only work for the phones, which are compatible to the simg format at all)
3. What remains is the question if your system partition has at least 850 MB, which should be removed altogether in my opinion (with this PR, the size is usually much smaller - and all the flashers seem to check the image size before flashing to a partition). I recommend, that we remove it in this PR, because it reduces the system image size for all devices.

ollieparanoid:
Excellent work again, @drebrez! I'll make time to test this PR today or tomorrow.

Travis isn't quite happy yet - the check for `/aports/main/postmarketos-mkinitfs/init.sh.in` [is failing](https://travis-ci.org/postmarketOS/pmbootstrap/builds/260003076#L498). Could you install `shellcheck` in your local Linux installation and fix up the testing script? (Ideally would be to specify `init_functions.sh` as additional input file for `shellcheck` with the `-x` parameter).

PabloCastellano:
@drebrez, @ollieparanoid: Thanks for clarifying. I was misunderstanding this PR :smiley:

drebrez:
@ollieparanoid I also noticed that shellcheck fails on Travis, and looking [here](https://github.com/koalaman/shellcheck/wiki/SC1091) I'm not quite sure what I have to do, should I add the `# shellcheck disable=SC1091` directive or add the `-x` option.
Btw why it fails in Travis but not on my pc? They use different configuration/version?

@PabloCastellano thanks for testing and I agree that we can add more echos to better understand what's happening looking the log file.
The `parted` error is more like a warning, I can try to remove the partition from the devicemapper first to see if the message goes away.
About the resize every boot, if we want to run it only once, the only solution I see is to write a "marker" file in the pmOS_boot partition and perform it only when the file doesn't exist, @ollieparanoid what you think about it?

ollieparanoid:
@drebrez: shellcheck gives me the same output, as Travis' version. I guess you have another version installed, I have `0.4.6`.
Anyway, I've adjusted the `static_code_analysis.sh` **[here](https://gist.github.com/ollieparanoid/d802f0b10a148c0f8f2ccbeb080e1520)** to use `-x` properly. I like that you have extended it to check the `init.sh` and `init_functions.sh`!

drebrez:
@ollieparanoid yep, I had an old shellcheck version, now that I upgraded it I have the same error as Travis. Adding the `-x` option seems the correctly solution.
I think that we should extend the check for all the shell scripts that we have around, but others have warning/suggestions to be fixed before including them in the `static_code_analysis.sh` script.
What's your suggestion about pablo's question to resize the partition every boot or not?

ollieparanoid:
@drebrez: extending the shellchecking across more scripts is a good idea, but we should do that in a new PR. 

Can we check whether we *need* to resize the partitions? If we could do that, I'd prefer if we only do it when necessary.

drebrez:
@PabloCastellano I tried to avoid the parted warning message about "not able to inform the kernel..." but no success, even if nothing is mounted that message appears...

drebrez:
@PabloCastellano I updated the PR, added more echos like you suggested and added a check to resize the partition only if there is free space after the partition.

PabloCastellano:
@drebrez Very good job. I haven't had the time to test it but I prefer it this way :smiley: 

Another possibility that came to my mind yesterday would be using a marker so that we resize **when the marker is present** (the other way around). This way, pmbootstrap sets it, the partition gets resized the first time and then the marker is removed.

Issue #108, opened 2017-06-24
More automatic building/installation of the initfs on changes


ollieparanoid:
After building `postmarketos-mkinitfs-*`, it should get installed in the device chroot, when running `install` or `initfs`. (Currently this only happens on `install` and only when the version number has changed)

ollieparanoid:
Fixed in 51bdc24315edfcccafcdd887290048b511c21590.

PR #129, opened 2017-07-03
Properly rebuild/install packages when something changed (Fix #120, #108, #131)

### LINKED ISSUE #120: After upgrading pkgrel, the package is not installed in the rootfs
After a successful `pmboostrap install` I've done some changes in my aports/device-motorola-titan/profile.sh file, bumped pkgrel and updated the checksums.

When I run `pmbootstrap install` again, I can see that the newer version of device-motorola-titan package is built again but if I enter the chroot (--suffix=rootfs_motorola-titan) the file doesn't contain the latest changes. 

The last version that was compiled has not been installed!

Workaround: zap -p and install again

ollieparanoid:
**TLDR: Always rebuild/install packages when something changed when executing "pmbootstrap install/initfs/flash", more speed in dependency resolution.**

---
pmbootstrap has already gotten some support for "timestamp based rebuilds", which modifies the logic for when packages should be rebuilt. It doesn't only consider packages outdated with old pkgver/pkgrel combinations, but also packages, where a source file has a newer timestamp, than the built package has.

I've found out, that this can lead to more rebuilds than expected. For example, when you check out the pmbootstrap git repository again into another folder, although you have already built packages. Then all files have the timestamp of the checkout, and the packages will appear to be outdated. While this is not largely a concern now, this will become a problem once we have a binary package repository, because then the packages from the binary repo will always seem to be outdated, if you just freshly checked out the repository.

To combat this, git gets asked if the files from the aport we're looking at are in sync with upstream, or not. Only when the files are *not* in sync with upstream *and* the timestamps of the sources are newer, a rebuild gets triggered from now on.

In case this logic should fail, I've added an option during "pmbootstrap init" where you can enable or disable the "timestamp based rebuilds" option.

In addition to that, this commit also works on fixing #120: packages do not get updated in "pmbootstrap install" after they have been rebuilt. For this to work, we specify all packages explicitly for abuild, instead of letting abuild do the resolving. This feature will also work with the "timestamp based rebuilds".

This commit also fixes the working_dir argument in pmb.helpers.run.user, which was simply ignored before.

Finally, the performance of the dependency resolution is faster again (when compared to the current version in master), because the parsed apkbuilds and finding the aport by pkgname gets cached during one pmbootstrap call (in `args.cache`, which also makes it easy to put fake data there in testcases).


PabloCastellano:
(Not sure if it is a bug with Alpine Edge or related to this branch)

I've tried to test it by building my kernel, then making some change and then rebuilding. Unfortunately this error shows up the second time I rebuild the kernel.

```
(031810) [23:16:37] WARNING: armhf package 'so:libpython3.so' installed version 3.6.1-r4 is newer, than the version in the repositories: 3.4.5-r2
(031810) [23:16:37] (rootfs_motorola-titan) install
(031810) [23:16:37] (rootfs_motorola-titan) % apk --no-progress add -u /home/user/packages/user/armhf/linux-motorola-titan-3.4.42-r1.apk python3.4-dev pkgconfig python3.4
  python3-3.6.1-r4:
    conflicts:
               python3.4-3.4.5-r2[so:libpython3.so=0]
    satisfies: world[python3]
               mkbootimg-7.1.2_r6-r1[python3]
  python3.4-3.4.5-r2:
    conflicts:
               python3-3.6.1-r4[so:libpython3.so=0]
    satisfies: world[python3.4]
               python3.4-dev-3.4.5-r2[python3.4=3.4.5-r2]
```

ollieparanoid:
*erm* failing travis ... wait a moment^^

PabloCastellano:
Issue #131 has appeared to me with this branch

ollieparanoid:
Status update: rebased to master, implemented parsing of the `APKINDEX.tar.gz` files in the order they are written in the `/etc/apk/repositories` file to see if that fixes #131. In case it does, there is still some clean up left to do, before this can be merged (deleting APKINDEX files in `zap` is not required anymore, ...).

PabloCastellano:
After latest changes #131 is fixed. Ready to be merged :)

Issue #133, opened 2017-07-05
run() missing 1 required positional argument: 'flavor' in flash_system


antonok-edm:
Calling `pmbootstrap flasher flash_system` fails and generates the following log output:
[error.txt](https://github.com/postmarketOS/pmbootstrap/files/1124460/error.txt)

ollieparanoid:
Thanks to @MartijnBraam for fixing this in #134!

PR #134, opened 2017-07-05
Added flavor to the flash_system command for consistency

MartijnBraam:
The changes in flash_kernel broke the flash_system command (Issue #133). This makes the flavor argument optional for flasher.run()


ollieparanoid:
Thanks!

PR #136, opened 2017-07-05
[i9070][i9100] Build kernels natively

ollieparanoid:
This PR removes the need of the `_pmb_build_in_native_chroot` variable inside the APKBUILDs. Instead of compiling inside a native `armhf` chroot, to get Alpine's `busybox-static` binary, the `armhf` version gets installed inside the `x86_64` chroot. Just like `musl-armhf`.

This speeds up compiling a lot and removes the need for the `_pmb_build_in_native_chroot` hack, while still using Alpine's busybox instead of the pre-compiled blob shipped with the kernel sources.


drebrez:
Tested it and works everything. 
For me you can close it.

Issue #110, opened 2017-06-24
Automatic syntax formatting for all shell scripts


ollieparanoid:
That would be really helpful, something like `autopep8`, and an automatic check, like `flake8`. Suggestions welcome :)

ollieparanoid:
[shfmt](https://github.com/mvdan/sh) seems to be the right one for the job. However, it does not seem to be commonly packaged in Linux distributions, and with that program, we would introduce another dependency to the static code analysis script, making it even harder to install everything to run it locally before submitting a PR (which is required to keep development enjoyable).

I guess I will use it personally for formatting, but not let Travis require this for now.
It would be more user friendly to package all dependencies for static code analysis for Alpine Linux, and then use pmbootstrap to install them inside the native chroot, and do the static code analysis from there. That way, the host system does not get touched, and users do not need to worry about the dependencies.

If someone thinks, that this is a bad idea, please speak up here in the thread. This is how I would do it:

* Package shfmt, [shellcheck](https://bugs.alpinelinux.org/issues/7105) (this one has 11 (!) new Haskell dependencies), autopep8, flake8 for Alpine Linux
* Rewrite the static analysis script to:
  * Install all tools in the native chroot
  * Copy the pmbootstrap source folder into the chroot (in a temporary folder)
  * Run all linters (fail on error)
  * Run all formatters
  * Copy the modified source back

mvdan:
Let me know if you need any help packaging `shfmt` :)

Issue #88, opened 2017-06-14
Indicate initramfs status with LED


ollieparanoid:
Idea from @MartijnBraam, who said, that Jolla also uses the LED for initramfs status.
> For the led you need to glob through the /sys/class/led directories looking for one with RGB support

MartijnBraam:
Maybe create some small C executable to abstract away the status led?

ata2001:
On amami, I couldn't find anything with RGB support, there are directories for every supported color:
```
/sys/class/leds/led:rgb_blue/
/sys/class/leds/led:rgb_green/
/sys/class/leds/led:rgb_red/
```

MartijnBraam:
Maybe the led path should be added to the deviceinfo as an array since there isn't realy a naming standard for RGB leds (But they are working on a RGB led interface: https://patchwork.kernel.org/patch/8470681/).

```bash
# amami
deviceinfo_status_led  = "
/sys/class/leds/led:rgb_red
/sys/class/leds/led:rgb_green
/sys/class/leds/led:rgb_blue
"

# n900
deviceinfo_status_led  = "
/sys/class/leds/lp5523:r
/sys/class/leds/lp5523:g
/sys/class/leds/lp5523:b
"
```

ollieparanoid:
I don't think the python code can handle multiline arrays just yet in deviceinfo files (but we could add that).

Anyway, how about storing the names of the LED files, assuming, that they are all under `/sys/class/leds/`? That would be way shorter:
`deviceinfo_status_led = "lp5523:r lp5523:g lp5523:b"`

MartijnBraam:
only led names would work. But I found out that some led controllers need more than the brightness setting to change the value. For exampe:

> Simply set the brightness, then set on_off_ms, then trigger via rgb_start. Note that two transactions will need different numbers to be sent to rgb_start

https://stackoverflow.com/questions/24736562/nexus-5-cannot-change-led-colour-through-sysfs-interface

drebrez:
the samsung i9070 doesn't even have a notification led, but as an alternative we can use the button backlight:
`` /sys/class/leds/button-backlight/brightness``

MartijnBraam:
Isn't the notification light required in the android device standard?

PabloCastellano:
Hey.

I've started working on this and I have a proof of concept at https://github.com/PabloCastellano/pmbootstrap/tree/feature/indicate-initramfs-status-led

[...]

PR #138, opened 2017-07-06
Add kernel modules to device Samsung i9070

drebrez:
Add the compilation/installation of the kernel modules containing the wifi firmware from the vendor files repository.
Also add the modules-load.conf file to specify that the wifi module should be loaded at boot time.


drebrez:
@MartijnBraam Sure I can move it, it was the same location as in the vendor files repository.
Is it ok to have subfolders there? like `/lib/firmware/wifi` ?

MartijnBraam:
You can create subdirectories, but in the upstream firmware package the broadcom drivers into `/lib/firmware/brcm`

drebrez:
@MartijnBraam you mean I have to move them from `/lib/firmware/wifi` to `/lib/firmware/brcm`?

MartijnBraam:
It doesn't matter that much, I let @ollieparanoid decide. Maybe we should use /lib/fimware/pmos to avoid (future) conflicts with the linux-firmware package

ollieparanoid:
I like, that you've included the usual comment at the beginning of `modules-load.conf` and how you made small patches for the firmware paths :+1:

You have changed tabs to spaces in the `APKBUILD`s. Please don't do that, it makes the diff hard to read. In this pull request, and also when comparing to Alpine's original `APKBUILD` (in case of `linux-samsung-i9070` vs. `linux-vanilla`). Furthermore I'd like to have the `APKBUILD`s (and shell scripts) follow the same formatting style throughout the entire code base, so it looks consistent. (For Python we already follow the PEP8 standard - which actually uses whitespaces instead of tabs - and TravisCI checks for that. If you want to automatically format your `APKBUILD`s/shell scripts before making a PR, I recommend [shfmt](https://github.com/postmarketOS/pmbootstrap/issues/110) - and we might create automatized checks for that, too).

About the firmware path for `bcmdhd_sta.bin`: I'd prefer the non-conflict solution with `/lib/firmware/postmarketos/`.

I am not sure about the best location for the `nvram_net.txt`. If we treat it  like a config file, it should go to a folder in `/etc/`. I've seen, that you can configure the [mac address](https://github.com/Epirex/android_vendor_samsung_u8500/blob/cm-13.0/janice/system/etc/wifi/nvram_net.txt#L13) there? But then again, network managers and tools like `macchanger` (packaged for Alpine) should be able to [set the mac](https://github.com/alobbs/macchanger/blob/master/src/netinfo.c#L93) to any value at runtime, without touching that config file - the standard way it works for all Linux wifi drivers. So if users don't need to change this file at all, I think we should also put it to `/lib/firmware/postmarketos/`. If you want to, you could try whether the `macchanger` tool is working on your phone, to help making this decision. What do you guys think about this?

I like, that you have added a colorized prompt to the shell in `profile.sh`. But if we do that in the device-specific package, it will look different for every device, or be lots of maintenance effort to keep in sync between the packages. So how about putting that in `postmarketos-base` instead? We could even put a  generic "battery" function in there, if we put the path to the battery in the 
deviceinfo file. I suggest discussing this in a separate issue/PR.

One more thing, I realize, that there are aliases, that are user specific, such as 'ping -c 5', I haven't seen that before, but I can understand if that is useful for you. Maybe we could add a mechanism, so that you could include your user-specific configs during `pmbootstrap install` in the root image, but it does not get versioned along with pmbootstrap, such as a creating a user-specific package and blacklisting that in the gitignore? This also needs more discussion in a separate issue.

**Thank you very much for your hard work on this. It's obvious, that you know what you are doing and this wifi support alone brings the project forward by a great deal!**

PS: Why have you *reduced* the lzo compression, is it to save time during package building?

drebrez:
> You have changed tabs to spaces in the APKBUILDs.

I wasn't sure about it, I followed the Python standard. 
But I will revert it back to tabs.
(an automated check will be useful)

> About the firmware path for bcmdhd_sta.bin: I'd prefer the non-conflict solution with /lib/firmware/postmarketos/.

I'm fine, I will change it. 
Same question again, should we structure the firmware files in subfolders there? 
And for the wifi is better to have a generic `wifi` folder or we should use a `brcm` folder?

> I am not sure about the best location for the nvram_net.txt. If we treat it like a config file, it should go to a folder in /etc/. I've seen, that you can configure the mac address there?

For me make sense to leave it beside the firmware, since users doesn't have to change it. 
The mac address specified there seems ignored because when I load the module and bring the wlan0 interface up, the mac address is 00:00:00:00:00. I'm not sure why.
I will try to see if the `macchanger` tool works.

> profile.sh

I can remove it from this PR and we can discuss about putting a generic profile script in postmarketos-base

> battery function

Same as the profile script, I will remove it and we can create an issue/PR to discuss it. (maybe is related to #88)

> Why have you reduced the lzo compression, is it to save time during package building?

There is a comment about it in the APKBUILD, it's because the `lzop` inside the chroot was throwing and error during the compilation. This was in the armhf chroot, didn't tried if there is the same issue now that I changed it back to native chroot. Is the busybox installed there the same? 
It depends how the busybox was built (https://github.com/mirror/busybox/blob/master/archival/lzop.c#L1058-L1068)

ollieparanoid:
> Same question again, should we structure the firmware files in subfolders there?
> And for the wifi is better to have a generic wifi folder or we should use a brcm folder?

I think doing it without subfolders is enough. Then we don't need to think about the correct subfolder name, and we should not have many files in there anyway (before we get to mainline kernels, it will only be all firmware files for one device).

> For me make sense to leave it beside the firmware, since users doesn't have to change it.

:+1: (Trying out `macchanger` isn't necessary anymore, because you've already said, that the mac address configured in that file does not matter.)

> `lzop`

I've overlooked the comment. Thanks for sharing the link to the source. Yes, busybox in `x86_64` is compiled with the same config, as the `armhf` version in Alpine.

PR #144, opened 2017-07-10
Close #22: deviceinfo: add heimdall partitions / make flasher variables optional

### LINKED ISSUE #22: Boot process (Samsung I9070)
The Samsung I9070 does not have a separate recovery partition because it uses an Sony Ericsson SOC (U8500). This means I will have to do some work to boot pmOS on this device. However it is unclear to me how the boot process works.

The way I think it works:
Upon booting the device will load the Samsung initramfs from a hardcoded block offset. This initramfs contains an init script which you use to load the pmOS initramfs from the recovery partition. Your initramfs will load the actual kernel and boot pmOS.

Now since I don't have a recovery partition, I'm not sure what would be the best way for me to continue. Can I freely use another partition (such as KERNEL2) for the initramfs?

ollieparanoid:
This patch creates two new variables for the deviceinfo file:
* `deviceinfo_flash_heimdall_partition_kernel`
* `deviceinfo_flash_heimdall_partition_initfs`

The idea is, that we can use `pmbootstrap flasher flash_kernel` for both the `i9100` and the `i9070` (which has different partition names) then. I have taken the partition names for the `i9070` from the [device page](https://github.com/postmarketOS/pmbootstrap/wiki/Samsung-i9070-(Samsung-Galaxy-S-Advance)). I have tested it on the `i9100`.

@drebrez: Could you please do a review and check if this works in the `i9070`, too?
Thank you!


drebrez:
@ollieparanoid I didn't manage to test it properly, I keep getting this error while flashing:
`ERROR: libusb error -7 whilst receiving bulk transfer. Retrying...` 

We need to add `--no-reboot --resume` arguments to the first flash operation otherwise it reboots or it refuses to get other downloads.

What you think about flashing both file at once?
`heimdall flash --"$INITFS_PARTITION" /tmp/initramfs.lzo --"$KERNEL_PARTITION" "$KERNEL"`
with this we can remove the 20 seconds sleep and avoid possible errors.

Here the patch files of both changes: [heimdall_fixes.zip](https://github.com/postmarketOS/pmbootstrap/files/1146799/heimdall_fixes.zip)

In the meantime I'm trying to figure out the libusb error...

ollieparanoid:
Thanks for testing! There are lots of results about the libusb error, but I don't really have time right now to read through them - if all else fails, you could make an upstream bugreport at Heimdall's tracker.

I would love to do it without reboot or even flashing both images at once! Although when I had tested it some time back, both versions did not work - only the one with the reboot and sleep in between worked reliable for me. But I will test this again, a new version of Heimdall has been released in the meantime (already packaged for pmOS), and if it does not work, I'll report this upstream.

ollieparanoid:
There's a huge [issue](https://github.com/Benjamin-Dobell/Heimdall/issues/209), which starts out with the `ERROR: libusb error -7 whilst sending bulk transfer. Retrying...` message.

This is what I found interesting:
* *For what it is worth, I had this issue when my device was plugged into a USB3 port. Moving it to a USB2 port fixed it.*
* *If anyone still has problems, please open a new issue.*
* [detailed instructions](https://github.com/Benjamin-Dobell/Heimdall/issues/209#issuecomment-46785928) on how to create a report suitable for debugging (warning: requires proprietary software)

Anyway, this is an issue with heimdall, not that much with that PR (which moves the partition names to the deviceinfo file). How about we merge this PR and open up another issue in the pmbootstrap tracker about heimdall not working on your pc/with the i9070 (or are there any other issues with this PR)?
This would be very helpful to fix heimdall upstream I think.

ollieparanoid:
Thank you for your improvements, flashing in one go is much better. Tested on my `i9100` again, and everything works :+1:

Issue #146, opened 2017-07-11
Add configfs based usb network configuration


ollieparanoid:
@Decatf wrote in IRC:
> On newer kernels android uses configfs usb. Has there been any devices ported that succesfully use this?

Looks like we just need to extend the initramfs init script to the networking, like it is done [here](https://github.com/mer-hybris/hybris-boot/pull/79) and it should work.

ollieparanoid:
@Decatf also wrote, that he can only test this after he gets usb networking with that kernel to work in Android. So I'm setting this to "future" for now.

Decatf:
The mainline for Nexus 7 2013 kernel by John Stultz also uses configfs usb (judging from the defconfig). So perhaps anyone who is porting that device or some variant which uses that kernel could also work on this.

Decatf:
@ollieparanoid:
The ConfigFS USB network config is working on Android for my device. USB tethering is indeed working on Android. It only required some defconfigs and android init stuff.

For postmarketOS I tried applying applying the [config method from hybris](https://github.com/mer-hybris/hybris-boot/pull/79/files#diff-06a4f1d6faa676d41fc4b6a039d118c5L194) that you showed me in IRC. Here's the [init_functions.sh](https://github.com/Decatf/pmbootstrap/commit/c109d91ce64d6cf79f252578cefe4771cde2f33f) I'm using now.

With these changes there's no USB device showing up in dmesg on my PC yet. On the device kernel log there are now these two lines.
```
[   19.197869] using random self ethernet address
[   19.198039] using random host ethernet address
```

For reference [this](https://github.com/Decatf/android_device_samsung_p4-common/blob/aosp-7.1-upstream_kernel/rootdir/etc/init.p3.usb.rc) is what is the p4wifi device specific usb init looks like for Android.

ollieparanoid:
Great news, that it's working on Android!

Here is something different from the hybris-boot code (`USB` instead of `UDC`), is that intentional?
```
echo "$(ls /sys/class/udc)" > /config/usb_gadget/g1/USB
```

If that doesn't help it, you could dump the working config from Android while it is running (the configfs folder in this case) - maybe there is something different, that isn't clear from reading the Android usb init script.

ollieparanoid:
Thanks again for your pull request!

PR #148, opened 2017-07-12
Add ConfigFS based usb network configuration

Decatf:
This extends the initramfs init script to setup USB networking on kernels
using USB_CONFIGFS.

This addresses issue #146.


aports/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -92,14 +95,31 @@ find_root_partition() {
...
+	elif [ -d "$CONFIGFS" ]; then
+		mkdir -p $CONFIGFS/g1
+		printf "%s" "18D1" >"$CONFIGFS/g1/idVendor"
+		printf "%s" "D001" >"$CONFIGFS/g1/idProduct"
+
+		mkdir -p $CONFIGFS/g1/strings/0x409
+
+		mkdir -p $CONFIGFS/g1/functions/rndis.usb0
+
+		mkdir -p $CONFIGFS/g1/configs/c.1
```
drebrez: this line is not really necessary since `mkdir -p` takes care of creating any non-existent parent directory if necessary

aports/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -92,14 +95,31 @@ find_root_partition() {
...
-	[ -e "$SYS" ] || return
+	CONFIGFS=/config/usb_gadget
 
 	# Do the setup
-	printf "%s" "0" >"$SYS/enable"
-	printf "%s" "18D1" >"$SYS/idVendor"
-	printf "%s" "D001" >"$SYS/idProduct"
-	printf "%s" "rndis" >"$SYS/functions"
-	printf "%s" "1" >"$SYS/enable"
+	if [ -e "$SYS" ]; then
```
drebrez: we can use the same conditional expression for both checks. 
Don't know if it's better to use `-d` or `-e` .

Decatf:
- The configfs code is now in a new function called `setup_usb_network_configfs`.

 - I've removed the use of `mkdir` with the `-p` option. A config item is created with `mkdir` so now each line will do explicit creation of items instead having some intermediate item creation possibly happening with `mkdir -p`.

 - Both these bash conditionals will work. I've used the same as in `setup_usb_network_android` now.
```
-d file
    True if file exists and is a directory.

-e file
    True if file exists.
```
https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html

aports/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -102,6 +105,26 @@ setup_usb_network_android() {
...
+	mkdir $CONFIGFS/g1
+	printf "%s" "18D1" >"$CONFIGFS/g1/idVendor"
+	printf "%s" "D001" >"$CONFIGFS/g1/idProduct"
+
+	mkdir $CONFIGFS/g1/strings/0x409
+
+	mkdir $CONFIGFS/g1/functions/rndis.usb0
+
+	mkdir $CONFIGFS/g1/configs/c.1
+	mkdir $CONFIGFS/g1/configs/c.1/strings/0x409
```
ollieparanoid: This will fail, because the `$CONFIGFS/g1/configs/c.1/strings/` folder does not exist yet.
Please adjust this (re-add `-p` here or add another `mkdir` line), build the package and test if it works on your test device.

ollieparanoid:
Thank you again for making the changes!

Could you fix the `mkdir` pointed out above and increase the `pkgrel` to `0.1.5`? Then it would be ready for merge from my point of view :+1: 

(Please note, that in postmarketOS and Alpine Linux all scripts run with Busybox' POSIX compatible `ash` shell instead of `bash`. But the syntax you have pointed out work the same way in regular POSIX shells.)

aports/postmarketos-mkinitfs/init_functions.sh
Decatf: The strings folder is created internally by configfs code when a new folder is added to `$CONFIGFS/g1/configs/`.

```
localhost:/home/user# mkdir /sys/kernel/config/usb_gadget/g1/configs/c.2
localhost:/home/user# ls /sys/kernel/config/usb_gadget/g1/configs/c.2
MaxPower      bmAttributes  strings
```

ollieparanoid: Oh, I stand corrected. Thanks for the explanation!

PR #155, opened 2017-07-15
[RX51] Account for uInitrd filename change in u-boot script

### LINKED ISSUE #152: rootfs not found on rx-51 (Nokia N900)
Building pmos for the rx-51, and installing it to an sdcard (./pmbootstrap install --sdcard /dev/sdX) results in a kernel panic due to the kernel not being able to find a root filesystem (see screenshot)

My N900 is using u-boot from here: http://maemo.org/downloads/product/Maemo5/u-boot-flasher/

img_20170714_185727

Sorry for the fuzzy photo, the watchdog timer on the N900 reboots the device fairly soon after the kernel panics, so I had to be fast.

craftyguy:
The uInitrd filename is appended with 'flavor' so, for example, with the
rx-51 the generated file is named 'uInitrd-nokia-rx51'. This corrects
the rx51's u-boot script to account for this change in filename.

This fixes #152

### DIFF
--- a/aports/device-nokia-rx51/APKBUILD
+++ b/aports/device-nokia-rx51/APKBUILD
@@ -1,6 +1,6 @@
 pkgname=device-nokia-rx51
 pkgver=1
-pkgrel=5
+pkgrel=6
 pkgdesc="Nokia N900"
 url="https://github.com/postmarketOS"
 arch="noarch"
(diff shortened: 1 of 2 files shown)


MartijnBraam:
Seems that this was broken in https://github.com/postmarketOS/pmbootstrap/commit/71f4308fc9a8968dc27f01619e9b1568f79246f7#diff-b4785c1c08837b227fee56e773497250 (and that PR was approved by me, I should probably have cleared the whole SD before testing that name change)

PR #156, opened 2017-07-15
[RX51] Add touchscreen calibration udev rule

craftyguy:
This adds a udev rule for calibrating the touchscreen on the RX51.
The values used were obtained using weston-calibrator on my device.


craftyguy:
Hmm... without this, my touch screen is upside down (and yes, I'm holding the device correctly!) I wonder what could be going on here..

craftyguy:
Wait, did you figure out why it was upside down for you?

MartijnBraam:
I have no idea, If I calibrate it again I get the same values but still upside down, might have merged this one slightly too quickly

PR #157, opened 2017-07-15
Enable appended DTB image on mozilla flame device

yuvadm:
Also rebases @kskarthik's work onto the latest master, I hope that's acceptable on a pending device branch. If not I'll create a new PR with a plain merge.


yuvadm:
On second though, this isn't elegant at all. Will open a new PR.

PR #158, opened 2017-07-15
[RX51] Add linux-firmware package

craftyguy:
This removes the call to `firmware_install` in the APKGBUILD for
linux-nokia-rx51, and adds the `linux-firmware` package as a dependency
for device-nokia-rx51.

There are two immediate benefits for this:
1) wifi now works

2) the FM radio device was causing the boot to take ~2 minutes longer
than it should, now that the firmware is present there's no delay in
initializing this device


MartijnBraam:
Hmm this is indeed way neater, as long as we don't get issues with firmware for linux 4.11 on a 4.6 kernel

craftyguy:
If #159 (move to mainline) is approved, that wouldn't be a problem. In any case, it seems to be extremely rare that firmware in linux-firmware breaks device drivers. That's based on my limited experience though, the big ones (e.g. iwlwifi) will select the desires firmware version to load and installing a newer linux-firmware package has no effect on them.

PR #159, opened 2017-07-16
[RX51] Update to mainline kernel

craftyguy:
A couple of issues are resolved by moving to the mainline kernel (from Pali's
kernel):

1) 'reboot' command now works as intended
2) boot is *much* faster
3) charging LED seems to be working now

@MartijnBraam I would appreciate your feedback!


MartijnBraam:
This works great, but it also conflicts with your firmware PR

craftyguy:
Yea I know, I'll resolve the conflict once one gets merged..

PR #160, opened 2017-07-16
Dictionary lookup of event loop actions

cclauss:
This was kind of an experiment that adds lots of blank lines to this script -- feel free to reject it.  (102 additions / 47 deletions :-( )

The observation that triggered this was that in main() all actions can execute using just one parameter, args.  This means that it is possible to use args.action to do a dict lookup to get the corresponding function.  This approach adds tons of blank lines to the script in exchange for a rapid seek to the action implementation and a substantial reduction of the McCabe complexity of main().

$ __flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics__
```
./pmbootstrap.py:43:1: C901 'main' is too complex (23)
./pmb/aportgen/core.py:38:1: C901 'rewrite' is too complex (17)
[ ... ]
```


ollieparanoid:
Thanks for the PR, I like the idea! The only thing I'd change is removing the redundant `action_dict` and calling the functions directly by their name (how about an `action_` prefix for the functions?).
[Here](https://stackoverflow.com/a/834451) is how to call functions, when you store the name in a variable (which would be `"action_" + args.arction` then).

ollieparanoid:
Thanks again!

PR #170, opened 2017-07-17
Add acpid package

craftyguy:
This commit adds the acpid daemon. This requires a kernel with NETLINK
support compiled in before daemon will run. acpid will be extremely
helpful for devices that have external buttons which need event
handling. The nokia-rx51 is full of such devices/switched!


MartijnBraam:
The chmod inside the package step seems out of place, also I think the manpages should be handled by the automatic docs package stuff from abuild.

If this package is complete it can probably be upstreamed to alpine.

craftyguy:
I agree. As discussed in IRC, this is a 'first take' implementation meant to make this package available to us for device support. Definitely needs more polish before upstreaming it though..

aports/acpid/APKBUILD
```diff
@@ -0,0 +1,30 @@
+# $Id$
```
ollieparanoid: I suggest removing the Contributors lines, the whole header in fact. Because then no one will ask the Arch contributors about problems with this pmOS/Alpine package (this is how I've handled it with `pmbootstrap aportgen`, it also removes the maintainers/contributors). You could give credit to them/to Arch in the commit message.

craftyguy:
@ollieparanoid 

Thanks for the feedback and help with the -docs fiasco. I've made the changes you requested and tested that the package builds as expected.

PR #173, opened 2017-07-17
Fix PR #160

craftyguy:
Probably meant to call 'globals()' and not 'locals()'. As @MartijnBraam
pointed out in IRC, probably best to have these functions in a separate
module so they can be parsed with getattr


ollieparanoid:
I screwed up testing in #160, must have tested on the old code without the PR :-\ Thanks for the immediate fix! *testing now*

PR #175, opened 2017-07-17
Fix 145: add deviceinfo_flash_offset_base again

ollieparanoid:
Tested with `lg-mako`, and the initramfs generation still works.
@drebrez review is important to see if this works for his HTC desire.
It would be cool if @PabloCastellano or @MartijnBraam could also review this :+1:


drebrez:
found a few issues, here is a patch file with my suggested fixes:
[fix_PR_175.patch](https://github.com/postmarketOS/pmbootstrap/files/1154174/fix_PR_175.patch.txt)

ollieparanoid:
@drebrez Thanks for your patch! It looks good to me, only setting the defaults for `deviceinfo` variables is something we should move to `pmb/config/__init__.py` in the long run. But that will be obvious when implementing a linter for the `deviceinfo` files (as it is on my personal milestones plan), so I suggest we keep it that way for now.

Could you push your patch as commit to this branch, and could @MartijnBraam and @PabloCastellano  do the review then? Thanks a lot! :+1:

drebrez:
@ollieparanoid just pushed the fixes, @PabloCastellano and @MartijnBraam could you please have a look an test it? thanks :wink:

VictorNine:
Tested on HTC Desire HD and works

PR #177, opened 2017-07-17
[RX51] Enable NETLINK in kernel (WIP)

craftyguy:
This IPC interface to the kernel is needed for acpid to function.

@MartijnBraam, FYI!


craftyguy:
don't merge this one just yet, I may teed to tweak it a little

craftyguy:
I don't think this is needed anymore. Sorry for the noise!

PR #179, opened 2017-07-17
Two trivial changes (typo + info message)

PabloCastellano:


pmb/flasher/frontend.py
```diff
@@ -62,6 +63,13 @@ def kernel(args):
     else:
         logging.info("(native) flash kernel " + flavor)
         pmb.flasher.run(args, "flash_kernel", flavor)
+    logging.info("You will get an IP automatically assigned to your "
+                 "USB interface shortly.")
+    logging.info("Connect to the telnet session and type your LUKS "
+                 "password to boot postmarketOS:")
```
MartijnBraam: This should probably only be printed if luks is actually enabled.

PabloCastellano: That was my first approach but args.full_disk_encryption is only available when `pmbootstrap install` is run. This is `pmbootstrap flasher`

ollieparanoid: you could add `(not necessary when full disk encryption disabled):` at the end of the sentence.

PabloCastellano:
Changes done

PR #180, opened 2017-07-18
Add asus-grouper

antonok-edm:
This PR includes pablog's new patches for weston. They worked really well for me, but I suppose it's worth checking with him before merging. I've added the PR in case it's ready to go.


aports/device-asus-grouper/deviceinfo
```diff
@@ -0,0 +1,25 @@
...
+deviceinfo_modules=""
+deviceinfo_modules_initfs=""
+deviceinfo_external_disk="false"
+deviceinfo_external_disk_install="false"
+deviceinfo_flash_methods="fastboot"
+deviceinfo_arch="armhf"
+
+# Fastboot related
+deviceinfo_generate_bootimg="true"
+deviceinfo_flash_offset_base="0x00000000"
```
drebrez: @antonok-edm we have an open PR that adds again the support of this variable, which means that at the moment it was ignore and was using the default value of "0x10000000", if your device has the default base address you should remove it from the deviceinfo, otherwise you can set the correct value there

PabloCastellano:
Good work!

For the sake of simplicity, could you rebase your PR from upstream/master? (being upstream postmarketOS/pmbootstrap repository)

```
git rebase upstream/master
git push -f
```

PabloCastellano:
When you are asked about the conflict (because https://github.com/postmarketOS/pmbootstrap/pull/183 is not merged into master), simply type `git rebase --skip`

antonok-edm:
All set.

aports/device-asus-grouper/deviceinfo
drebrez: @antonok-edm the `deviceinfo_flash_offset_base` variable is now supported in mainline branch, please merge it and verify the correct value you need in your deviceinfo.
If your device has an offset base of `0x10000000` you can simply remove the `deviceinfo_flash_offset_base` variable.

ollieparanoid:
Thank you very much for the PR!

PR #184, opened 2017-07-18
[RX51] Set keyboard layout for weston

craftyguy:
This configured weston.ini to include the appropriate keyboard layout
for the RX51, so now we can enter numbers in weston!

@MartijnBraam, FYI

### DIFF
--- a/aports/device-nokia-rx51/APKBUILD
+++ b/aports/device-nokia-rx51/APKBUILD
@@ -1,6 +1,6 @@
 pkgname=device-nokia-rx51
 pkgver=1
-pkgrel=8
+pkgrel=9
 pkgdesc="Nokia N900"
 url="https://github.com/postmarketOS"
 arch="noarch"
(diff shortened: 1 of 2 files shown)


ollieparanoid:
Looks good to me! @MartijnBraam : please test, review, merge if it works for you :+1:

craftyguy:
Sigh. Don't merge this yet.

Apparently it conflicts with postmarketos-base, which provides weston.ini (for some unknown reason). @MartijnBraam had a good idea of allowing layout to be set in deviceinfo, and having some package parse that to set weston.ini. I don't know what package this would be, seems like it should be weston since weston.ini is very specific to weston. Alternatively, let device packages own weston.ini since settings may be device-specific.

In any case, this PR is trending towards being useless.

craftyguy:
Alright, this PR has changed a bit, but depends on PR #191 to make any sense.

craftyguy:
Since #191 was killed/replaced, this PR now depends on #201 to make any sense.

craftyguy:
@MartijnBraam please test this out, it has been updated to be compatible with #201

PR #191, opened 2017-07-18
Support setting weston keymap options from deviceinfo

craftyguy:
Since postmarketos-base owns weston.ini, this modifies the post-install
hook for this package to support deviceinfo_weston_keymap options for
setting weston.ini. Please test this on devices that do not have these
options in deviceinfo!

A follow-on PR will prompt users for the layout value (e.g. 'us' or
'de', etc) and dump this into deviceinfo when running 'pmbootstrap
init'.

It's worth mentioning that if these deviceinfo variables are not set, then an empty [keyboard] section will be added to weston.ini. Weston will still start successfully, and without errors, it simply just ignores this empty section.


craftyguy:
This PR has been replaced by #201

Issue #178, opened 2017-07-17
Create generic interface for handling device-specific ACPI events in user-definable ways


craftyguy:
There should be a way to handle ACPI events from things like button presses & switches on a device in a way that is device agnostic and supports users re-defining actions.

ACPI events are not generic, but seem to be specific to different devices even though the 'button' is shared. One example is a 'power button', which might generate different ACPI events on different devices, but the desired behavior for such event might the same across different devices (e.g. to turn off the display or power off the device)

One idea of how this flow would play out, using an example of a keyboard slider that generates a specific, unique (to the specific device) ACPI event:

1) user slides out keyboard on device 

2) acpid receives `KBD_SLIDE/00BB0000` (I made this up, but the point is this can be specific to devices/hardware and is not generic)

3) acpid runs generic handler which matches `KBD_SLIDE/00BB0000` with `KEYBOARD_SLIDE_OUT` (a 'generic' event to associate with what just happened) 

4) The generic handler calls `xdg-open pmarketos:keyboard_slide_out` to run whatever the user wants

5) pmarketos:ACTIONs are defined in `/etc/xdg/mimeapps.list`, device maintainers can provide default settings for their respective devices (e.g. `device-nokia-rx51` provides some defaults for acpi events specific to the rx51)

6) The device-specific-ACPI-event to generic-event mapping is made in some .conf file accessible by the handler, defined by device maintainers, and provided in the device-specific package (e.g. `device-nokia-rx51`)

MartijnBraam:
This stuff is mostly handled by intents in android but linux doesn't really have something like this as far as I know. Also do you have any idea how things like long pressing the power button are handled? can we fire a seperate script with acpid for that?

craftyguy:
@MartijnBraam if it generates a separate acpid event it would be trivial to support a different action. You bring up a good point though, I don't know. Now that I have a kernel which supports NETLINK and a working acpid package, I can test this out later when I get more time. If it does not end up generating a different event, then maybe my suggestion is a no-op :(

ollieparanoid:
I like this concept! @MartijnBraam raises a good concern though with the long presses, but you can test that. We might even extend acpid at some point to support long presses, if necessary, but it would be a good start even without the long presses!

For the Android devices I have tested, the power button turns off the device regardless of the OS running after pressing it for a few seconds.

BTW: I've realized that we will need an OpenRC service for acpid in the long run (I think Gentoo already has one).

[...]

PR #192, opened 2017-07-19
[RX51] Add support for physical events w/ acpid

craftyguy:
This adds a way to handle physical events like keypad slide switch,
screen lock, camera lid, and others. This uses acpid from busybox (I was
wrong about the real acpid package), and an acpi.map file to map events
to 'scripts'. The scripts are symlinked to /etc/acpi/handler.sh,
automatically by the new post-install script, where
they are handled based on which script was called. This allows for
easily adding whatever is necessary for #178.

If this PR is merged, I will create a new wiki page detailing how others
can use this framework on other devices.


craftyguy:
@MartijnBraam 

Thanks for the feedback. I have discovered that it is futile to start busybox's acpid on bootup, because it starts before the kernel has finished modprobing devices (like gpio_keys) and the current implementation of busybox acpid essentially ignores any new devices that come up after it has started. I have reached out to the busybox mailinglist for further assistance (hopefully with the decision to patch it).

In the meantime, I propose that this PR is merged so I (and maybe others) can continue to implement the "not implemented" features in the acpi handler here. I have also added volume keys, headphone jack, and power button events.

ollieparanoid:
Great work! Also how you've handled this with talking to upstream :)

If I understand you correctly, the `acpid` package is not needed anymore? In my opinion we should delete it if we have no use case for it. Otherwise it will become outdated over the time and have security holes.

In case you want to, I could open up another repository for discarded packages and move it there.

MartijnBraam:
Well maybe the complete acpid daemon fixes the problems with busybox acpid having race conditions with loading modules.

craftyguy:
It does, it uses inotify to monitor for new devices I believe..
however I couldn't figure out how to wire it up to the buttons/keys
since it monitors /dev/acpi (or something similar) and ARM doesn't
actually have ACPI.

ollieparanoid:
So it doesn't make sense for our use case and we could remove the package?

craftyguy:
I think so, but want to investigate a little more just to be sure.

PR #197, opened 2017-07-19
[RX51] Enable DRM in kernel

craftyguy:
This enables DRM (Direct Rendering Manager) support in the kernel for
the RX51. This opens up the possibility of using the Weston DRM backend,
which in turn allows for DPMS (I hope) and other cool tricks.

n900:~# glxinfo|head
name of display: :0
display: :0  screen: 0
direct rendering: Yes
server glx vendor string: SGI
server glx version string: 1.4

n900:~# glxgears
94 frames in 5.0 seconds = 18.772 FPS

Note, `CONFIG_DRM_OMAP_CONNECTOR_ANALOG_TV` is required to work around the issue identified here: https://lkml.org/lkml/2017/6/30/191


craftyguy:
@MartijnBraam I would appreciate you checking this one out!

MartijnBraam:
I'm building this branch now, hmm I think I got more than 18 FPS in maemo

craftyguy:
>  I think I got more than 18 FPS in maemo

Yea.. the performance isn't great because I *think* they used a binary blob driver for the PowerVR SGX530 graphics.. I guess we can talk about whether or not we want to include this blob here for performance, assuming it even works with the mainline kernel now since it likely hasn't been updated for years. I'm not even sure where you can find it now.

MartijnBraam:
I've booted maemo and I get 22 FPS with all the binary drivers so it's not so bad as I thought.

MartijnBraam:
It's working for me, and somehow my touchscreen now also works but that might be a side effect from something else

craftyguy:
> and somehow my touchscreen now also works 

Haha, yea that's probably unrelated since the only things being changed are related to DRM and the LCD panel (not the touchscreen controller). Then again, I had to enable some completely unrelated CONFIG_DRM_OMAP_CONNECTOR_ANALOG_TV option to get omap_drm to load successfully, so what do I know! :P

PR #201, opened 2017-07-19
Handle generating weston.ini in post-install script, based on deviceinfo options

craftyguy:
This removed the weston.ini file and automatically creates it in the
postmarketos-base post-install script based on options set in the
deviceinfo file for a particular device. This replaces #191.

(Optional) options supported in this PR for deviceinfo are:

deviceinfo_weston_core_modules - defaults to **NONE** if none
specified

deviceinfo_weston_core_backend - defaults to fbdev backend if none
specified

deviceinfo_weston_keymap_rules - keymap rule, found under /usr/share/X11/xkb/rules/

deviceinfo_weston_keymap_model - keymap model


aports/postmarketos-base/postmarketos-base.post-install
```diff
@@ -46,3 +48,46 @@ if ! apk audit /etc | grep -q etc/motd; then
...
+
+weston_config="/etc/xdg/weston/weston.ini"
+
+# Clear out weston.ini if it exists for some reason, or create a new file
+mkdir -p /etc/xdg/weston
+if [ -f $weston_config ]; then
+        echo '' > $weston_config
+else
+        touch $weston_config
+fi
```
ollieparanoid: We shouldn't just delete files from the file system in my opinion. As a user I'd be really angry if a package decides to delete my config. Package managers never overwrite manually made changes in /etc.
I think you can get rid of the while if.. block, the touch is redundant, too (or is it not?).

aports/postmarketos-base/postmarketos-base.post-install
```diff
@@ -46,3 +48,46 @@ if ! apk audit /etc | grep -q etc/motd; then
...
+
+### [Core]
+echo "[core]" >> $weston_config
+# modules
+if [ -n "$deviceinfo_weston_core_modules" ] && [ "$deviceinfo_weston_core_modules" != "" ]; then
```
ollieparanoid: I'd write that as:
```
if [ -z "$deviceinfo_weston_core_modules" ]; then
# the default
else
# use it
fi

aports/postmarketos-base/postmarketos-base.post-install
craftyguy: I considered that.. so should I just return/ignore options in deviceinfo and not write anything to weston.ini if that file already exists?

ollieparanoid:
Thanks for making the changes! As you know, I only looked at it quickly yesterday. I've realized now, that we would get breakage from all devices, that have XWayland disabled by defaulting to XWayland. So I must suggest another change, sorry.

For all devices, that have [this hack](https://github.com/postmarketOS/pmbootstrap/blob/faaec71289cb0a744f349a6d8b6c6d5ff6969500/aports/device-lg-mako/profile.sh#L5) in their `profile.sh`:
* remove that ugly hack, the whole if ... fi thing (yay!)
* make sure in the deviceinfo, that it does not default to xwayland (maybe it 
would be better to specify "" as default?)
* regenerate all device-* checksums, where you've made changes
* please try to test if the disabled xwayland works in advance (if the 
weston.ini gets generated correctly, you could modify the rx51 deviceinfo to 
test that quickly).

craftyguy:
Great feedback, thank you. I'll make those changes, and test as best as I can on the RX51, but you may want to mention other folks who work on devices that would be affected to also test the PR before merging it.

craftyguy:
@ollieparanoid 
All of your requested changes have been implemented.

When no modules are specified, this entry is not set in weston.ini (tested on my RX51), xwayland does NOT load in this case.

I also tested that populating the new deviceinfo fields results in those fields being input into weston.ini by installing the postmarketos-base package.

The 'dirty hack' has been removed from lg-hammerhead, lg-mako, and motorola-titan. Please have folks associated with those devices test this PR to make sure it works as intended!

craftyguy:
> It creates the correct weston.ini for my n900, but the n900 doesn't seem to have the keymap config yet.


Right, I haven't merged the PR to modify the N900's deviceinfo to add the settings for keymap. I wanted this PR to be merged before I do that, since I will also need to add weston config to that other PR.

MartijnBraam:
Yeah I thought so, that's why it's still an approved review.

ollieparanoid:
I'll test this on my `lg-mako` and then merge it :+1:

ollieparanoid:
It works for me, almost there! Could you increase the `pkgrel` for the changed `device-` packages? Then it's ready to ship! 🚢

craftyguy:
@ollieparanoid `pkgrel` has been bumped in all device- APKBUILDS!

Issue #59, opened 2017-06-04
Small Onscreen-Keyboard for initramfs to enter LUKS password


lawl:
We need a small (as in binary size and dependencies) on-screen keyboard to enter the LUKS password to decrypt the root partition.

Potential starting point to fork off: https://github.com/zrafa/onscreenkeyboard

Other pointers to code/tools we could use welcome.

dzervas:
:+1:

ollieparanoid:
Some recommendations from research I made a while ago:
* When writing a new keyboard from scratch:
  * Try to use an established keyboard layout format (something from GNU/Linux or TWRP/AOSP?)
  * Try to support framebuffer/dri/x11/wayland if possible. Pay attention to DRI at least, because that is what should be used in favor of framebuffer in mainline Linux.
    * it might make sense to statically link it against SDL, so you'd get all these platforms and it shouldn't be *that* big
* It may make sense to have the keyboard in ladscape-view in the initramfs
* There is an existing GNU/Linux on screen keyboard standard with dbus. Certainly nothing for the initramfs, but if you intend your keyboard to be usable outside of the initramfs, you might be able to dynamically load dbus `.so` library and integrate into the desktop, if you want?

With that being said, even the simplest keyboard, that just works, would be a huge improvement compared to the telnet typing we do now. Keep it simple :)

MartijnBraam:
It's also important to watch what gets started in initramfs since that is not in a crypto partition but running processes stay loaded after unlocking, If someone flashes a new boot partition without the phone owner knowledge it could insert a rootkit in the initrd that has access to the unlocked rootfs.

ollieparanoid:
*This is a bit off-topic, but I think when an attacker is able to flash a new boot partition without the owner's knowledge, then it does not matter what was or was not on the boot partition.*
I'm all in for stopping what is running in the initramfs before entering the normal init though.

We can discuss this more on IRC, if you guys want to, but let's keep this out of the ticket.

ollieparanoid:
Thanks to @MartijnBraam's efforts, we have a device with a keyboard working now. It would be nice, if the unlocking screen also supported typing in text with a real keyboard (but if it can't be done easily, then ignore this - we could also disable showing the on-screen-keyboard unlocking method, when the keyboard is set to true in the deviceinfo).

MartijnBraam:
[...]

PR #206, opened 2017-07-20
Created splash screen generator

MartijnBraam:
This package creates a splash screen based on info in the deviceinfo file in it's postinstall hook.

Example:

test

This provides the make-splash.py script:

```
usage: make-splash.py [-h] [--landscape] [--text TEXT] width height filename

Splash generator for postmarketOS

positional arguments:
  width
  height
  filename

optional arguments:
  --landscape
  --text TEXT  Additional text for the splash screen
  --config FILENAME Config file for style overrides
  --center  Center the text underneath the logo
```

The image above is created with:

```bash
$ make-splash.py --text="On-screen keyboard is not implemented yet, Plug in a USB cable and run on your PC: telnet 172.16.42.1" 800 480 test.ppm
```

## ToDo

- [x] Figure out how to make mkinitfs use the generated files


MartijnBraam:
Hmm flake8 complains that a line is missing while it exists...

drebrez:
@MartijnBraam can you also create the second splash screen with the script? the one with "Please wait while..." displayed after you unlock the partition

MartijnBraam:
It can generate splashes with any text you specify (with the --text parameter)

MartijnBraam:
Here is it running on my n900:

splash

ollieparanoid:
### Initramfs integration
I recommend this:
* `postmarketos-mkinitfs` depends on this package
* the splash screen gets generated in the post-install hook of `postmarketos-splash`, according to the resolution specified in the deviceinfo
* `mkinitfs.sh` would add that file to the initramfs

### Python 3
I'm thinking a bit about having Python 3 in the base system though - we have it anyway in most cases, because for `mkbootimg` we use the Python 3 version provided by AOSP currently. But [this](https://github.com/osm0sis/mkbootimg) could be a suitable replacement (maintained more regularly, standalone repo, might start tagging releases soon, at least I have requested it via mail) and as it happens to be a fork of the C version, we would have Python 3 out of the base system again (@drebrez found this fork :)).

The Alpine devs would try to keep the base system as small as possible, and I'm also a fan of that.

Anyway, the only satisfying alternative I can think of (which does not need Python 3) would be writing a tiny C program, that generates the splash screen on the fly and displays it (with fbdev or dri!). Then we could also put fancy animations in there ;)

But that is a lot more effort to get right than writing a short and maintainable Python program, so **in conclusion I think this is a good solution for now!** (Besides, we want to have a on-screen-keyboard for initramfs in the long term #59, so maybe it makes more sense to throw a splash screen into that code later.)

### Suggestions
* Directly write a .ppm.gz file, it saves some space (ppm is really verbose) and is directly ready to use in `mkinifs` then.
* Save the generated splash image under a less generic name, e.g. `/usr/share/postmarketos/telnet.ppm.gz`
* Also generate the "Loading..." splash? -> `loading.ppm.gz`
* Use a lowercase 'p' character in *... yet, plug in ...*
* Center the text below the logo
* Own line for the telnet command (if easily possible)
* The ubuntu font is pretty characteristic, I'd rather have it a bit more differentiated. How do the 
[other fonts](https://pkgs.alpinelinux.org/packages?name=ttf*&branch=edge&repo=main&arch=armhf&maintainer=) look like?

\o/ Thank you for making this!

MartijnBraam:
The zipping from the gz can be easily done with gzip after generating the splash. If we add this as dependency for postmarketos-makeinitfs then it can call make-splash.py directly and pass it the correct paths to save it in the initramfs (and call gzip after).

The --text command does support newlines for the USB command.

centering the text would be kinda difficult since this is already manually doing word-wrapping for the line breaks.

MartijnBraam:
Here are a few tests:

Logo: Droid black, text: Droid regular

screenshot

Logo: Droid light, text: Droid regular

screenshot

Everything Dejavu regular

screenshot

kskarthik:
Dejavu regular is quiet good :)

MartijnBraam:
I like the second one the most myself, dejavu doesn't have a lot of character, it should've been called generic-font

ollieparanoid:
I like both the last and the middle version. So what you are saying @MartijnBraam is, we would get rid of the post-install hook and directly generate the splash from the `mkinitfs` script? Sounds good to me.
Then I'd suggest removing the hook from this PR, and after this PR is through modifying the mkinitfs package accordingly.

MartijnBraam:
Hmm maybe I can make the font paths configurable (since this is linux) so end users can change the splash config. 

So `/etc/postmarketos/splash.ini` with:

```ini
[background]
color="#000000"

[name]
text=postmarketOS
font="/usr/share/fonts/TTF/Roboto-Light.ttf"
color="#ffffff"

[logo]
text="♻"
font="/usr/share/fonts/ttf-dejavu/DejaVuSans.ttf"
color="#009900"

[text]
font="/usr/share/fonts/TTF/Roboto-Light.ttf"
color="#999999"
```

Since we're not creating ios but something customizable. This change is also quite easy to implement with configparser in python3

ollieparanoid:
The config sounds good! So here are the TODOs, as I see it:
* implement that cool config
* remove the post install hook (because we'll be calling this from mkinitfs)
* ship it! (please merge, when you have implemented the above two points and Travis' check runs through)

...and in another PR:
* make `postmarketos-mkinitfs` depend on `postmarketos-splash`
* generate the splashes during initramfs generation

aports/device-nokia-rx51/deviceinfo
```diff
@@ -14,3 +14,8 @@ deviceinfo_external_disk_install="true"
 deviceinfo_flash_methods="0xFFFF"
 deviceinfo_generate_legacy_uboot_initfs="true"
 deviceinfo_arch="armhf"
+
+# Weston Config
```
ollieparanoid: I think this shouldn't be in the diff. Thanks for the code update, I'll test it later and write feedback.

MartijnBraam: I rebased that file to origin/master to avoid a conflict when merging

MartijnBraam:
Well look at that, I made a mess of git again. It should be neat again when squashed on merge

drebrez:
@MartijnBraam @ollieparanoid I suggest having a third splash screen with a message to inform the user that the system partition is missing, because right now it shows nothing for that case...

ollieparanoid:
Good idea @drebrez! (but we'll need to do that in the second PR, that adjusts the mkinitramfs)

aports/postmarketos-splash/APKBUILD
```diff
@@ -0,0 +1,23 @@
...
+subpackages=""
+source="make-splash.py"
+
+build() {
+	return 0
+}
+
+package() {
+	install -D -m755 "$srcdir"/make-splash.py \
+                "$pkgdir"/usr/bin/make-splash.py
```
ollieparanoid: I find the installation path name very generic. How about `pmos-make-splash`? (I'd keep `make-splash.py` as file name in the aport though)

*(I've used `mkinitfs` in `postmarketos-mkinitfs` to stay compatible with Alpine's `mkinitfs`, which gets called from `linux-vanilla`'s post-install script for example - that was before I realized we'll roll our own mainline kernel package anyway.)*

aports/postmarketos-splash/APKBUILD
```diff
@@ -0,0 +1,23 @@
...
+pkgrel=0
+pkgdesc="Splash screen for postmarketOS"
+url="https://github.com/postmarketos"
+arch="all"
+license="mit"
+depends="ttf-dejavu py3-pillow"
+makedepends=""
+install=""
+subpackages=""
+source="make-splash.py"
```
ollieparanoid: Maybe this got lost during rebase? The example config is missing in `source` and `install` now.

aports/postmarketos-splash/make-splash.py
```diff
@@ -0,0 +1,97 @@
...
+        else:
+            lines.append('{word} '.format(word=word))
+            line_length = length + space_length
+
+    if response_type == 'list':
+        return [line.strip() for line in lines]
+    elif response_type == 'str':
+        return '\n'.join(line.strip() for line in lines)
+    else:
+        raise ValueError('Invalid response type. Valid values are "list" and "str".')
```
ollieparanoid: This got lost:
```python
if __name__ == '__main__':
    import argparse

    parser = argparse.ArgumentParser(description="Splash generator for postmarketOS")
    parser.add_argument('width', type=int)
    parser.add_argument('height', type=int)
    parser.add_argument('filename')
    parser.add_argument('--landscape', action='store_true')
    parser.add_argument('--text', type=str, help='Additional text for the splash screen')
    args = parser.parse_args()

    make_splash(args.width, args.height, args.filename, args.landscape, args.text)
```

aports/postmarketos-splash/APKBUILD
MartijnBraam: No it's not missing, I just added that file as reference. The generator works fine without supplying a config file as it has internal defaults for everything and providing an actual config file in the package will make it conflict with other packages that might overwrite it.

MartijnBraam:
The tool also supports centering text now, which is kinda important to have a neat "Loading..." screen.

This is an example with some settings changed with the config file:

test

aports/postmarketos-splash/APKBUILD
ollieparanoid: Reference configs installed into the system are useful, so it's easy to edit them right away.
Why would another package provide a config here?

ollieparanoid:
I love being able to customize stuff in my Linux distros, perfect! Thanks for adding the center option.

Almost ready to ship! (The checksums are wrong though.)

aports/postmarketos-splash/APKBUILD
MartijnBraam: Maybe you want to have the text font and background color different if you install postmarketos-ui-plasma to make it integrate it better (might be a horrible example). The config can be included somewhere in `/etc` or `/usr/share/doc`.

ollieparanoid: Like Ubuntu, Kubuntu etc. all have different boot screens? That makes sense to me.
But what if the user wants to install multiple `postmarketos-ui` packages? Then you would have the conflict there, so I don't see any improvement compared to just installing the config from this package.
How about we install it for now, and if we notice that it needs to be changed because such a use case comes up, we change it afterwards?

MartijnBraam:
I've implemented the changes and added the config file to `/etc/postmarketos/splash.ini`

ollieparanoid:
It is really fun to play with, thanks :>

Issue #209, opened 2017-07-20
Cross-compiling fails with C++ code


PureTryOut:
`pmbootstrap.py` tries to cross-compile packages, and although C code works fine, C++ doesn't. It falls back to GCC in qemu, and since Plasma Mobile is basically all C++ code, this makes compiling *really* slow.

`pmbootstrap.py log_distccd` reports (for example):
```
distccd[10023] (dcc_job_summary) client: 127.0.0.1:34472 COMPILE_ERROR exit:1 sig:0 core:0 ret:0 time:420ms g++ /home/user/build/src/solid-5.31.0/src/solid/devices/frontend/camera.cpp
```

An example APKBUILD (does not matter what package in practice, as long as it contains C++ code):
```
pkgname=karchive
pkgver=5.31.0
pkgrel=1
pkgdesc='Qt 5 addon providing access to numerous types of archives'
arch="all"
url='https://community.kde.org/Frameworks'
license="LGPL"
depends="qt5-qtbase"
makedepends="extra-cmake-modules doxygen qt5-qttools-dev"
source="https://download.kde.org/stable/frameworks/${pkgver%.*}/${pkgname}-${pkgver}.tar.xz"
validpgpkeys="53E6B47B45CEA3E0D5B7457758D0EE648A48B3BB" # David Faure <faure@kde.org>

prepare() {
  mkdir -p "$srcdir/build"
}

build() {
  cd "$srcdir/build"
  cmake ../${pkgname}-${pkgver} \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_INSTALL_PREFIX=/usr \
    -DKDE_INSTALL_LIBDIR=lib \
    -DBUILD_TESTING=OFF
  make
}

package() {
  cd "$srcdir/build"
  make DESTDIR="${pkgdir}" install
}
sha512sums="4a82b91956bd52a7b9da6c40bcaa0340b3e1b0e964ce44be8993a6a62abde7514d449df1748a79a4bf7c8a7aacaaf4bc3f2b125a7748dde5f02a9ad5c49efb48  karchive-5.31.0.tar.xz"
```

ollieparanoid:
[Heimdall](https://github.com/Benjamin-Dobell/Heimdall/) is C++ code and I think it works there, so it might be QT specific? I'll check this out, thanks for making this bug report!

PR #216, opened 2017-07-21
Various distccd related improvements, mostly respect --verbose

ollieparanoid:
I've done some refactoring while debugging #209.
* Unused file `pmb/build/crosscompiler.py` removed (that was a
  left over from the now obsolete `_pmb_build_in_native_chroot` hack)
* Do verbose logging in distccd, when `pmbootstrap --verbose` is
  being invoked
* Restart distccd, when the commandline has changed (e.g. when the
  currently running version was not verbose, and the new one is
  verbose.) Prior to this change, it only got restarted, when the
  architecture changed (so it did not allow changing the job count
  on the fly for example).
* Insert missing whitespace in arguments help.


ollieparanoid:
Thanks for the review! Yeah, it's necessary, because it gets written into the info file (as plain string). But as you say, it never gets executed that way.

PR #225, opened 2017-07-21
Close #195: Ignore parameters in dependencies

### LINKED ISSUE #195: pmbootstrap.py can't handle dependencies like "py-alabaster<0.8"
It just complains it can't find `py-alabaster<0.8`, while it should find any py-alabaster version lower than  0.8, which is in fact available in the repositories.

```
(022745) [16:14:59] Build is neccessary for package 'extra-cmake-modules': Binary package out of date (binary: 5.31.0-r0, aport: 5.36.0-r0)
(022745) [16:14:59] Calculate depends of packages ['py-sphinx', 'cmake'], arch: x86_64
(022745) [16:14:59] ERROR: Could not find package 'py3-alabaster<0.8' aport and could not find it in any APKINDEX
```

ollieparanoid:
Please test this branch for a while, doing a full installation or so, so there is enough time to uncover potential bugs. It works fine for me.
@PureTryOut: I'd also be happy if you could test it, because it should fix the bug you reported :)

---
* Ignore `>`, `<`, `=` and `!` operators, when they are specified in
  the dependencies. This was the desired behavior before, but it was
  not implemented correctly (so it wouldn't ignore them everywhere).
  Of course the real fix would be to honor these operators like apk
  does. But this isn't feasible right now, and it should work for
  most, if not all, our use-cases. I have documented this in the wiki
  under build internals and if we happen to need correct operator
  handling, we should do it then.

Minor other changes:
* `pmbootstrap parse_apkindex`: support optional package parameter to
  only show the parsed content for one package.
* Support building most python APKBUILDs by replacing ${pkgname#py-}
  properly


ollieparanoid:
If you have a fast PC: `pmbootstrap zap -p`, `pmbootstrap install` and see if everything runs through

ollieparanoid:
Thank you very much for reviewing and testing! Now we only need to wait until @PureTryOut confirms, that this fixes his use case and then we can merge it.

PureTryOut:
This seems to work fine! The package that caused this problem (`py-sphinx`) is now being found, and although now a different error comes up (yay...), that's unrelated to this PR. So LGTM!

PR #227, opened 2017-07-22
Close #194: Support subfolders in aports

### LINKED ISSUE #194: Separate folders for devices & packages under aports
Since, number of supported devices are growing, also more packages are planned to be added in future (plasma-mobile) I suggest that we need to separate the device builds from the packages. 

Does it create any problems from code level?

ollieparanoid:
This is heavily work in progress and does not patch up all places yet.
To continue this, basically grep the code for `args.aports` and find
the places, where it does not use `pmb.build.other.find_aport()` yet.
Then adjust the testcases, and test everything manually.


ollieparanoid:
Alright, the PR is **ready to be reviewed and tested**. The testsuite runs through and all places, where `args.aports` have been used were modified to use the new folder structure. To test, please run `aports_subfolders.sh`, to adjust the folder structure (I have not done that yet to keep the diff in this PR clean - I would do that just before the final rebase and merge).

This is a deep architectural change for my taste, so it should get tested a lot. That's why I've assigned all of you (which does not mean that everyone will find the time of course, and that is not necessary - one or two good tests would be enough).

[Here](https://github.com/postmarketOS/pmbootstrap/wiki/Migration-to-aports-subfolders) are migration instructions, that I wrote down. Please try to improve them. I will write the link to these instructions in the channel's topic, once this PR is through.

Issue #91, opened 2017-06-15
"The Mainline Kernel"


ollieparanoid:
One of the goals of pmOS is to unify multiple devices in using one and the same kernel binary.
For a few selected devices, this goal is coming unexpectedly close already - **in the meaning of: the kernel should boot, and we should get output on the serial interface**, allowing further development from there:
* ~~[nokia-rx51 (N900)](https://github.com/postmarketOS/pmbootstrap/wiki/nokia-rx51-%28Nokia-N900%29) ([most hardware is actually working!](http://elinux.org/N900#Kernel_Status))~~
* [lg-hammerhead (Nexus 5)](https://github.com/postmarketOS/pmbootstrap/wiki/lg-hammerhead-%28Google-Nexus-5%29)
* [lg-mako (Nexus 4)](https://github.com/postmarketOS/pmbootstrap/wiki/lg-mako-%28Google-Nexus-4%29)? (assuming it is similar enough to the nexus 7, [ticket](https://github.com/postmarketOS/pmbootstrap/issues/12))
* ["Nexus 7"](https://lwn.net/Articles/680109/) and more devices listed there in the comments ([asus-grouper](https://github.com/postmarketOS/pmbootstrap/wiki/asus-grouper-%28Google-Nexus-7-2012%29) would be supported?)
* [QEMU](https://github.com/postmarketOS/pmbootstrap/pull/56) ;)
* [samsung-apexqtmo (Galaxy S Relay 4G)](https://cascardo.eti.br/blog/News_on_Debian_on_apexqtmo/)

We won't be able to use Alpine's `linux-vanilla` for these devices, because:
* We really need the latest kernel (as support for these devices is being actively developed on)
* We will need custom patches, until they are all mainlined
* We will need different config options for the kernel.

Tasks:
* [x] Create a `linux-postmarketos` aport based on `linux-vanilla`
* [x] Get it working with qemu (should be the easiest)
* [x] Get it working with nokia-rx51
* [x] Get it working with lg-hammerhead
* [x] Close this ticket and continue documentation in the [wiki](https://github.com/postmarketOS/pmbootstrap/wiki/The-mainline-kernel)

PabloCastellano:
[...]

Issue #64, opened 2017-06-05
Binary repo / Continuous integration should run the testsuite


ollieparanoid:
- [x] "lazy reproducible builds" (apks won't be fully reproducible, but all files inside the apks should have the same content, and each apk should contain the same files. the exception is the signature file, which is of course different.)
  - [x] generate a buildinfo.json file with all relevant versions of all dependencies, recursively
  - [x] implement `pmbootstrap challenge`, which makes sure, that the exact dependencies are installed, rebuilds the package, compares all file's contents inside the apks
  - [x] make sure, that all pmbootstrap packages are reproducible (`linux-` packages don't work yet, but they are not required for the testsuite. All other packages work.)
  - [x] Merge `lazy-reproducible` builds into master
- [x] Automatic challenging of every build before publishing:
  - [x] Building and signing on a dedicated machine ("pmbuilder", I wanted to use a libreboot Thinkpad, but that's too much work as of now, might do that later though).
  - [x] That machine gets a dedicated GitHub account (@pmbuilder) and uploads its binaries into [pmOS-binary-packages-staging](https://github.com/pmbuilder/pmOS-binary-packages-staging)
  - [x] Write a script, that finds the next package, that needs to be updated, builds it, and uploads it to a dedicated repo: [pmbuilder](https://github.com/pmbuilder/pmbuilder)
  - [x] Travis CI job, that challenges the build and - on success only - pushes the signed build from @pmbuilder (and *not* its own build!) to the [binary-package-repo](https://github.com/postmarketOS/binary-package-repo), together with the change APKINDEX
- [ ] Stabilization phase:
  - pmbootstrap
    - [x] Include the `.buildinfo.json` file for each package, when releasing it
    - [x] Verify the APKINDEX
      - Usage: `pmbootstrap challenge APKINDEX.tar.gz`
      - Verify, that all files referenced in the APKINDEX exist in the same folder
      - Verify, that there is no extra file in the repo, that is not mentioned in the APKINDEX
    - [x] Testcases for all challenge related functions
    - [x] Add switch `--experimental-binary-pmOS-packages` or something, to be used by the Travis CI  job until it is stable
    - [x] When a package is up-to-date in the binary repo, install it from there and don't build it locally!
[...]

PR #228, opened 2017-07-22
Unified linux-postmarketos kernel for qemu and n900

MartijnBraam:
The creates the `linux-postmarketos` package which for now supports the n900 (and probably the other maemo devices) and qemu with the vexpress-a9 machine simulation.

I've put the generated dtbs in `/usr/share/dtb` for now and set the dtb field in the deviceinfo for the n900 and my new qemu device.

## ToDo

- [x] Append the dtb to the kernel somewhere in the buildsystem (instead of in the kernel package)
- [ ] Integrate launching qemu somehow
- [x] qemu graphics support
- [x] qemu storage support
- [ ] qemu keyboard support
- [x] qemu networking support
- [ ] qemu mouse support
- [x] Rebase this stuff for the newer device packages in the master branch


aports/linux-postmarketos/APKBUILD
```diff
@@ -9,8 +9,8 @@
 # Changes:
 # - enable devtmpfs (needed for udev -> touch support in weston)
 
-_vendor=nokia
-_flavor=nokia-rx51
+_vendor=postmarketos
```
ollieparanoid: The `_vendor` variable isn't used anywhere, so we don't need it.

aports/linux-postmarketos/APKBUILD
```diff
@@ -9,8 +9,8 @@
 # Changes:
 # - enable devtmpfs (needed for udev -> touch support in weston)
```
ollieparanoid: Please adjust the whole comment block here, so it really describes the changes compared to `linux-vanilla`.

ollieparanoid:
I am glad, that you are working on this!

However, I think we need to discuss the handling of the board architectures a bit more. The two of us [agreed on](https://github.com/postmarketOS/pmbootstrap/issues/91#issuecomment-316793280) having multiple builds for different boards is a good idea (dear reader, if you think otherwise, please let us know).

So with the current PR you're just ignoring all other boards, but that won't scale. I have two approaches in mind:
1. Have a single package, and generate the other kernel binaries as subpackages (so as output we'd have `linux-postmarketos-omap3`, `linux-postmarketos-otherboard`, ...).
* Advantages:
  * Just one package to maintain, that builds all versions
* Disadvantages:
  * Building that package will take forever once we have a reasonable amount of boards. (This will probably become problematic for #64 , which will utilize Travis to rebuild and verify these packages - and Travis has a maximum runtime of 50 minutes.!)
   * When compiling/packaging one board fails, you must rebuild and repackage all of them, which will take lots of time!

2. Have one package *per board*.
* Advantages:
  * shorter compile/package time (see above)
* Disadvantages:
  * a lot more maintenance effort to keep in sync

---
Another idea I have is writing some Python code, so we can have `pmbootstrap aportgen linux-postmarketos-omap3`. Then we could do the second approach easily, and it would scale - we would only need to maintain the unified aportgen code for all boards and generate the rest from there.

So my proposal would be 2. + aportgen, what do you think, everyone?

MartijnBraam:
My current plan was ignoring the size of the kernel until it becomes a real problem since not a lot of devices have mainline kernel support yet. The devices that have very tiny boot partitions aren't devices that have mainline support at all. Splitting this up in seperate packages for different board architectures is relatively easy (just duplicating the package and changing the config) or can be done with the subpackage support in alpine.

I'm currently modifying the APKBUILD to be closer to the linux-vanilla one again since we have module loading working for this so it also generates a linux-postmarketos-dev package.

ollieparanoid:
That sounds like a good plan to me!

(In my opinion, maintaining different packages becomes a problem when shipping a security patch for example (or if we decided to also build `-hardened` versions, with Alpine's grsec fork).)

aports/linux-postmarketos/APKBUILD
```diff
@@ -21,11 +21,11 @@ case $pkgver in
 	*.*.*)  _kernver=${pkgver%.*};;
 	*.*) _kernver=$pkgver;;
 esac
-pkgrel=3
+pkgrel=1
 
 arch="armhf"
-pkgdesc="Mainline Linux for the Nokia N900"
-url="https://github.com/pali/linux-n900"
+pkgdesc="Mainline Linux for OMAP chipsets"
```
ollieparanoid: With the comments in mind, I suggest:
`pkgdesc="Mainline Linux configured for pmOS supported chipsets (OMAP)"`

craftyguy:
Just to confirm, the direction is to ultimately create one unified kernel for all devices (based on mainline) rather than having multiple mainline kernels for multiple devices?

MartijnBraam:
Yeah the grand plan is have one kernel for everything (like on desktop linux). We might have to settle for a few seperate linux kernels (build from the same tree so no extra maintenance time) so the kernel binary doesn't grow too big.

MartijnBraam:
@ollieparanoid do you know where and how in the build process the dtbs should be appended now?

ollieparanoid:
In `mkinitfs`: before creating the `uinitrd` and `bootimg` I suggest a new function for appending the dtb file (so [here](https://github.com/postmarketOS/pmbootstrap/blob/14f164b8a27e4ba1f037983114013907bafc9b13/aports/postmarketos-mkinitfs/mkinitfs.sh#L190)).

We could do one of the following:
1. Create a `vmlinuz+dtb` file in mkinitfs and adjust all code paths to use `vmlinuz+dtb` instead of `vmlinuz` when it is present.
2. In the mainline kernel APKBUILD, do not install the kernel as `vmlinuz`, but as `vmlinuz-without-dtb` or something, then create a `vmlinuz` file in mkinitfs, when the `vmlinuz-without-dtb` file is present.

I think the second solution is more elegant, what do you think?

ollieparanoid:
I realize that the mkinitfs does a lot more now than making the initramfs. Maybe we should rename that program. But this is where I think it belongs.

ollieparanoid:
And about how to append: `cat vmlinuz dtbfile > vmlinuz+dtb` should do it.

MartijnBraam:
I think it's better to use `vmlinuz-dtb` since that's used by more linux mainline stuff and the current stuff we use. Also all current images are suffixed with the flavor so I think you would get:

For devices with dtb:

```
vmlinuz-postmarketos-dtb
vmlinuz-postmarketos
```

and for devices without dtb (I don't think any mainline devices would boot without dtb) you just don't have the -dtb file.

Then `create_bootimg()` needs to use the -dtb if the deviceinfo_dtb variable is set and otherwise use the plain kernel.

The non-mainline devices currently don't have the deviceinfo_dtb set and write the image+dtb to the plain vmlinuz-flavor file so they keep working without a change.

MartijnBraam:
In d307488 I've written some (untested) code that implements my idea above.

MartijnBraam:
I'm hitting an issue now with mkinitfs. It's called twice directly in succession with different parameters.

log:
```
(171/171) Installing xorg-server-xwayland (1.19.3-r2)
Executing busybox-1.27.0-r2.trigger
Executing dbus-1.10.18-r0.trigger
Executing kmod-24-r0.trigger
Executing fontconfig-2.12.4-r0.trigger
Executing mkfontscale-1.1.2-r0.trigger
Executing mkfontdir-1.0.7-r0.trigger
Executing postmarketos-mkinitfs-0.1.8-r0.trigger
==> initramfs: creating /boot/initramfs-postmarketos
install: can't stat 'drm': No such file or directory
install: can't stat 'drm_kms_helper': No such file or directory
install: can't stat 'ext4': No such file or directory
Creating 800x480 splashscreen
Creating 800x480 splashscreen
==> kernel: appending device-tree omap3-n900
==> initramfs: creating uInitrd
Image Name:   uInitrd
Created:      Sun Jul 23 14:57:48 2017
Image Type:   ARM Linux RAMDisk Image (uncompressed)
Data Size:    7063001 Bytes = 6897.46 kB = 6.74 MB
Load Address: 00000000
Entry Point:  00000000
==> kernel: creating uImage
Image Name:   postmarketos
Created:      Sun Jul 23 14:57:48 2017
Image Type:   ARM Linux Kernel Image (uncompressed)
Data Size:    3826793 Bytes = 3737.10 kB = 3.65 MB
Load Address: 80008000
Entry Point:  80008000
Executing glib-2.52.1-r0.trigger
Executing shared-mime-info-1.8-r0.trigger
Executing gdk-pixbuf-2.36.6-r1.trigger
Executing gtk-update-icon-cache-2.24.31-r0.trigger
OK: 627 MiB in 187 packages
(032481) [16:57:52] Calculate depends of packages ['postmarketos-mkinitfs'], arch: armhf
(032481) [16:57:52] Calculate depends of packages [], arch: armhf
(032481) [16:57:52] (rootfs_nokia-rx51) mkinitfs postmarketos
(032481) [16:57:52] (rootfs_nokia-rx51) % mkinitfs -o /boot/initramfs-postmarketos 4.12.0
==> initramfs: creating /boot/initramfs-postmarketos
install: can't stat 'drm': No such file or directory
install: can't stat 'drm_kms_helper': No such file or directory
install: can't stat 'ext4': No such file or directory
Creating 800x480 splashscreen
Creating 800x480 splashscreen
==> kernel: appending device-tree omap3-n900
==> initramfs: creating uInitrd
Image Name:   uInitrd
Created:      Sun Jul 23 14:57:59 2017
Image Type:   ARM Linux RAMDisk Image (uncompressed)
Data Size:    7062918 Bytes = 6897.38 kB = 6.74 MB
Load Address: 00000000
Entry Point:  00000000
==> kernel: creating uImage
Image Name:   postmarketos
Created:      Sun Jul 23 14:57:59 2017
Image Type:   ARM Linux Kernel Image (uncompressed)
Data Size:    3826793 Bytes = 3737.10 kB = 3.65 MB
Load Address: 80008000
Entry Point:  80008000
(032481) [16:57:59] Calculate depends of packages ['postmarketos-mkinitfs'], arch: armhf
(032481) [16:57:59] Calculate depends of packages [], arch: armhf
(032481) [16:57:59] (rootfs_nokia-rx51) mkinitfs postmarketos-dtb
(032481) [16:57:59] ERROR: [Errno 2] No such file or directory: '/home/martijn/.local/var/pmbootstrap/chroot_rootfs_nokia-rx51/usr/share/kernel/postmarketos-dtb/kernel.release'
(032481) [16:57:59] Run 'pmbootstrap log' for details.
```

MartijnBraam:
The appending works now, now I only need @ollieparanoid's opinion to check another box in the start post.

MartijnBraam:
This kernel seems to work way better on the n900 than the one currently in the master branch (because the display boot bug is dependend on some race condition and this magically makes it work better)

The x86 qemu build can't run yet because of #260 and the qemu armfh build doesn't have input working but still boots fine.

@ollieparanoid how about merging the progress so far to improve the development for the n900 (and let more people mess with qemu)

aports/device-qemu-amd64/APKBUILD
```diff
@@ -0,0 +1,23 @@
+pkgname=device-qemu-amd64
```
ollieparanoid: Clever how you've used `qemu` as vendor and `amd64` as device name!

aports/linux-postmarketos/APKBUILD
```diff
@@ -0,0 +1,182 @@
+# APKBUILD loosly based on linux-vanilla aport. Changes:
```
ollieparanoid: Please correct this to `loosely`.

ollieparanoid:
*took (why can't I edit the review, GitHub?)

...and Travis CI is failing. Please fix that, too.

MartijnBraam:
You can change the review comment (at least I can change yours). you get a edit icon beside the view changes on hover

ollieparanoid:
BTW: I didn't think of putting "-dtb" in the flavor, great solution!

ollieparanoid:
(Deleted the branch, I suggest making a new branch for the next PR - please tell me if you think that this does not make sense.)

MartijnBraam:
yeah the branch should be deleted I think with github flow

PR #231, opened 2017-07-22
Finish packaging base of plasma-mobile

PureTryOut:
Everything now compiles. Let's start testing it!


ollieparanoid:
So stoked to see the first photo of this actually running on a device :>

PR #233, opened 2017-07-23
De-couple weston from postmarketos-base

craftyguy:
This commit:

- UI is selectable with `pmbootstrap init`. Currently only 'weston' and 'none' are options. It'll automatically pick up any new `postmarketos-ui-*` packages added at later dates.

- splits off weston packages install from postmarketos-base and puts
them into postmarketos-ui-weston. Also note that NO weston packages are
installed by "pmbootstrap install" by default unless the user selects a ui in the `init`

- configuration of weston is now in postmarketos-ui-weston. 

- the demos have been spun off to `postmarketos-demos`, and `postmarketos-ui-weston` lists this package as a dependency.


aports/postmarketos-demos/APKBUILD
```diff
@@ -5,7 +5,7 @@ pkgdesc="Simple touch menu for a few demo programs"
 url="https://github.com/postmarketOS"
 arch="all"
 license="GPL3+"
-depends="weston-clients"
+depends="postmarketos-ui-weston"
```
MartijnBraam: I don't think you should create a circular dependency between postmarketos-demos and postmarketos-ui-weston, the dependency on weston-clients is probably fine as-is.

Also I think postmarketos-demos depends on weston-xwayland for the main window so that's a dependency that was missing from the original package.

pmb/config/init.py
```diff
@@ -37,6 +38,12 @@ def init(args):
     cfg["pmbootstrap"]["device"] = pmb.helpers.cli.ask(args, "Device",
                                                        None, args.device)
 
+    # UI selection
+    ui_list = pmb.helpers.ui.list(args)
+    logging.info("Available UIs (" + str(len(ui_list) - 1) + "): " + ", ".join(ui_list))
+    cfg["pmbootstrap"]["ui"] = pmb.helpers.cli.ask(args, "Enter a valid UI package to install:",
```
MartijnBraam: This should validate if the user chose an UI that actually exist instead of asking not to mess up.

craftyguy: I completely agree with you, and actually started implementing this, but then realized that NONE of the input into `init` is validated. I dont' want to also implement "actually validate user input into `pmbootstrap init`" in the same PR since this PR is already pretty busy.

I think user validation into `init` should be implemented, just not in this PR. So, I am following the exact same methodology that existing input into `init` follows.

aports/postmarketos-demos/APKBUILD
craftyguy: @MartijnBraam: the problem with adding weston apps as dependencies into other packages (e.g. postmarketos-demos) is that now postmarketos-ui-weston is no longer splitting off weston from other postmarketos components. Either `postmarketos-demos` and `postmarketos-ui-weston` need to be merged into a single package, which I can do, or everything that needs weston should depend on `postmarketos-ui-weston`. The circular dependency here is because @ollieparanoid wanted the demos installed by default when a user selects the 'weston' UI. I couldn't think of a better way to handle this, since there are cases where a user may select *no* UI and the later try to install the postmarketos-demos packages by itself (which would result in a 'broken' weston installation since postmarketos-ui-weston would not be pulled in)

MartijnBraam: Yeah but `postmarketos-demos` doesn't actually depend on the `ui-weston` package and the demo package might be added as dependency for another ui (like sway) as test.

I don't think getting a broken deskop by installing only the demo package is a big problem, If you install firefox on a barebone linux install you also don't get a desktop manager, just the xorg client components so you have to install your window manager to use it.

pmb/config/init.py
craftyguy: @MartijnBraam I just filed #242, maybe I, or someone else, can implement user input validation in the `pmbootstrap init` function..

aports/postmarketos-demos/APKBUILD
craftyguy: Ah, I see, so just depend on `weston-xwayland` in `postmarketos-demos`, and if the user installs the demos without installing the `postmarketos-ui-weston` package then it's OK they will end up with a broken desktop for the reason you stated.. (firefox on bare linux system)

MartijnBraam: Yeah that seems right to me, I think that anyone that is randomly installing packages on their phone does know enough to not destroy everything, also not installing postmarketos-ui-weston doesn't really break anything except that the demo's cannot be started.

craftyguy: Ok, change made :)

craftyguy:
@ollieparanoid do you have any feedback here?

ollieparanoid:
I'll have a look :)

aports/postmarketos-base/APKBUILD
```diff
@@ -10,14 +10,11 @@ tmux htop nano postmarketos-mkinitfs eudev sudo"
 makedepends=""
```
ollieparanoid: please increase the package version.
This is a major architectural change, so I suggest: `pkgver=2`, `pkgrel=1`.

aports/postmarketos-demos/APKBUILD
```diff
@@ -5,7 +5,7 @@ pkgdesc="Simple touch menu for a few demo programs"
 url="https://github.com/postmarketOS"
```
ollieparanoid: Please increase the `pkgrel`.

pmb/config/init.py
ollieparanoid: I agree with @MartijnBraam but for me it's okay if we do this in a second PR like @craftyguy proposed.
With the other questions I have followed a theme of having a long explanation on top, and then a short name before the input field. How about this?

`Available user interface packages ($count): ...`
`Interface:`

pmb/helpers/ui.py
```diff
@@ -0,0 +1,33 @@
...
+along with pmbootstrap.  If not, see <http://www.gnu.org/licenses/>.
+"""
+import os
+import glob
+
+
+def list(args):
+    """
+    Get all UIs, for which aports are available
+    :returns: ["None", "postmarketos-ui-one", "psotmarketos-ui-two", ...]
```
ollieparanoid: "None" spelled with an uppercase "n" here, please adjust that.

ollieparanoid:
How about making `weston` the default UI? because that is the default now, and if someone doesn't know about the user interfaces, weston is better than no UI if you want to have the demos.
`none` is something for the next step, when you have weston working and want to experiment I think (unless we got plasma-mobile or another, real phone UI working in the future - then one of those should be the default in my opinion).

ollieparanoid:
From the testing side it looks good. But I'd like to test again after the changes are addressed. Thanks for making this PR 👍

aports/postmarketos-base/APKBUILD
craftyguy: done!

aports/postmarketos-demos/APKBUILD
craftyguy: done!

craftyguy:
@ollieparanoid, feedback incorporated:

1) Weston is the default

2) prompt changes to be in line with other prompts in `init`

3) fixed "None" --> "none" 

4) increased pkgver & pkgrel for postmarketos-base

5) increased pkgrel for postmarketos-demos

ollieparanoid:
Tested and working, boots straight up to weston like it should on my `i9100`. Thanks for making this, great improvement!

PR #235, opened 2017-07-23
Ignore directories while searching for loop devices

cmdr2:
Fixes an issue on alpine-vanilla, where `/dev/loop` is a directory (containing `/dev/loop/0`, `/dev/loop/1` ..etc), and causes `./pmbootstrap install` to fail.

Ignoring the `/dev/loop` directory works because `/dev/` also contains files named `/dev/loop0`, `/dev/loop1` ..etc, which the new code will still pick up.

The `install` command passes after this fix on my alpine-vanilla.


ollieparanoid:
Tested and working. Thank you!

Issue #107, opened 2017-06-23
pmbootstrap does not work with the hardened Alpine kernel on host system


ollieparanoid:
This is quite ironic. It appears to be working with the hardened kernel from Arch Linux though.

rfs613 reported in IRC:
> Just as an FYI... while pmbootstrap installs an alpine linux chroot, this doesn't work if your host OS is Alpine.
More specifically, if you use the default grsec-enabled Alpine.
the chroot command works, but inside the chroot, you cannot install setuid binaries. So "apk add sudo" fails with a permission error.
This is apparently a grsec feature. It can be disabled, or you can run vanilla kernel (no grsec) to get around it.

ollieparanoid:
Looks like it works in Arch, because it only uses 2% or so from the original grsec patchset. ([source](https://www.reddit.com/r/archlinux/comments/68b2jn/linuxhardened_in_community_repo_a_grsecurity/) as found by rfs613).

PabloCastellano:
cmdr2 reported in IRC that `sysctl -w kernel.grsecurity.chroot_deny_chmod=0` does disable grsec in Alpine and pmbootstrap works.

ollieparanoid:
I think we could check for that option, before executing any command in a chroot the first time in a pmbootstrap session (the lifetime of one pmbootstrap command, such as "chroot" or "install") and throw an error, telling the user how to turn this off. Then we could close this issue :)

cmdr2:
To comprehensively disable grsec until the next reboot, I suggest:
  ls -1 /proc/sys/kernel/grsecurity | xargs -I '___' sudo sysctl -w kernel.grsecurity.___=0

The previous sysctl command I suggested only disables one flag, and the build fails later due to some other flag blocking it. That throws a 'mv ..... : could not preserve permissions' error (something like that, can't find the error now).

ollieparanoid:
Thanks for the details, @cmdr2 and @PabloCastellano!

@cmdr2, could you test if the following script raises an exception, when the option is enabled, and runs through without complaining when it is not?

Then we could add a check like that to pmbootstrap and close this ticket :+1: 

```python
#!/usr/bin/env python3
import os

def grsec_check():
    path = "/proc/sys/kernel/grsecurity/chroot_deny_chmod"
    if not os.path.exists(path):
        return
    
    with open(path, "r") as handle:
        status = handle.readlines()[0].rstrip()
    
    if status != "0":
[...]

PR #236, opened 2017-07-23
Update runtime error message thrown for grsec/hardened kernels

cmdr2:
Currently the error message says pmbootstrap can be made to work by disabling grsec, which is not true anymore (the build fails later since qemu is not working correctly with a hardened kernel).

This PR updates the message thrown.

More investigation at #107


ollieparanoid:
Thanks!

PR #240, opened 2017-07-23
New splash images in mkinitfs

MartijnBraam:
This makes `postmarketos-mkinitfs` call the new `pmos-make-splash` command to generate the splash images instead of using the prepackaged ones.


aports/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -181,9 +181,18 @@ install -Dm755 "/usr/share/postmarketos-mkinitfs/init.sh.in" \
 install -Dm755 "/usr/share/postmarketos-mkinitfs/init_functions.sh" \
 	"$tmpdir/init_functions.sh"
 
-for i in /usr/share/postmarketos-mkinitfs/splash*.ppm.gz; do
-	install -Dm644 "$i" "$tmpdir"/"$(basename $i)"
-done
+# generate the splash screens
```
ollieparanoid: Could you please put that in its own function `generate_splash_screens()`?

MartijnBraam:
The splash images also haven't been testted on other devices than the n900 yet so they might be rotated on other devices (but I have added a --landscape parameter, without any code yet).

ollieparanoid:
Yeah you are right, it is cut-off on the `samsung-i9100`, so the telnet message can't be read anymore and the postmarketOS text is also cut-off. I really do not want to ship half-broken features, so I suggest the following (it is some work to do, I know and sorry for that):
* Search for the real resolutions for all supported devices and put them in the deviceinfo
* Adjust your code to automatically rotate the splash screen, when the width is shorter than the height (i.e. detect landscape automatically - maybe the option is redundant then?)
* Actually test this yourself on both the n900 and another device, such as hamerhead
* Update the wiki with the new deviceinfo fields

But although it is cut off it looks very promising. Can't wait until we have this 👍

MartijnBraam:
The rotation doesn't really depend on the width and height but more how the hardware is scanned out since fbsplash writes directly to the framebuffer memory. I think the output from weston is rotated 90 degrees to the fbsplash output I guess?

Also your i9100 has the 480x800 resolution but my fallback is 720x1280.

ollieparanoid:
Sorry if this was confusing from me, the output is rotated as expected I think - the postmarketOS text is in parallel to the Samsung logo on the top of the phone.
But still, the text is cut off, please implement the bullet points above so we don't have breakage on the devices where the current splash screen is readable (even if breakage just means having a broken boot splash screen).

MartijnBraam:
It should work perfectly now on your i9100 (and the other devices). I still have to charge my hammerhead so I can't test that right away.

ollieparanoid:
I will test this now. Please increase the pkgrel of the `device-` packages.

ollieparanoid:
Tested and working. Looks really awesome now \o/

PR #243, opened 2017-07-23
Fix Plasma Mobile installation

PureTryOut:
Even though it all compiled, there was a problem which prevented `plasma-mobile` and `plasma-phone-components` to install, which is now fixed.

I've also gone through and added all subpackages I could find (`-doc` and `-lang`).


aports/baloo/APKBUILD
```diff
@@ -6,10 +6,12 @@ arch="all"
 url='https://community.kde.org/Frameworks'
 license="LGPL"
 depends="kfilemetadata kio lmdb hicolor-icon-theme"
-depends_dev="kfilemetadata lmdb-dev kidletime"
-makedepends="$depends_dev extra-cmake-modules kdoctools python doxygen qt5-qttools-dev"
+makedepends="$depends_dev extra-cmake-modules kdoctools python doxygen
```
ollieparanoid: As I see it, you have deleted the $depends_dev variable above, but used it in $makedepends. Please also remove it from $makedepends then.

aports/farstream/APKBUILD
```diff
@@ -1,14 +1,13 @@
 pkgname=farstream
 pkgver=0.2.8
-pkgrel=1
+pkgrel=2
```
ollieparanoid: You have only increased the pkgrel in some APKBUILDs. But these aports are not in master yet, and there are so many of them, I think it doesn't necessarily need to be changed (might cause problems for people who have already built the last version though, because then they don't build the newer versions).

aports/baloo/APKBUILD
PureTryOut: Actually I meant to move it to makedepends. I misunderstood the purpose of depends_dev, and what's in there should be in makedepends instead for this package.

aports/plasma-phone-components/APKBUILD
```diff
@@ -31,7 +30,7 @@ build() {
 }
 
 package() {
-	cd "$srcdir/build"
```
craftyguy: Same as with the previous comments I have, needed to revert this change to get this package to build

PureTryOut:
Well I did the changes required, but as I (obviously) didn't really change the specific lines @craftyguy highlighted, Github thinks I haven't changed anything and won't close the change request...

ollieparanoid:
Thanks for reviewing @craftyguy! I'll merge this, as @PureTryOut has changed what you and I requested.

@PureTryOut Yay, another upstreamed patch 👍  Thanks again for your hard work on this! I bet we have the first photo of plasma mobile running on pmOS soon! \o/

After merging, I will rebase this to master, so it has the latest bug fixes (regarding @MartijnBraam's comment about having old bugs in this branch).

PR #245, opened 2017-07-23
[WIP] Fairphone 2 support

z3ntu:
It's not tested yet because I can't flash the system.img via fastboot.
```
[20:23:30] (native) flash system image
target reported max download size of 536870912 bytes
Invalid sparse file format at header magi
erasing 'system'...
OKAY [  0.293s]
sending sparse 'system' 1/2 (524285 KB)...
OKAY [ 16.480s]
writing 'system' 1/2...
FAILED (remote: Unknown chunk type)
finished. total time: 16.782s
```
It's maybe a problem with the lk/fastboot implementation on the FP2 but I can't say for sure.

> I feel like we did look at that, but it wasn't necessary, as we only ran into it when flashing userdata with a lot of things in it

(quote from a fairphone dev)

But everything in these two packages compiles fine.


ollieparanoid:
I've merged it into a new `device-fairphone-fp2` branch for now, and it will need some clean up once you get it booting and find out what is needed (e.g. the hack in the `profile.sh` copied from `lg-mako` is probably not necessary). Thanks for making this!

z3ntu:
Yeah, I copied everything from hammerhead 😉

Issue #246, opened 2017-07-23
Multiple runs of `pmbootstrap init` should update aports location in config file


craftyguy:
Re-running `pmbootstrap init` should update the aports location in ~/.config/pmbootstrap.cfg. While this may not do anything for normal users, for developers who are working across different forks (and want to maintain them in different directories), being able to re-use the config file (preserving device, parallel compile settings, etc) would be desirable.

ollieparanoid:
Thanks for resolving this in #247!

PR #247, opened 2017-07-23
Update local aports location in 'init'

craftyguy:
This addresses #246 by updating the local aports folder location every
time `pmbootstrap init` is called. This location would change if an
existing config file is in ~/.config and pmbootstrap's location has
changed (e.g. when switching to a different fork in a different location
on disk). With this commit, you just have to run `pmbootstrap init` and
change (or accept) the existing config values and aports location is
updated automatically.


ollieparanoid:
Actually saving the aports folder in the config does not really have a benefit right now, it is more of an unwanted side-effect by how the defaults for the command line options are written.

Right now, the aports folder is always relative to the pmbootstrap folder (this may change at some point in the future, it might be cleaner to put it in an extra repository at some point).

Could you change your PR, so it does not write the aports folder into the config at all, and always uses the default (except when overridden via command line)?

Thank you for making this PR!

craftyguy:
Sure, that's even better!

craftyguy:
@ollieparanoid I'm now deleting the aports config before writing the config to disk, so that it is not saved in the cfg file

ollieparanoid:
Works, thanks!

PR #250, opened 2017-07-24
Various improvements for 'pmbootstrap export' (Fix #232)

### LINKED ISSUE #232: "pmbootstrap export" without building system image: 'Namespace' object has no attribute 'add'
Reported by Bloo on IRC.
As I see it now, it shouldn't even try to make the system partition in case it is missing (because then the --add parameter would only work the first time). Instead it makes sense to me to simply not export it, if it is missing (and maybe write a notice on how to generate it).

ollieparanoid:
* Do not try to generate the system image, but print a notice that
  it was not generated yet
* Display export path
* New default path: /tmp/postmarketOS-export (instead of current
  working dir)
* Create the export folder, if it does not exist yet
* Prettier output in the export message


ollieparanoid:
Thanks for reviewing to both of you! Adjusted the install message, as @PabloCastellano recommended.

PR #251, opened 2017-07-24
[RX51] Add fstab for N900

craftyguy:
This PR: 

1) Adds a fstab for the N900 with /boot and swap defined

2) Enables swap on boot


craftyguy:
Expanding scope of this PR per convo in Matrix about replacing fstab and adding /boot

craftyguy:
@MartijnBraam: This is ready to merge now if it passes your review. It mounts /boot and sets up & starts swap on device startup.

aports/device-nokia-rx51/device-nokia-rx51.post-install
```diff
@@ -8,4 +8,14 @@ done
 # Enable acpid
 rc-update add acpid default
 
+# Add /boot mount point in fstab
+if [ -z "$(grep /boot /etc/fstab)" ]; then
+        echo "/dev/dm-0 /boot   ext4    defaults        0       0" >> /etc/fstab
```
MartijnBraam: Isn't the boot partition ext2? It complains about a missing journal for me at least.

craftyguy:
Fixed now, when I was looking for the /boot partition filesystem I accidentally looked at the wrong partition, and did not notice the message about there not being a journal when testing it.

PR #252, opened 2017-07-24
Fix #226: Building busybox recurses forever.

### LINKED ISSUE #226: Unable to build busybox with "pmbootstrap build"
I am trying to test out a patch I am writing for busybox, and have tried to build busybox through pmbootstrap for the armhf architecture. Here are the steps I followed:

1) Download busybox aports folder from upstream Alpine

2) copy it to pmbootstrap/aports/busybox

3) run `pmbootstrap build --arch armhf`

4) Full error log is in this gist: https://gist.github.com/craftyguy/bbf1b2dbb87003e98000b79db7e00c43

Running the build with 'native' (x86_64) succeeds.

ollieparanoid:
If you really want to build busybox, I recommend turning the
"timestamp based rebuilds" feature off, otherwise it will build
for all architectures all the time whenever you change something,
even if you do not increase the version number (that's the idea
of that feature). This is, because busybox is a dependency for
basiscally everything, so it must get updated whenever you install
something, in case it was out of date.

It is easier to simply rename the package.

### DIFF
--- a/pmb/build/init.py
+++ b/pmb/build/init.py
@@ -33,7 +33,8 @@ def init(args, suffix="native"):
         return
 
     # Initialize chroot, install packages
-    pmb.chroot.apk.install(args, pmb.config.build_packages, suffix)
+    pmb.chroot.apk.install(args, pmb.config.build_packages, suffix,
+                           build=False)
 
     # Fix permissions
     pmb.chroot.root(args, ["chmod", "-R", "a+rw",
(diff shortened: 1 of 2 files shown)


craftyguy:
Thanks, I will try this out.

> It is easier to simply rename the package.

Unless I'm misunderstanding what you mean by this, it actually turned out to be harder from what I remember.. things broke pretty hard when I tried renaming the package from 'busybox', and I wouldn't be able to replace busybox in my device chroot either since it would conflict with the 'busybox' package, right?

ollieparanoid:
If you actually want to replace it, then you shouldn't rename it. But disable the "timestamp based rebuilds" unless you want it to rebuild for x86_64 all the time.

PR #253, opened 2017-07-24
Low hanging fruit basket (Close #220, Close #239)

### LINKED ISSUE #220: Support specifying multiple packages in checksum/build/aportgen commands
So you could do: `pmbootstrap checksum aports/device-*` for example.
Relevant files:
`pmb/parse/arguments.py`
`pmb/helpers/frontend.py`

### LINKED ISSUE #239: Add shortcuts for chroots to the chroot command
Everytime I need to chroot into any of the non-native chroots I have to search through the wiki to find what value to use for the `--suffix` parameter. Since most of the time you want to enter the rootfs chroot for the current device or the buildroot for the current architecture. I think this can be made quicker by adding two arguments to the chroot command:

```
--rootfs: Enter the rootfs for the current device
--buildroot: Enter the foreign architecture build chroot (not sure about this name)
```

ollieparanoid:
* postmarketos-splash: change arch from "all" to "noarch"
* lg-mako: use .tar.gz instead of .zip (because that's the reference APKBUILD)
* #220: Allow specifying multiple packages for checksum, build, aportgen
* #239: Add chroot shortcuts (--rootfs and --buildroot)


pmb/helpers/frontend.py
```diff
@@ -35,30 +35,50 @@
 import pmb.parse
 
 
+def _parse_suffix(args):
+    if (args.suffix and (args.rootfs or args.buildroot) or
```
MartijnBraam: You can define mutually exclusive argument groups with argparse instead of handeling this yourself.

ollieparanoid:
Thanks, I didn't know that `argparse` could do that. Much nicer now, as `argparse` directly prints out that it is exclusive with `--help`!

Is it ready to ship now? 🛥

pmb/helpers/frontend.py
```diff
@@ -35,30 +35,46 @@
...
 def checksum(args):
-    pmb.build.checksum(args, args.package)
+    for package in args.packages:
+        pmb.build.checksum(args, package)
 
 
 def chroot(args):
-    pmb.chroot.apk.check_min_version(args, args.suffix)
-    pmb.chroot.root(args, args.command, args.suffix, log=False)
+    suffix = _parse_suffix(args)
```
PabloCastellano: Looks good to me :+1: 

One last thing I would add is printing to stdout the suffix before entering the chroot. Something like "Entering chroot rootfs_motorola-titan"

ollieparanoid: Thanks a lot for reviewing!

Well, there is a message in `pmbootstrap log`, is that enough? Or do you also want to have a message in the terminal, where the chroot command was entered?

PabloCastellano: Yes, I mean in the terminal where you run the chroot.
I find it useful. One use case is that you might be working on multiple devices and you think you have chroot'd to the rootfs of the current device but instead you chroot'd to another one.

But as you want :)

ollieparanoid:
I've adjusted it to show the command before entering the chroot (can be silenced with `-q` if necessary: `(buildroot_armhf) % sh`).

PR #257, opened 2017-07-25
Split up initramfs to make it smaller (Close #211)

### LINKED ISSUE #211: Split up initramfs to make it smaller
The idea is to have a small "real"/"small" initramfs, and a "bigger" initramfs file on the boot partition, that gets loaded from the "small" initramfs. The big one can contain more binaries, such as `resize2fs`, `parted`, `cryptsetup` and an on screen keyboard. Here are the TODOs, how it could be done. It will be some effort though!

* generate a second initramfs file in `postmarketos-mkinitfs`, which is also compressed
* write a new `init.sh` script for the little initramfs:
** find the boot partition (with subpartition support, use kpartx like done now)
** mount it
** extract the second initramfs and pass control ([e.g. like done here, but from a file, not from a partition](https://github.com/postmarketOS/pmbootstrap/blob/e241cca904a1177f0e684b5d469339ae9b0ba283/aports/linux-samsung-i9100/init#L20-L29))
* only include the following files in the small initramfs: busybox, kpartx

From there on we could optimize it further (compile minimal busybox, statically link busybox and kpartx against musl).

drebrez:
Add generation of initramfs-extra with additional binaries
Extract both initramfs with `pmbootstrap initfs extract`
Add new splashscreens for missing partitions/files errors
Changes in init script:
 - use busybox findfs applet to find boot partition
 - mount boot partion
 - extract initramfs-extra
 - show error splashscreens accordingly and loop forever
 - start usb unlock directly from unlock_root_partition (hook removed)

(Close #211)


drebrez:
I removed the usb encryption unlock as an hook because it requires the cryptsetup, and I added the extraction of the initramfs-extra right after the hooks.

This is necessary to show the various splashscreens with the errors, because without the device-specific hook that configures the framebuffer/resolutions we are not able to display anything.

A possible solution was to add an additional hook that runs after the device-specific hook and before the usb-unlock. But since there is a variable replace during the mkinitfs creation, it wasn't nice to replace it in the hook directly.
And we also want to execute all the hooks before doing the forever_loop in case of an error.

aports/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -92,6 +93,29 @@ find_root_partition() {
...
 }
 
+find_boot_partition() {
+	findfs LABEL="pmOS_boot"
+}
+
+mount_boot_partition() {
+	partition=$(find_boot_partition)
+	if [ -z "$partition" ]; then
+		show_splash /splash-noboot.ppm.gz
```
ollieparanoid: Please also print out a text message for serial debugging at this point.

aports/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -92,6 +93,29 @@ find_root_partition() {
...
+		loop_forever
+	fi
+	mount -r -t ext2 "$partition" /boot
+}
+
+# $1: initramfs-extra path
+extract_initramfs_extra() {
+	initramfs_extra="$1"
+	if [ ! -e "$initramfs_extra" ]; then
+		show_splash /splash-noinitramfsextra.ppm.gz
```
ollieparanoid: Please also print out a text message for serial debugging at this point.

aports/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -190,7 +241,13 @@ show_splash() {
...
 echo_connect_ssh_message() {
-    echo "Your root partition has been decrypted successfully!"
-    echo "You can connect to your device using SSH in a few seconds:"
-    echo "ssh user@$IP"
+	echo "Your root partition has been decrypted successfully!"
+	echo "You can connect to your device using SSH in a few seconds:"
+	echo "ssh user@$IP"
+}
+
+loop_forever() {
```
ollieparanoid: Good idea, makes the code very readable! :+1:

aports/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -92,6 +93,29 @@ find_root_partition() {
 	echo "$DEVICE"
 }
 
+find_boot_partition() {
+	findfs LABEL="pmOS_boot"
```
ollieparanoid: `findfs` is much better than what we had before :+1:

pmb/chroot/initfs.py
```diff
@@ -46,33 +46,46 @@ def extract(args, flavor, suffix, log_message=False):
     """
```
ollieparanoid: Thanks for adjusting the `extract()` function!
* Please execute this function twice and add a parameter for differentiating between the normal intiramfs and the `-extra` version
* Could you also adjust the `pmbootstrap initfs ls` output to include the extra initramfs' content?

aports/postmarketos-mkinitfs/init_functions.sh
drebrez: We can use it also for the `pmOS_root` partition, but it doesn't work for the `crypto_LUKS` because that's not the label but the type attribute.

ollieparanoid:
Makes sense with dropping to shell. I've put this in an extra ticket #277, so we can get this PR through faster (another initramfs PR #127 depends on this one, and we should get the initramfs changes done before the aports subfolders change #227).
I've rebased the code to master.

ollieparanoid:
I'm pretty confident about this PR now, and I'd like to merge it as soon as possible (because it blocks your other fantastic initramfs PR, and both of them should be done before we do the aports subfolders PR in my opinion).

I've tested the following with this PR:
* [x] boot partition not found
* [x] booting through with FDE
* [x] booting through without FDE

What I did not test:
* system partition not found

Would you like to refactor the `extract()` function as I've requested above by yourself?
If not, I could do it and then this would be ready to ship!

PR #258, opened 2017-07-25
Enable syslog on boot

craftyguy:
This enables the busybox syslog implementation, which logs to
/var/log/messages, on boot. This is very helpful for debugging
applications which write to syslog (e.g. NetworkManager)

### DIFF
--- a/aports/postmarketos-base/APKBUILD
+++ b/aports/postmarketos-base/APKBUILD
@@ -1,6 +1,6 @@
 pkgname=postmarketos-base
 pkgver=2
-pkgrel=1
+pkgrel=2
 pkgdesc="Meta package for minimal postmarketos base"
 url="https://github.com/postmarketOS"
 arch="noarch"
--- a/aports/postmarketos-base/postmarketos-base.post-install
+++ b/aports/postmarketos-base/postmarketos-base.post-install
@@ -6,7 +6,7 @@
 for service in devfs dmesg; do
 	rc-update -q add $service sysinit
 done
-for service in hwclock modules sysctl hostname bootmisc wpa_supplicant wifi-handler; do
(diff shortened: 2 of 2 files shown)


ollieparanoid:
Thanks for making this PR.

But enabling syslog by default will **wear down the flash memory's life time**, so I would rather not have that enabled by default.

If it is needed for debugging, a developer can enable in manually with: `rc-update add syslog boot`

How about adding this command to the wiki somewhere instead? (e.g. to a NetworkManager page, when it is needed there.)

MartijnBraam:
Maybe we can figure some syslog daemon to only keep a small in-memory buffer for debugging (like android)

craftyguy:
> But enabling syslog by default will wear down the flash memory's life time, so I would rather not have that enabled by default.

I don't think this has been true for quite a while. Most modern (e.g. last 10 years) flash devices have been designed with uniform wear leveling logic, and are typically guaranteed to more read/writes than a million syslog instances running 24/7 could accomplish over some number of years. FYI, the kernel log is currently being written to disk at /var/log/dmesg :)

I would argue that syslog is an invaluable tool not just for development, but for supporting users. If users have a problem with pmos, asking for syslog output would be very beneficial in helping to determine where the problem is. Having it log to a ramdisk as @MartijnBraam suggested is possible, though if your system reboots shortly after an issue occurred then it wouldn't be any help.

craftyguy:
Ok ok, you can put your pitchforks away :)

Busybox syslogd supports logging to SHM, so this PR will configure syslog to log to SHM instead of disk. The downside to this, as I pointed out previously, is that the log will be lost 1) when the device is rebooted, and 2) if/when the syslog service is restarted.

PR #267, opened 2017-07-26
Fix #242: Validate input in pmb.helpers.cli

### LINKED ISSUE #242: init function should validate user input
It is currently possible to run `pmbootstrap init` and enter arbitrary values into the prompts. This function should validate that the user's input is a valid input for the given prompt, based on a set of acceptable values. Perhaps `pmb.helpers.cli.ask` can be extended to also include passing a set of valid input values, and in the case of an invalid value entered by user it can provide an error, reprint acceptable values, and re-prompt until the user gets it right.

ollieparanoid:
* Validate all inputs from `pmbootstrap init`
* Add a new `confirm()` function, that validates input of yes/no
  questions properly
* All questions loop until they have a valid answer now

### DIFF
--- a/pmb/helpers/cli.py
+++ b/pmb/helpers/cli.py
@@ -17,23 +17,51 @@ You should have received a copy of the GNU General Public License
 along with pmbootstrap.  If not, see <http://www.gnu.org/licenses/>.
 """
 import datetime
+import logging
+import re
 
 
-def ask(args, question="Continue?", choices=['y', 'n'], default='n',
-        lowercase_answer=True):
-    date = datetime.datetime.now().strftime("%H:%M:%S")
-    question = "[" + date + "] " + question
-    if choices:
-        question += " (" + str.join("/", choices) + ")"
-    if default:
-        question += " [" + str(default) + "]"
+def ask(args, question="Continue?", choices=["y", "n"], default="n",
+        lowercase_answer=True, validation_regex=None):
+    """
+    Ask a question on the terminal. When validation_regex is set, the user gets
+    asked until the answer matches the regex.
+    :returns: the user's answer
+    """
+    while True:
+        date = datetime.datetime.now().strftime("%H:%M:%S")
+        question_full = "[" + date + "] " + question
+        if choices:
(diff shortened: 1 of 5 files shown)


pmb/helpers/cli.py
```diff
@@ -17,23 +17,51 @@
...
+    asked until the answer matches the regex.
+    :returns: the user's answer
+    """
+    while True:
+        date = datetime.datetime.now().strftime("%H:%M:%S")
+        question_full = "[" + date + "] " + question
+        if choices:
+            question_full += " (" + str.join("/", choices) + ")"
+        if default:
+            question_full += " [" + str(default) + "]"
```
MartijnBraam: Maybe to make this consistent standard linux utilities you can make the chose list `[yn]` instead of `(y/n}`. The default is then represented with a capital `[yN]` or `[Yn]`

ollieparanoid: Good idea. But we have a problem when using `[yN]` with the *non*-yes-no-questions, where we use `[]` already as default value. For example: `User interface [weston]:`


It's hard to do research on that topic - is there a standard or something? All I could find is how `apt` handles this output, and it looks like this ([source](https://askubuntu.com/questions/292840/term-for-modified-by-you-or-by-a-script-since-installation-dialogue)):

```
Configuration file `/etc/...'
 ==> Modified (by you or by a script) since installation.
 ==> Package distributor has shipped an updated version.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : background this process to examine the situation
 The default action is to keep your current version.
*** interfaces (Y/I/N/O/D/Z) [default=N] ? N
```

...which is very similar to how pmboostrap handles it, except we don't print `default=` before the default value. But also this was not a yes/no question.


If you can find a link, that recommends how it should be done on POSIX, and has it consistent for yes-no questions as well as other questions, then I'd be happy to change this. Otherwise I suggest you merge it for now, and we change it afterwards if someone with more knowledge of how this should be handled comes along.

PR #274, opened 2017-07-27
Fix #166: pmbootstrap shutdown: umount deep folder levels first

### LINKED ISSUE #166: umount: target is busy
Randomly and suddenly, I got the following error from `pmbootstrap install`. `zap` and `shutdown` also subsequently provided the same error.
```
(005120) [15:05:39] % sudo umount /home/antonok/.local/var/pmbootstrap/chroot_native/mnt/rootfs_asus-grouper
umount: /home/antonok/.local/var/pmbootstrap/chroot_native/mnt/rootfs_asus-grouper: target is busy
        (In some cases useful info about processes that
         use the device is found by lsof(8) or fuser(1).)
```
Couldn't find anything useful in lsof or fuser, but I am also not experienced with these commands.
No clue what might have contributed to this happening, but I did not use chroot manually.

ollieparanoid:
* Refactored `umount_all` to get the list of folders to be umounted from
  `umount_all_list`, so we can test that function in a test case.
* Adjusted `umount_all_list` to return the deep folder levels first
* Added a testcase for that
* Remove redundant calls to `umount_all()` (which were from a time before
  `pmbootstrap` was released, in which failing commands did not cause
  `pmbootstrap` to raise an exception)


ollieparanoid:
Awesome \o/

PR #276, opened 2017-07-27
Fix #269: Always use truncate to create empty image file

### LINKED ISSUE #269: zfs: image installation does not work
@locusf:
https://gist.github.com/locusf/3a285ff1b68bc516c8db20e95bc28f4a

@MartijnBraam 
> that should be a fixed issue because there is a fallback with truncate implemented #28
> I guess fallocate still creates an empty file so the fallback isn't used

ollieparanoid:
I've realized that truncate is provided by busybox anyway, so no need to install coreutils for this.

This should fix creating system image files on ZFS, where fallocate is not implemented.

@locusf: Could you test if this works on your system with ZFS? Thanks!


locusf:
Me too.

PR #280, opened 2017-07-28
[RX51] Work around acpid detection of new input devices

craftyguy:
This is a dirty workaround for busybox's acpid not detecting new devices
that are added to the system after acpid has started. On the N900,
this is a problem because acpid starts earlier in the boot process than
the kernel loading modules like gpio_keys (which supplies a number of
input devices we want to handle with acpid).
Yes, this udev rule will result in acpid restarting many times during
the boot. I did say "dirty workaround"!


ollieparanoid:
Sorry for the merge conflict, that you have now. Here's the migration guide:
https://github.com/postmarketOS/pmbootstrap/wiki/Migration-to-aports-subfolders

If you need any help, just ask in the channel, as usual.

craftyguy:
Oh no, you broke my PR!!!

I've updated it to use the new aports format :smile:

craftyguy:
Good news everyone! I seem to have found a way to solve this once and for all (some more testing may be needed to confirm)! I will close this PR and roll in the solution into a new PR along with some other RX51-specific tweaks

PR #284, opened 2017-07-28
Package Plasma Mobile (All credit to PureTryOut!)

ollieparanoid:
*NOTE: Plasma mobile compiles, but we didn't get it to run yet.*

Here is @PureTryOut's amazing porting and packaging work of Plasma Mobile.
It is almost ready to be integrated into `master` now, with the aports subfolders implemented.

However, it still contains two `.tar.gz` files with the source code directly created from git commits, instead of downloading it from somewhere: for `plasma-phone-components` and `plasma-mobile`. Maybe @bhush9 could create release versions of these two components? That would be greatly appreciated!

@PureTryOut: You should have write access to the `feature/plasma-mobile` branch now, so you can rebase it, and push future fixes/improvements there (such as fixing the `.tar.gz` issue ideally).

We're getting a bit closer to Plasma Mobile on postmarketOS every day!


kskarthik:
Yay !

PR #285, opened 2017-07-29
Update to Linux Kernel 4.12.4

craftyguy:
This updates the linux-postmarketos kernel to 4.12.4, and fixes an issue
where the APKBUILD was pulling down the 4.12 source but not the 4.12.x
patches. I think that pulling down the major.minor source and a sep.
package for patches to update to the major.minor.minor(er?) release adds
unnecessary complexity since kernel.org already hosts the sources with
these patches applied (e.g. in this case the 4.12.4 source)


ollieparanoid:
> I think that pulling down the major.minor source and a sep.
package for patches to update to the major.minor.minor(er?) release adds
unnecessary complexity since kernel.org already hosts the sources with
these patches applied (e.g. in this case the 4.12.4 source)

@MartijnBraam inherited that from Alpine's `linux-*` APKBUILDs. I don't really know why they do it that way though, maybe to save disk space on developer machines, because the most of the kernel source can be re-used that way (@ncopa, do you want to explain it?). It's fine with me if we directly use the patched source.

craftyguy:
The APKBUILD we are shipping now is not the complete APKBUILD in
upstream alpine, it is missing the portion which
downloads/extracts/applies the patchset. This results in our current
APKBUILD just installing the major.minor version (e.g.  4.12).

ncopa:
> I don't really know why Alpine Linux does it that way, maybe to save disk space on developer machines, because the most of the kernel source can be re-used that way (@ncopa, do you want to explain it?)

Correct. We do it to save space and bandwidth usage. the 4.9 kernel has had 40 updates (4.9.40) the full linux source is around 90MB so the 40 releases would use 3.6GB in total. multiply that with number of build servers for the various architectures.

By adding a few lines of shell code that bandwidth and disk usage could easily be saved.

MartijnBraam:
But the travis ci builds don't keep state between the build so the base .tar.xz would be downloaded anyway

ncopa:
that was done long before travis got involved

ollieparanoid:
thanks for the explanation :)

craftyguy:
This actually makes a lot of sense. Our APKBUILD was broken (it pulled X.Y kernel source but not patches). Once the current PRs around the linux-postmarketos are merged, I can re-implement this patch mechanism in our APKBUILD

PR #286, opened 2017-07-29
Move wpa_supplicant to default runlevel

craftyguy:
Having it in boot doesn't actually work, the service is being started
too early and fails. This change moves it to the default runlevel. This
change also removes "wifi-handler" since that service doesn't actually
exist.


ollieparanoid:
I'm a bit confused about the `wpa_supplicant` service in relation to using NetworkManager - do we even need the enable the dedicated OpenRC service then? As I understand it, NetworkManager starts `wpa_supplicant` on its own.

craftyguy:
NetworkManager requires wpa_supplicant, but it does not start the
service on its own.

MartijnBraam:
NetworkManager could start wpa_supplicant if the service manager has a dbus interface to start it which openrc doesn't have. Maybe there exists a daemon that implements the dbus listener for service management?

craftyguy:
Maybe, but I don't see the problem with starting wpa_supplicant
anyways with openrc. In fact, pmos was trying to start wpa_supplicant
on startup by putting it into the boot runlevel.. it just didn't work
:)
Other mechanisms for networking (i.e. using the
openrc networking service) benefit from having wpa_supplicant running too.

ollieparanoid:
OpenRC also has service dependencies, which should be enough for this. @MartijnBraam, the NetworkManager -> wpa_supplicant use case doesn't sound relevant enough for me to add such a daemon, as it works the same when we start wpa_supplicant as dependency from NetworkManager, at least that's how I understand it. But please correct me if I'm wrong.

MartijnBraam:
Yeah that works way easier, It might be an option when more services require something like that

ollieparanoid:
Scratch that, the `post-install` file is not in the checksums. But the pkgrel increase is something we should do in my opinion.

craftyguy:
Ugh, you'd think that I would eventually learn to bump pkgrel.. but nooooo I always forget :P 

pkgrel updated!

craftyguy:
> Scratch that, the post-install file is not in the checksums.

FYI I just filed #288 to point out that this should be changed :)

PR #295, opened 2017-07-30
Allow 'pmbootstrap menuconfig' without specifying "linux-"

PabloCastellano:
We can allow both ways:
* `pmbootstrap menuconfig linux-motorola-titan`
* `pmbootstrap menuconfig motorola-titan`

The former will show a tip about the second
This also prevents users from running menuconfig on aports like
'device-motorola-titan' or 'mkbootimg'

### DIFF
--- a/pmb/build/menuconfig.py
+++ b/pmb/build/menuconfig.py
@@ -29,6 +29,13 @@ import pmb.parse
 
 
 def menuconfig(args, pkgname, arch):
+    if pkgname.startswith("linux-"):
+        pkgname_ = pkgname.split("linux-")[1]
+        logging.info("PROTIP: You can simply do 'pmbootstrap menuconfig " +
+                     pkgname_ + "'")
+    else:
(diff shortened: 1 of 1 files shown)


pmb/build/menuconfig.py
```diff
@@ -29,6 +29,13 @@
 
 
 def menuconfig(args, pkgname, arch):
+    if pkgname.startswith("linux-"):
+        pkgname_ = pkgname.split("linux-")[1]
+        logging.warn("PROTIP: You can simply do 'pmbootstrap menuconfig " +
```
ollieparanoid: For consistency, I'd use `logging.info` here (we do that usually, where we log to the currently active console, and `logging.debug` for logging to the `pmbootstrap log`).

PabloCastellano:
@ollieparanoid You have convinced me! I'll make the branches in this repo from now on :+1: and it uses `logging.info` now

PR #297, opened 2017-07-30
Add possibility to export Odin flashable tar

drebrez:
for devices where the flasher method is heimdall-isorec or heimdall-bootimg


pmb/flasher/export.py
```diff
@@ -65,13 +73,91 @@ def export(args, flavor, folder):
...
+        odin_initfs_md5 = partition_initfs + ".bin.md5"
+        odin_device_tar = args.device + ".tar"
+        odin_device_tar_md5 = args.device + ".tar.md5"
+
+        handle.write(
+            "#!/bin/sh\n"
+            "cd " + temp_folder + "\n")
+        if method == "heimdall-isorec":
+            handle.write(
+                # copy the kernel image
```
ollieparanoid: In my opinion, having a comment above every line is a bit verbose. (We also start comment lines with capital letters). Suggestion with fewer comment lines:
```python
# Kernel: copy and append md5
 "cp /boot/vmlinuz-" + flavor + " " + odin_kernel_md5 + "\n"
"md5sum -t " + odin_kernel_md5 + " >> " + odin_kernel_md5 + "\n"
# Initramfs: recompress with lzop, append md5
"gunzip -c /boot/initramfs-" + flavor + " | lzop > " + odin_initfs_md5 + "\n"
"md5sum -t " + odin_initfs_md5 + " >> " + odin_initfs_md5 + "\n")
```

pmb/flasher/export.py
```diff
@@ -65,13 +73,91 @@ def export(args, flavor, folder):
...
+                "cp /boot/vmlinuz-" + flavor + " " + odin_kernel_md5 + "\n"
+                # append md5sum to the kernel binary
+                "md5sum -t " + odin_kernel_md5 + " >> " + odin_kernel_md5 + "\n"
+                # recompress initramfs with lzop
+                "gunzip -c /boot/initramfs-" + flavor + " | lzop > " + odin_initfs_md5 + "\n"
+                # append md5sum to the initfs binary
+                "md5sum -t " + odin_initfs_md5 + " >> " + odin_initfs_md5 + "\n")
+        elif method == "heimdall-bootimg":
+            handle.write(
+                # copy the boot.img
```
ollieparanoid: Suggestion to merge the comment lines: `# boot.img: copy and append md5` (I've stated `boot.img` with a lowercase letter here, because it is the same as the file name.)

pmb/flasher/frontend.py
```diff
@@ -56,7 +56,6 @@ def kernel(args):
     pmb.chroot.initfs.build(args, flavor, "rootfs_" + args.device)
 
     # Generate the paths and run the flasher
-    pmb.flasher.init(args)
```
ollieparanoid: Why is this redundant? `pmb.flasher.init()` builds and installs the required programs (e.g. heimdall), and sets up the mount points. I think this is needed before flashing the kernel - or is it called elsewhere already?

pmb/flasher/frontend.py
```diff
@@ -86,7 +85,6 @@ def system(args):
     if not os.path.exists(args.work + "/chroot_native" + img_path):
         raise RuntimeError("The system image has not been generated yet,"
                            " please run 'pmbootstrap install' first.")
-    pmb.flasher.init(args)
```
ollieparanoid: Why is this redundant?

pmb/flasher/export.py
```diff
@@ -24,13 +24,21 @@
...
 import pmb.config
 import pmb.flasher
+import pmb.helpers.file
 
 
 def export(args, flavor, folder):
+    logging.info("Export symlinks to: " + folder)
+    if args.odin_flashable_tar:
+        odin_flashable_tar(args, flavor, folder)
+    else:
```
ollieparanoid: As a user, I would expect to *additionally* have the odin tar, when the `--odin` flag is specified. So I recommend removing the else here, and always executing the symlinks function.

pmb/flasher/frontend.py
drebrez: @ollieparanoid because `pmb.flasher.init()` is already called inside `pmb.flasher.run()` (see [here](https://github.com/postmarketOS/pmbootstrap/blob/master/pmb/flasher/run.py#L24)), which is called 2 lines below.
Also in the `system` function is the same, I removed the `pmb.flasher.init()` call because it was just called before calling `pmb.flasher.run()`

ollieparanoid: That makes sense, thanks!

pmb/flasher/export.py
drebrez: That makes sense, I also thought about it but I wasn't sure, I will do like you suggest ;)

PR #298, opened 2017-07-30
Implemented checksum test in the travis script

MartijnBraam:
This checks the checksum for all sources for all modified packages in the aports directory


test/check_checksums.py
```diff
@@ -0,0 +1,61 @@
...
+        subprocess.check_output(command)
+    except subprocess.CalledProcessError as e:
+        print("Something gone wrong in pmbootstrap. Log:")
+        logfile = os.path.expanduser("~/.local/var/pmbootstrap/log.txt")
+        with open(logfile) as log:
+            print(log.read())
+
+    result = check_output_always(['git', 'status']).decode()
+
+    if 'working tree clean' in result:
```
PabloCastellano: My git version (2.7.4) does say `working directory clean` instead. We should check both forms.

test/check_checksums.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python3
+import os
+import subprocess
+
+
+def get_changed_files():
+    try:
+        raw = subprocess.check_output(['git', 'diff', '--name-only', os.environ['TRAVIS_COMMIT_RANGE']])
+    except subprocess.CalledProcessError as e:
```
PabloCastellano: As `TRAVIS_COMMIT_RANGE` does not exist in our local environments, it will complain

I think `except (KeyError, subprocess.CalledProcessError) as e:` will suffice to handle both cases.

test/check_checksums.py
MartijnBraam: How about checking for `nothing to commit`?

PabloCastellano: It would not work in my case, since I have also `.txt` files where I take notes. In this case, it shows: `nothing added to commit but untracked files present (use "git add" to track)`

MartijnBraam: Ok didn't think of that case, replaced it with `r'working (tree|directory) clean'` now

PabloCastellano:
Oh wait, I found one solution:

```
$ git status -u no
On branch feature/allow_menuconfig_without_prefix
Your branch is up-to-date with 'origin/feature/allow_menuconfig_without_prefix'.
nothing to commit, working directory clean
```

ollieparanoid:
Great feature and great review by @PabloCastellano, I agree with you!

MartijnBraam:
What solution do you prefer? The one in 177f9ec which matches both variants with a regex or using `git status -u no`?

ollieparanoid:
I prefer not to parse the output, if it is not necessary.

MartijnBraam:
I've found what's probably the best solution. `git status -u no --porcelain` which is actually made for parsing and produces an emty result if there are no changes.

PR #304, opened 2017-07-31
init: Allow to specify extra packages that will be always installed to rootfs

PabloCastellano:


MartijnBraam:
You don't seem to allow an empty list in the init command

PabloCastellano:
Oops! Fixed

pmb/config/init.py
```diff
@@ -92,6 +92,12 @@ def init(args):
                                      default=args.timestamp_based_rebuild)
     cfg["pmbootstrap"]["timestamp_based_rebuild"] = str(answer)
 
+    # Extra packages to be installed to rootfs
+    logging.info("Extra packages that will be installed to rootfs. Specify"
```
ollieparanoid: This may sound stupid, but I'd like to start this line with different words, than the next line (e.g. "Additional packages" here and "Extra packages" in the next one, like it already is).
That way it is less irritating to read for me (and the other questions follow this pattern).
Of course I'm open to discuss this and have my opinion changed.

PabloCastellano:
@ollieparanoid That makes totally sense. I'll work on it

PabloCastellano:
Done last requested changes :+1:

ollieparanoid:
Thanks, this works great!

PR #308, opened 2017-07-31
Travis checksum pr fix

MartijnBraam:
Test for the PR builds in travis


test/check_checksums.py
```diff
@@ -56,6 +60,9 @@ def check_checksums(package):
 if __name__ == "__main__":
     if 'TRAVIS_COMMIT_RANGE' in os.environ:
         print('Checking commit range: {}'.format(os.environ['TRAVIS_COMMIT_RANGE']))
+    if 'TRAVIS_PULL_REQUEST_BRANCH' in os.environ:
+        print('Checking PR branch: {}'.format(os.environ['TRAVIS_PULL_REQUEST_BRANCH']))
```
PabloCastellano: Looks like this variable [is always present](https://travis-ci.org/postmarketOS/pmbootstrap/builds/259308421#L504). I would also check that is not the empty string

MartijnBraam: Its not present when you run it outside of travis, that's why there is a check around it. I still have the line visible to make the log explicit about the fact that the variable is empty

PabloCastellano: Ok!

PR #310, opened 2017-07-31
Fix 'install: can't stat' message, when no hook is installed

ollieparanoid:
Fix for this report by @PureTryOut in the channel:
> latest master reports this when flashing the kernel:
```
install: can't stat '/etc/postmarketos-mkinitfs/hooks/*.sh': No such file or directory
```


ollieparanoid:
Thanks for testing! Rebased and merged.

PR #316, opened 2017-08-01
Minor fixes

PabloCastellano:


pmb/config/init.py
```diff
@@ -76,9 +76,7 @@ def init(args):
     cfg["pmbootstrap"]["work"] = ask_for_work_path(args)
 
     # Parallel job count
-    default = args.jobs
-    if not default:
-        default = multiprocessing.cpu_count() + 1
+    default = getattr(args, "jobs", multiprocessing.cpu_count() + 1)
```
ollieparanoid: That looks like a much better way to write it, but it isn't quite the same (and Travis "catched" that by looping forever on `yes "" | pmbootstrap init`):
```
[14:24:08] How many jobs should run parallel on this machine, when compiling?
[14:24:08] Jobs: [14:24:08] ERROR: Input did not pass validation (regex: [1-9][0-9]*). Please try again.
... (input validation error gets printed forever)
```
`args.jobs` is always set, defaulting to an empty string, so `gettattr` does not work as expected here.
You could reproduce this locally by deleding your `~/.config/pmbootstrap.cfg`.

PabloCastellano:
Oops, sorry for not testing enough! I have pushed only the first change in documentation this time.

ollieparanoid:
I gave it a closer look, and actually we can simplify the jobs default value code by specifying it in `pmb/config/__init__.py`. I can't formerly request a review from you through GitHub, because you made the PR - but do you approve this? If so, I'd merge it.

PabloCastellano:
@ollieparanoid yes, approved! :)

PR #317, opened 2017-08-01
Close #314: Add --details-to-stdout parameter

### LINKED ISSUE #314: Add flag to print output during long builds for Travis CI
Travis will kill the job, if there's no output for 10 minutes.
For #64 we need to build packages in pmbootstrap, which will not have any output for that long.
It would be useful to have a flag, that when enabled, prints out a dot every minute or so to let Travis know, that pmbootstrap is still doing what it should.

ollieparanoid:
When this parameter is set, `args.logfd` (the log file descriptor)
gets set to `sys.stdout`, so all logging gets directly sent to the
active pmbootstrap session, and not to a log file.

This will be used for Travis jobs, because they get killed after
10 minutes without any output (and builds may take longer).


ollieparanoid:
Uh, added all of you as reviewers. One review would be enough though!

craftyguy:
Works great! I could really use this feature too since it eliminates the "open new term and look at log" step when there's a failure.

One question though.. should the new option also log to logfile in addition to stdout? there's a lot of text that blows by on the screen, and having it also preserved in a log file would be useful for debug (e.g. reporting new issues, etc)

ollieparanoid:
This PR doesn't do it, because it was harder to implement and it doesn't matter for the Travis CI use case. As a workaround, you could use `tee` for now, to both output to terminal and to write a file.

Issue #323, opened 2017-08-02
[mozilla-flame] Append `dtb` file to `boot.img` (qcdt)


ollieparanoid:
From what @yuvadm told us in the channel, the `flame` comes with its own version of `fastboot`/`mkbootimg`, which appends the `dtb` file at the end of the `boot.img` (and not at the end of the kernel, as it is with all other Android devices). The kernel is part of the `boot.img` file though.

I'm not sure what the best way would be to handle this right now (suggestions welcome!) - @yuvadm, could you point us to the source code of these mozilla specific forks?

yuvadm:
@ollieparanoid Thanks for opening an issue on this topic. Also CC'ing @z3ntu here as well since he provided some initial patches.

Essentially we need to pick a `mkbootimg` that supports the `--dt` flag which is able to append the device tree files to the `boot.img`. This is needed not only for the Flame but also for another device I believe @z3ntu is working on porting.

There are several different patched `mkbootimg` floating around, I'll paste a few that I ran into here, and then we can decide on the merits of each one:

 - https://github.com/efidroid/build/blob/master/tools/mkbootimg
 - https://github.com/LineageOS/android_system_core/blob/cm-14.1/mkbootimg/mkbootimg

Once we figure this out, we need to **conditionally** add the `--dt` flag to `mkbootimg`. An initial breaking patch looks like this:

```diff
--- a/aports/postmarketos-mkinitfs/mkinitfs.sh
+++ b/aports/postmarketos-mkinitfs/mkinitfs.sh
@@ -171,6 +171,7 @@ create_bootimg()
                --ramdisk_offset "${deviceinfo_flash_offset_ramdisk}" \
                --tags_offset "${deviceinfo_flash_offset_tags}" \
                --pagesize "${deviceinfo_flash_pagesize}" \
+               --dt "${kernelfile/initramfs-/dt.img-}" \
                -o "${outfile/initramfs-/boot.img-}"
 }

```

z3ntu:
I have a version of the last patch on my laptop which is non-breaking but I'm on my phone atm

z3ntu:
Also the method used is called qcdt and I would use the lineage source of the tools if available. @m1cha pushed the important changes from efidroid to Lineage

yuvadm:
@z3ntu good to know! Would love to get the updated version whenever you have a chance to post it. Also curious if this worked for me. I'm still running into problems with the flame `dt.img`.

[...]

PR #335, opened 2017-08-05
Initial support for the Mozilla Flame device

yuvadm:
Fixes all steps mentioned in #323 - enables building a proper bootable `boot.img`.

All commits have been rebased, squashed and organized.

Next step is to review the system image and debug some flashing issues. For now it would be great to push the work done so far upstream.

Thanks to @kskarthik for the initial work on the flame, @z3ntu for the initial mkbootimg/dtbtool patches and the rest of the devs on IRC for the constant support.


kskarthik:
Wish my flame is alive now 🙄 miss you guys

aports/device/linux-mozilla-flame/APKBUILD
```diff
@@ -0,0 +1,121 @@
...
+"
+subpackages=""
+license="GPL2"
+
+_abi_release=${pkgver}
+_carch="arm"
+HOSTCC="${CC:-gcc}"
+HOSTCC="${HOSTCC#${CROSS_COMPILE}}"
+
+ksrcdir="$srcdir/codeaurora_kernel_msm-03396c93577aafae7ca3603de1c74eafd30e7d54"
```
MartijnBraam: This should probably use ${_hash} from the variables above to prevent the build from breaking when a new version is released of the kernel

aports/main/postmarketos-mkinitfs/APKBUILD
```diff
@@ -27,4 +27,4 @@ package() {
 }
```
MartijnBraam: You forgot to increate the pkgver or pkgrel for mkinitfs.

yuvadm:
@z3ntu I stored the generated DTBs in `/usr/share/dtb` for posterity's sake, but do we need anything more than that? + @ollieparanoid

PabloCastellano:
DTB files go in /usr/share/dtb when they are generated by the mainline kernel (which is the case for n900 and qemu so far). Otherwise your kernel normally generates a zImage file without dtb and a zImage-dtb file with dtb appended and then you just need to adjust package() in your APKBUILD to pick the -dtb one.

yuvadm:
@MartijnBraam resolved both of your comments, thanks for the feedback!

yuvadm:
@PabloCastellano that makes sense, so AFAIU no more action is neccesary in this regard?

aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -195,6 +195,7 @@ create_bootimg()
 		--ramdisk_offset "${deviceinfo_flash_offset_ramdisk}" \
 		--tags_offset "${deviceinfo_flash_offset_tags}" \
 		--pagesize "${deviceinfo_flash_pagesize}" \
+		${deviceinfo_bootimg_qcdt:+ --dt /boot/dt.img} \
```
z3ntu: Maybe add $flavor to the dt.img filename?

yuvadm: I guess I can do that. Theoretically I don't really care about `dt.img`, it's an intermediate file only needed to create the `boot.img`. Stashing it in `/boot` is what makes the most sense, but I'm open to suggestions.

z3ntu: I think it's good in /boot but adding the flavor would make it more in line with the zImage and ramdisk

yuvadm:
@ollieparanoid thanks for the feedback, will make the neccessary changes. As for the device tree blob, just so I'm clear you're talking about all the `*.dtb` files which I currently stash in `/usr/share/dtb`?

ollieparanoid:
I'm merging this into `device-mozilla-flame` *first*, so I can make sure, that `postmarketos-mkinitfs` gets merged properly (GitHub says it can do it, but then we will probably end up with a broken APKBUILD again..). It will shortly be merged into master! Thanks for your hard work on this @yuvadm!

PR #336, opened 2017-08-05
Increase minimal system image size by 15% to have some free space

drebrez:
We need a better computation of the minimal partition size because right now it just computes the total required size using du (see [here](https://github.com/postmarketOS/pmbootstrap/blob/master/pmb/install/install.py#L48)) and simply adds 50MB to the root partition have some free space (see [here](https://github.com/postmarketOS/pmbootstrap/blob/master/pmb/install/install.py#L140)) and 5MB to the boot partition.

In this way we might end up with a partition with no space left because the ext4 partition has a default of 5% reserved blocks for root processes and also the filesystem metadata takes some place.

With this PR I changed the computation to add 15% of the size to the computed one, instead of the 50MB.

I did some tests with/without the change in this PR and I got these results:

with the addition of 50MB:

| img file size | boot partition size | root partition size | root filesystem (size/used/avail) |
|-|-|-|-|
| 333 | 19 | 313 | 296/249/**28** |
| 404 | 19 | 384 | 364/306/**35** |
| 598 | 19 | 578 | 553/533/**0** |

and adding a file with 500MB of random data into the device rootfs, it fails creating the system image with `cp: write error: No space left on device`

with the 15% addition:

| img file size | boot partition size | root partition size | root filesystem (size/used/avail) |
|-|-|-|-|
| 325 | 19 | 305 | 288/249/**20** |
| 407 | 19 | 387 | 367/306/**38** |
| 630 | 19 | 610 | 585/533/**8.9** |
| 900 | 19 | 880 | 851/769/**22** |

It's not the best solution since it depends a lot on the files that gets written but at least it doesn't fail with `no space left`.

Do you have other ideas/suggestion on how to better compute the minimal size and still have some free space inside?

### DIFF
--- a/pmb/install/install.py
+++ b/pmb/install/install.py
@@ -137,7 +137,7 @@ def install(args):
     for flavor in pmb.chroot.other.kernel_flavors_installed(args, suffix):
         pmb.chroot.initfs.build(args, flavor, suffix)
 
-    size_image = str(int(get_chroot_size(args)) + 50) + "M"
+    size_image = str(int(float(get_chroot_size(args)) * 1.15)) + "M"
     size_boot = str(int(get_chroot_boot_size(args)) + 5) + "M"
 
     # Finally set the user password


ollieparanoid:
Anyone got any thoughts on that? If not, I would merge it.

*At least have a quick look at those beautiful markdown tables!*

PabloCastellano:
So far 15% solves the issue but I think we should increase it in the future.

This is running pmOS in qemu without the fix:

```
localhost:~$ df -h
Filesystem                Size      Used Available Use% Mounted on
/dev/hda2               608.0M    595.2M         0 100% /
/dev/hda2               608.0M    595.2M         0 100% /
tmpfs                    99.8M    152.0K     99.6M   0% /run
dev                      10.0M         0     10.0M   0% /dev
shm                     498.7M         0    498.7M   0% /dev/shm
cgroup_root              10.0M         0     10.0M   0% /sys/fs/cgroup
```

This is with this fix applied:

```
localhost:~$ df -h
Filesystem                Size      Used Available Use% Mounted on
/dev/hda2               647.4M    593.9M      6.3M  99% /
/dev/hda2               647.4M    593.9M      6.3M  99% /
tmpfs                    99.8M    152.0K     99.6M   0% /run
dev                      10.0M         0     10.0M   0% /dev
shm                     498.7M         0    498.7M   0% /dev/shm
cgroup_root              10.0M         0     10.0M   0% /sys/fs/cgroup
```

Those extra 6.3MB make the difference!

drebrez:
@PabloCastellano since the qemu uses the img file directly, if you do `truncate -s +100MB /path/to/qemu.img` when you start it the initramfs script will expand the internal partition and filesystem automatically.
This is a simple workaround to have some more space for the qemu.

The problem I think is the `du` command used to compute the file sizes inside the chroot folder, maybe a difference between the logical size (the expected data size) and the physical size (the actual recorded data).

Issue #338, opened 2017-08-05
Split initramfs causes Sony Z2's networking to break.


opendata26:
When i connect my device to my computer in the current master nothing happens at all, the splash screen still correctly appears though.

With #257 reverted my device appears in dmesg and i'm able to connect to it with telnet.

ollieparanoid:
First I thought this could have happened, because the system partition was not reflashed. But you have said, that you did that, and your symptoms are different.

According to the code, the boot partition (which is a subpartition of the system image!) gets mounted *after* setting up the networking code, and the networking code is what fails you.

So the only change in the networking code has been in `init_functions.sh`, function `setup_usb_network_android()`:
```diff
-	echo "$(ls /sys/class/udc)" > $CONFIGFS/g1/UDC
+	ls /sys/class/udc > $CONFIGFS/g1/UDC
```
[Location in the full diff](https://github.com/postmarketOS/pmbootstrap/pull/257/files#diff-6513c915087191b94b978efc08f9240eL125)

Could you revert just that single line, and check if it works then?
Thank you very much!

opendata26:
that worked!

ollieparanoid:
Sorry, I did not manage to make a PR for [this diff](https://github.com/postmarketOS/pmbootstrap/issues/338#issuecomment-320502311) today (for which we know it caused a regression). I'll do it soon, but you're welcome to make a PR your self, so it gets fixed faster.

Thanks for your patience in helping to debug and fix this :)

PR #339, opened 2017-08-05
Add Xperia Z2 Tablet WiFi

opendata26:
Needs #257 reverted for the adapter to show up.
Has working ssh, virtual terminal, drm, framebuffer and  #weston.

See #338


aports/device/device-sony-castor_windy/initfs-hook.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+# set framebuffer resolution
```
ollieparanoid: Is this empty hook intentional? If you don't need it, please remove it (remove this file, remove it from the APKBUILD's sources and remove the install line in the `package()` function).

opendata26: Nope, will remove it

ollieparanoid:
Could you rebase to master and check if all network issues are gone?
...and sorry that I just see this now, please rename the aports to use `-` instead of `_` as per our [package naming convention](https://github.com/postmarketOS/pmbootstrap/wiki/Package-naming-convention).
I hope we can merge this soon, thanks for all your hard work on this!

MartijnBraam:
Looks good to merge now, Thanks

PR #346, opened 2017-08-07
Adding device semc-anzu

rendeko:
Testing device had permanent locked bootloader, so flashing could not be tested. However, everything built successfully. Further testing on a device with an unlocked bootloader will be needed to know what works and how to continue.


aports/device/device-semc-anzu/deviceinfo
```diff
@@ -0,0 +1,33 @@
...
+deviceinfo_name="Xperia Arc"
+deviceinfo_manufacturer="Sony Ericsson"
+deviceinfo_date=""
+deviceinfo_keyboard="false"
+deviceinfo_nonfree="????"
+deviceinfo_dtb=""
+deviceinfo_modules_initfs=""
+deviceinfo_external_disk="true"
+deviceinfo_external_disk_install="true"
+deviceinfo_flash_methods="heimdall-isorec"
```
ollieparanoid: The flasher is probably wrong (`heimdall` is only for Samsung devices), I guess it is `fastboot`. See also:
https://github.com/postmarketOS/pmbootstrap/wiki/deviceinfo_flash_methods

aports/device/linux-semc-anzu/APKBUILD
```diff
@@ -0,0 +1,99 @@
+# APKBUILD based on linux-lg-mako.
+# Kernel config based on: arch/arm/configs/semc_anzu_defconfig
+
+# Unsure if pkgrel is correct...
```
ollieparanoid: The `pkgrel` is fine, so you can remove this comment. Simply increase it whenever modifying the package (and *not* upgrading the `pkgver`).

aports/device/device-semc-anzu/deviceinfo
ollieparanoid: Scratch that, from what you told us in the channel, the correct flasher would be ["flashtool"](https://github.com/Androxyde/Flashtool) which we have not even packaged yet. Exciting! I hope you get it working :>

ollieparanoid:
I'll merge this PR and change the flash method afterwards myself (it doesn't take long and then we have one less open PR).
Thanks for your work on this port! I hope we can see the kernel actually booting at some point (*context: currently the bootloader is locked and only allows kernels signed by the vendor*).

mirh:
Just for the records of you gentlemen, I wanted to let you know that all the 2011 xperia (manufactured before a not-so-clear 2012 week), are capable of \*unofficial\* bootloader unlock. @rendeko 
https://forum.xda-developers.com/showthread.php?t=1285285
https://forum.xda-developers.com/showthread.php?t=1551494
https://forum.xda-developers.com/showthread.php?t=1730874
https://forum.xda-developers.com/showthread.php?t=2768514
https://forum.xda-developers.com/showthread.php?t=2070325

Also worth noting that there's possibly also a 3.10 branch [in the wild](https://github.com/LegacyHuawei/local_manifests/issues/9#issuecomment-168169631)

ollieparanoid:
@BeskosTakis: We have well documented [installing](https://wiki.postmarketos.org/wiki/Installation_guide) and [porting](https://wiki.postmarketos.org/wiki/Porting_to_a_new_device) instructions (just have a quick look at them) and the sd card installation is really easy (just pass `--sdcard` to the install command). It would be awesome if you could give it a try, and we'll help you if you get stuck.

@mirh: Thanks for the links! I've added them [to the device's wiki page](https://wiki.postmarketos.org/wiki/Sony_Ericsson_Xperia_Arc_(semc-anzu)).

BeskosTakis:
@mirh have you managed to boot either android either postmarketOS using 3.10 kernel ? i've tried multiple times to boot android using the 3.10 kernel and i was never successfull.

@ollieparanoid Thank you for the information, i've managed to build it successfully 5 days ago, but i didn't have a card reader , so i couldn't flash the os. I've tried to connect the phone via usb  and i also tried to insert the sdcard to an external camera, but i got an error saying that /dev/sdd1 (this is where the sd was mounted) is not an sdcard. I will try soon with a usb card reader and i hope that it won't happen again, as i am stuck there

mirh:
I was reportedly told @Blefish had got (even?) 4.4 to actually boot on his huawei (I think U8800). 
Then maybe for as much as the kernel would be supposed to be unified with legacyxperia, maybe a little adjustment or two are still needed.

BeskosTakis:
I would be really interested if there is a bootable 3.10 or newer kernel for anzu. It is my old device and i think it could be great for experimenting hobbists like me :) 

maybe @blefish and @mikeNG can share some info about msm7x30 kernels

Blefish:
It's true that 4.4 kernel boots. It's completely rewritten, so it uses upstream drivers + stuff that I have forward ported to new interfaces and device tree. Unfortunately it's still proof-of-concept as some drivers require more work and lately I have been busy with other stuff. The reason why I stopped was MTD NAND driver. It kind of worked, but it went nuts when I performed a firmware flash and I ended up with bad blocks on NAND. I think they were not marked properly.

The old MTD driver is dependant on old DMA driver, so it's quite a headache.

As for 3.10, we forward ported it based on 3.4 and it worked quite well, including camera.

mirh:
Mhh, I'm delighted to hear that (and I'd be almost tempted to ping Rob Clark, should he *still* be interested on getting aligned A2xx to mainline) - but where would this support be in your msm7x30 4.4 branch?
I see nothing different from normal. 

@thewisenerd I know you love MTD - add that to your checklist

ollieparanoid:
@BeskosTakis: glad that you've tried it out! Just in case you can't get a SD card reader any time soon, the recovery zip install method should also work in theory, as long as you have something like TWRP installed on your device.

```shell
$ pmbootstrap install --no-fde --android-recovery-zip
$ pmbootstrap flasher sideload
```

PR #348, opened 2017-08-08
Fix refresh rate on sony-amami

ata2001:
See https://github.com/postmarketOS/pmbootstrap/wiki/sony-amami-(Xperia-Z1-Compact)#issues

Makes weston fall back to default refresh rate (60 Hz).

I know this is an awful hack, but I don't think this kernel/driver worth the time debugging, since the panel driver could easily be ported to mainline based on robclark's work.


MartijnBraam:
Hmm interesting, the mdss video driver for the hammerhead has the same line in it (and also has refresh rate issues)

ollieparanoid:
I think it is highly possible, that this hack will only work with weston.
Because weston [falls back to 60 Hz](https://patchwork.freedesktop.org/patch/150943/) when the driver says, that it has less than 1 Hz. I do not think, that other wayland compositors will act the same way.

How about directly forcing 60 Hz instead? According to [framebuffer.txt](https://www.mjmwired.net/kernel/Documentation/fb/framebuffer.txt#265), the value would be specified in pico seconds. So as I understand it, the correct value would be `1000000 / 60` (which calculates to `16667`, but I would leave the `/ 60` in place, so it is easier to understand). Could you try that?

It is super useful, that you found the place in the code where we actually can fix/workaround this
issue, thanks a lot!

Regarding issue #349: Travis should run through as soon as you rebase your code on `master`. Sorry for that!

ata2001:
@ollieparanoid Thanks for doing the math! It works as expected.

PR #351, opened 2017-08-08
Added parser for bash variables

MartijnBraam:
This adds a parser for bash global variables. It supports some string interpolation methods:

- FOO=BAR
- FOO=$BAR
- FOO=${bar}
- FOO="Awesome${bar}test"
- FOO="${${NAME}}"
- FOO="${HOSTCC#${CROSS_COMPILE}}"
- FOO="${FOO:-default}"
- FOO="${FOO=default}"
- FOO="${FOO:=default}"

It doesnt support:

- multiline variables
- FOO="${$NAME}"
- variables defined in case blocks


MartijnBraam:
I haven't added it to any code in pmbootstrap yet because it still fails a few testcases, most importantly it cannot parse the switch..case blocks used for the kernel version in the apkbuild.

I'll add some more documentation to the code and rename the module/class

ollieparanoid:
Okay, cool! Yeah, I suggest we wait with the merge until it parses at least as good as what we have now, and passes all testcases.

MartijnBraam:
It looks like this parser already supports more than the current apkbuild parser (which doesn't try to read the kernelver case blocks)

ollieparanoid:
That's great! You could simple remove the `dont_merge_pr` label when it is ready.
(I've noticed that your new files did not include a license header yet, please also add it and set the copyright to yourself.)

test/test_shellparser.py
```diff
@@ -0,0 +1,129 @@
...
+def test_simple(shell_script, expected):
+    parsed = ShellParser(shell_script)
+    for key, value in expected.items():
+        assert parsed.variables[key] == value
+
+
+def test_apkbuild():
+    fixture = "../aports/main/hello-world/APKBUILD"
+
+    env = {
```
ollieparanoid: I think this leads to problems. The thing is, that we don't know the CARCH, CBUILD_ARCH and CROSS_COMPILE variable unless we have parsed the `APKBUILD` at least once. Then we decide what to pass in these variables at build time (see `pmb/build/autodetect.py`).
So the initial APKBUILD parsing logic must work without specifying those (maybe it makes sense to always default to an empty string? I have not put that much thought into it yet).

pmb/parse/apkbuild.py
```diff
@@ -19,53 +19,7 @@
...
-    if ret["pkgname"].startswith("py-"):
-        replacement = ret["pkgname"][3:]
-        for var in ["depends", "makedepends", "subpackages"]:
-            for i in range(len(ret[var])):
-                ret[var][i] = ret[var][i].replace(
-                    "${pkgname#py-}", replacement)
-
-    ret["makedepends"] = replaced
-    return ret
+from pmb.parse.shell import ShellParser
```
MartijnBraam: I can fix the import line but I don't agree with using snake case for class names, that rule is in pep8 for a reason, classes don't behave the same as modules and I'm shocked that flake8 doesn't throw a tantrum about it.

pmb/parse/shell.py
```diff
@@ -0,0 +1,197 @@
...
+
+You should have received a copy of the GNU General Public License
+along with pmbootstrap.  If not, see <http://www.gnu.org/licenses/>.
+"""
+
+import re
+
+
+class ShellParser:
+    """Parser for global variables in shell script files"""
```
MartijnBraam: The single line doctring is also reST style, classes just don't have more information by default and this is the way it should be formatted for single line docstrings ([PEP-257](https://www.python.org/dev/peps/pep-0257/))

test/test_shellparser.py
MartijnBraam: I've put in these defaults because those values don't actually matter for the current APKBUILD parser implementation. I've thought about adding a flag to the parser to not throw exceptions on undefined variables but just use blank values for them like bash would do without additional settings.

pmb/parse/apkbuild.py
ollieparanoid: Thanks for telling me about snake case classes being invalid in PEP8, of course we should not do it then.
We could use [pep8-naming](https://pypi.python.org/pypi/pep8-naming) to check for it in flake8, it seems (do you want to make a PR about that after this one is through?)

pmb/parse/shell.py
ollieparanoid: Learned something again, thanks for clarifying.

ollieparanoid: (It would be nice if we also had a flake8 plugin for that.)

ollieparanoid:
*Just labeled, because it became clear in the chat that it isn't easy to see which PRs are ready to merge and which are not. Please remove the label when you think it's ready! Also we might continue it if you can't find any time for it.*

ollieparanoid:
Thanks for the update! I think the proper way to test this PR is to `pmbootstrap zap -p` and `pmbootstrap install` everything from scratch. At least that must work. So I did that, and got the following: `ERROR: 'Could not find variable pkgver%.*'`

There are lots of PKGBUILDs we have, that use "pkgver%" (just grep for it), 
weston also uses `${_pkgver%%.*}`.


```
(008101) [20:27:40] Add user
(008101) [20:27:40] (native) % adduser -D user -u 12345
(008101) [20:27:40] (native) % chown -R user:user /home/user
(008101) [20:27:44] Calculate depends of packages ['cryptsetup', 'util-linux', 
'e2fsprogs', 'parted'], arch: x86_64
(008101) [20:27:44] ERROR: 'Could not find variable pkgver%.*'
(008101) [20:27:44] Run 'pmbootstrap log' for details.
(008101) [20:27:44] See also: <https://postmarketos.org/troubleshooting>
(008101) [20:27:44] Traceback (most recent call last):
  File "/home/user/code/pmbootstrap/pmb/__init__.py", line 53, in main
  File "/home/user/code/pmbootstrap/pmb/helpers/frontend.py", line 136, in 
install
    pmb.install.install(args)
  File "/home/user/code/pmbootstrap/pmb/install/install.py", line 246, in 
install
    build=False)
  File "/home/user/code/pmbootstrap/pmb/chroot/apk.py", line 197, in install
    strict=True)
  File "/home/user/code/pmbootstrap/pmb/parse/depends.py", line 68, in recurse
    aport = pmb.build.find_aport(args, pkgname_depend, False)
  File "/home/user/code/pmbootstrap/pmb/build/other.py", line 56, in find_aport
    apkbuild = pmb.parse.apkbuild(args, path_current)
  File "/home/user/code/pmbootstrap/pmb/parse/apkbuild.py", line 61, in apkbuild
    parsed = ShellParser(handle, environment=env)
  File "/home/user/code/pmbootstrap/pmb/parse/shell.py", line 52, in __init__
    self._parse()
  File "/home/user/code/pmbootstrap/pmb/parse/shell.py", line 61, in _parse
    result = self._process_substitutions(value)
  File "/home/user/code/pmbootstrap/pmb/parse/shell.py", line 128, in 
_process_substitutions
    label_value = self._process_substitution_value(label)
  File "/home/user/code/pmbootstrap/pmb/parse/shell.py", line 214, in 
_process_substitution_value
    raise KeyError("Could not find variable {}".format(label))
KeyError: 'Could not find variable pkgver%.*'
```

(Please remove the `dont_merge_pr` label when you think it is ready.)

MartijnBraam:
I've ran a full zap-and-install cycle and it seems to work fine now. I've also added a testcase that throws all the current APKBUILD files at the parser and sees it it raises an exception.

ollieparanoid:
Excellent! Now there's that `from pmb.parse.shell import ShellParser` which you commented would be OK for you if changed to what we use elsewhere in the code (only `import pmb.parse.shell` instead).
And I feel like I should do one more code review to actually understand what this piece of code does before I merge it. I'll try to do it rather sooner than later. Thank you for all your work on this, this will make `pmbootstrap`'s APKBUILD parsing robust like a mountain.

On a side note, have you noted any speed differences with parsing?

MartijnBraam:
I haven't done any benchmark but I guess this code is slower because it implements a whole stateful parser instead of the C regex engine

ollieparanoid:
Thanks for making the changes! I'll try to get this reviewed and merged in the some time in the next days (additional reviews welcome!).

(Oh and [that](https://github.com/postmarketOS/pmbootstrap/pull/351#discussion_r132819618) could still be improved.)

ollieparanoid:
TODOs:
  - [ ] Instead of `from pmb.parse.shell import ShellParser` and `ShellParser(...)` use `import pmb.parse` and `pmb.parse.shell(...)` ([see here](https://github.com/postmarketOS/pmbootstrap/pull/351#discussion_r132819618))
  - [ ] Fix these regressions: [regression_test.txt](https://github.com/postmarketOS/pmbootstrap/files/1300238/regression_test.txt)

Script, that generated `regression_test.txt`:
```shell
#!/bin/sh -e
# NOTE: you need to add sort_keys=True to json.dumps() in
# pmb/helpers/frontend.py:parse_apkbuild()!

NEW=/home/user/code/pmbootstrap
OLD=/home/user/code/pmbootstrap_backup
TMP=/tmp/apkbuild_parse_regression_test

# Copy pmbootstrap code
[ -d "$TMP" ] && rm -rf "$TMP"
mkdir -p "$TMP"
cp -r "$OLD" "$TMP/pmbootstrap_old"
cp -r "$NEW" "$TMP/pmbootstrap_new"

# Make sure we have the same aports (no pkgrel increase etc)
rm -r "$TMP/pmbootstrap_new/aports"
ln -s "$TMP/pmbootstrap_old/aports" "$TMP/pmbootstrap_new/aports"

# Compare parsing of each APKBUILD
cd "$TMP/pmbootstrap_old/aports"
for i in *; do
	cd "$TMP/pmbootstrap_old/aports/$i"
	for j in *; do
		echo "$i/$j"
		for k in old new; do
			"$TMP/pmbootstrap_$k/pmbootstrap.py" -q parse_apkbuild "$j" \
				> "$TMP/$k"
		done

		diff --label "old" --label "new" -u "$TMP/old" "$TMP/new" || true
	done
done
```

PureTryOut:
Will this be finished some day? The last activity for this PR being a comment on the 13th of September, it seems quite abandoned...

ollieparanoid:
Feel free to re-open when working on this again.

PR #353, opened 2017-08-09
Configfs network setup was broken (related to #338)

### LINKED ISSUE #338: Split initramfs causes Sony Z2's networking to break.

ollieparanoid:
@opendata26: Please test this branch, does it partially fix #338?
@PureTryOut: Does this fix your network issue, too?


PureTryOut:
Nope 😞 I'm now going back in history to see where it broke exactly.

ollieparanoid:
@opendata26 confirmed in the channel, that it partially resolves #338, so I'm merging it.
@PureTryOut: thanks for testing - we'll fix the issue you're having in another PR then!

PR #359, opened 2017-08-10
Close #326: Implement command to retrieve and set configuration values

### LINKED ISSUE #326: "pmbootstrap config" to retrieve and set configuration values
I would like to see a `pmbootstrap config` command (as in `git config`). This way we could change pmbootstrap configuration values with simple commands:

```bash
$ pmboostrap config ui
weston
$ pmboostrap config ui none
$ pmboostrap config ui
none
```

PabloCastellano:


PabloCastellano:
I've disabled the logging, otherwise we would get outputs like:

```
$ pmb config ui
weston
[23:28:33] Done
```

PabloCastellano:
Implemented:

- check variable name is valid
- show full configuration file when you don't specify 'name'

ollieparanoid:
Tested and working, thank you!

PR #364, opened 2017-08-11
Fix #342: don't use distutils.version.LooseVersion anymore

### LINKED ISSUE #342: Version checking: don't use distutils.version.LooseVersion anymore
Right now we use `distutils.version.LooseVersion` from python, because it behaves like the version comparison from abuild (which is the same as in Gentoo, because once upon a time Alpine was built from Gentoo).

The problem is, that this version check throws exceptions in some corner cases:
```python
>>> from distutils.version import LooseVersion
>>> LooseVersion("22.7-r1") < LooseVersion("22.7.3-r1")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/distutils/version.py", line 52, in __lt__
    c = self._cmp(other)
  File "/usr/lib/python3.6/distutils/version.py", line 337, in _cmp
    if self.version < other.version:
TypeError: '<' not supported between instances of 'str' and 'int'
>>>
```

...so it is not really suitable for what we are doing. To be fair, this isn't a public python API, but some internal code we shouldn't be using anyway.

* [Python Bugreport](https://bugs.python.org/issue30272)

There is already a testcase for version checking, that actually uses `abuild`'s testcase. Right now we ignore all cases, where we get an exception thrown.

- [ ] Create `pmb/helpers/version.py`
- [ ] Move the [`compare_version`](https://github.com/postmarketOS/pmbootstrap/blob/5eb9b92e7b02e2aa6017dba4e641a9fe9193af6e/pmb/parse/apkindex.py#L27) function there
- [ ] Rewrite all calls to that function to use the new path
- [ ] Rewrite the function to not use `distutils`
- [ ] Test all cases, do not allow exceptions to pass the case

ollieparanoid:
Previously, distutils.version.LooseVersion was used, because it was
sort of close enough to how Alpine parses versions.

This new version uses the exact same algorithm, as `apk` does, and
it passes *all* of `apk`'s testcases for version checking (previously
we simply skipped the ones, that did not pass).

**One proper review + test is enough.**


test/test_version.py
```diff
@@ -40,30 +40,39 @@ def args(request):
...
-                                # TypeError exception.
-                continue
-            assert(real == expected)
+    print("---")
+    print("total: " + str(count))
+    print("errors: " + str(len(errors)))
+    print("---")
+    for error in errors:
+        print(error)
+    assert(errors == [])
```
MartijnBraam: Assert is a statement not a function so it doesn't need the parenthesis, also it is inconsistent with the assert in line 70 with a space between the assert and the parenthesis.

PR #366, opened 2017-08-12
Make gcc-aarch64 reproducible

ollieparanoid:
This will fix https://github.com/postmarketOS/binary-package-repo/issues/1.

GCC generates hardlinks between files `A` and `B` in its `make install` step. The problem is, that `tar` randomly packages `A` as full binary, and links `B` to `A`, or the other way around! I was able to reproduce this issue consistently when re-building `gcc-aarch64` on Travis CI (interestingly, this did not appear for `gcc-armhf`).

The fix is, to delete `B` and create a symlink `B` that points to `A` instead.

### How to test this PR
```shell
pmbootstrap zap -p
pmbootstrap build hello-world --arch=armhf
pmbootstrap build hello-world --arch=aarch64
```

### DIFF
--- a/aports/cross/gcc-aarch64/APKBUILD
+++ b/aports/cross/gcc-aarch64/APKBUILD
@@ -12,16 +12,39 @@ LANG_FORTRAN=false
 LANG_ADA=false
 options="!strip !tracedeps"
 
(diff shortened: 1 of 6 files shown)


ollieparanoid:
*Restoring and keeping the branch, because the Travis CI job uses it right now.*

PR #367, opened 2017-08-12
Refactor README

yuvadm:
Kept pretty much the same content from the README but slightly more readable, and added usage as a basic walkthrough of how to build an image for a supported device (which is probably most users will want to do to start hacking on pmOS)


README.md
```diff
@@ -1,56 +1,79 @@
...
+
 Sophisticated chroot/build/flash tool to develop and install postmarketOS.
 
-[Static code analysis](https://github.com/postmarketOS/pmbootstrap/blob/master/test/static_code_analysis.sh) status: [![travis badge](https://api.travis-ci.org/postmarketOS/pmbootstrap.png?branch=master)](https://travis-ci.org/postmarketOS/pmbootstrap)
+For in-depth information please refer to the [postmarketOS wiki](https://github.com/postmarketOS/pmbootstrap/wiki).
 
 ## Requirements
-* Linux distribution
-* Python 3
+* Linux (`x86_64` only)
```
ollieparanoid: `aarch64` is also supported, and I would like to see "Linux distribution" here, because Linux is just the kernel, and you need a few userspace programs for pmbootstrap, such as `chroot`.

README.md
```diff
@@ -1,56 +1,79 @@
 # pmbootstrap
```
ollieparanoid: I like how you've improved the first section, it looks a lot better! :+1:

README.md
```diff
@@ -1,56 +1,79 @@
...
-* Python 3
+* Linux (`x86_64` only)
+* Python 3.4+
 * OpenSSL
 
-## Important links
-* [postmarketOS introduction](https://ollieparanoid.github.io/post/postmarketOS)
-* [Security warning](https://ollieparanoid.github.io/post/security-warning/)
-* [Porting progress](https://github.com/postmarketOS/pmbootstrap/wiki/Devices)
+## Usage
```
ollieparanoid: This feels like a mix of the porting guide and the (unfinished) [installation guide](https://github.com/postmarketOS/pmbootstrap/wiki/Installation-Guide) in the wiki. I'd rather have people pointed to the wiki, because it is easier to edit the instructions there, so I think that it will lead to more actual edits.

How about putting few common instructions in here (check out the repo, run pmbootstrap --help for usage information) and then placing two big links to the porting guide / installation guide?

It might make more sense to move obscure commands from the examples we have right now into the new [Tips and tricks](https://github.com/postmarketOS/pmbootstrap/wiki/Tips-and-tricks) page.

yuvadm: @ollieparanoid Indeed, tha idea is for the README to function as it is intended - standalone :)

I personally prefer the README to have everything I need to start hacking on a repo without having to visit other links, but I'll accept any other decision made by the core team. So if we want to keep the wiki as the main documentation, we'll just have pointers to it.

README.md
yuvadm: Fair enough, fixed

README.md
ollieparanoid: > but I'll accept any other decision made by the core team.

Sorry for the lengthy response on these few words, but this is important to me: [There is no core team.](https://github.com/ollieparanoid/ollieparanoid.github.io/commit/047c3a3dde9a651ad68a5beef4470107f12ab786) We have a community where every contributor and helping hand is equally important. The people in the GitHub team are there for practical reasons (such as directly collaborating on the homepage in your case, or getting selected as PR reviewer, which GitHub does not allow otherwise), but not because they have a higher status in the project.

> I personally prefer the README to have everything I need to start hacking on a repo without having to visit other links

That is super useful and I love this in other projects. But for `pmbootstrap` I think the whole topic is too complex and you *need* to visit the wiki to make sense of it. Even if you just explain how to do the installation, there are steps very different depending on your device (e.g. the N900 needs uboot to be installed and a sdcard installation, where for others you need to flash an image file). So as I see it there is really no way around clicking other links - but I'm all with you in making the information as accessible as possible.

Let's consider that we do want to add a full guide in the `README.md` - where do we draw the line? You have written *you can build and flash a postmarketOS image by running through the following steps. For new devices check the porting guide* at the beginning. So this means, we would only add the information for the installation to an already ported device - but why would a first time user want to modify the kernel if there is an already known-to-work configuration available? I think that shouldn't be in there, if we did put the installation guide in the `README.md`.

I'd love to hear your opinion on that, and if you think the guide format is the best way, please do an improved version, that replaces the installation guide. I might as well be completely wrong with this and the way you would present the information really made it a lot easier to jump right into development.

Thanks for taking the time and trying to improve the `README.md` file. After all, this is the first thing a new user sees in the repository 👍

yuvadm: @ollieparanoid I apreciate the detailed reply, as well as the sense of equality within the community :) I think it's super important for the project in the long run and I'm glad that the other contributors feel the same way.

As for the README vs. wiki, I very much understand the need to dig deeper to actually find pmbootstrap useful for a specific device. I'm just trying to balance it with getting new developers more involved, and indeed that might be a target that is very hard to reach at this point. So I'm definitely not dogmatic about it.

What I tried to do in the README is run through a typical use case so that devs understand what this repository does. It's there to provide **context**, I don't expect anyone to actually run through those commands to flash a device.

Alternatively, if we just decribe the usage as it is today, just explain it a bit better, I think it can serve the same purpose. So I'm very much open to any decision here. Mostly the formatting changes are important at this point :)

drebrez: @ollieparanoid @yuvadm what you think about writing all the steps/commands of an existing ported device from the `git clone` to the final `ssh` command?
There will be something like 4-5 commands and a random user might be interested to check if his device is already ported and try it on his phone.
Personally when I find a readme page with a lot of commands, the first thought is, it seems very complicated and complex. 
Instead with few commands, which in this case is just so, it feels more attractive.

drebrez: - git clone
- pmbootstrap init
- pmbootstrap install
- pmbootstrap flash_kernel
- pmbootstrap flash_system
- telnet to unlock partition
- ssh into device

yuvadm: @drebrez 👍 that sounds like what I was going for. @ollieparanoid does that make sense to you?

ollieparanoid: That sounds like a nice solution! :+1:

yuvadm:
Updated and rebased, per your comments :)

README.md
```diff
@@ -1,56 +1,74 @@
...
+
 Sophisticated chroot/build/flash tool to develop and install postmarketOS.
 
-[Static code analysis](https://github.com/postmarketOS/pmbootstrap/blob/master/test/static_code_analysis.sh) status: [![travis badge](https://api.travis-ci.org/postmarketOS/pmbootstrap.png?branch=master)](https://travis-ci.org/postmarketOS/pmbootstrap)
+For in-depth information please refer to the [postmarketOS wiki](https://github.com/postmarketOS/pmbootstrap/wiki).
 
 ## Requirements
-* Linux distribution
-* Python 3
+* Linux distribution (`x86_64` or `aarch64` only)
```
ollieparanoid: Sorry for picking on this line again, but could we remove the `only`? Two architectures is not that bad, I think Android only supports `x86_64` as host arch.

README.md
```diff
@@ -1,56 +1,74 @@
...
 
+Assuming you have a supported device, you can build and flash a postmarketOS image by running through the following steps. For new devices check the [porting guide](https://github.com/postmarketOS/pmbootstrap/wiki/Porting-to-a-new-device).
 
-## Usage
-**Check out the [porting guide](https://github.com/postmarketOS/pmbootstrap/wiki/Porting-to-a-new-device) for a practical start!**
+First, clone the git repository and initialize your pmbootstrap environment:
+
+```bash
+$ git clone https://github.com/postmarketOS/pmbootstrap
+$ cd bootstrap
```
ollieparanoid: That should be `cd pmbootstrap`.

README.md
```diff
@@ -1,56 +1,74 @@
...
-* [Porting progress](https://github.com/postmarketOS/pmbootstrap/wiki/Devices)
+## Usage
 
+Assuming you have a supported device, you can build and flash a postmarketOS image by running through the following steps. For new devices check the [porting guide](https://github.com/postmarketOS/pmbootstrap/wiki/Porting-to-a-new-device).
 
-## Usage
-**Check out the [porting guide](https://github.com/postmarketOS/pmbootstrap/wiki/Porting-to-a-new-device) for a practical start!**
+First, clone the git repository and initialize your pmbootstrap environment:
+
+```bash
```
ollieparanoid: Would you mind changing the `bash` to `shell` everywhere?

ollieparanoid:
I like it! There are very minor things, I've commented, and except for them I think it's perfect.
Is everyone else also happy with it? 😄

yuvadm:
@ollieparanoid done! happy to work on it :)

MartijnBraam:
:+1:

ollieparanoid:
Soo.... the issue was open for four days now (long time in the fast moving world of pmOS :p), I'm merging it. If someone wants to see something changed, just open an issue/write in the chat/make a new PR 👍

PR #379, opened 2017-08-14
Add setting for device keymap for devices with keymaps

MartijnBraam:
This makes `pmbootstrap init` ask which keymap you want if the deviceinfo contains a `deviceinfo_keymaps` setting.


pmb/config/init.py
```diff
@@ -58,6 +59,21 @@ def ask_for_ui(args):
                       " one from the list above.")
 
 
+def ask_for_keymaps(args, device):
+    info = pmb.parse.deviceinfo(args, device=device)
+    if 'keymaps' not in info:
+        return
+    options = info['keymaps'].split(' ')
```
ollieparanoid: For consistency with the other code, I'd prefer double quotes `""`.

pmb/helpers/keymap.py
```diff
@@ -0,0 +1,29 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with pmbootstrap.  If not, see <http://www.gnu.org/licenses/>.
+"""
+import pmb.chroot.root
+
+
+def setup_keymap(args):
```
ollieparanoid: If there are no plans to add more functions here, I would put that 5 line function in `pmb/helpers/other.py`.

pmb/config/init.py
```diff
@@ -58,6 +59,21 @@ def ask_for_ui(args):
...
+    options += [""]
```
ollieparanoid: As discussed in the channel, it would be nice if it defaults to the first layout of the device and does not present an empty option here, so the user will always have a good layout.

pmb/helpers/keymap.py
ollieparanoid: ...on second thought, in `pmb/install/install.py`, because it is installation specific.

ollieparanoid:
Travis is not quite happy yet.

craftyguy:
Don't forget to update the N900 wiki page 'keymap' section once this is merged

ollieparanoid:
As discussed in the chat, this PR will crash in `pmbootstrap init` right now, when specifying a new device (although it will crash with a nice message explaining which files need to be created).

@MartijnBraam said in the chat when asked about how to handle it:
> Skip the keymaps stuff for news devices, but we should probably validate the keymaps in the install step
> If you pull in keymap changes then it might break now

ollieparanoid:
*Just labeled, because it became clear in the chat that it isn't easy to see which PRs are ready to merge and which are not. Please remove the label when you think it's ready! Also we/I could finish it, if you can't find time for it.*

craftyguy:
@ollieparanoid 

I think this is ready to go, I've implemented:

1) the validation during init when the device is new

2) validation during install that keymap actually exists for the device

Note that this now depends on PR #418

ollieparanoid:
I did two minor changes, and I'll merge it as soon as Travis looked over it.
* fix setting default keymap to the first keymap of the deviceinfo,
  in case no keymap has been selected before (we need to check for
  "" instead for None)
* show the available keymap count, just like all other commands show
  the available counts (interfaces, devices)

PR #380, opened 2017-08-14
Update plasma-framework to 5.37.0 and package dbus-launch without x11

### LINKED ISSUE #377: Package dbus-launch without X11 support
The `dbus-launch` we currently use is from the dbus-x11 package, while it works it is very slow to start since it tries to connect to an X server. After one minute the X connection times out an dbus-launch does its thing anyway.

The `dbus-launch` can be compiled without X11 support to skip that timeout. Maybe we should provide a dbus-tools or dbus-wayland package with the same tools without X11 support?

PureTryOut:
KDE released plasma-framework 5.37.0, so this PR updates it. Also added *a ton* of missing subpackages and makedepends. Still missing a few though, I'll go through each package zapping the chroots before building them soon.

Also packaged dbus-launch without X11 (just install `dbus`), which resolves #377.


ollieparanoid:
I wasn't sure whether this would automatically build `dbus-nox11` when installing `weston` (due to some dependency). So I tested it, and it does *not*, everything is fine.
Thanks for updating this!

PR #385, opened 2017-08-15
Resolve #361 by zapping existing chroots after init

### LINKED ISSUE #361: pmbootstrap init: Explain, that old user interfaces are still installed when the rootfs chroot exists
This is unintuitive if not explained, as aleph- let us know in the IRC:
> Also fwy, choosing none in the initialization process, for some reason weston is still selected instead of just plain framebuffer.

craftyguy:
This extends zap() to add a 'no_confirm' option (False by default), and
zap() is now called by init with no_confirm=True to automatically zap
any existing chroots after the user runs init. This helps insure that
what is installed in the chroots is exactly what the user expects after
setting options in init.


PabloCastellano:
This is just my opinion and I am happy to change it, but I have used `pmbootstrap init` a lot because I have worked with multiple devices and I don't really want nor expect the `init` command to zap my chroots.

It is also true that I have not had the need to change the user interface and so, I've never faced this problem. For me this change will only suppose having to wait more time while changing from one device to another and creating the system image.

Besides that, I've recently implemented the `pmbootstrap config` command so that now `init` is not the only way to change the pmbootstrap configuration. I could use this new command but still I feel like sometimes I'd prefer to use `init` and in this case it would be doing something that I don't expect.

May we consider an option `zap_before_install` or `zap_after_init` for people that really want to do it?

PabloCastellano:
Other options might be:

- Detecting somehow that you have chosen a different user interface since last time you built the chroot?
- Force uninstalling all user interface packages and installing only the chosen one

craftyguy:
@PabloCastellano you have a very developer-oriented (and perfectly valid!) use case. I wouldn't expect users to have more than one device, but may re-run init to change config (as what happened in #361). 

What about a `--skip-zap` option to `init` to skip zapping it? This would make zapping chroots default behavior for `init`, but allow you to preserve chroots for developing on multiple devices. I still think zapping chroots after init should be the default behavior since I would expect that users would want the config they just set to actually be applied and requiring an additional step to "really now apply it" seems silly.

ollieparanoid:
As proposed in #361, we could also ask the user if the chroot(s) should be removed, in case they already exist (I would limit that to one question only, and default to `y`). So people who know what they are doing and want to save time can press `n`, and the default will be what most people expect, which is zapping the folders. How about that, @craftyguy and @PabloCastellano?

pmb/config/init.py
```diff
@@ -102,9 +103,13 @@ def init(args):
     # Save config
     pmb.config.save(args, cfg)
 
-    logging.info(
```
ollieparanoid: As much as I would love to remove the zapping warning, we still do no automatically check if the chroots need an update or not once a day. So I think it is still important to warn the user, that one should not run `init` once and then keep using the `chroot` for months without updating.

pmb/config/init.py
```diff
@@ -102,9 +103,13 @@ def init(args):
...
-        "WARNING: The applications in the chroots do not get updated automatically.")
-    logging.info("Run 'pmbootstrap zap' to delete all chroots once a day before"
-                 " working with pmbootstrap!")
-    logging.info("It only takes a few seconds, and all packages are cached.")
+    # Zap any existing chroots
+    logging.info("Zapping any existing chroots")
+    # Read in device info
+    setattr(args, "deviceinfo", pmb.parse.deviceinfo(args))
+    # Do not zap any existing packages or cache_http directories
+    setattr(args, "packages", False)
```
ollieparanoid: I think it would be more elegant, if we could pass `packages` and `http` as parameters to `zap`, instead of faking the `args` variable for this case. (The `args.deviceinfo` is a bit special here, because it does not come from a commandline argument, unlike most other `args` options).

pmb/config/init.py
```diff
@@ -102,9 +103,13 @@ def init(args):
...
 
-    logging.info(
-        "WARNING: The applications in the chroots do not get updated automatically.")
-    logging.info("Run 'pmbootstrap zap' to delete all chroots once a day before"
-                 " working with pmbootstrap!")
-    logging.info("It only takes a few seconds, and all packages are cached.")
+    # Zap any existing chroots
+    logging.info("Zapping any existing chroots")
+    # Read in device info
+    setattr(args, "deviceinfo", pmb.parse.deviceinfo(args))
```
ollieparanoid: We need to check whether the deviceinfo file exists, before trying to load it (use case: add a new device).
And it would be nice if we checked if any chroot_ folder exists, before actually calling zap (you could do that with `len(glob.glob(args.work + "/chroot_*"))` for example).

craftyguy:
That seems like a good compromise to me, something like:

> A previous device chroot was detected, would you like to remove this to apply any new configuration set here? [y]

pmb/config/init.py
craftyguy: So modify zap further to have `packages` and `http` parameters and not try to read them from `args` like it does now? That does seem better, I was trying to refrain from modifying zap too much, which is why I went with this approach.

pmb/config/init.py
craftyguy: ok I will re-add the warning. #390 for hopefully a good way to address chroot package updates

craftyguy:
@PabloCastellano do you agree with my comment? If so, I can get this implemented ASAP!

PabloCastellano:
@craftyguy Yes, that looks a proper solution to me! :+1:

craftyguy:
@ollieparanoid 

I've implemented the prompt (defaulting to "y" as discussed) and have made all requested changes!

pmb/chroot/zap.py
```diff
@@ -23,7 +23,7 @@
 import pmb.helpers.run
 
 
-def zap(args):
+def zap(args, skip_confirm=False, packages=False, http=False):
```
ollieparanoid: Could you change that to `confirm=True`, so we don't unnecessarily invert the argument here (and adjust it below in the function accordingly)?

pmb/config/init.py
```diff
@@ -102,9 +103,20 @@ def init(args):
     # Save config
     pmb.config.save(args, cfg)
 
+    if pmb.helpers.cli.confirm(args, "Would you like to zap any existing chroots to apply any new configuration set here?", default=True):
```
ollieparanoid: I would change that to the following (shorter message, only display when 
needed):
```python
if glob.glob(args.work + "/chroot_*") and
	pmb.helpers.cli.confirm(args, "Zap existing chroots to apply configuration?",
				      default=True)":
```

pmb/config/init.py
```diff
@@ -102,9 +103,20 @@ def init(args):
...
+        if not os.path.exists(args.aports + "/device/device-" + args.device + "/deviceinfo"):
+            setattr(args, "deviceinfo", None)
+        else:
+            setattr(args, "deviceinfo", pmb.parse.deviceinfo(args))
+        # Zap any existing chroots
+        logging.info("Zapping any existing chroots")
```
ollieparanoid: The `logging.info` here and the comment above seem to be a bit redundant, because you have asked just before whether we should do the zapping. How about removing that?

craftyguy:
@ollieparanoid requested changes have been made!

ollieparanoid:
As discussed in the channel, this is just a tiny `glob` away from getting merged.

craftyguy:
Sorry about missing the glob comment. It's implemented now!

craftyguy:
@ollieparanoid 

 I updated the glob thing, but it seems that init is creating the native chroot before the condition statement, so even if no chroot existed beforehand, there's always at least one (the native one) available and the user gets prompted. Maybe this doesn't happen if the device is brand new (e.g. not in the list of devices) ?

ollieparanoid:
You are right, the `chroot_native` always got created. We did that to verify write access in the `$WORK` folder. I've changed it, so it writes the `cache_http` folder instead. Waiting for Travis, then it's finally ready to go! :tada:

PR #387, opened 2017-08-15
Allow users to copy ssh pubkey to device

craftyguy:
@ollieparanoid 

I would like to implement this, but I'm unsure how to actually perform the `cp ~/.ssh/id_rsa.pub <chroot_rootfs>/home/user/ssh/authorized_keys` step since there doesn't seem to be an obvious mechanism in `pmb` to facilitate copying arbitrary files from outside of a chroot into a chroot. Do you have any recommendations on how to proceed? 

Why this PR/effort? SSH password auth is old school, key auth is where it's at! Seriously though, there are many benefits to SSH key authentication compared to password auth!


drebrez:
You can try using this function
https://github.com/postmarketOS/pmbootstrap/blob/master/pmb/helpers/run.py#L67,
for example here it's used in the opposite direction
https://github.com/postmarketOS/pmbootstrap/blob/742e4624e8bb475f2aa5316a140b5a1abc7ccd38/pmb/build/menuconfig.py#L72
to copy the new config outside the chroot. Should work also for you.

pmb/install/install.py
```diff
@@ -140,9 +140,21 @@ def install(args):
     size_image = str(int(float(get_chroot_size(args)) * 1.15)) + "M"
     size_boot = str(int(get_chroot_boot_size(args)) + 5) + "M"
 
-    # Finally set the user password
+    # Set the user password
     set_user_password(args)
 
+    # If user has a ssh pubkey, offer to copy it to device
+    user_ssh_pubkey = os.path.expanduser("~/.ssh/id_rsa.pub")
+    if os.path.exists(user_ssh_pubkey):
```
PabloCastellano: I would also add an else case that shows a message like "couldn't find ssh key, you will have to login using password" :+1:

craftyguy:
@drebrez & @PabloCastellano 

Thank you both for the help! I've made the changes and it works now. Ready to merge if anyone else wants to review :)

pmb/install/install.py
craftyguy: Done!

PabloCastellano:
@drebrez better use logging.info instead of print :+1:

craftyguy:
@PabloCastellano oops, good catch. fixed! (old habits die hard)

pmb/install/install.py
```diff
@@ -140,9 +140,24 @@ def install(args):
     size_image = str(int(float(get_chroot_size(args)) * 1.15)) + "M"
     size_boot = str(int(get_chroot_boot_size(args)) + 5) + "M"
 
-    # Finally set the user password
+    # Set the user password
     set_user_password(args)
 
+    # If user has a ssh pubkey, offer to copy it to device
```
ollieparanoid: I would prefer if you put the whole block of new code into a new function in this file.

Please note, that the way you have implemented it right now, it won't work, because the contents of the home folder in the chroot will *not* end up in the final image. The reason is, that it contains empty folders for mountpoints, which do not make sense on the actual device.
So to make it work, I suggest calling that new function after `fix_mount_folders(args)` below, and writing to `/mnt/install/home/user/.ssh/...` instead of writing to the chroot folder.

(Right now the code is a bit awkward and copies everything first, then deletes the home folder. #389 will fix this.)

craftyguy:
@ollieparanoid I just tested this on an actual device and it works as intended, I think it's ready to go if it passes your review!

ollieparanoid:
Thanks!

PR #389, opened 2017-08-15
Fix #363: Omit /home/user when calculating system image size

### LINKED ISSUE #363: Calculating system image size does not ignore home folder
@opendata26 reported in IRC:
> the system image size also includes things in the home folder [in the rootfs chroot]

The thing is, that we do *not* copy this folder to the actual image (and this isn't really documented anywhere, which makes it totally understandable that we didn't pay attention to this), so the size calculation is slightly wrong, when there's a big file in the home folder.

@drebrez: Since you have implemented the minimum size calculation, would you like to take a look at this?

ollieparanoid:
* The system image size is now calculated as: root size - home size.
* New function in `pmb/helpers/other.py`: `folder_size()`, with a
  testcase.
* Instead of copying everything to the system image folder, and
  deleting the home folder afterwards, do not copy the home folder
  in the first place.
* Added `pmbootstrap -s` to skip generating the initramfs for faster
  debugging.
* Set the default value in the "are you sure, that your partition has
  at least..." to "y", so we can run `yes '' | pmbootstrap install`
  to make it run through the whole installation process.

---
@opendata26, you have reported the bug. Could you test if this PR fixes it for you?


pmb/install/install.py
```diff
@@ -26,65 +26,78 @@
...
+    full = root - home
 
-    # Run the du command
-    result = pmb.chroot.root(args, ["sh", "-c", "du -cm ./boot | grep total$ | cut -f1"],
-                             working_dir=mountpoint, return_stdout=True)
-    return result
+    # Add some free space, see also:
+    # https://github.com/postmarketOS/pmbootstrap/pull/336
+    full *= 1.15
+    boot += 15 * 1024 * 1024
```
ollieparanoid: I thought a few extra free MB would be nice, in case the kernel/initramfs/initramfs-extra size increases, or you even want to install another kernel.

The idea would be to mount that partition with write access to `/boot`, so one could upgrade the kernel simply with `apk`. This is already possible for the N900, because they mount `/boot` in the `fstab` (generated in `post-install`) and actually boot the kernel from that partition.

But you are right, we should also add those 15 MB to the full size, if we did it that way.

About increasing it in percent: we could also do that, but I don't see much benefit here.

So would it be OK for you, if we used 15 instead of 5 MB and also increased the "full" size with 15 MB?

pmb/parse/arguments.py
```diff
@@ -103,6 +103,9 @@ def arguments():
     parser.add_argument("-j", "--jobs", help="parallel jobs when compiling")
     parser.add_argument("-p", "--aports",
                         help="postmarketos aports paths")
+    parser.add_argument("-s", "--skip-initfs", dest="skip_initfs",
```
ollieparanoid: I think you are totally right with this, good point!

The problem is, that it is not that easy to tell when the initramfs needs to be regenerated. In early (but public) versions of `pmbootstrap`, the initramfs would not get generated often enough, and this made development harder. So I decided to always re-generate it as a workaround.

I've added ticket #411 to track this.

pmb/install/install.py
drebrez: @ollieparanoid it's fine for me to have additional space in the boot partition, instead of adding 15 MB in the full, we can change it to 1.20 which is an additional 5%.
With some real values will be:

|full|additional 5% in MB|
|-|-|
|300|15|
|400|20|
|500|25|
|600|30|
|700|35|
|800|40|

ollieparanoid:
Thanks for the reviews! :+1:

PR #393, opened 2017-08-15
Add firmware-adreno package

opendata26:
This adds the firmware needed to use adreno gpus


aports/device/firmware-adreno/APKBUILD
```diff
@@ -0,0 +1,43 @@
...
+pkgrel=0
+pkgdesc="Firmware for adreno gpus"
+url="https://github.com/postmarketOS"
+arch="armhf"
+license="proprietary"
+depends=""
+makedepends=""
+install=""
+subpackages=""
+source="a225_pm4.fw a225_pfp.fw a225p5_pm4.fw a300_pfp.fw a300_pm4.fw a330_pfp.fw a330_pm4.fw"
```
ollieparanoid: All sources should be downloaded from external sources (except for patches, and small scripts, which do not have their own repository).

Here is an example download link for the firmware files:
`https://github.com/TheMuppets/proprietary_vendor_sony/raw/cm-14.1/shinano-common/proprietary/etc/firmware/a225_pm4.fw`

This repository contains all other files needed for this package, too:
https://github.com/TheMuppets/proprietary_vendor_sony/tree/cm-14.1/shinano-common/proprietary/etc/firmware

aports/device/firmware-adreno/APKBUILD
```diff
@@ -0,0 +1,43 @@
...
+options="!check"
+
+build() {
+	return 0
+}
+
+package() {
+        install -D -m644 "$srcdir/a225_pm4.fw" \
+                "$pkgdir/lib/firmware/a225_pm4.fw"
```
ollieparanoid: Please put firmware files in `/lib/firmware/postmarketos`, so we don't have any collisions with `/lib/firmware` from the `linux-firmware` package.

aports/device/firmware-adreno/APKBUILD
ollieparanoid: Oh and make sure, that these are actually the same files, by comparing their hashes. I've read in another ticket, that they may be different!

opendata26:
Done

PR #394, opened 2017-08-16
Add deviceinfo to initramfs

PabloCastellano:
It was mentioned in https://github.com/postmarketOS/pmbootstrap/issues/88#issuecomment-318635417 and other places I don't remember now.

The `deviceinfo` file contains variables that may be useful for initramfs too. At the moment this file is usually smaller than 1KB.

This way the variables can be used in initramfs and we don't have to replace them in place.


PabloCastellano:
Is it a problem if I change the name of initramfs-extra image from `initramfs-$flavor-extra` to just `initramfs-extra`?

This way we could also get rid of `@INITRAMFS_EXTRA@` easily

drebrez:
@PabloCastellano when I created the initramfs-extra, @ollieparanoid asked me to add the flavor, and also the initramfs has the flavor. 
Since we are adding the deviceinfo in the initramfs we cannot have a single initramfs shared for all the devices, but the initramfs-extra at the moment is not device specific and can be shared.
Or we accept multiple initramfs for each device and a single initramfs-extra or we add the flavor value in a variable inside the initramfs or even in the deviceinfo that gets written to the initramfs.
@ollieparanoid do you have any other ideas?

PabloCastellano:
@drebrez It is true that initramfs-extra cannot be shared between devices but since both initramfs and initramfs-extra are placed inside the device's chroot, it is easily identifiable which device the initramfs is for:

```
ls -l /var/pmbootstrap/chroot_rootfs_motorola-titan/boot/
total 24260
-rw-r--r-- 1 root root 10885120 Aug 17 13:13 boot.img-motorola-titan
-rw-r--r-- 1 root root  1599655 Aug 17 13:13 initramfs-motorola-titan
-rw-r--r-- 1 root root  3071950 Aug 17 13:13 initramfs-motorola-titan-extra
-rw-r--r-- 1 root root  9280854 Aug 17 12:59 vmlinuz-motorola-titan
```

So I don't think it should not be a problem if these are named without the `$flavor`, but please correct me if I am wrong since you are more expert in initramfs than I am :smiley:

drebrez:
@PabloCastellano what I was saying is that only the initramfs is device specific, since it contains the modules to load and now the deviceinfo, the initramfs-extra only contains additional tools and nothing related to the device itself, this means that the initramfs-extra is exactly the same for every device.
But you are also right that all the files (boot.img, initramfs, vmlinux, ...) are stored in the device's chroot, and therefore there will be device specific files anyway.
Don't know why we need to differentiate the files with the flavors, maybe in future @ollieparanoid wanted to have all of them in the same place?

ollieparanoid:
First of all, thanks for tackling this issue! 

### Reason for `$-flavor`
The kernel and initramfs in `/boot` in traditional Linux distributions have such a flavor appended, because this allows to install multiple kernels side-by-side and choose between them in a boot menu. Alpine also has this concept, you could install `linux-vanilla` next to `linux-hardened` if you wanted to.

It's obvious for the kernel (vmlinuz) to have an own name. But it is also necessary for the initramfs, because the initramfs in traditional Linux distributions contain kernel modules (which are always specific to one kernel!).

### postmarketOS relation
My idea was to also support this use-case in postmarketOS. Specifically to have the mainline kernel next to the outdated kernel, that the device shipped with (but maybe also for a `lts` kernel at some point in the future).

Just like in regular Linux distributions, our `initramfs` can also contain kernel modules (this is the idea for the mainline kernel, and we use it there already, Android kernels usually do not come with module support).


### `initramfs-extra`
Right now, the `intiamfs-extra` does *not* contain any *kernel-specific* files, such as kernel modules. It made sense to me to add the `flavor`, so when you run `mkinitfs`, you will get a new `initramfs-extra` for the kernel you are currently using, and not unexpectedly overwrite another 

But as you rightfully point out, this also makes it more complex, because we need to put the name of the `initramfs-extra` file into the `initramfs`.

**I think it is okay to have just one `/boot/initramfs-extra`** (without the flavor appended)!

What I still dislike is, that it gets overwritten for all kernels, when you run `mkinitfs` for one specific kernel. To solve that, I propose to generate the `/boot/initramfs-extra` file when the package gets compiled, so it really only gets done once. But that would be the task of another PR. If you guys agree with that, I would open a ticket about that. (This is inspired by what @drebrez said about having it the same on all devices... that would really be useful!)


### Bonus remarks: choosing the kernel
Right now we do not have a bootloader in most devices, where you can actually choose the kernel (exception is the N900 with uboot). As you know already, we put the initramfs + kernel in a `boot.img` and flash/boot that with fastboot in most cases (it's a bit different on `heimdall-isorec`, but you also can't choose the kernel there).

But that does not mean, that this will be the case in the future - maybe we will get to put such a bootloader on the device at some point (e.g. via kexecboot or by porting u-boot, ... this all has been done before for specific Android devices).

What *is* supported (but untested) right now on Android is flashing one kernel, and `fastboot boot`ing another kernel - without collisions in the `/boot` folder, where the kernel related files are stored.

ollieparanoid:
*Just labeled, because it became clear in the chat that it isn't easy to see which PRs are ready to merge and which are not. Please remove the label when you think it's ready!*

PabloCastellano:
Understood. Ok, what do you think about this setup?

On every `install` generate the images with the kernel version appended, as:
- `boot.img-motorola-titan-3.4.113-2`
- `initramfs-motorola-titan-3.4.113-2`
- `initramfs-motorola-titan-extra-3.4.113-2`
- `vmlinuz-motorola-titan-3.4.113-2`

Then create a shorter symlink for each file:
- `boot.img` ->`boot.img-motorola-titan-3.4.113-2`
- `initramfs` -> `initramfs-motorola-titan-3.4.113-2`
- `initramfs-extra` -> `initramfs-motorola-titan-extra-3.4.113-2`
- `vmlinuz` -> `vmlinuz-motorola-titan-3.4.113-2`

So the symlinks always point to the last generated (and default) images and you can even create the images for newer versions of kernels without overwritting the previous ones.

I think we can leave this for another issue and merge the part that doesn't need discussion.

ollieparanoid:
> I think we can leave this for another issue and merge the changes currently in the branch that don't need a discussion.

That sounds good! So for now we keep that one variable with the `initramfs-extra-...` value in this PR?

PabloCastellano:
@ollieparanoid Yes, the `@INITRAMFS_EXTRA@` placeholder is kept: https://github.com/postmarketOS/pmbootstrap/pull/394/files#diff-2c637a1fe9cbc6f096c4e8e0f8f306daL145

ollieparanoid:
Thanks for making this PR, merged!

PR #395, opened 2017-08-17
Yet another minor fixes PR

PabloCastellano:
- postmarketos-splash: print filename to log file
- postmarketos-mkinitfs: Fixed wrong variable name


aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -25,7 +25,7 @@ parse_commandline()
 	kernel=$3
 	modules_path="/lib/modules/${kernel}"
 
-	if [ ! -d ${abi_path} ]; then
+	if [ ! -d "${modules_path}" ]; then
```
ollieparanoid: Thanks for noticing!
**This check is broken and would introduce breakage if it worked, so it should be entirely removed**:
Just about all Android kernels do not have modules enabled, which means that they do not have a `/lib/modules/...` folder after installation. So from reading the code `mkinitfs` should `exit 1` for them - but it somehow does not, I don't even see right now what is the reason for that (thought it was the misplaced `!` in the brackets, but I couldn't confirm it with shell testing).

aports/main/postmarketos-splash/make-splash.py
```diff
@@ -6,7 +6,7 @@
 
 
 def make_splash(width, height, filename, landscape=False, text="", config_file=None, center=False):
-    print("Creating {}x{} splashscreen".format(width, height))
+    print("Creating ({}x{}) splashscreen {}".format(width, height, os.path.basename(filename)))
```
ollieparanoid: Thanks for that, it's a detail that I also wanted to have 👍

ollieparanoid:
*Just labeled, because it became clear in the chat that it isn't easy to see which PRs are ready to merge and which are not. Please remove the label when you think it's ready!*

PabloCastellano:
Done

ollieparanoid:
@PabloCastellano thanks! there are two things blocking this PR (this might have gotten lost in the noise):
* https://github.com/postmarketOS/pmbootstrap/pull/395#discussion_r133759529
* merge conflict (rebase on master)

PabloCastellano:
@ollieparanoid Oh, I missed that. Do you prefer commenting out the code + leaving a reference to the comment where you explain the issue, or simply remove the check?

ollieparanoid:
@PabloCastellano I strongly prefer *not* leaving commented out code anywhere, please remove it.

PabloCastellano:
Conflict resolved

PR #404, opened 2017-08-17
Recovery installer zip

ata2001:
Work in progress.

TODO:

- [ ] Samsung isorec flashing
- [ ] More elegant way to handle recovery install options
- [ ] Improve support for multiple flash methods in pmbootstrap
- [ ] Compare the device codename with the one in the system.prop, to avoid flashing to wrong device


ata2001:
The test failure is likely a flake8 bug, PEP8 even recommends breaking before binary operators. https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

pmb/install/recovery.py
```diff
@@ -0,0 +1,39 @@
...
+                           str(args.full_disk_encryption).lower())]))
+
+    commands = [
+        # Move config file from /tmp/ to zip root
+        ["mv", "/tmp/install_options", "install_options"],
+        # Copy boot.img to zip root
+        ["cp", rootfs + "/boot/boot.img-" + args.device, "boot.img"],
+        # Create tar archive of the rootfs
+        ["tar", "-pczf", "rootfs.tar.gz", "-C",
+         rootfs, "."],
```
drebrez: in the tar you should exclude the home directory (/home/user) like it is done for the normal install (see [here](https://github.com/postmarketOS/pmbootstrap/blob/5144323137c267324510613c9c248fc94ec8bedd/pmb/install/install.py#L85-L87), there is already a PR to skip it instead of copy and delete it)

pmb/flasher/export.py
```diff
@@ -75,6 +78,19 @@ def symlinks(args, flavor, folder):
         pmb.helpers.file.symlink(args, file, link)
 
 
+def android_recovery_zip(args, folder):
+    """
+    Export android recovery compatible zip.
+    """
+    recovery_zip = "pmos-" + args.device + ".zip"
```
drebrez: if the user didn't generated the recovery zip what happen? maybe you can show a message like it's done for the system img (see [here](https://github.com/ata2001/pmbootstrap/blob/99561c6d06c10cfe9b840fe9723126856dc473ee/pmb/flasher/frontend.py#L115))

pmb/install/recovery.py
```diff
@@ -0,0 +1,39 @@
...
+    # Create config file for the recovery installer
+    with open(args.work + "/chroot_" + suffix + "/tmp/install_options",
+              "w") as install_options:
+        install_options.write(
+            "\n".join(['DEVICE="{}"'.format(args.device),
+                       'FLASH_BOOTIMG="{}"'.format(
+                           str(args.recovery_flash_bootimg).lower()),
+                       'INSTALL_PARTITION="{}"'.format(
+                           args.recovery_install_partition),
+                       'CIPHER="{}"'.format(args.cipher),
```
drebrez: how do you specify the cipher? seems missing from the arguments parser

ata2001: It's there: https://github.com/ata2001/pmbootstrap/blob/99561c6d06c10cfe9b840fe9723126856dc473ee/pmb/parse/arguments.py#L181

ollieparanoid:
You are right, Python developers recommend the opposite now, as they did before, and somehow the flake8 on Travis CI doesn't reflect that change. I've opened #405 for that, so we can fix it in one go across the whole code base. For consistency reasons I hope it is okay if you could change it to use what passes flake8 until then.

I'm so happy to see your work on the recovery installer, and that it will finally be integrated in pmbootstrap \o/

pmb/install/recovery.py
drebrez: @ata2001 sry, didn't know it was already present in the master branch, that's the reason I didn't noticed it in the changed files.

aports/main/postmarketos-android-recovery-installer/build_zip.sh
```diff
@@ -0,0 +1,36 @@
...
+
+BINARIES="/sbin/cryptsetup /sbin/kpartx /usr/sbin/parted /usr/sbin/partprobe"
+LIBRARIES=$(lddtree -l "$BINARIES" | awk '/lib/ {print}' | sort -u)
+copy_files "$BINARIES" bin/
+copy_files "$LIBRARIES" lib/
+check_whether_exists rootfs.tar.gz
+if [ "$FLASH_BOOT" = "true" ]
+then
+	check_whether_exists boot.img
+fi
```
ollieparanoid: How about putting that in one line?

```shell
[ "$FLASH_BOOT" = "true" ] && check_whether_exists boot.img
```

aports/main/postmarketos-android-recovery-installer/APKBUILD
```diff
@@ -0,0 +1,34 @@
...
+source="build_zip.sh
+	update-binary
+	pmos_install
+	pmos_install_functions
+	pmos_setpw"
+arch="noarch"
+license="GPL3"
+
+package() {
+	mkdir -p "$pkgdir/usr/sahre/postmarketos-android-recovery-installer/"
```
ollieparanoid: mkdir -p "$pkgdir/usr/sahre/postmarketos-android-recovery-installer/"

typo: `sahre` instead of `share`

Also this line is not necessary, because you have another `mkdir -p` below, which contains this folder

aports/main/postmarketos-android-recovery-installer/pmos_setpw
```diff
@@ -0,0 +1,11 @@
+#!/sbin/ash -e
+# shellcheck source=pmos_install_functions
+. /tmp/postmarketos/bin/pmos_install_functions
+
+extract_partition_table
+set_subpartitions
+echo "Set the password of the encrypted rootfs!"
+set_password
```
ollieparanoid: This is a one line function anyway, which only gets called in `pmos_setpw`, and you directly call `cryptsetup` for removing the key. So I would also directly put the `cryptsetup luksAddKey` command here, and remove the `set_password` function.

aports/main/postmarketos-android-recovery-installer/pmos_setpw
```diff
@@ -0,0 +1,11 @@
...
+#Remove temporary keyfile
```
ollieparanoid: You could add a space after the `#`.

pmb/config/__init__.py
```diff
@@ -292,6 +292,14 @@
...
         },
     },
+    "adb": {
+            "depends": ["android-tools"],
+            "actions":
+            {
+                "list_devices": [["adb", "devices"]],
+                "sideload": [["adb", "sideload", "$RECOVERY_ZIP"]],
+            }
+    },
```
ollieparanoid: This is so beautiful! \o/

pmb/install/install.py
```diff
@@ -143,15 +148,36 @@ def install(args):
     # Finally set the user password
     set_user_password(args)
 
+    if args.android_recovery_zip:
+        logging.info("*** (3/{}) CREATING RECOVERY-FLASHABLE ZIP ***".format(
+                     steps))
+        buildroot = "buildroot_" + args.deviceinfo["arch"]
```
ollieparanoid: How about renaming `buildroot =` to `suffix =`?

pmb/install/install.py
```diff
@@ -143,15 +148,36 @@ def install(args):
     # Finally set the user password
     set_user_password(args)
 
+    if args.android_recovery_zip:
```
ollieparanoid: The `install()` function is getting pretty long. I would like to split the lines below into two new functions: `install_system_image()` and `install_android_recovery_zip()`. The `install` function would call the appropriate function here, depending on `args.android_recovery_zip`. What do you think about that?

ollieparanoid:
*Just labeled, because it became clear in the chat that it isn't easy to see which PRs are ready to merge and which are not.*

pmb/install/install.py
ata2001: That variable was already declared in `install()`, so I wanted to avoid confusion. https://github.com/ata2001/pmbootstrap/blob/99561c6d06c10cfe9b840fe9723126856dc473ee/pmb/install/install.py#L126

That however is not going to be a problem, once we've split `install()`.

pmb/install/install.py
ata2001: Do you think after splitting `install()`, `create_zip()` still needs an own file/function?

ollieparanoid: yes, I'd still keep the separate function and file for `create_zip()`. The `install.py` is pretty full already, and it's a good logical split in my opinion.

aports/main/postmarketos-android-recovery-installer/build_zip.sh
```diff
@@ -0,0 +1,34 @@
...
+		echo "$1 not found"
+		return 1
+	fi
+}
+
+# shellcheck disable=SC1091
+. ./install_options
+
+BINARIES="/sbin/cryptsetup /sbin/kpartx /usr/sbin/parted /usr/sbin/partprobe"
+# shellcheck disable=SC2086
```
ata2001: ash doesn't support arrays, see https://github.com/koalaman/shellcheck/wiki/SC2086

ata2001:
Thank you @drebrez and @ollieparanoid for reviewing! I believe I made the requested changes.

ollieparanoid:
Great work! I'll test this on my `lg-mako` now.

ollieparanoid:
TWRP version ([here](https://dl.twrp.me/mako/)):
> twrp-3.1.1-0-mako.img	9M	2017-05-17 20:12:29 EDT

What I've tested:
* `pmbootstrap flasher --method=adb sideload` without generating the image first -> prints out the error with instructions correctly
* `pmbootstrap install --android-recovery-zip`
  * I got `error: closed` the first time, but after rebooting TWRP it started sideloading
  * Then it failed during extraction somewhere below 50%, all the Linux terminal says is: `Total xfer: 1.03x`
  * in TWRP, it says:
```
Starting ADB sideload feature...
Installing zip file '/sideload/package.zip'
postmarketOS recovery installer
Extracting partition info from fstab...
Entering working directory...
Extracting files...
# Linux PC says, that it's serving the "pmos-lg-mako.zip", but it stops before it reaches 50%.
Symlinking .so files to /lib/...
Unmounting /system...
Creating partition table on /dev/block/mmcblk0p21...
Failed. Check /pmos.log for more info!
Updater process ended with ERROR: 1
```
* Could it be, that the RAM is full or something?
* There is no `pmos.log` in `/` (checked with `ls /` in TWRP terminal and `adb shell`), and `/sideload` is empty.
* I have tried this 3 times, and it fails there consistently.


Now that I have tried it out, I also have some more code remarks (sorry, I know it's annoying when you don't get the feedback in one go) - I'll comment them in the code.

I have also started updating the wiki for this method, please review if this is correct and do necessary changes (and maybe extend it a bit more if it makes sense): https://github.com/postmarketOS/pmbootstrap/wiki/deviceinfo_flash_methods#recovery-mode-adb



I've also had issues with `adb` not wanting to start inside the chroot. I guess, that is because I `killall`'d it. But it works when I started the `adb` server outside of `pmbootstrap`. Inside `pmbootstrap` I got the following:
```
* daemon not running. starting it now on port 5037 *
* daemon started successfully *
** daemon still not running
```


What I will try now, is zapping everything, and re-building the image with the `none` interface, so it is smaller and see if it works then, and add my code remarks. Let's work a bit more on this, so we can ship this awesome feature in the quality it deserves!

aports/main/postmarketos-android-recovery-installer/build_zip.sh
```diff
@@ -0,0 +1,34 @@
...
+LIBRARIES=$(lddtree -l $BINARIES | awk '/lib/ {print}' | sort -u)
+copy_files "$BINARIES" bin/
+copy_files "$LIBRARIES" lib/
+check_whether_exists rootfs.tar.gz
+[ "$FLASH_BOOT" = "true" ] && check_whether_exists boot.img
+zip -r "pmos-$DEVICE.zip" .
```
ollieparanoid: How about generating the zip file in `/var/lib/android-recovery-zip` (or any other folder in `/var/lib`)? Because `/var/lib` is for [variable state information](http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s08.html) and `/usr/share/` is for [read only data](http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html).

You could `cd` into that folder, before executing the `zip`, and then get the same folder structure.

pmb/install/install.py
```diff
@@ -226,3 +195,68 @@ def install(args):
...
+
+    # Flash information
+    logging.info("*** (4/4) FLASHING TO DEVICE ***")
+    logging.info("Run the following to flash your installation to the"
+                 " target device:")
+    logging.info("* pmbootstrap flasher --flash-method adb sideload")
+    logging.info("  Flashes the installer zip to your device:")
+
+    # Export information
+    logging.info("* If the above steps do not work, you can also create a"
```
ollieparanoid: There's only one step, so that could be reworded as `"* If this does not work, ..."`.

pmb/config/__init__.py
```diff
@@ -285,6 +285,14 @@
                 ["heimdall", "flash", "--$PARTITION_KERNEL", "$BOOT/boot.img-$FLAVOR"]],
         },
     },
+    "adb": {
+            "depends": ["android-tools"],
+            "actions":
+            {
+                "list_devices": [["adb", "devices"]],
+                "sideload": [["adb", "sideload", "$RECOVERY_ZIP"]],
```
ollieparanoid: All other flash commands wait for the device so far, and `adb` seems to be also able to do that with `wait-for-usb-sideload`. What do you think about using that here?

pmb/config/__init__.py
```diff
@@ -285,6 +285,14 @@
                 ["heimdall", "flash", "--$PARTITION_KERNEL", "$BOOT/boot.img-$FLAVOR"]],
         },
     },
+    "adb": {
+            "depends": ["android-tools"],
+            "actions":
+            {
+                "list_devices": [["adb", "devices"]],
```
ollieparanoid: Does starting the `adb` server withing `pmbootstrap` work for you? It always fails for me. After `zap` I get:
```
* daemon not running. starting it now on port 5037 *
ADB server didn't ACK
* failed to start daemon *

* daemon not running. starting it now on port 5037 *
ADB server didn't ACK
* failed to start daemon *
```

aports/main/postmarketos-android-recovery-installer/build_zip.sh
ollieparanoid: Another point: How about using `zip -0`, so it does not try to compress the already compressed rootfs again, and therefore saving some time? It will get a bit bigger, because the other files are not compressed (which are relatively small compared to the rootfs), but in my opinion that is a good trade-off.

pmb/install/recovery.py
```diff
@@ -0,0 +1,39 @@
...
+    Create android recovery compatible installer zip.
+    """
+    zip_root = "/usr/share/postmarketos-android-recovery-installer/"
+    rootfs = "/mnt/rootfs_" + args.device
+
+    # Install recovery installer package in buildroot
+    pmb.chroot.apk.install(args,
+                           ["postmarketos-android-recovery-installer"],
+                           suffix)
+
```
ollieparanoid: Something like `logging.info("(native) create recovery zip")` would be nice here, because the last print on the normal terminal is the installation of `postmarketos-android-recovery-installer` and it will take a bit of time until the zip creation below finishes.

ata2001:
@ollieparanoid Thank you so much for testing!

> I got error: closed the first time, but after rebooting TWRP it started sideloading

Even without pmos.log I can tell, it's a bug in the new error handling system. parted returns nonzero, when creating subpartitions, so we need to catch and ignore them. I've used an sdcard for testing, that's why it was working for me. Sorry about this.

> There is no pmos.log in / (checked with ls / in TWRP terminal and adb shell), and /sideload is empty.

Oops, it's a typo in the fail message. It's left there, from when I started writing this script. The installer now does everything inside /tmp/postmarketos/. Sorry about this too.

> I have also started updating the wiki for this method, please review if this is correct and do necessary changes (and maybe extend it a bit more if it makes sense): https://github.com/postmarketOS/pmbootstrap/wiki/deviceinfo_flash_methods#recovery-mode-adb

Thank you, I will check it out and edit if necessary.

The error came from the lack of testing, I forgot to test it with --recovery-install_partition=system. It's an easy fix however, I fix everything, and please test it again!

(I will write down, how to test everything, when updating it in the future, to avoid errors like this.)

ata2001:
@ollieparanoid it's --flash-method in the current branch, but renaming that --method is a good idea

pmb/config/__init__.py
```diff
@@ -289,6 +290,16 @@
                 ["heimdall", "flash", "--$PARTITION_KERNEL", "$BOOT/boot.img-$FLAVOR"]],
         },
     },
+    "adb": {
+            "depends": ["android-tools"],
+            "actions":
+            {
+                "list_devices": [["adb", "-P", "5038", "devices"]],
```
ata2001: changed port, so adb server running on host won't cause any problem

ata2001:
I made the requested changes.

ata2001:
Results:

compression | zip size | time           |
-------------------|------------|------------------|
without -0     | 131MB | 2 min 50 s  |
with -0          | 136MB | 2 min 2 s   |

ollieparanoid:
I'd prefer saving 48 seconds for 5 MB, what do you think?
Also thanks for updating the PR. I don't have time *right now* to test it, but I'll try to do it soon!

ollieparanoid:
Thanks again for upgrading the code. I just ran it on my `lg-mako`, and it fails with ([full log](https://github.com/postmarketOS/pmbootstrap/files/1239758/pmos.txt)):
```shell
+ mount -t ext2 -rw /dev/block/mmcblk0p21p1 /system/boot
mount: mounting /dev/block/mmcblk0p21p1 on /system/boot failed: Operation not supported on transport endpoint
```

pmb/install/install.py
```diff
@@ -247,3 +213,71 @@ def install(args):
...
+    logging.info("*** (3/4) CREATING RECOVERY-FLASHABLE ZIP ***")
+    suffix = "buildroot_" + args.deviceinfo["arch"]
+    mount_device_rootfs(args, suffix)
+    pmb.install.recovery.create_zip(args, suffix)
+
+    # Flash information
+    logging.info("*** (4/4) FLASHING TO DEVICE ***")
+    logging.info("Run the following to flash your installation to the"
+                 " target device:")
+    logging.info("* pmbootstrap flasher --flash-method adb sideload")
```
ollieparanoid: `--method`

ata2001:
@ollieparanoid Seems like an issue with the kernel you're using. http://www.link2sd.info/faq (scroll to the bottom). Maybe it doesn't support ext2. Can you try manually creating and mounting an ext2  partition?

ollieparanoid:
There are two topics I would like to discuss, which I mentioned earlier, but had no time to elaborate them properly.

### Put `adb` in flash_methods in `deviceinfo`?
Right now you could specify `--method=adb` for *any* device, even if it is completely unsupported (like the N900). I would prefer if we only allowed that if it was listed in the `deviceinfo` (maybe with an explanation how to add it, in case the user calls it anyway). What do you think about that?

### Make `export` behave like with the system image
The recovery zip gets generated as alternative to the system image in the `pmbootstrap install` step.
With the system image, export works like the following:
* `pmbootstrap flasher export` automatically creates a symlink to the system image in case it exists
* if it does not exist, a "NOTE:" gets printed out with a message how you would generate that image (which we could also do in case `adb` is a flash method, if we put it in the deviceinfo)
* there's no need to specify an extra parameter

I know that you have modeled it after the Odin export. In my opinion it makes more sense for the Odin part to do it with the extra parameter, because you can't generate the odin image in `pmbootstrap install` (in case you agree and want to make it more like the system image, and @drebrez  is also on board with this, it might sense to adjust the Odin image generation to behave the same way in a new PR).

I'd also like your opinion on that.

Thanks again for all the hard work you have put into this feature \o/

ata2001:
> Put adb in flash_methods in deviceinfo?

I've been thinking about this. The problem is, adb is not a device specific feature, but a protocol, that any device could implement with a working usb interface/networking (which I believe every of the currently supported devices has), so it has no place in the `deviceinfo` IMO.

ata2001:
> Make export behave like with the system image

We should definitely do that. I'd also like to hear @drebrez 's opinion on the odin and zip export issue.

ollieparanoid:
> The problem is, adb is not a device specific feature, but a protocol, that any device could implement with a working usb interface/networking

This is true, but you could say the same about `fastboot`.

> (which I believe every of the currently supported devices has)

The N900 and all future non-Android devices will probably not have this method. The `i9100` and `i9070` have `adb` in the recovery system, but they are not supported by the recovery zip yet (because of the `isorec` stuff), so at least for me it would make sense not to specify `adb` as supported flash method in the `deviceinfo`.

---
About the mount issue: We're pretty sure, that this is `ext2` related, right? How about we use `ext4` in `/boot` instead with your method, this should be compatible for all (or at least for more) Android kernels  if I'm not mistaken. In case it turns out, we really need to be able to choose between ext2 and ext4, we could add this as option later.

ata2001:
@ollieparanoid Have you seen my edit? You're right about fastboot, however that is something available on the device by default, but these recoveries are something made by the community (even if the manufacturer installs one, you can probably overwrite it, what is not the case with loki/lk (unless you're McBitter). Just because someone installed pmOS successfully on some version of some recovery doesn't mean it will work for other people with the same device.

If that would make the python code less complicated, we could add adb to the deviceinfo every time someone successfully installs it with an android recovery on a device, but it would still make no sense to me.

ata2001:
> About the mount issue: We're pretty sure, that this is ext2 related, right? How about we use ext4 in /boot instead with your method, this should be compatible for all (or at least for more) Android kernels if I'm not mistaken. In case it turns out, we really need to be able to choose between ext2 and ext4, we could add this as option later.

It could fall back to ext4 in that case. But then a recovery install would become different than the image generated by pmbootstrap. Do we really want that? It could cause issues in the future, and slow postmarketos progress down.

ollieparanoid:
Sorry, I did not see your edit.

> Just because someone installed pmOS successfully on some version of some recovery doesn't mean it will work for other people with the same device.

This is also true, but at least it makes it likely to succeed (in the sense of: if it works for me with TRWP, it is very likely that it will work on the same device with (a future/different version of) TWRP for someone else).

I understand what you mean, you'd rather have only the kind of information of the device in the file, which never change. But we have already diverged from that. We store kernel module names and dtb file names in there, which in case of Android are also something made by the community (the people who upstreamed the drivers to Linux often chose the names, especially for the `dtb` files). So I personally see that file as a mix of what is unchangeable information about the device, like the resolution or device name, and information according to real world usage, with all the software we have around.

>If that would make the python code less complicated, we could add adb to the deviceinfo every time someone successfully installs it with an android recovery on a device, but it would still make no sense to me.

We don't have to put it in there if you don't like it, I just want to make sure that we make a conscious decision here as community (more opinions from other people welcome!) and talk about the pros- and cons. If we put it in there, I would also be fine with putting it in all Android device's `deviceinfo` file, where we expect it to be working even without testing, if you prefer that.

---

> It could fall back to ext4 in that case. But then a recovery install would become different than the image generated by pmbootstrap. Do we really want that? It could cause issues in the future, and slow postmarketos progress down.

I would prefer that actually, as I do not see any issue with having another filesystem on /boot for pmOS development, or even slowing it down (do you have an example?). Using `ext4` here (or at least falling back to it) simply means more compatibility with this method. I don't see how else it could be used on the `lg-mako`, other than building my own TWRP image with a custom kernel. Which is not convenient at all and means a lot of extra work for porting. For me, the big advantage of TWRP here is, that it is practically available for almost every Android device.

ata2001:
> I would prefer that actually, as I do not see any issue with having another filesystem on /boot for pmOS development, or even slowing it down (do you have an example?). Using ext4 here (or at least falling back to it) simply means more compatibility with this method. I don't see how else it could be used on the lg-mako, other than building my own TWRP image with a custom kernel. Which is not convenient at all and means a lot of extra work for porting. For me, the big advantage of TWRP here is, that it is practically available for almost every Android device.

Right, but it could still print a warning like 'failed to mount ext2 formatted partition, reformatting it in ext4' or something.

ata2001:
> We don't have to put it in there if you don't like it, I just want to make sure that we make a conscious decision here as community (more opinions from other people welcome!) and talk about the pros- and cons. If we put it in there, I would also be fine with putting it in all Android device's deviceinfo file, where we expect it to be working even without testing, if you prefer that.

I always wanted to tell you, how I appreciate, that you're never trying to force your design decisions on the community, and that even if you started this project, you aren't trying to be a boss, that there is no hierarchy in this project, and write access means no power, but only responsibility. If this project is going to be successful (it already is in my eyes), that will be  a proof, that people don't need bosses and hierarchy to create something good and useful.

ollieparanoid:
Very kind words, thank you very much! 🎉

drebrez:
@ollieparanoid @ata2001 I'm thinking about where to place this functionality and the odin zip also.
The odin zip it seems correct in the export because like the kernel and the system image, a user can export in order to flash/copy/install/.. it in his device manually or through other softwares. For the recovery zip method I see a slightly different situation because it also contains a part of the install process (disk encryption stuff).
I also see 2 separated functionality here, because the zip file for the recovery can also be installed directly from the device, without the sideload method.
What about adding the system.img file directly inside the zip? if that is possible then it will be more like an export functionality.

ata2001:
> What about adding the system.img file directly inside the zip?

It could be useful, but I wouldn't want to just throw away the recovery installer scripts I wrote to be honest.

ata2001:
It will be confusing, when sideload doesn't flash the same zip as you would export. If we want to keep both methods, then we should add an argument to `install` instead.

drebrez:
@ata2001 I'm not saying to throw it away, it's fine to create the partitions and encrypt them from there. The only mixed part that I see is copying the rootfs into the zip file to copy it later in the partitions created directly in the device. I was thinking something like opening the system.img inside the recovery script to copy out the rootfs files.
At that point the user will have to do the normal `pmbootstrap install` to generate the image, then he can export it for the twrp zip

ata2001:
The point of the recovery installer is to create zips, that you can share on the internet, so users don't have to use an external computer to install pmos.

drebrez:
@ata2001 not sure it is possible to open the system.img directly in the recovery, but with that you can still share the zip on the internet.

ata2001:
But the image is already encrypted, you can't really open it in the recovery without a password.

drebrez:
@ata2001 I was also thinking about that combination. If the system.img inside the zip is not encrypted you think it's possible to open it to extract the files? because if it's not possible we can already discard my idea :)

drebrez:
@ata2001 Do you know if it's possible to prompt the user if he want the encrypted disk or not during the recovery install?

ata2001:
@drebrez you can mount it I guess, but what's the point?

ata2001:
@drebrez `--no-fde` it won't prompt, but does the same

drebrez:
@ata2001  because like I mentioned before, is see the recovery zip more like a way to install it on the device, not a different way to create a system image. all the other methods (fastboot, heimdall, odin, ...) they all start from the already created system and kernel image files, they don't create them again.

ata2001:
@drebrez my approach is rewriting part of pmbootstrap in shell. This is the only way you can install it on a device without using an external computer. It's more than a flashing method. We could of course just put the two image files in a zip, and flash them. If you'd like to do that, adding an option to `pmbootstrap install` would be perfect IMO.

drebrez:
@ata2001 I know that the recovery zip is the only way to install it without a computer, and the install part inside the recovery is good like it is. We are just discussing how to fit it properly in the pmbootstrap point of view, right?

ata2001:
@drebrez yes

my solution requires a different install method, I don't believe it could be done only in export/flasher (you need it in both anyway)

drebrez:
the use cases i'm thinking are...

user with pmbootstrap does:
- pmbootstrap install (with or without disk encryption based on what he want)
- pmbootstrap export recovery zip
- transfer the zip to the device/twrp or use sideload feature
- during install it shouldn't prompt again to encrypt or not the disk

user without pmbootstrap (prebuilt zip):
- download recovery zip 
- transfer the zip to the device/twrp or use sideload feature
- prompt the user if he want encryption or not
- install it

ata2001:
@drebrez but you still need to generate the prebuilt zip somehow, and where else to put it if not in pmbootstrap? Also, if this script is working fine, is maintaining two different methods for recovery install necessary?

ata2001:
Like I've said before, we could add `--encrypt-in-recovery` option to install, so it would only generate a 'prebuilt' zip, when the user specifies it on the command line.

ollieparanoid:
@drebrez: At first sight, using the system image inside the recovery zip looks like a nice solution to make the recovery zip and odin export more consistent. But practically it will not work here, because:
* the recovery installation has the ability to encrypt the partition itself
* it therefore needs the rootfs files in an unencrypted state

So we would need the user to either do `install --no-fde` or decrypt the partition first during the recovery zip installation (-> in TWRP somehow). These two options are way more inconsistent than what we have now, so I strongly recommend *not* to do any of that. Plus it would really mean, that a 
lot of code would need to be rewritten in this PR.

**Instead I suggest we get this merged as soon as possible.** So we have a first working version, which we can improve later on if necessary. (@drebrez: for example you could make a PR in the future, which asks the user if full disk encryption should be turned on or not during the instalation in TWRP, I think from your use-case comment that this is a feature you would like to have).


As I see it, we have the following blockers left:
 - [ ] Support `ext4` as boot partition (@ata2001 prefered to try `ext2` first, then fall back to `ext4`)
 - [ ] Make `export` (of the recovery zip) [consistent with the export of the system image](https://github.com/postmarketOS/pmbootstrap/pull/404#issuecomment-323817484) (and less consistent with the odin export, yes, but as outlined above that is different anyway, because it wraps the pre-built system image, which we can not do with the recovery zip)

This PR got really filled up with comments by now. If you want to discuss anything of what I've said, please do. But open a new issue for that in the tracker, and reference this PR (we even have a `discussion` label).

**Let's get this awesome feature shipped! :ship:**

ollieparanoid:
Tested again on my `lg-mako` and it's working now \o/
I'll be very happy to merge this as soon as the conflicts files are resolved (and sorry for them, I know they keep coming all the time due to integrations of other code, and that's annoying for other PRs) 👍

ata2001:
@ollieparanoid The copyright stuff is still missing from the new files. Should I add myself as the copyright holder, or what should I do?

ollieparanoid:
Oh, good catch! Yep, that would be the convention.

aports/main/postmarketos-android-recovery-installer/pmos_setpw
```diff
@@ -0,0 +1,29 @@
+#!/sbin/ash -e
+
+#Copyright 2017 Attila Szöllősi
```
ollieparanoid: Space after the `#` would be more consistent with other comments.

pmb/parse/arguments.py
```diff
@@ -181,6 +184,16 @@ def arguments():
                          " added to the rootfs (e.g. 'vim,gcc')")
     install.add_argument("--no-fde", help="do not use full disk encryption",
                          action="store_false", dest="full_disk_encryption")
+    install.add_argument("--android-recovery-zip",
+                         help="generate TWRP flashable zip",
+                         action="store_true", dest="android_recovery_zip")
+    install.add_argument("--recovery-flash_bootimg",
```
ollieparanoid: Could we replace the `_` with `-`?

pmb/parse/arguments.py
```diff
@@ -181,6 +184,16 @@ def arguments():
...
+                         help="include kernel in recovery flashable zip",
+                         action="store_true", dest="recovery_flash_bootimg")
```
ollieparanoid: Would you mind making this `store_false`?

pmb/config/__init__.py
```diff
@@ -289,6 +290,16 @@
...
+                "sideload": [["adb", "-P", "5038", "wait-for-usb-sideload"],
```
ollieparanoid: This would be even more consistent with the rest if you don't mind, because it displays a message just before waiting:
```diff
             "actions":
             {
                 "list_devices": [["adb", "-P", "5038", "devices"]],
-                "sideload": [["adb", "-P", "5038", "wait-for-usb-sideload"],
+                "sideload": [["adb", "-P", "5038", "start-server"],
+                             ["echo", "< wait for any device >"],
+                             ["adb", "-P", "5038", "wait-for-usb-sideload"],
                              ["adb", "-P", "5038", "sideload",
                               "$RECOVERY_ZIP"]],
             }
```

pmb/parse/arguments.py
ata2001: The reason I went with `store_true`, is that right now it is only able to flash boot.imgs, and they aren't going to work on older samsung devices, that require isorec flashing.

I will change it to `flash-kernel` in the future.

pmb/config/__init__.py
ata2001: You've also added `start-server` to the commands. Why is that necessary?

pmb/parse/arguments.py
ollieparanoid: Thanks for explaining, that is a very good reason!

pmb/config/__init__.py
ollieparanoid: Otherwise the starting server messages would get printed after the "wait for any device" message. This way it is always at the bottom, until a device gets connected.

PR #414, opened 2017-08-18
Detect and resize root partition on QEMU

drebrez:
This will detect and resize the partition inside the QEMU and resize it.

To increase the size of the partition you simply increase the size of the image using `truncate -s +200M /path/to/qemu.img` for example and then start QEMU


ollieparanoid:
Thank you very much! I've [adjusted the wiki accordingly](https://github.com/postmarketOS/pmbootstrap/wiki/Common-Qemu-Notes).

PR #417, opened 2017-08-19
Alias: "pmbootstrap export" for "pmbootstrap flasher export"

ollieparanoid:
I kept writing `pmbootstrap export` instead of `pmbootstrap flasher export`. With this PR we have an alias, and with `argparse` it is quite easy actually.


drebrez:
why not separating completely the export from the flasher? remove the `pmbootstrap flasher export` and only have the `pmbootstrap export` call?

ollieparanoid:
That sounds reasonable. But in my opinion we should wait with that until #404 (recovery install) is through, or it will break completely (assuming we also move the export code out of the `pmb/flasher` folder).

I would have kept the `flasher export` for compatibility, but since the project is in an early state and nobody seems to miss it, we might as well remove the `flasher export` argument, and only use `export` as @drebrez suggested.

Thanks for the review @PabloCastellano!

PabloCastellano:
#404 is merged already, this PR can continue!

ollieparanoid:
@PabloCastellano: right, I've updated it, ready for the next iteration of reviewing and testing! 👍 
@drebrez: I've moved the export code to pmb/export, and the odin code to pmb/export/odin. Could you also review/test this, and most importantly: What should I put in the copyright line of `odin.py` (because you wrote it all)?

ollieparanoid:
@drebrez: good call, just verified it now and noticed that the log output was in the wrong order. That is fixed now. Also the message for exporting the Android recovery zip (which doesn't have its own parameter anymore, that was from before the PR was merged).

PR #418, opened 2017-08-19
Disable zap on init for new devices

craftyguy:
The device chroot shouldn't exist yet anyways. Also add a 'deviceExists'
variable that can be used to determine if the device entered earlier in
the init has a deviceinfo under aports/device/device-*


pmb/config/init.py
```diff
@@ -72,6 +72,8 @@ def init(args):
     cfg["pmbootstrap"]["device"] = pmb.helpers.cli.ask(args, "Device",
                                                        None, args.device, False, "[a-z0-9]+-[a-z0-9]+")
 
+    deviceExists = os.path.exists(args.aports + "/device/device-" + cfg["pmbootstrap"]["device"] + "/deviceinfo")
```
ollieparanoid: Please use snake case for variables, instead of camel case (for consistency with the rest of the code).

craftyguy:
Done! I meant to do snake case but I'm going mad switching between C/C++ and python 'standards'!

PR #423, opened 2017-08-20
if there is no resolv.conf, create an empty file.

indefini:


craftyguy:
what purpose does this serve? I think most (all?) dhcp clients will create this file when they run, and most (all?) GUI network manager things will create this if you're configuring static network assignment. Is there something that fails when /etc/resolv.conf doesn't exist?

indefini:
yes "./pmbootstrap.py install" fails because the "cp" command fails.
If the file is created with dhcp then we can remove the touch command.

PR #430, opened 2017-08-20
Adjust luksFormat options to reduce time to open luks device on boot

craftyguy:
As pointed out by @ollieparanoid in Matrix, creating the luks root device on a super fast modern PC results in a passphrase the requires many many hash iterations to verify and on the much slower target devices for pmos this results in a VERY long time on some devices to execute a luksOpen.

This PR introduces a new option to call cryptsetup with:

`--iter-time 200` - Default in cryptsetup 1.7 is 2000ms. This results in a very large number of iterations on a modern desktop PC, and takes a long time for older devices to generate. The number of iterations are ["not very security relevant" (FAQ section 3.4)](https://gitlab.com/cryptsetup/cryptsetup/wikis/FrequentlyAskedQuestions#5-security-aspects). This value could be lowered a bit further since they seem to recommend at least 10k iterations and this setting results in nearly 20k iterations on a Core i7 CPU.

Here's some datapoints using the new defaults this introduces on a Nokia N900. Previous open time with previous defaults (iter-time=2000ms & hash=sha256) was around 20-30 seconds.

```
/ # time cryptsetup luksOpen /dev/dm-1 root  
Enter passphrase for /dev/dm-1:              
real    0m 6.02s                             
user    0m 5.00s                             
sys     0m 0.07s                             
/ # cryptsetup luksDump /dev/dm-1            
LUKS header information for /dev/dm-1        

Version:        1                            
Cipher name:    aes                          
Cipher mode:    cbc-plain64                  
Hash spec:      sha1                         
Payload offset: 4096                         
MK bits:        256                          
MK digest:      4d d5 d8 49 e1 c7 b5 d5 39 7f bc 67 1b ab e0 5b c6 2a 09 83               
MK salt:        9c 16 c5 ab 55 82 50 d3 b9 ac 3e 38 ff e4 08 6c                           
                4a 38 20 cb 3e b7 9c 3d 0e 82 bc 55 13 b2 f5 91                           
MK iterations:  19475                        
UUID:           3b5c0365-46d0-4642-8748-62c3f1e7297b                                      

Key Slot 0: ENABLED                          
        Iterations:             155386       
        Salt:                   f6 ca 67 7e 66 8f d1 39 6a 2f 00 ce 4c e9 35 e5           
                                d6 a9 a8 76 b2 81 99 96 ef b1 55 91 24 53 4a e2           
        Key material offset:    8            
        AF stripes:             4000         
Key Slot 1: DISABLED                         
Key Slot 2: DISABLED                         
Key Slot 3: DISABLED                         
Key Slot 4: DISABLED                         
Key Slot 5: DISABLED                         
Key Slot 6: DISABLED                         
Key Slot 7: DISABLED                         
/ # cryptsetup benchmark                     
# Tests are approximate using memory only (no storage IO).                                
PBKDF2-sha1        26640 iterations per second for 256-bit key                            
PBKDF2-sha256      49423 iterations per second for 256-bit key                            
PBKDF2-sha512      11307 iterations per second for 256-bit key                            
PBKDF2-ripemd160   32443 iterations per second for 256-bit key                            
PBKDF2-whirlpool    7803 iterations per second for 256-bit key                            
Required kernel crypto interface not available.                                           
Ensure you have algif_skcipher kernel module loaded. 
```


MartijnBraam:
I don't think we should hardcode a time value in pmbootstrap since the time depends on the host machine building it. I think the best solution is running `cryptsetup benchmark` on the device as part of the porting process and add the sha256 time to the deviceinfo, then you can calculate the iterations necessary to unlock the luks volume in x amount of seconds on the phone hardware.

craftyguy:
that's a great suggestion! I'll add that tomorrow

craftyguy:
Oh, one other bit, the results from cryptsetup benchmark aren't super accurate.. it took 6 seconds to open when 19k sha1 iterations were used, even though benchmark indicates that thin should have been much faster.

ollieparanoid:
I don't really like the idea of having the iteration count depending on the installation PC at all, so I recommend specifying the absolute iteration count instead of a time value. Seems like we [can do that](https://superuser.com/a/766613).


My suggestion:
* Always specifying `-i` in `cryptsetup luksFormat`
* Set a sane default value, so we don't have an extra step in the porting guide 
just to benchmark a good value
* Make it possible to override that in the deviceinfo for older phones (in case it is slow)
* Make it possible to override that as command-line argument

craftyguy:
@ollieparanoid

I think stackowerflow is lying to you. The cryptsetup manpage suggests -i is synonymous with --iter-time:

```
--iter-time, -i
The number of milliseconds to spend with PBKDF2 password processing. This option is only relevant to the LUKS operations as luksFormat or luksAddKey.
```

There doesn't seem to be a way to do what you are suggesting, at least not documented in their docs. Maybe there was at one point?

craftyguy:
I found this feature request: https://gitlab.com/cryptsetup/cryptsetup/issues/280

Maybe it's best to wait for that to be implemented before moving forward with this PR. (If y'all agree) Sorry for the premature PR in this case!

ollieparanoid:
Damn, I should have double checked that. Thanks for finding that link, that feature request looks exactly like what we need, and solving it upstream seems to be the best way indeed! You have already offered to do it there, which is very nice of you, so thanks! \o/

craftyguy:
As discussed in chat today, using this PR as is is a reasonable workaround since upstream cryptsetup doesn't seem interested in adding a way to limit hashes.

ollieparanoid:
The reson why the issues on gitlab get ignored is stated in the [FAQ](https://gitlab.com/cryptsetup/cryptsetup/wikis/FrequentlyAskedQuestions):
> 1.6 Where is the project website?
> 
> There is the project website at https://gitlab.com/cryptsetup/cryptsetup/ 
> Please do not post questions there, nobody will read them. Use the mailing-list
> instead.


So I've written to the mailing list, with you added in CC. The e-mail should pop-up here soon: http://www.saout.de/pipermail/dm-crypt/

In the meantime, we can use this workaround. Thanks @craftyguy!

craftyguy:
Oh wow, guess I should have started with the FAQ :(

Great find though, and thank you for sending the note to their mailing list!

PR #435, opened 2017-08-21
Better Mozilla Flame device support

yuvadm:
Did some cleanup on the Flame, especially with installing device-specific firmware. Please take a look and let me know if this PR makes sense.

Firmware blobs were copied to an external repository and are loaded from there. This is an alternative to actually linking the blobs from the special partitions (which I mount in fstab anyway, for now).

* No need to append DTB to kernel
* Remove kernel module build for now
* Add device-specific firmware package
* Add special partitions to fstab
* Fix typo in screen width


aports/device/linux-mozilla-flame/APKBUILD
```diff
@@ -20,7 +20,7 @@ case $pkgver in
 	*.*.*)  _kernver=${pkgver%.*};;
 	*.*) _kernver=$pkgver;;
 esac
-pkgrel=4
+pkgrel=2
```
PabloCastellano: Typo? Shouldn't it be 5?

aports/device/linux-mozilla-flame/APKBUILD
```diff
@@ -95,31 +92,16 @@ build() {
...
-		KBUILD_BUILD_VERSION="$((pkgrel + 1 ))-Alpine" CONFIG_NO_ERROR_ON_MISMATCH=y \
-		INSTALL_MOD_PATH="${pkgdir}" firmware_install
-
-	rm "${pkgdir}/lib/modules/${pkgver}/build"
-	install -Dm644 "${srcdir}"/build/Module.symvers "${pkgdir}/lib/modules/${pkgver}/build/Module.symvers"
-	install -Dm644 "${srcdir}"/build/System.map "${pkgdir}/lib/modules/${pkgver}/build/System.map"
-	install -Dm644 "${srcdir}"/build/modules.builtin "${pkgdir}/lib/modules/${pkgver}/build/modules.builtin"
-	install -Dm644 "${srcdir}"/build/modules.order "${pkgdir}/lib/modules/${pkgver}/build/modules.order"
-	cd ../..
+  make ARCH="$_carch" CC="${CC:-gcc}" \
```
PabloCastellano: indentation

aports/firmware/firmware-mozilla-flame/APKBUILD
```diff
@@ -0,0 +1,28 @@
+pkgname=firmware-mozilla-flame
+pkgver=1
+pkgrel=1
+pkgdesc="Mozilla Flame Firmware"
+url="https://github.com/yuvadm/mozilla-flame-firmware"
+arch="armhf"
+license="MIT"
```
PabloCastellano: Should be `proprietary`, right?

aports/firmware/firmware-mozilla-flame/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+depends=""
+makedepends=""
+install=""
+subpackages=""
+_hash=503a4052a8bfdc3b4c06533f2d08fe4d6a923842
+source=$pkgname-$_hash.zip::$url"/archive/${_hash}.zip"
+options="!check !strip"
+
+build() {
```
PabloCastellano: I've tested that you can just remove `build()` and nothing bad happens. I don't know why a lot of aports are defining it as `return 0` instead of just removing it :thinking:

aports/device/linux-mozilla-flame/APKBUILD
yuvadm: I made a backwards incompatible change to prevent pkgrel overflow :) It's weird incrementing pkgrel on every single build, so I'm trying to keep it consistent and increment on every release. If this is a problem I'll revert back to `5`.

ollieparanoid: Please only increment this number.

ollieparanoid:
@PabloCastellano did a good review, nothing to add there :)
Just one question: What's the advantage of mounting the folders?

yuvadm:
@ollieparanoid I'm still not sure whether there are any hardcoded paths in any drivers that expect those partitions to be mounted. I'll remove them once I'm positive they're not needed anymore.

yuvadm:
Fixed and updated

yuvadm:
@ollieparanoid Thanks for the indentation fix, indeed having proper indentation rules for text editors would be a good idea.

PR #440, opened 2017-08-21
Plasma Shell

PureTryOut:
It now runs!

[Plasma shell running on postmarketOS](https://zippy.gfycat.com/LikablePoliteHog.webm)

As you can see in the video, there is still some work to do, but we're getting really close!

Should be as easy to install as selecting `plasma-mobile` while running `./pmbootstrap.py init`.

If your device doesn't support OpenGL yet, it's recommend to create the file `~/.config/kdeglobals` and put the following content into it:

```
[QtQuickRendererSettings]
SceneGraphBackend=software
```

Please refer to the [Wiki page](https://github.com/postmarketOS/pmbootstrap/wiki/Plasma-Mobile) for debugging Plasma Mobile.

- [x] Get it running on Qemu without OpenGL warnings
- [x] Get the background to be controlled by something (Dolphin?)
- [ ] Have it run on any other architecture than x86_64


PureTryOut:
The latter. I'll remove the label once it's ready for review.

z3ntu:
> Furthermore, it is now possible to use ConsoleKit2 instead of logind for setting up the Wayland session, extending the number of supported platforms.
> 
> \- Plasma 5.11 (actually 5.10.95) release notes

Does that mean some patch isn't required anymore?

PureTryOut:
Once Plasma 5.11 is actually released, yes it does. It's still in beta though, so we'll stay on 5.10.4 for the time being (although I *could* update to 5.10.5).

PureTryOut:
So, right now we got a pretty good shell running.

Plasma Mobile shell

However, this is *only* on x86_64. Even though the original goals I've set for this PR to get merged are basically reached, I feel like this is not enough to merge it to master. So I've updated the goals to include getting it running on other architectures. Just one (armhf probably) would be enough for now.

ollieparanoid:
Some notes about related KDE work in the Alpine environment:
* [Adélie Linux](https://adelielinux.org/) (Alpine derivative by @awilfox) is packaging the LTS version of KDE and will sooner or later upstream it directly into Alpine.
* Plasma mobile requires the latest stable or git version of "plasma workspace and friends" in its current state
* We could in theory build directly on the LTS KDE framework packages though (benefits: possibly better stability, less maintenance effort for us, we can concentrate more on getting plasma mobile ready to use)

Thanks to @bhush9 from plasma mobile for confirming that information and to @awilfox for providing this opportunity!

@PureTryOut: In my opinion, that makes a lot of sense. But only if it actually saves work. If it would be more work to rebase on the LTS stuff right now, then I'd recommend that we wait until the next KDE framework update comes around (which forces to change all the aports anyway), and instead of upgrading the KDE framework, actually downgrade to the LTS version that @awilfox maintains.

PureTryOut:
From what I understand, there is no such thing as a “LTS KDE Framework” right? Only Plasma has a LTS version as far as I know. Because of this I already wanted to go with whatever Adélie was going to upstream, and just keep the Plasma packages updated ourselves till LTS 5.12 comes around.

ollieparanoid:
Not sure if you've seen this in `#postmarketOS`:
> @PureTryOut: btw, what do you think about merging plasma-mobile into master as soon as the binary repo PR is merged? then we'll have official builds of the packages and I'll try to do them for x86_64, aarch64, armhf and pretty much as soon as they come out. also I'm noticing the build errors more directly if I have to build them myself :p 
> I think it makes testing easier, because then everything is configured to download the binary packages without extra steps

PureTryOut:
Fine to me I guess. Should make it easier for everyone to test.

I'll make a new issue for the LibreSSL error on armhf I getting somewhere this weekend.

awilfox:
The KDE Frameworks are basically "always LTS".  They have ABI guarantees and each monthly release is basically considered a minor/bugfix release.

ollieparanoid:
Alright, the binary repo has been merged. I think there are two things left to do, before merging this:
- [ ] Change the `pkgdesc` of `postmarketos-ui-plasma-mobile` to add a warning that compiling takes forever (which we remove once it is in the binary repo)
- [ ] Don't bundle the source .tar.gz of "plasma-mobile" and "peruse" in the aports folder.

The first one is easy, but the second one is not. Some background:
* We need to build from git commits, there are no official releases yet
* Phabricator (where KDE hosts its git repositories) can't produce stable tarballs ([repo](https://phabricator.kde.org/source/plasma-mobile/))
* GitHub produces stable tarballs, but plasma's automatic github mirroring is currently outdated ([repo](https://github.com/KDE/plasma-mobile))

Solutions I can think of:
1. Get deterministic archive support into phabricator (only real fix, but too time consuming for us to be feasible)
2. Kindly ask @bhush9 if someone could make the mirror sync again
3. Set up a GitHub mirror in a new organization (so we don't clutter our real repositories list) where everyone who works on plasma mobile in postmarketOS gets write access so we can pull the sources from Phabricator. We wouldn't change the source, patches go in our `aports` folder if necessary.
4. Download tarballs from phabricator (that is possible, right?) and run `strip-nondeterminism` before making the checksums (we can overwrite `fetch()` and similar functions within `APKBUILDs`) (This will take some coding effort to get right.)

My opinion: Don't block ourselves with this, so I would set up solution 3 tomorrow unless KDE's repo is syncing properly again already (if it is, we can simply remove our own repo again).

What do you think @PureTryOut and everyone who is interested?

PureTryOut:
@bhush9 has already asked the KDE sysadmin before to fix mirroring. I'm not sure whatever came from that, but that would be my preferred solution. If we don't want to wait for that, I'd go for option 3 as well (mirror the code ourselves temporarily).

ollieparanoid:
@PureTryOut: Here we go. You should have write access in these repos (might need to click a confirmation mail link or something): https://github.com/pmsourcedump

One more thing, right now the `pkgver` of `plasma-mobile` is `5845.8005327e`. How about using `0_git20171120` (date of the commit we're packaging) instead, so it can be updated properly once a release (higher than 0) comes out?

ollieparanoid:
I've made the following changes to get it ready for merging:
* Added a compiling warning to the `pkgdesc` of `postmarketos-ui-plasma-mobile`
* `plasma-mobile` downloads a tarball from our workaround repo now (as discussed above)
* Patched `peruse` to download the tarball as well, but then decided to move it into its own branch for now (see #926)

While doing that, I did some test compiling. And yeah I think we need to fix how pmbootstrap handles dependencies first (in other words, this is a **blocker** now: https://github.com/postmarketOS/pmbootstrap/issues/824#issuecomment-346100442), I'll try to get this done the next days.

Thanks for all your work on the plasma packaging @PureTryOut and in case you don't like the changes I made, let me know!

PureTryOut:
After a discussion with @bhush9 yesterday, `plasma-mobile` was deemed unnecessary. I had it removed it locally already, just hadn't pushed it yet.

Other changes seem fine to me, although I'd say we drop `peruse` entirely for now. I guess we should fix #824, and then we can merge it.

ollieparanoid:
Ah, that's what he meant with "drop it"... I somehow thought that meant: drop the source to that other repo :p Is everything `plasma-mobile` related now part of the official packages?

PureTryOut:
Yup.

ollieparanoid:
Sorry, I've mixed this up in the chat: the `_git`... thing goes in the `pkgver`, not in the `pkgrel`:
https://wiki.postmarketos.org/wiki/Create_a_package#Packaging_git_commits

When that is resolved, I think it's good for a merge 👍

PureTryOut:
Fixed 👍

Issue #383, opened 2017-08-14
Split pmbootstrap and aports in two repositories


ollieparanoid:
There are different release cycles, that make sense for `aports` and `pmbootstrap`.
* `aports`: always up-to-date
  * we could provide different branches like `edge`, `3.6`, `3.5` which follow Alpine's branching model
  * but these branches need to be always up to date
* for `pmbootstrap` we could have a stable release every now and then
  * that way we could package it for `pip` and for some Linux distributions (`AUR`, Gentoo overlay, ...)
  * it would be a better experience for new users (install `pmbootstrap` through their distribution/pip, more stability)

A great benefit, that @MartijnBraam pointed out:
> if you ship pmbootstrap seperately, then you can make `pmbootstrap init` manage the aports git repo and make it selectable between "edge", which is the latest `master` or a specific tag

yuvadm:
I'd be interested to start working on this packaging task. Would love to hear your thoughts on how we should handle this. Specifically the biggest pain point is the assumption that `aports` is under the working directory.

In this future split, where would be the right place to have `aports`? Within `.var/local/pmbootstrap` perhaps? Should `pmbootstrap` handle the loading of `aports`?

ollieparanoid:
Tough question and I'm almost out of postmarketOS time for today :/

I would like pmbootstrap to assist the user in getting the `aports` repository, so that no manual step is involved. It would need to happen before the first question in `init`, because that shows the list of devices, which we can only do when the `aports` folder is available.

The work path seems to be suitable for aports, so we would need to ask for that first.

Here's a rough draft from a user's point of view.
* pmbootstrap init
* ask for work folder
* aports not found (specify with `-p`)! do you wish to download it to: $work/aports? [y]
  * pulls it with git
  * display instructions how to update them (make that convenient with a pmbootstrap shortcut?)
* which device do you want to port?
* ...

kskarthik:
Label this as Long term priority until we get plasma working on already green ports :) Then we are good to go

ollieparanoid:
It has a `future` label already :) But if @yuvadm or someone else wants to work at this right now, why not? It is a blocker for getting `pmbootstrap` into `pypi` #327 after all and making the development more robust #382.

marmistrz:
How about a very simple workflow: if the aports directory is missing, we just show the `git clone ...` command to the user. 

This looks like a very small change which I could do myself, we'd just need to decide the default path for aports.

[...]

Issue #327, opened 2017-08-03
Add setup.py (setuptools)


PabloCastellano:
With setuptools we can have pmbootstrap available in PyPi and have pmbootstrap installed to $PATH easily (I'd also like to have a `pmb` alias)

ollieparanoid:
I think before publishing to pypi, we should split `pmbootstrap` and the `aports` into different repositories. Or does anyone see it differently?

PureTryOut:
If not doing that means the `aports` folder ends up in pypi as well, then yes, it should probably be moved first.

MartijnBraam:
hmm it shouldn't end up on pypi if you don't explicitly add the aports folder in the setup.py

PR #439, opened 2017-08-21
Write custom os-release (closes #324)

PabloCastellano:
This is the content of /etc/os-release generated using this branch

```
PRETTY_NAME="postmarketOS 0.2.0"
NAME="postmarketOS"
VERSION_ID="0.2.0"
VERSION="0.2.0-783a0dde"
ID="postmarketos"
ID_LIKE="alpine"
HOME_URL="http://www.postmarketos.org/"
SUPPORT_URL="https://github.com/postmarketOS/"
BUG_REPORT_URL="https://github.com/postmarketOS/"
PMOS_HASH="783a0dde1645b2d6c9ebaa434206291b304c324c"
```

PR #443, opened 2017-08-22
Add initial setup.py (#327)

yuvadm:
Started to prepare an initial `setup.py` according to [Python packaging best practices](https://packaging.python.org/tutorials/distributing-packages/)

- Supports package installation with a `pmbootstrap` entry point executable
- Supports `pytest` testing
- Moved all entry point logic into `pmb` module, had to fix some imports
- Dynamically resolves version from `pmb/config/__init__.py` using regex (see https://packaging.python.org/guides/single-sourcing-package-version/#single-sourcing-the-version)

Running `pmbootstrap` after installing it system-wide (or into a virtualenv) currently **requires** the usage of the `-p aports` flag. But this discussion is probably more relevant in #383.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python3
+
+from setuptools import setup, find_packages
+from setuptools.command.test import test as TestCommand
+
+from codecs import open
+from os import path
+
+here = path.abspath(path.dirname(__file__))
+
+with open(path.join(here, 'README.md'), encoding='utf-8') as f:
+    long_description = f.read()
+
+
+class PyTest(TestCommand):
+    user_options = [('pytest-args=', 'a', 'Arguments to pass to pytest')]
+
+    def initialize_options(self):
+        TestCommand.initialize_options(self)
+        self.pytest_args = ''
+
+    def run_tests(self):
+        import shlex
+        import pytest
+        errno = pytest.main(shlex.split(self.pytest_args))
+        sys.exit(errno)
+
+setup(
+    name='pmbootstrap',
+    version='0.0.1',
+    description='A sopisticated chroot / build / flash tool to develop and install postmarketOS',
+    long_description=long_description,
+    author='postmarketOS Developers',
+    author_email='info@postmarketos.org',
+    url='https://www.postmarketos.org',
+    license='GPLv3',
+    python_requires='>=3.4',
+    classifiers=[
+        'Development Status :: 3 - Alpha',
+        'Intended Audience :: Developers',
+        'License :: OSI Approved ::GNU General Public License v3 (GPLv3)',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.4',
+        'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
+    ],
+    keywords='postmarketos pmbootstrap',
+    packages=find_packages(exclude=['aports', 'build', 'dist', 'keys', 'tests']),
+    tests_require=['pytest'],
+    cmdclass = {'test': PyTest},
+    entry_points={
+        'console_scripts': [
+            'pmbootstrap=pmb:main',
+        ],
+    },
+)
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setup(
     classifiers=[
         'Development Status :: 3 - Alpha',
         'Intended Audience :: Developers',
-        'License :: OSI Approved ::GNU General Public License v3 (GPLv3)',
+        'License :: OSI Approved :: GNU General Public License v3 (GPLv3)',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ setup(
     keywords='postmarketos pmbootstrap',
     packages=find_packages(exclude=['aports', 'keys', 'test']),
     tests_require=['pytest'],
-    cmdclass = {'test': PyTest},
+    cmdclass={'test': PyTest},
     entry_points={
         'console_scripts': [
             'pmbootstrap=pmb:main',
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def run_tests(self):
+        import shlex
+        import pytest
+        errno = pytest.main(shlex.split(self.pytest_args))
+        sys.exit(errno)
+
+setup(
+    name='pmbootstrap',
+    version='0.0.1',
```
MartijnBraam: Aren't we actually on 0.0.2 already since we've tagged 0.0.1 just before the aports subfolder migration?

setup.py
```diff
@@ -0,0 +1,56 @@
...
+    description='A sopisticated chroot / build / flash tool to develop and install postmarketOS',
+    long_description=long_description,
+    author='postmarketOS Developers',
+    author_email='info@postmarketos.org',
```
MartijnBraam: Does this e-mail address work?

setup.py
```diff
@@ -0,0 +1,56 @@
...
+    url='https://www.postmarketos.org',
```
MartijnBraam: Do we still need to prefix urls with `www.`?

yuvadm: Naked root currently redirects to `www.` so I kept the canonical form.

setup.py
yuvadm: Probably not, placeholder.

setup.py
yuvadm: We only have on pre-release 0.1.0, so I can change to `0.1.0rc1`. Or can we jump to `0.2.0`?

PabloCastellano: Should use `pmb.config.version` for coherency.

setup.py
```diff
@@ -0,0 +1,56 @@
...
+    def run_tests(self):
+        import shlex
+        import pytest
+        errno = pytest.main(shlex.split(self.pytest_args))
+        sys.exit(errno)
+
+setup(
+    name='pmbootstrap',
+    version='0.0.1',
+    description='A sopisticated chroot / build / flash tool to develop and install postmarketOS',
```
PabloCastellano: Typo: sophisticated

setup.py
```diff
@@ -0,0 +1,56 @@
...
+    classifiers=[
+        'Development Status :: 3 - Alpha',
+        'Intended Audience :: Developers',
+        'License :: OSI Approved ::GNU General Public License v3 (GPLv3)',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.4',
+        'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
+    ],
+    keywords='postmarketos pmbootstrap',
```
PabloCastellano: Keywords expects a list like ['postmarketos', 'pmbootstrap']

yuvadm: Not according to the docs https://packaging.python.org/tutorials/distributing-packages/#keywords and the sample project

PabloCastellano: Oh, right! Either something has recently changed or I have been doing it wrong for a while :D

setup.py
```diff
@@ -0,0 +1,65 @@
...
+        import shlex
+        import pytest
+        errno = pytest.main(shlex.split(self.pytest_args))
+        sys.exit(errno)
+
+
+here = path.abspath(path.dirname(__file__))
+_version_re = re.compile(r'version\s+=\s+(.*)')
+
+with open(path.join(here, 'pmb/config/__init__.py'), 'rb') as f:
```
PabloCastellano: Since \_\_init\_\_.py is a python file, you don't need to parse it. Simply `import pmb.config` and use `pmb.config.version` as in https://github.com/postmarketOS/pmbootstrap/pull/439

setup.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+
+import re
```
PabloCastellano: The tests are complaining because `sys` is not imported

setup.py
yuvadm: Take a look at https://packaging.python.org/guides/single-sourcing-package-version/#single-sourcing-the-version This method might a problem if you don't take good care of which dependencies you use. Notice that almost no python packages actually import their own code in setup.py

setup.py
yuvadm: Yeam seems weird for me as well :)

setup.py
yuvadm: Fixed

yuvadm:
@ollieparanoid Since you and @MartijnBraam are expected to be the only developers with access to pushing releases to Pypi, would you like me to just put your name and e-mail there? I've seen this in other projects that don't a proper foundation behind the development.

I'll also add an `aports` check, thanks for the suggestion!

yuvadm:
@ollieparanoid How would you expect the `aports` resolution to happen? Right now if I install `pmbootstrap` to a virtualenv, the expected `aports` path is e.g. `/home/yuval/.virtualenvs/pmb/lib/python3.6/site-packages/pmbootstrap-0.2.0-py3.6.egg/aports/device/device-mozilla-flame/deviceinfo`.

Maybe we need some better behavior here? Or maybe we can track this in #383?

ollieparanoid:
(The `aports` check is not pushed to this branch yet, I think)

I'd be fine if you put my e-mail address there for now, I think that is better than a non-working address.
But please don't name me as author, I am not the only one working on it, we have a great community now!

Regarding pushing to `pypi`, I recommend we do this after discussing making the `aports` userfriendly with `pypi` in #383 and implementing that in a follow-up PR.

yuvadm:
Of course, we're definitely not ready for Pypi yet, #383 is blocking that.

What I meant regarding the `aports` problem is that we solve it in a more holistic way that should be discussed in the context of #383. Do you have a better suggestion re: my previous comment?

ollieparanoid:
@yuvadm: I don't see how we can provide a better default for the `aports` folder if it was installed this way until we resolve #383. So I'd rather keep this broken path (along with the check, that tells the user to use `-p`) for now, than adding code to figure out some useful default path, which we will need to discard after implementing #383 anyway.

yuvadm:
Added a missing aports check as requested, sorry for the delay on this small fix.

PR #448, opened 2017-08-23
[WIP] Initial work on sony aries

kskarthik:


ollieparanoid:
Great work!

Currently we have the policy to put devices, where graphical output does not work, into an extra device branch.

However, it turned out, that this will have a big rebasing overhead for all changes, and these extra packages do not hurt in my opinion. So I would suggest we directly integrate this into master. Thoughts on that, everyone?

drebrez:
@ollieparanoid maybe we can change that policy to only put devices in master with a working graphical output, since we also have to init option to not install any UI.
A suggested policy for devices in master is that at least pmOS boot and ssh (usb network) works?

kskarthik:
@drebrez kenel compile works with telnet. No GUI yet. 

@ollieparanoid in that case, I suggest the stable device ports to `aports/device/stable` & WIP one's to `aports/device/wip` 

But again, this may need code level changes.

kskarthik:
@drebrez kenel compile works with telnet. No GUI yet. 

@ollieparanoid in that case, I suggest the stable device ports to `aports/device/stable` & WIP one's to `aports/device/wip` 

But again, this may need code level changes.

kskarthik:
B2G project used to maintain a separate repo for device manifests. There is a pmbootstrap type program but we can edit the repo links from where it should clone the device manifest. Its better we rip off the entire device part to a new repo & maintain it. Again, this also needs  relevant code changes.

kskarthik:
pmbootstrap is ongoing heavy dev. Its not ideal to keep device configs inside it. 

Also, we also need to relocate the wiki. We talked about this on IRC today.

ollieparanoid:
@kskarthik: You're completely right about the wiki and about having device configs (or rather the whole aports folder) not in the tree. See #452 and #383 respectively. I'd rather not implement support for sub-subfolders in `aports` right now, as this takes development time again and we just changed the directory structure.

@drebrez wrote:
> maybe we can change that policy to only put devices in master with a working graphical output, since we also have to init option to not install any UI.
> A suggested policy for devices in master is that at least pmOS boot and ssh (usb network) works?

That makes sense to me.
I'll throw this PR in the channel to possibly get a few more opinions, and if no one has anything against it, I would merge this to master.

ollieparanoid:
No one complained, so I merged it. Thanks for your work on this @kskarthik!

kskarthik:
☺

Issue #444, opened 2017-08-22
RX-51. Install to sdcard fails


zenety:
Would love to bring back my N900 a bit back to life. I ran the pmbootstrap init for the rx51. Then ran the pmbootstrap install --sdcard /dev/sdb (in my case).

It errors out on.
`
pmbootstrap.py install --sdcard /dev/sdb[00:47:24] *** (1/5) PREPARE NATIVE CHROOT ***
[00:47:28] *** (2/5) CREATE DEVICE ROOTFS (nokia-rx51) ***
[00:47:28] Register qemu binfmt (arm)
[00:47:31] (native) build x86_64/binutils-armhf-2.28-r2.apk
[00:48:49] NOTE: The failed command's output is above the ^^^ line in the logfile: /home/joost/.local/var/pmbootstrap/log.txt
[00:48:49] ERROR: Command failed: (native) % cd /home/user/build && su user -c 'CARCH=x86_64 abuild -d'
[00:48:49] Run 'pmbootstrap log' for details.
[00:48:49] See also: <https://postmarketos.org/troubleshooting>
`

The pmbootstrap log option gives me this:

```
mv -f .deps/tilegx.Tpo .deps/tilegx.Po
mv -f .deps/mips.Tpo .deps/mips.Po
make[4]: Leaving directory '/home/user/build/src/binutils-2.28/gold'
make[3]: *** [Makefile:941: all-recursive] Error 1
make[3]: Leaving directory '/home/user/build/src/binutils-2.28/gold'
make[2]: *** [Makefile:692: all] Error 2
make[2]: Leaving directory '/home/user/build/src/binutils-2.28/gold'
make[1]: *** [Makefile:6080: all-gold] Error 2
make[1]: Leaving directory '/home/user/build/src/binutils-2.28'
make: *** [Makefile:851: all] Error 2
>>> ERROR: binutils-armhf: all failed
(002898) [00:48:49] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(002898) [00:48:49] NOTE: The failed command's output is above the ^^^ line in the logfile: /home/joost/.local/var/pmbootstrap/log.txt
(002898) [00:48:49] ERROR: Command failed: (native) % cd /home/user/build && su user -c 'CARCH=x86_64 abuild -d'
(002898) [00:48:49] Run 'pmbootstrap log' for details.
[...]

PR #454, opened 2017-08-23
Added memory requirements to README

MartijnBraam:
This adds the minimal amount of memory to the requirements section of the readme.


craftyguy:
Maybe it should say "2GB of system RAM" since "memory" can be a bit ambiguous (e.g. it can sometimes refer to "flash memory")

craftyguy:
Maybe it should say "2GB of system RAM" since "memory" can be a bit ambiguous (e.g. it can sometimes refer to "flash memory")

ollieparanoid:
I see it like @craftyguy and edited the text accordingly. Please note that in case of #444 one GB for a full Ubuntu (which eats up several hundreds MB of RAM alone) + compiling in `pmbootstrap` was not enough, and that it was "enough" with 2 GB. So the actual value may be much lower, but recommending 2 GB seems like a good idea to me (added the "recommended" word). Thanks for making the PR @MartijnBraam!

Issue #450, opened 2017-08-23
Add a -y (--asume-yes) parameter to automatically answer "yes" to questions 


PabloCastellano:
Actually I miss a `-y` argument that allows me to run `pmbootstrap install` automatically accepting the warning that my system partition is at least XXX MB and I want to include my ssh key in the device.

Same happens when I run `pmbootstrap zap` and I want to delete all my chroots but I have to answer 'y' several times.

ollieparanoid:
Workaround: `yes | pmbootstrap`. But I agree, that would be nice to have!

craftyguy:
> it will also set empty passwords

This is probably obvious, but I want to ask so it's explicitly part of the 'requirements' for this request:

Should `-y` for `install` still prompt for passwords? Or is this request to implement a mode for `install` and `zap` such that they are completely hands-off?

PabloCastellano:
My idea is that `-y` automatically answers yes to yes/no questions. If pmbootstrap asks for a password, then the user still has to enter it. What do you think?

MartijnBraam:
I would `-y` to only answer yes to all answers, for full automated running you would have a `-q` option or something

craftyguy:
Ok, I agree with both of y'all, just want to make sure I understood
the requirements :)

ollieparanoid:
FYI: we already have `-q`, which silences the output for now. I think it makes sense to add a different switch to skip passwords (in case you want to have the output, but want to skip passwords).

craftyguy:
I'll take a shot at implementing this, since I'm now annoyed that I have to spam 'y' :)

PabloCastellano:
@craftyguy Great that you got motivated about this ux feature! :)

I've seen your current PR and looks good. But here's a more generic approach so that we can make this flag more generic and valid for more pmbootstrap commands. Tell me what you think and give it a try if you like it:

- Check `args.assume_yes` inside pmb.helpers.cli.confirm`
- Add the `-y` argument to the main parser (I mean, in the same level as `--version`)

This way `-y` would also work for:
- initfs: confirm overwriting extract folder
- zap: confirm removing chroot
- init: confirm use timebased builds and zap existing chroots

craftyguy:
Oh, that's a great idea. I hadn't thought about other commands. I will modify my PR to include this functionality if we are at a consensus on this

craftyguy:
@PabloCastellano 

[...]

PR #467, opened 2017-08-25
Add -y/--assume-yes option to install (#450)

craftyguy:


pmb/helpers/cli.py
```diff
@@ -62,6 +62,8 @@ def confirm(args, question="Continue?", default=False):
     Convenience wrapper around ask for simple yes-no questions with validation.
     :returns: True for "y", False for "n"
     """
+    if (args.assume_yes):
+        return True
```
ollieparanoid: How about printing the question anyway?
```python
logging.info(question + " (y/n) [" + default_str + "]: y")
```
(`default_str` needs to be moved up for that.)

craftyguy: Does it matter? It would just be informative I guess, but there's nothing the user could do to prevent the action from being taken at that point. I guess I don't see any reason to do that, but I'm totally open to it if you think it's useful in some way.

ollieparanoid: Yeah for information, so you see more of what it is doing.
When running `zap` it could be that it zaps no chroots or 5 of them, and the only way to see it (without `pmbootstrap log`) is the question that asks if you want to remove it.

craftyguy:
@ollieparanoid 

The question is now printed, here's what it looks like for `zap`. I added the `(y/n) [n]: y` to help make it clear that some options default to `n` but here it's being told `y` because this new option (`-y`) was passed to pmbootstrap. 


```
$ ./pmbootstrap.py -y zap                                                                    
[07:58:55] Unregister qemu binfmt (arm)      
[07:58:55] Shutdown complete                 
[07:58:55] Remove /media/data/clayton/src/pmbootstrap-out/chroot_native? (y/n) [n]: y     
[07:58:55] Remove /media/data/clayton/src/pmbootstrap-out/chroot_buildroot_armhf? (y/n) [n]: y                                                                                      
[07:58:55] Remove /media/data/clayton/src/pmbootstrap-out/chroot_rootfs_nokia-rx51? (y/n) [n]: y                                                                                    
[07:58:55] Done

```

PR #474, opened 2017-08-26
Implement zapping of apk packages with ver different than aports (466)

craftyguy:
This adds a new option to `zap`: `-b / --mismatch-bins`

When set, any binary apks in the work directory packages folder will be
removed if their version differs from the version in the relevant
APKBUILD in aports.


pmb/chroot/zap.py
```diff
@@ -44,3 +45,23 @@ def zap(args, confirm=True, packages=False, http=False):
         for match in matches:
             if not confirm or pmb.helpers.cli.confirm(args, "Remove " + match + "?"):
                 pmb.helpers.run.root(args, ["rm", "-rf", match])
+
+    if mismatch_bins:
```
ollieparanoid: I recommend moving the whole block into a new function in the same file.

pmb/chroot/zap.py
```diff
@@ -44,3 +45,23 @@ def zap(args, confirm=True, packages=False, http=False):
...
+                bin_pkgname = bin_apks[bin_apk]["pkgname"]
+                bin_version = bin_apks[bin_apk]["version"]
+                bin_apk_path = arch_pkg_path + "/" + bin_pkgname + "-" + bin_version + ".apk"
+                aport = pmb.build.other.find_aport(args, bin_pkgname)
+                if aport:
+                    apkbuild = pmb.parse.apkbuild(args, aport + "/APKBUILD")
+                    aport_version = apkbuild["pkgver"] + "-r" + apkbuild["pkgrel"]
+                    # Clear out any binary apks that do not match what is in aports
+                    if pmb.parse.version.compare(bin_version, aport_version) and os.path.exists(bin_apk_path):
+                        logging.info("Removing mismatched APK for " + bin_pkgname + "-" + bin_version)
```
ollieparanoid: * For consistency with other messages, I would write `Remove` instead of `Removing`.
* How about specifying the arch, that gets removed? Consider `noarch` packages, which would print the same line three times (x86_64, armhf, aarch64).
* How about specifying the version found in the aports folder?

With the bullet points above I would recommend:
```python
logging.info("Remove mismatched binary package (aports version: " + aport_version + "): " + arch + "/" + bin_pkgname + "-" + bin_version + ".apk")
```

pmb/parse/arguments.py
```diff
@@ -151,6 +151,9 @@ def arguments():
                      " the precious, self-compiled packages")
     zap.add_argument("-hc", "--http", action="store_true", help="also delete http"
                      "cache")
+    zap.add_argument("-b", "--mismatch-bins", action="store_true", help="also delete"
```
ollieparanoid: `-m` would seem more intuitive to me, is there a reason why you chose `-b`? (We could also add both versions I think.)

pmb/chroot/zap.py
```diff
@@ -44,3 +45,23 @@ def zap(args, confirm=True, packages=False, http=False):
...
+    if mismatch_bins:
+        for arch_dir in os.scandir(os.path.realpath(args.work + "/packages/")):
+            arch = arch_dir.name
+            arch_pkg_path = os.path.realpath(args.work) + "/packages/" + arch
+            bin_apks = pmb.parse.apkindex.parse(args, arch_pkg_path + "/APKINDEX.tar.gz")
+            for bin_apk in bin_apks:
+                bin_pkgname = bin_apks[bin_apk]["pkgname"]
+                bin_version = bin_apks[bin_apk]["version"]
+                bin_apk_path = arch_pkg_path + "/" + bin_pkgname + "-" + bin_version + ".apk"
+                aport = pmb.build.other.find_aport(args, bin_pkgname)
```
ollieparanoid: For me it failed here with:
> ERROR: Could not find aport for package: weston-shell-fullscreen

This isn't because of this PR, but because our current APKBUILD parser can't resolve the subpackages correctly from the weston package (if you look at the APKBUILD, they are quite complex!). But it should not crash here in my opinion.
How about:
* `pmb.build.other.find_aport(args, bin_pkgname, False)` setting the `must_exist` to false
* adding these lines (advantage is, that you catch the error and show a message and do not need to intend the following code):
```python
if not aport:
    logging.warning("WARNING: Could not resolve aport for package " + bin_apk_path)
    continue
```

pmb/chroot/zap.py
craftyguy: I will point out that "removing" is correct grammar in this instance, but I've changed it to "removing" nonetheless!

pmb/parse/arguments.py
craftyguy: I used -b for "binaries"! but yes, if the whole theme is "mismatched-binaries" then your suggestion makes sense.

pmb/chroot/zap.py
craftyguy: done!

craftyguy:
@ollieparanoid requested changes have been made

Also, should I remove the "dont_merge_pr" label after making changes requested for a review, or just leave it?

ollieparanoid:
Please remove the label when there's nothing left to discuss/change.

Regarding the incorrect grammar with "remove" instead of "removing": You are right, that is a bit strange. If you prefer it, please open a ticket so we can change it across the whole codebase ("(native) install ..." becomes "(native) installing ...") etc.

I have tested your PR again and everything is working as expected.
Thanks again for this contribution \o/

Issue #453, opened 2017-08-23
Improve Qemu video performance


PureTryOut:
Right now, Qemu doesn't have OpenGL support whatsoever, which causes some issues with Plasma Mobile (and probably future DE's and potential apps developers). I believe we should look into using the QXL video driver for the guest (instead of what we use now, vga I guess?), together with SPICE for viewing the VM.

I use this method right now for a Windows VM I need for work, and although it isn't great, there is basic video acceleration, and it might be enough for app development and Plasma. It also allows for better integration with the host by means of shared clipboard and no locked mouse cursor.

It'll require the user to have installed SPICE on his system, so maybe we can let pmbootstrap check for SPICE first and use vga as a fallback?

[QXL and SPICE](https://www.linux-kvm.org/page/SPICE)

ollieparanoid:
Since this is network based, maybe it makes sense to use the `qemu` from within the chroot then.

Just in theory, I think it would not be impossible to use `spice` from the chroot too (although it needs to talk to X/wayland compositor of the host system). In theory, preloading all the required `.so` files and executing Alpine's `spice` binary from outside the chroot should be enough.

MartijnBraam:
We could also provide a statically linked spice client to execute on the host I guess.

ollieparanoid:
Yes, but then we would need to compile it by ourselves.

PabloCastellano:
I did some research.

Actually `pmbootstrap qemu` runs the following command:

```
qemu-system-x86_64 -kernel /var/pmbootstrap/chroot_rootfs_qemu-amd64/boot/vmlinuz-postmarketos -initrd /var/pmbootstrap/chroot_rootfs_qemu-amd64/boot/initramfs-postmarketos -append "console=tty1 console=ttyS0 PMOS_NO_OUTPUT_REDIRECT" -m 1024 -redir tcp:2222::22 -serial stdio -drive file=/var/pmbootstrap/chroot_native/home/user/rootfs/qemu-amd64.img,format=raw -enable-kvm
```

Following @PureTryOut's linked article, I have tried appending the following arguments to the previous command:

```
-vga qxl -spice port=5900,addr=127.0.0.1,disable-ticketing
```

And I can successfully connect to it using `spicy` (from the `spice-client-gtk` package in Ubuntu)

```
spicy -h 127.0.0.1 -p 5900
```

First I tried without QXL support and weston didn't show up. Instead, the motd message was printed in a loop every few seconds (is it expected?)

[...]

PR #481, opened 2017-08-29
Qemu support for the QXL driver and SPICE

PabloCastellano:
## How to test

Requirements:

* You need qemu installed in your host machine
* You need a spice client installed in your host machine (supported: `spicy` and `virt-viewer`)

```
pmbootstrap config qemu-amd64
pmbootstrap -y install --no-fde
pmbootstrap qemu --spice
```

* Is performance better or worse compared to `pmbootstrap qemu` (without `--spice`)?
* Does the mouse pointer moves more smoothly?

## Summary

Fixes https://github.com/postmarketOS/pmbootstrap/issues/453.

This is a step forward to get better graphics performance in QEMU.

Glossary:
* **QXL** is a paravirtual graphics driver with 2D acceleration support
* The **SPICE** project aims to provide a complete open source solution for remote access to virtual machines in a seamless way

This PR enables QXL driver support in the linux kernel configurations so that we can also use SPICE to connect to the VM if a client is available in the host machine. A new `--no-spice` argument is available to force not using the QXL driver. When using the QXL driver, qemu runs without showing any window and silently listens on port 8077/tcp waiting for a spice client to connect (similar to vnc).

Fun fact: I've seen that is somewhat common to use the port 5900 but this is already used by VNC and would cause a conflict if it is already running in the host machine. Instead, I chose 8077, which doesn't belong to any well-known service and represents 'PM'(bootstrap) in decimal.

At the moment, only `spicy` and `remote-viewer` spice clients are supported but adding support for more clients can be easily done.

For this purpose, I've also added support in pmbootstrap to run processes in _background_. This way we can run qemu in background and then run a spice client (in foreground) to connect to the virtual machine.

References:
* https://www.linux-kvm.org/page/SPICE
* https://wiki.archlinux.org/index.php/QEMU#qxl
* https://wiki.archlinux.org/index.php/QEMU#SPICE


PureTryOut:
So far performance seems to vary per PC, but it isn't really an improvement either way. On my desktop, the mouse lags behind literally a second. On my laptop it isn't so bad, but it's not an improvement from the old setup either. In both setups Kwin still complains about lacking OpenGL support, so I guess there is something going wrong still. Somebody could check with glxinfo I guess.

Maybe look into virtio-gpu after all then?

PabloCastellano:
Yup, seems like we have very different results :disappointed: 

In my case the first thing I noticed using the qxl driver is that the mouse moves smoothly, as opposed to when not using it. We need more testers in order to identity what causes it. Sad news is that OpenGL is still not supported apparently by qxl (confirmed by @PureTryOut on Matrix -- or could it be related to his graphics issues?) so I'll continue the investigation with the `virtio-gpu` driver.

FWIW, in my case I'm running
* qemu 2.5.0
* remote-viewer 1.0.0
* laptop with integrated graphics card using FOSS `i915_bpo` kernel module

@PureTryOut said in Matrix he is running:
* qemu 2.9.0
* tried with remote-viewer and spicy, same results
* amd graphics card with FOSS drivers (he can play games fine, so the drivers are not the problem)

PureTryOut:
Good to note, on both PC's I run Qemu 2.9.0 and used both spicy and remote-viewer. One PC's runs on a RX480 with the FOSS Mesa drivers (the Radeonsi backend) and the other on Intel Integrated Graphics 530, also FOSS Mesa drivers (the i965 backend). The laptop with the integrated graphics actually ran better than the dedicated graphics card.

PureTryOut:
Looking into the issue a bit, I found [this page](http://www.linux-kvm.org/page/HOWTO_VMGL). The first paragraph ends with "Warning: Usermode network stack cripples performance, use Tap for now.".

Seeing I can run `./pmbootstrap.py qemu` without root, I'm guessing we're running in the usermode network stack?

ollieparanoid:
@PabloCastellano: thanks for making this PR and thanks for the detailed description! The 8077 explanation is pretty clever.
So how do we move on in ths PR with the information @PureTryOut provided?
*Adding `dont_merge_pr` label until this is clear, please remove it when it is ready to merge.*

PabloCastellano:
@ollieparanoid I've written my plan in https://github.com/postmarketOS/pmbootstrap/issues/453#issuecomment-326260053

PabloCastellano:
Rebased and made SPICE not run by default. So the functionality is same as before, but you can specify `--spice` to connect to the VM using a SPICE client.

pmb/qemu/run.py
```diff
@@ -141,23 +190,41 @@ def run(args):
...
                  " following services:")
     logging.info("(ssh) ssh -p " + str(args.port) + " user@localhost")
     logging.info("(telnet) telnet localhost " + str(args.port + 1))
     logging.info("(telnet debug) telnet localhost " + str(args.port + 2))
-    pmb.helpers.run.user(args, command)
+
+    if not run_spice:
+        logging.warning("WARNING: Could not find any SPICE client in your PATH"
+                        ", or --spice was not specified, so Qemu will run"
+                        " without some features as 2D acceleration.")
```
ollieparanoid: that sounds funny, how about `" without some features, such as 2D acceleration."`?

pmb/qemu/run.py
```diff
@@ -62,7 +62,49 @@ def which_qemu(args, arch):
...
+def which_spice(args):
+    """
+    Finds some SPICE executable or raises an exception otherwise
+    :returns: tuple (spice_was_found, path_to_spice_executable)
+    """
+    executables = ["remote-viewer", "spicy"]
+    for executable in executables:
+        if shutil.which(executable):
+            return True, executable
+    return False, ""
```
ollieparanoid: What's the advantage of using a tuple here? I would have implemented it like `shutil.which` and returned `None` in case it was not found.

PabloCastellano: You're right. Looks like the result of several refactoring

pmb/qemu/run.py
PabloCastellano: :+1:

ollieparanoid:
First of all, thank you for implementing this feature!

I've tested it now and it works as expected. Also I've changed the logging a bit, so it only prints a *warning*, when SPICE was not found but the user requested it, and a *note* otherwise.

If you agree with this change, I'd like to get this PR merged ASAP.

---

Speaking of merging, you have asked me in the channel whether we could merge this PR without "squash and merge". I see that you have put a lot of effort in the commit messages, and I can assure you, that these will not be lost, even if we did a squash and merge (you probably know that, just making sure). So I've given this some thought and I do not want to step on your toes here, I'm sure we can figure a way out that works for both of us.

As I see it, we have two options now, I'll try to outline the characteristics of both, that I can think of.

### 1. Use "squash and merge"
The big reason to choose this is, that the reviewer usually tests and reviews the last commit in the branch only. After all, that is the commit, that the users will run. Testing everything inbetween would take a lot more time, and I think we both can agree that it is not that useful, and the time could be spent better elsewhere in the project. If we would just "squash and merge" your PR, we would get:
* Consistency with all other PRs
* Code is git-bisectable (you suggested that it might be possible to only bisect on master, and not the other branch - but git is made for throwing away branches, and that is what we do, otherwise we would have about hundred branches by now for all the PRs - correct me if I'm wrong please!)
* Less noise in the git log (also useful for git blame)

### 2. Cherry pick commits
This seems to be the cleanest variant when not squashing the whole PR. The first three commits could be cherry picked, and we could smash the other two "minor correction" commits into one, and give it a commit message that contains the PR ID (so the history and this conversation can be looked up on GitHub), as well as a meaningful message that it relates to this QEMU stuff.

This is more work for the person that merges it, because it involves running a handful of git commands, and again I think this time might be more useful spent elsewhere in the project.

### What now?
My recommendation is, that you choose between 1. and 2. for this PR, and that for the future if you feel like multiple commits should be merged without being squashed, please make multiple PRs for them.

For example: You could have made a single PR, that changes the kernel configs and deviceinfo files (your 1st commit) in parallel with a second PR, that enables background tasks (2nd commit).

You would need to wait until both of them were merged for the final commit, but because both of them are smaller, they are easier to review. And especially when you note somewhere, that you have another PR coming, that depends on the other two, I would do my best to prioritize getting those two PRs in, so you are not blocked.

The best part is, that corrections would end up in the right PR without an extra "make requested changes" commit afterwards, that we have to deal with now if we don't squash and merge.

---

The wall of text means that I care and that I'm trying hard to find a solution here. Please tell me if my suggestions don't work for you and you want to have another solution.

PabloCastellano:
Hey!

I have just rebased the branch from master.
Since you agreed to merge it asap (I've been busy since then), let's merge it asap now.

About the squashing VS merging discussion I appreciate a lot your interest in finding a consensus. That really says a lot from you. I'm OK with squashing now and opening multiple PRs next time if I find it necessary :+1:

PR #482, opened 2017-08-30
qemu: improve networking

### LINKED ISSUE #433: Qemu ARM warnings
When running the ARM Qemu image, several warnings are printed out. They might not block anything, but it would probably be good to get them worked out anyway.

Our current networking solution is great, but deprecated it seems.

```
qemu-system-arm: -redir tcp:2222::22: The -redir option is deprecated. Please use '-netdev user,hostfwd=...' instead.
```

Since we don't specify the image format, Qemu starts guessing and mentions it's dangerous. This probably affects performance quite a lot.
```
WARNING: Image format was not specified for '/home/bart/.local/var/pmbootstrap/chroot_native/home/user/rootfs/qemu-vexpress.img' and probing guessed raw.
         Automatically detecting the format is dangerous for raw images, write operations on block 0 will be restricted.
         Specify the 'raw' format explicitly to remove the restrictions.
```

Last but not least some PulseAudio errors:

```
pulseaudio: set_sink_input_volume() failed
pulseaudio: Reason: Invalid argument
pulseaudio: set_sink_input_mute() failed
pulseaudio: Reason: Invalid argument
```

PabloCastellano:


MartijnBraam:
you can't use the e1000 on the ARM architecture

drebrez:
@PabloCastellano @MartijnBraam what you think about redirecting also the ports used in the initramfs for telnet? port 23 and 24?

MartijnBraam:
that would be very handy

ollieparanoid:
Added `dont_merge_pr`, because Travis fails and it does not work for @PureTryOut.

PabloCastellano:
Ok, let me some time to find another workaround that works for all platforms (or share yours if you have any!). Redirecting ports 23 and 24 make sense too.

PabloCastellano:
I've rebased the branch and made the following improvements related to networking:

- Use hostfwd instead of -redir, which is deprecated
- qemu devices: add interfaces configuration and virtio_net module
- redirect telnet ports 23 and 24 (debug)

However telnet connects but the connection seems to get stuck and you cannot really use it (maybe related to not using tun/tap?)

```
$ telnet localhost 2224
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
```

Now it works for arm. I cannot test it for aarch64 since I'm using an older version of qemu that prevents it to boot (black screen).

About the `format=raw` warning, I'm unable to remove the warning because the `-sd` parameter doesn't support specifying the format.

IMHO it is ready to merge.

ollieparanoid:
Thanks for updating the PR! Sorry that I did not mention this earlier, but: Ubuntu 16.05 is shipped with qemu 2.5, which does not have the `-netdev` option (as @MartijnBraam wrote somewhere, and just confirmed in the chat).

So how do we proceed here? I see the following options:
* get qemu from the Alpine chroot working (see #453) instead of using the host's qemu
* offer multiple networking configurations for qemu with automatic version detection

PureTryOut:
The former option sounds the best to me, it makes pmbootstrap as independent from the host as possible, and will reduce created issues due to users using outdated versions.

PabloCastellano:
@ollieparanoid I'm using qemu 2.5 from Ubuntu 16.04 and the `-netdev` option does work :thinking:

ollieparanoid:
@MartijnBraam: could it be that you tested it on another distribution?

MartijnBraam:
Oh might have been my old mint box, which is on Ubuntu 1404 for qemu

PR #483, opened 2017-08-30
Fix firmware-adreno and move it to firmware

opendata26:
Should probably have tested the original pr ☺️


MartijnBraam:
Looks good :+1:

PR #489, opened 2017-08-31
Added LG G Watch R  (lg-lenok) Support

JamesHealdUK:
Hope Im doing it right...


aports/device/device-lg-lenok/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+	install -D -m644 "$srcdir"/deviceinfo \
+		"$pkgdir"/etc/deviceinfo
+	install -D -m644 "$srcdir"/initfs-hook.sh \
+		"$pkgdir"/etc/postmarketos-mkinitfs/hooks/00-${pkgname}.sh
+	install -D -m644 "$srcdir"/90-android-touch-dev.rules \
+		"$pkgdir"/etc/udev/rules.d/90-android-touch-dev.rules
+	install -D -m644 "$srcdir"/profile.sh \
+		"$pkgdir"/etc/profile.d/00-$pkgname.sh
+	install -D -m644 "$srcdir"/fb_fix.sh \
+		"$pkgdir"/etc/profile.d/00-$pkgname.sh
```
drebrez: @JamesHealdUK you set twice the same name `00-$pkgname.sh` and overriding the `profile.sh` script. If it works it probably means that you can directly remove the `profile.sh`

aports/device/linux-lg-lenok/APKBUILD
```diff
@@ -0,0 +1,106 @@
+# APKBUILD based on linux-vanilla aport. Changes:
+# - disabled module installation
+# - add !check !tracedeps
+# - package: just install zimage and kernel.release, because the kernel config
+#	does not generate modules or dtb files
+# - do not create -dev subpackage (makes no sense without module support)
+#
+# Changes:
+# - enable devtmpfs (needed for udev -> touch support in weston)
```
drebrez: @JamesHealdUK can you please describe the changes you've done? see [example](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/device/linux-huawei-y530/APKBUILD#L6-L11)

aports/device/linux-lg-lenok/APKBUILD
```diff
@@ -0,0 +1,106 @@
...
+pkgname=linux-${_flavor}
+pkgver=3.10.105
+case $pkgver in
+	*.*.*)  _kernver=${pkgver%.*};;
+	*.*) _kernver=$pkgver;;
+esac
+pkgrel=4
+arch="armhf"
+pkgdesc="G Watch R kernel"
+url="https://github.com/vt0r/kernvatore-lenok"
```
drebrez: @JamesHealdUK the url doesn't match with the one used to download the sources, please set the correct url.
In the source variable below, if you want you can do something like this:
`$pkgname-$_hash.tar.gz::${url}/archive/${_hash}.tar.gz`

aports/device/device-lg-lenok/APKBUILD
JamesHealdUK: Woops... Just solved this, thanks for pointing it out

JamesHealdUK:
Thanks for pointing those out dreb. Should have fixed them all now, given it a test on my side and everything looks to be A-OK!

JamesHealdUK:
Damn.. Really sorry guys, Im not very good with Git, hopefully my latest commit should fix that!

aports/device/device-lg-lenok/profile.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+if [ -e /lib/udev/v4l_id ]; then
```
ollieparanoid: This file originates from device-lg-mako and has two purposes:

1. An ugly workaround for "video4linux", which a smartwatch probably does not use at all.
2. launch the demos, because xwayland and therefore postmarketos-demos does not work.

I really do not think you need the 1. hack. The second one makes sense if xwayland is broken, or simply because you couldn't start the demos manually because the screen is too small.

What do you think about removing 1. (and maybe also 2.)?

JamesHealdUK: Sounds good, Il remove the profile file all together I think! Dont think any of it is needed.

Thanks for the heads up!

JamesHealdUK:
Thanks @ollieparanoid ! Hope this turns out OK

ollieparanoid:
Perfect, thanks!

PR #491, opened 2017-08-31
Flavor instead of device name when creating recovery zip

### LINKED ISSUE #490: Recovery zip creation fails on maguro (it expects device name to equal kernel name)
When running `pmbootstrap install --android-recovery-zip`, it gives the following error:

```
...
(026064) [13:37:44] (buildroot_armhf) create recovery zip
(026064) [13:37:44] (buildroot_armhf) % cd /var/lib/postmarketos-android-recovery-installer/ && mv /tmp/install_options install_options
(026064) [13:37:45] (buildroot_armhf) % cd /var/lib/postmarketos-android-recovery-installer/ && cp /mnt/rootfs_samsung-maguro/boot/boot.img-samsung-maguro boot.img
cp: can't stat '/mnt/rootfs_samsung-maguro/boot/boot.img-samsung-maguro': No such file or directory
(026064) [13:37:45] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(026064) [13:37:45] NOTE: The failed command's output is above the ^^^ line in the logfile: /home/kevin/.local/var/pmbootstrap/log.txt
(026064) [13:37:45] ERROR: Command failed: (buildroot_armhf) % cd /var/lib/postmarketos-android-recovery-installer/ && cp /mnt/rootfs_samsung-maguro/boot/boot.img-samsung-maguro boot.img
(026064) [13:37:45] Run 'pmbootstrap log' for details
...
```

Going into the chroot reveals that the file it's trying to copy is actually called `boot.img-samsung_tuna`because the Galaxy Nexus (`device-samsung-maguro`) uses the kernel `linux-samsung-tuna`. (I'm not sure if the fact that they're named differently is a bug in and of itself.) The installer, however, expects the device name and kernel name to be the same, so it crashes.

(As an aside on the naming: according to https://groups.google.com/forum/#!msg/android-building/qzSxFEqWvNg/qanfxMkzXHMJ, `tuna` is the generic name for all Galaxy Nexus variants, while `maguro` is the GSM/HSPA+ one and `toro` is the CDMA model.)

ata2001:
Fixes #490 

Although not sure, what happens, when you have multiple flavors installed in the device rootfs chroot.
Is that a normal case?

### DIFF
--- a/pmb/install/recovery.py
+++ b/pmb/install/recovery.py
@@ -27,6 +27,8 @@ def create_zip(args, suffix):
     """
     zip_root = "/var/lib/postmarketos-android-recovery-installer/"
     rootfs = "/mnt/rootfs_" + args.device
+    flavor = pmb.chroot.other.kernel_flavor_autodetect(args,
+                                                       "rootfs_" + args.device)
 
     # Install recovery installer package in buildroot
     pmb.chroot.apk.install(args,
@@ -52,7 +54,7 @@ def create_zip(args, suffix):
         # Move config file from /tmp/ to zip root
         ["mv", "/tmp/install_options", "install_options"],
         # Copy boot.img to zip root
-        ["cp", rootfs + "/boot/boot.img-" + args.device, "boot.img"],
+        ["cp", rootfs + "/boot/boot.img-" + flavor, "boot.img"],
         # Create tar archive of the rootfs
         ["tar", "-pczf", "rootfs.tar.gz", "--exclude", "./home/user/*",
          "-C", rootfs, "."],
--- a/pmb/install/recovery.py
+++ b/pmb/install/recovery.py
@@ -19,6 +19,7 @@ along with pmbootstrap.  If not, see <http://www.gnu.org/licenses/>.
 import logging
 
 import pmb.chroot
+from pmb.flasher.frontend import parse_flavor_arg
 
 
 def create_zip(args, suffix):
@@ -27,8 +28,7 @@ def create_zip(args, suffix):
     """
     zip_root = "/var/lib/postmarketos-android-recovery-installer/"
     rootfs = "/mnt/rootfs_" + args.device
-    flavor = pmb.chroot.other.kernel_flavor_autodetect(args,
-                                                       "rootfs_" + args.device)
+    flavor = parse_flavor_arg(args)
 
     # Install recovery installer package in buildroot
     pmb.chroot.apk.install(args,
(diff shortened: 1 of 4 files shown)


MartijnBraam:
The multiple flavors happen when you try mainlining a device, then you have the linux-somebrand-somedev flavor and the linux-postmarketos flavor

ata2001:
Added an option to specify flavor when calling `install`.

drebrez:
@ata2001 I think something like this should work => https://github.com/postmarketOS/pmbootstrap/blob/master/pmb/chroot/initfs.py#L96-L100

ata2001:
@drebrez I think `parse_flavor_arg` does the same, but raises an exception too, when no kernel with the flavor is installed.

pmb/install/recovery.py
```diff
@@ -19,6 +19,7 @@
 import logging
 
 import pmb.chroot
+from pmb.flasher.frontend import parse_flavor_arg
```
ollieparanoid: Please use `import pmb.flasher.frontend` and `pmb.flasher.frontend.parse_flavor_arg()` here, for consistency with the rest of the code.

Also if we're using that function in install, I would move it to `pmb.helpers.frontend._parse_flavor()` if you don't mind.

pmb/install/recovery.py
```diff
@@ -27,6 +28,7 @@ def create_zip(args, suffix):
     """
     zip_root = "/var/lib/postmarketos-android-recovery-installer/"
     rootfs = "/mnt/rootfs_" + args.device
+    flavor = parse_flavor_arg(args)
```
ollieparanoid: We need to add the flavor argument to `pmbootstrap install`, otherwise this will break in `parse_flavor_arg`.
https://github.com/postmarketOS/pmbootstrap/blob/master/pmb/flasher/frontend.py#L39

In my opinion, the `--flavor` description should mention, that this is only relevant for the recovery zip.

ata2001: @ollieparanoid sorry, forgot to `git add`

it mentions it already, please wait till I push another commit

ata2001:
@ollieparanoid made the changes

pmb/parse/arguments.py
```diff
@@ -192,6 +192,9 @@ def arguments():
                          " added to the rootfs (e.g. 'vim,gcc')")
     install.add_argument("--no-fde", help="do not use full disk encryption",
                          action="store_false", dest="full_disk_encryption")
+    install.add_argument("--flavor",
+                         help="Specify kernel flavor to include in recovery"
+                              "flashable zip", default=None)
```
ollieparanoid: space missing: `" flashable zip"`

ata2001:
added missing whitespace

ata2001:
ping @ollieparanoid

PR #493, opened 2017-09-01
Added Chocolate Doom

JamesHealdUK:
Hope this is alright, have tested it all and it seems to build OK. Havent been able to get it running on a device yet due to a lack of keyboard and a screen size big enough.

My initial efforts with getting it to run on my LG Lenok have failed, says no video device available whenever I try. Even in a profile.d file.

However, I've tried in Qemu and it launched, I get a black screen (but this might be due to the lack of video accel in Qemu).


JamesHealdUK:
Hang on, checked out from the wrong branch :(

PR #503, opened 2017-09-03
Update links to wiki for new wiki

JamesHealdUK:


JamesHealdUK:
Hang on, I should probably do this on the master branch through my local machine rather than on Github itself, itl save on the number of commits...

PR #514, opened 2017-09-04
[samsung-i9100] add wifi firmware

tssk:
Adding firmware files for samsung-i9100 to make wifi work.

But I did not test it (sorry, I know it is guide lines) - I manually placed files to /system/etc/wifi but the code uses /lib/firmware/postmarketos.

I need to start somehow - no problem if you reject this with feedback how to make it better, thanks.


aports/firmware/firmware-samsung-i9100/APKBUILD
```diff
@@ -0,0 +1,32 @@
+_files="bcmdhd_apsta.bin bcmdhd_mfg.bin bcmdhd_p2p.bin bcmdhd_sta.bin nvram_mfg.txt nvram_mfg.txt_murata nvram_net.txtnvram_net.txt_murata"
```
PabloCastellano: For readability sake I would write it as a column. See this as example: https://github.com/postmarketOS/pmbootstrap/blob/master/aports/device/linux-motorola-titan/APKBUILD#L35

aports/firmware/firmware-samsung-i9100/APKBUILD
```diff
@@ -0,0 +1,32 @@
...
+
+pkgname=firmware-samsung-i9100
+pkgver=1
+pkgrel=1
```
PabloCastellano: Minor issue, pkgrel should be 0 because it is the first version

aports/firmware/firmware-samsung-i9100/APKBUILD
```diff
@@ -0,0 +1,32 @@
...
+pkgdesc="Samsung Galaxy S II Firmware"
+url="https://github.com/TheMuppets/proprietary_vendor_samsung"
+arch="armhf"
```
PabloCastellano: should be "noarch" because it is not architecture dependent

aports/firmware/firmware-samsung-i9100/APKBUILD
```diff
@@ -0,0 +1,32 @@
...
+depends=""
+makedepends=""
+install=""
+subpackages=""
+for _i in $_files; do source="$source https://github.com/TheMuppets/proprietary_vendor_samsung/tree/cm-14.1/i9100/proprietary/system/etc/wifi/$_i"; done
+options="!check"
+
+build() {
+	return 0
+}
```
PabloCastellano: you can safely remove `build()` if only returns 0.

aports/device/device-samsung-i9100/APKBUILD
```diff
@@ -5,7 +5,7 @@ pkgdesc="Samsung Galaxy SII"
 url="https://github.com/postmarketOS"
 arch="noarch"
 license="MIT"
-depends="linux-samsung-i9100"
+depends="linux-samsung-i9100 firmware-samsung-i9100"
```
PabloCastellano: remember to bump pkgrel here since you did some change!

zhuowei:
You can change the default firmware path by changing the kernel config: there's instructions at https://wiki.postmarketos.org/wiki/Wifi

By the way, as an alternate approach, some devices e.g. [Nexus 5](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/device/device-lg-hammerhead/APKBUILD#L8) use the upstream linux-firmware package; there seems to be [one for the bcm4330](https://pkgs.alpinelinux.org/contents?file=*4330*&path=&name=linux-firmware&branch=edge&repo=main&arch=armhf) there, but I dunno if it'll work for your device, and you'll need to extract the nvram file anyways. Not sure which is the best approach.

ollieparanoid:
*Adding the `dont_merge_pr` label until the required changes have been made.*

Thanks so much for this, I also have a i9100 👍

tssk:
I do not undrestand why the Travic CI build fails? It works on my machine ok.

> $ test/check_checksums.py 
> Checking firmware-samsung-i9100 for correct checksums
> [23:00:22] (native) generate checksums for firmware-samsung-i9100
> [23:00:22] Done
> ** The checksums are correct

MartijnBraam:
Hmm looks like travis is having some networking issues, that might explain the checksum error

tssk:
To make this PR correct I think I should add linux patch to change the default path for loading of the firmware as zhuowei. But I do not know where to get the exact linux sources for the linux-samsung-i9100 package...

zhuowei:
@tssk The path is in the kernel config, not the source itself: see https://github.com/postmarketOS/pmbootstrap/blob/master/aports/device/linux-samsung-i9100/config-samsung-i9100.armhf#L1420 . Changing this should get it to load firmware and nvram from a different path: see the [Nexus 5 config](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/device/linux-lg-hammerhead/config-lg-hammerhead.armhf#L1514).

And if you do need the kernel source, see the aports/device/linux-samsung-i9100/APKBUILD file: it would be https://github.com/LineageOS/android_kernel_samsung_smdk4412 at commit 349a3e91e76d17e67ef6213e1f6712e700695631

tssk:
I think it should be ok now.

tssk:
@ollieparanoid That is a stupid mistake... Thanks a lot for finding out.

I will fix it and test on the device.

tssk:
I fixed to download path and tested the package manually on device (copied via ssh, install apk, reboot, set up wifi and connect to ap).

I do not know how to add the package to automatically include to device-samsung-i9100 package though.

aports/device/device-samsung-i9100/APKBUILD
tssk: I am not sure if it was necessary to add "firmware-samsung-i9100" to depends= ; I thought it will automatically include package to system image.

bhuvi8:
Hi, I tested the steps posted in the wiki page https://wiki.postmarketos.org/wiki/Samsung_Galaxy_SII_(samsung-i9100)
I am not able to get it to work...getting the below error on step 
sudo wpa_supplicant -i wlan0 -c wifi 
nl80211: Could not re-add multicast membership for vendor events: -2 (No such file or directory)

bhuvi8:
I'm not able to see the wlan0 interface after adding the firmware files and rebooting...

tssk:
@bhuvi8 Did you compile linux-samsung-i9100 and firmware-samsung-i9100 from my git branch? If not I think it is little bit offtopic here as comment to the PR should be.

Anyway I get this error too but the wifi works ok.

```
localhost:~$ sudo wpa_supplicant -i wlan0 -c XXX 
Successfully initialized wpa_supplicant
nl80211: Could not re-add multicast membership for vendor events: -2 (No such file or directory)
wlan0: Trying to associate with XXX (SSID='XXX' freq=2437 MHz)
wlan0: Associated with XXX
wlan0: CTRL-EVENT-SUBNET-STATUS-UPDATE status=0
wlan0: WPA: Key negotiation completed with XXX [PTK=CCMP GTK=TKIP]
wlan0: CTRL-EVENT-CONNECTED - Connection to XXX completed [id=0 id_str=]
```

PabloCastellano:
@tssk 
> I do not know how to add the package to be automatically included to device-samsung-i9100 package though.

Simply add the package name to the `depends=` line in other package's APKBUILD

PabloCastellano:
Just approved the changes because the PR looks good now. Unfortunately I don't own this device and I'm not able to test it

aports/device/device-samsung-i9100/APKBUILD
drebrez: @tssk that's correct, installing the `device-samsung-i9100` package it will also install the packages specified in the depends

tssk:
@PabloCastellano: I did add it to device-samsung-i9100/APKBUILD but it does not get cimpiled and installed automatically.

PabloCastellano:
@tssk This is the way to do it, see how other devices do it like this as well. Maybe you are not using timestamp based rebuilds?

ollieparanoid:
The code looks good now, and I have just built it. `device-samsung-i9100` does pull in `firmware-samsung-i9100` correctly. As @PabloCastellano points out, if you have timestamp based rebuilds disabled, your package might not have been rebuilt, that's why it may have not worked in your case.

I did not find to test this on my device (maybe later, really excited that you made this and that I can use wifi with it now!) - so I'll merge it now because the code looks good, compiles, and you have said it works - and even if it does not, it doesn't make the situation worse.

@tssk: Thank you very much!

PR #519, opened 2017-09-05
Working GUI for sony-aries

kskarthik:


aports/device/linux-sony-aries/APKBUILD
```diff
@@ -105,6 +105,6 @@ package() {
 }
```
PabloCastellano: Remember to bump pkgrel!

kskarthik:
Will be back with new PR

PR #521, opened 2017-09-05
linux-postmarketos: change kernel suffix to -postmarketOS (fixes #497)

### LINKED ISSUE #497: linux-postmarketos kernel should not have -Alpine as suffix
`-postmarketOS` would be better to avoid confusion with Alpine's official kernel.

PabloCastellano:


MartijnBraam:
Hmm isn't this just a case of changing `CONFIG_LOCALVERSION` in the kernel config. In that case it becomes 4.12.0-pmos1 or something. The build information that you changed is also very important to have reproducable builds of the kernel.

ollieparanoid:
@MartijnBraam: I don't quite understand what you mean, could you please elaborate?

MartijnBraam:
For example the uname output from my laptop:

```
Linux zenbook 4.12.4-1-ARCH #1 SMP PREEMPT Fri Jul 28 18:54:18 UTC 2017 x86_64 GNU/Linux
```

the -ARCH part is the CONFIG_LOCALVERSION. The `#1` part straight after that is the `KBUILD_BUILD_VERSION` set in this pr, which looks like it should just be an integer, not the pmos text

ollieparanoid:
Alright, looks like @PabloCastellano just followed Alpine's convention. I guess Alpine's primary goal with using `KBUILD_BUILD_VERSION` was to include the `pkgrel` there, which is just the value you would want to have. Then they probably added `-Alpine` there, so they had it all in one place.

I guess the correct way (as intended by the kernel developers) would be:
* put the `pkgrel` into `KBUILD_BUILD_VERSION`
* put the `-postmarketOS` in `CONFIG_LOCALVERSION`

However, we have a bit of a special situation here. There are *so many* different forks of Android kernels and it looks like folks put a meaningful value in `CONFIG_LOCALVERSION` [already](https://github.com/postmarketOS/pmbootstrap/blob/f7fa7ae1a91a6a6a4fdb2ee6477e5f8a8bceb8b6/aports/device/linux-lg-mako/config-lg-mako.armhf#L47). So I recommend we simply follow Alpine's (non-standard?) convention here. Then we would have the special Android fork name included, as well as a sign that this was built for postmarketOS. (If we do it that way, @PabloCastellano's patch would be perfect.)

So we could have: `3.12.4-1-lineageos 1-postmarketOS`

What would you say about that proposal?

MartijnBraam:
Yeah thats a better way to handle it with the current android kernel mess.

PR #522, opened 2017-09-06
add libsamsung-ipc

tssk:


Pneumaticat:
FYI it looks like [replicant has a more recently-updated fork](https://git.replicant.us/replicant/external_libsamsung-ipc). Not sure what the differences are, but it looks more maintained.

tssk:
@ollieparanoid Builds in my environment... `asm/types.h` seems to be some header from linux.

@Pneumaticat Thanks, I will check the Replicant sources.

ollieparanoid:
@tssk: I understand that it builds for you. The thing is, that `pmbootstrap` does not clear existing packages before building a new package. So it is probably the case, that it needs a dependency, which is not listed explicitly in `makedepends`. Could you try to `pmbootstrap zap` your chroot, then try to build it again? You should run into the same error.

By the way, `asm/types.h` is in the `linux-headers` package (which can be found with Alpine's awesome [package content search](https://pkgs.alpinelinux.org/contents?path=%2A%2Fasm&repo=main&page=3&file=types.h&arch=armhf&branch=edge)).

*I have been working on a `build --strict` feature, that clears the installed packages before each build to avoid such errors. It might be ready soon to be integrated in pmbootstrap.*

tssk:
@ollieparanoid I thought it might me the case. Thanks for confirming.

Would it be vise to make linux-samsung-i9100 package dependency for this package? Or should I better search in Alpine?

ollieparanoid:
It makes sense to add `linux-headers` to the makedepends of `libsamsung-ipc`.

Later *when we have it working*, we can add `libsamsung-ipc` as dependency of `device-samsung-i9100`. Right now it would be just one more package to build without any benefits so far, and if someone wants to test it, the package can be added in `pmbootstrap init` as extra package, or in `pmbootstrap install --add=libsamsung-ipc`.

tssk:
Very proud to say I successfully tested it on device. See https://wiki.postmarketos.org/wiki/Samsung_Galaxy_SII_(samsung-i9100)#Modem

But it still need some manual adjustments around modem firmware and nv_data. If you already have strategy for how this things include and where (I guess device-samsung-i9100 package) please let me know and I can prepare that.

ollieparanoid:
For your question on how to integrate this:
I would let `device-samsung-i9100` package depend on `libsamsung-ipc` (also increase `pkgrel`).

Then about the `/efs` partition: a quick research shows, that it has the IMEI of the phone and some MAC addresses. For now I recommend mounting that by creating a **read-only** entry in the fstab in the device package's post-install script like [here](https://github.com/postmarketOS/pmbootstrap/blob/594ce64321d4f775a81c4b0c527282b4803bd8c0/aports/device/device-nokia-rx51/device-nokia-rx51.post-install#L15-L18).

*Maybe we can look into IMEI/MAC spoofing at some point, to aid privacy - but that should be discussed in a new issue.*

In the wiki, you say that `ln -s /dev/mmcblk0p8 /dev/block/mmcblk0p8` is necessary. Sounds like that path is hardcoded in libsamsung-ipc (or in the kernel maybe?). For that I recommend to patch the path in the software in question, so it points to the real location without the symlink.

I am really impressed and amazed *_* awesome work @tssk!

ollieparanoid:
It compiles now, and the `APKBUILD` looks good! I'm merging this, as this is already very useful on its own. We can implement the rest in new PRs. Thank you very much @tssk!

Issue #501, opened 2017-09-02
Creating the rootfs image fails sometimes


ollieparanoid:
The following error just happened during `pmbootstrap install`:
`losetup: /home/user/rootfs/lg-mako.img: failed to set up loop device: No such file or directory`

When I run `pmbootstrap install` again, it gets created fine. This might be a regression in using `truncate` instead of `fallocate`? Or maybe it is related to the `loop` module not getting loaded properly?

I rand `lsmod | grep loop` directly after the failed `pmbootstrap install` and the `loop` module was loaded.

Full log:
```
(029356) [23:47:15] (native) % rm /home/user/rootfs/lg-mako.img
(029356) [23:47:15] (native) create lg-mako.img (435M)
(029356) [23:47:15] WARNING: Make sure, that your target device's partition table has allocated at least 435M as system partition!
[23:47:15] Continue? (y/n) [y] y
(029356) [23:47:18] (native) % su user -c 'mkdir -p /home/user/rootfs'
(029356) [23:47:18] (native) % truncate -s 435M /home/user/rootfs/lg-mako.img
(029356) [23:47:18] (native) mount /dev/install (lg-mako.img)
(029356) [23:47:18] (native) mount /home/user/rootfs/lg-mako.img (loop)
(029356) [23:47:18] % sudo mount --bind /dev/loop-control /home/user/.local/var/pmbootstrap/chroot_native//dev/loop-control
(029356) [23:47:18] (native) % losetup -f /home/user/rootfs/lg-mako.img
losetup: /home/user/rootfs/lg-mako.img: failed to set up loop device: No such file or directory
(029356) [23:47:19] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(029356) [23:47:19] NOTE: The failed command's output is above the ^^^ line in the logfile: /home/user/.local/var/pmbootstrap/log.txt
(029356) [23:47:19] ERROR: Command failed: (native) % losetup -f /home/user/rootfs/lg-mako.img
(029356) [23:47:19] Run 'pmbootstrap log' for details.
(029356) [23:47:19] See also: <https://postmarketos.org/troubleshooting>
(029356) [23:47:19] Traceback (most recent call last):
  File "/home/user/code/pmbootstrap/pmb/helpers/run.py", line 45, in core
    stderr=args.logfd)
[...]

PR #528, opened 2017-09-07
Fix 501: Always load the loop module so losetup can access the img file

BrianOtto:
This has been tested and confirmed working in a clean install, that was created after a reboot. Please see #501 for how I was testing things. FYI, the loop module had to be loaded before the `pmb.helpers.mount.bind_blockdevice` in `init()` for this solution to work.


benwaffle:
Why doesn't the existing check work though?

BrianOtto:
On my machine, it is finding /dev/loop-control, which means the loop module will not get loaded.

PR #530, opened 2017-09-07
Fix #524: Init zapped the previous work contents

### LINKED ISSUE #524: Zap in 'pmbootstrap init' removes chroots in *previous* work folder
@BrianOtto wrote in https://github.com/postmarketOS/pmbootstrap/issues/501#issuecomment-327408580:
> Any time I ran `./pmbootstrap.py init` the **chroot_native** directory (e.g. **/home/pmos/.local/var/pmbootstrap-test4/chroot_native**) would get removed in my PREVIOUS test directory. For example, when I was working in **pmbootstrap-test5**, the **pmbootstrap-test4** one got removed. I thought that by creating a new directory and setting the `Work Path` to this, the changes would be isolated. So either I am misunderstanding how the process works, or the script is accessing previous build files when it shouldn't.

ollieparanoid:


ollieparanoid:
Thanks for testing!

PR #531, opened 2017-09-07
Fix #151: git ambiguous argument error

### LINKED ISSUE #151: Timestamp based build: ambiguous argument
When running `pmbootstrap.py build linux-asus-grouper --arch=armhf --force`, I get the following error

```
fatal: ambiguous argument 'origin': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```
from the command `git diff --name-only origin`

My local repository was set up with the following commands:
```
git remote rename origin upstream
git remote add origin https://github.com/antonok-edm/pmbootstrap
```

ollieparanoid:
We check if origin/HEAD is present. In case that reference is missing, we show a meaningful error message now, with an explanation on how to add it.

Thanks a lot to @antonok-edm  and @drebrez for debugging what caused this issue!
@antonok-edm: maybe you want to test this?

Close: #151


ollieparanoid:
Yeah, it makes sense to move that function to `pmb/helpers/git.py`.
Thanks for the review!

ollieparanoid:
Code has been moved, it's ready for another review now.

drebrez:
@ollieparanoid I tested it by removing the `refs` and shows the error message correctly. 

But executing the suggested command didn't fixed it immediately, I had to perform a `git fetch`.

Edit: tried it again by manually deleting the file `.git/refs/remotes/origin/HEAD` and error appeared, with the suggested command fixed it immediately. Now I don't know why it didn't worked before.

What you think? should we also suggest to do a `git fetch`?

ollieparanoid:
I've added the `git fetch` to the command, that gets printed out, so it will work for sure. Do you think it is fine now?

Thanks a lot for testing and reviewing @drebrez!

drebrez:
@ollieparanoid I think it should be fine now 👍

PR #535, opened 2017-09-08
Re-organize demo menu so that shutdown is first; add battery monitoring daemon

pavelmachek:
On n900, postmarketos-demos is too high, and does not fit on screen. You can't even shutdown your machine cleanly. Fix it by putting shutdown first.

Normally, hardware is responsible for shutting down the system when battery is low. But N900 does not have such hardware, and kernel is not doing the job, either. Protect battery by shutting down when voltage drops too much.


aports/main/batmond/APKBUILD
```diff
@@ -0,0 +1,18 @@
+pkgname=batmond
+pkgver=1
+pkgrel=1
+pkgdesc="Battery monitoring daemon"
+url="https://github.org/postmarketOS"
+arch="noarch"
+license="GPL"
+depends=""
```
drebrez: @pavelmachek I think you should add python as a dependency of this package. (I think the package name is `py3-pillow`)

aports/main/batmond/batmond.py
```diff
@@ -0,0 +1,111 @@
...
+        f = open(s, "r")
+        r = f.read()
+        f.close()
+        return r
+
+
+class Battery(Test):
+    hotkey = "b"
+    name = "Battery"
+    path = "/sys/class/power_supply"
```
drebrez: @pavelmachek this script probably only work with the nokia n900 since it has hardcoded paths for the battery, I suggest adding a variable in the deviceinfo or if the `/sys/class/power_supply` is the same for every device we can search for the `voltage_now`, `charge_current`, `current_limit`... paths.
@ollieparanoid @MartijnBraam what you think about it?

aports/main/batmond/APKBUILD
MartijnBraam: pillow is the image manipulation tool for python, not sure why its needed for a battery monitor. adding a depends on `python3` would be enough

aports/main/batmond/batmond.py
MartijnBraam: The android hardware have the battery controller power down the board instead of software

pavelmachek:
Yes, this should become generic. But it would be good to know what the
differences between devices are, first. Can you try to get it to work
on your device?

I created this:

https://gitlab.com/tui/tui/blob/master/ofone/hardware.py

Original idea was to have one python module that hides differences
between different phones. Battery differences are just a subset, I
believe that for example reading temperature and controlling RGB LED
will differ between devices, too.

Next question is if this maybe should be C library instead of Python.

Anyway, there's a room for improvement, but I believe this is good
first step.

pavelmachek:
Ok.

In addition to missing hardware to protect the battery...

On N900, you need Linux booted to be able to charge the
battery. And you need a bit of battery power left to be able to boot
the system.

If someone else has similar problem on Android device, this will be
useful for them.

aports/main/batmond/APKBUILD
drebrez: @MartijnBraam didn't know about the pillow, I just looked at the depends of the postmarketos-splash package. Thanks for the clarification

ollieparanoid:
Would it be possible if you closed this, and split it into two? Because there's not much to discuss for the `postmarketos-demos` package change, but the battery monitoring daemon does need some discussion apparently (and both packages are not really related).
Thank you for your enhancements!

aports/main/batmond/batmond.py
craftyguy: Why is the source for this application hosted in its aports folder? Wouldn't it make more sense to have this someplace else (e.g. github) and have the APKBUILD pull it down and install it?

pavelmachek:
Well, this originaly comes from https://gitlab.com/tui/tui
project. But I thought simple stuff is easier to maintain here in the
project.

ollieparanoid:
I think what I wrote came out wrong. What I meant was: could you make two PRs for:
1. Re-organize the postmarketos-demos menu
2. The battery monitoring daemon

... because `1.` could be merged already, but `2.` needs some discussion and is a completely different logical change compared to `1.`. It would also be cleaner in the change log (we always squash commits here). I understand that this extra work with splitting this up is annoying, sorry for that and thank you again for making your improvements in the first place!

pavelmachek:
Ok, I originally intended to submit two PRs. But github was stronger
than me :-).

pavelmachek:
There is now separate pull request for the menu reorganization. For battery monitoring... let me know what you want. I believe that maintinaining monitor in the project in python is the best solution for now. Long term, it probably wants to move into C and into externally-maintained library.

If python, in-tree solution is acceptable, let me know, and I'll create new pull request with just that.

ollieparanoid:
I have read through the `batmond` code, and it appears to be *very* N900 specific, so I don't think it makes sense to rewrite it now to support more phones, and most other phones handle this in hardware anyway as @pavelmachek and @MartijnBraam explained.

@pavelmachek: here are my recommendations on packaging `batmond`:

* host the script in a new repo at `https://gitlab.com/tui/nokia-rx51-batmond` (or n900-batmond, or whatever you prefer)
* add a short `README.md`, that describes, that it is a standalone version of the battery functions of `tui`
* add a proper license (you have written "GPL" in the APKBUILD, I would recommend a more recent version unless you need to use the GPL1)
* tag a release of the initial version (e.g. 0.1)
* create a new PR here in pmbootstrap:
  * new aport `aports/device/nokia-rx51-batmond`
  * the APKBUILD downloads the release you tagged and uses that as source
  * add an OpenRC init script, so we can run it properly as daemon
* when that PR is merged, start a new PR that makes device-nokia-rx51 depend on your new package, and enables the batmond during the installation

*From here on, one could rewrite the battery monitor daemon in C if desired, and you could just tag a new version when it is ready. And later, support for multiple devices could be added etc.*

@pavelmachek: So thanks again for creating the standalone version in the first place, I'm looking forward to integrate it! Closing this PR, but please answer if you have any comments.

pavelmachek:
I'm pretty sure something like batmond will be needed on more
devices. I'd guess N9, N950 at the very least.

Lets see.

Anyway, I could create a new repository (etc)... but maybe it will be
easier to do it right, and put it into kernel.

PR #537, opened 2017-09-08
Add possibility to specify the system partition name for heimdall flashing method

drebrez:
Also added default values for the others heimdall partitions:
 kernel => KERNEL
 initfs => RECOVERY
 system => SYSTEM


ollieparanoid:
I've added documentation to the [deviceinfo specification](https://wiki.postmarketos.org/wiki/Deviceinfo_reference#flash).

PR #539, opened 2017-09-08
aports: packaged abootimg (fixes #135)

### LINKED ISSUE #135: Package abootimg
This tool is already packaged in Debian.

It supports creating boot.img, descompressing and updating easily which can make development a lot easier.

* [Code](https://gitorious.org/ac100/abootimg)
* [README](https://gitorious.org/ac100/abootimg?p=ac100:abootimg.git;a=blob;f=README;h=8c1d6ddc005d22a1dd422ea98d2c4a9060315ad1;hb=HEAD)

PabloCastellano:


ollieparanoid:
Thanks for making this PR @PabloCastellano! I've made a few adjustments, please review them and tell me if this PR is ready to be merged. 👍

ollieparanoid:
You are right, that is kind of strange. I've changed it to `0.6-20150325`, which should make it obvious that this is not a tagged version. Do you like it?

PabloCastellano:
FTBFS: pkgver cannot include dashes

```
>>> abootimg: Checking sanity of /home/user/build/APKBUILD...
>>> ERROR: abootimg: 0.6-20150325 is not a valid version
(006535) [11:56:24] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(006535) [11:56:24] NOTE: The failed command's output is above the ^^^ line in the logfile: /var/pmbootstrap/log.txt
(006535) [11:56:24] ERROR: Command failed: (native) % cd /home/user/build && su user -c 'CARCH=x86_64 abuild -r'
```

ollieparanoid:
Now that's ironic, I've implemented a [100% compatible version parser](https://github.com/postmarketOS/pmbootstrap/blob/e421313632676cf2e332e62a0fe039a8499a9021/pmb/parse/version.py) by translating the C version of apk-tools *and* I have implemented the strict mode, yet I fail to see that this version is invalid and that it fails to build in strict mode. TBH I didn't even notice, that the strict mode also checks the version string (because I only changed `abuild` parameters related to installing dependencies). But it is great to know that now, and also that is what one would expect a strict mode to do.

I have changed it to `0.6_git20150325` now, which is an accepted format. (For some context, Alpine's accepted versions are based on Gentoo's version format, because once upon a time it was built from Gentoo. And there they have a few suffixes hardcoded, and "git" seems to be the most fitting one. Also `0.6 < 0.6_git1234` so that is just what we want. There are other suffixes such as `pre`, which make the version less than the version without the suffix.).

Thank you very much for pointing this out and checking it again!

PR #540, opened 2017-09-09
Fix confusing error message for invalid makedepends

ollieparanoid:
I ran into this, and I think @PureTryOut and @craftyguy did also have this issue?

---

This happens currently, when a makedepend is invalid:

```
Traceback (most recent call last):
  File "/home/user/code/pmbootstrap/pmb/__init__.py", line 53, in main
  File "/home/user/code/pmbootstrap/pmb/helpers/frontend.py", line 84, in build
    args.buildinfo)
  File "/home/user/code/pmbootstrap/pmb/build/package.py", line 62, in package
    pmb.chroot.apk.install(args, apkbuild["makedepends"], suffix)
  File "/home/user/code/pmbootstrap/pmb/chroot/apk.py", line 197, in install
    strict=True)
  File "/home/user/code/pmbootstrap/pmb/parse/depends.py", line 88, in recurse
    if pkgname is None and strict:
UnboundLocalError: local variable 'pkgname' referenced before assignment
```

With this patch, you get the meaning full error that should have been printed instead:
> ERROR: Could not find package 'invalid-package' in the aports folder and could not find it in any APKINDEX

### DIFF
--- a/pmb/parse/depends.py
+++ b/pmb/parse/depends.py
@@ -25,7 +25,7 @@ import pmb.parse.apkindex
 def recurse_error_message(pkgname, in_aports, in_apkindexes):
     ret = "Could not find package '" + pkgname + "'"
     if in_aports:
-        ret += " aport"
+        ret += " in the aports folder"
         if in_apkindexes:
             ret += " and could not find it"
     if in_apkindexes:
@@ -64,6 +64,7 @@ def recurse(args, pkgnames, arch=None, in_apkindexes=True, in_aports=True,
         # Get depends and pkgname from aports
         logging.verbose("Get dependencies of: " + pkgname_depend)
         depends = None
+        pkgname = None
         if in_aports:
             aport = pmb.build.find_aport(args, pkgname_depend, False)
             if aport:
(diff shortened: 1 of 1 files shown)


pmb/parse/depends.py
```diff
@@ -25,7 +25,7 @@
 def recurse_error_message(pkgname, in_aports, in_apkindexes):
     ret = "Could not find package '" + pkgname + "'"
     if in_aports:
-        ret += " aport"
+        ret += " in the aports folder"
```
craftyguy: This is hard to reproduce, so I'm trying to review this based on how I would interpret this to play out.

It seems like the following message could be constructed:

> Could not find package <pkgname> in the aports folder and could not find it

From line 26 + line 28 + line 30.

ollieparanoid: That message can not be constructed, because whenever the `" and could not find it"` part gets added, the `" in any APKINDEX"` part also gets added. It has the same condition `in_apkindexes` (see lines 29, 31).

About how to trigger the error message:
Modify `aports/main/hello-world/APKBUILD` and add `invalid-package` to the `makedepends`, then run `pmbootstrap build --force hello-world`. Without this PR, it crashes with the stack written in the initial comment, and with this PR it throws a meaningful error message.

Could you confirm that this is working / is there anything else that looks suspicious?

Thanks for the review!

craftyguy:
I now get:
> RuntimeError: Could not find package 'invalid-package' in the aports folder and could
not find it in any APKINDEX


That's a lot more clear than what it is without this PR (the crash and ambiguous message).

PR #541, opened 2017-09-09
Allow -y option for subcommands so that it is more 'natural'

PabloCastellano:
Allows user to run `./pmbootstrap install -y` besides `./pmbootstrap -y install`


ollieparanoid:
This makes it pretty inconsistent, as we can not specify the other arguments (`-j`, `-v`, ...) as parameters for the subcommands. Maybe there is a general way to allow these arguments to be placed before and after the action with argparse?

Thanks for this PR!

ollieparanoid:
So... I couldn't find out how to do this with Python's argparse in a way, that it works for all arguments. If someone does find a way and wants to add this feature, please make a new PR (this one has a conflict now). Thanks!

PR #553, opened 2017-09-10
Replaced Weston variables in deviceinfo for seperate weston.ini

### LINKED ISSUE #499: Weston config should be loaded on start
Currently the config of weston is loaded post-install. This is done in [postmarketos-ui-weston.post-install](https://github.com/postmarketOS/pmbootstrap/blob/314c17e03cf8cddfd0f385d9db2f23f76f9a0418/aports/main/postmarketos-ui-weston/postmarketos-ui-weston.post-install).

Whenever changing one of the weston config values in the deviceinfo files, you should explicitly reinstall `postmarketos-ui-weston` (this is not done automatically). However, this does not make much sense because the deviceinfo file is in a `device-*` package, so you suspect only having to rebuild the `device-*` package. 

Therefore I propose to load the weston config when starting weston, in the [startup_weston.sh script](https://github.com/postmarketOS/pmbootstrap/blob/314c17e03cf8cddfd0f385d9db2f23f76f9a0418/aports/main/postmarketos-ui-weston/start_weston.sh).

PureTryOut:
From now on, if a device needs custom Weston variables (different renderer backend, Xwayland disabled, etc), this will be done in `weston.ini` installed by the device package. The Weston launch script will automatically detect if a custom `weston.ini` is available and use it.

The defaults for now are the DRM backend, Xwayland enabled, and the evdev keyboard ruleset.

This change also makes Weston load the config on boot. It means that if you change variables in your device package, you only need to reinstall that package instead of postmarketos-ui-weston as well.

Resolves #499


drebrez:
@PureTryOut I counted 20 devices with the fbdev-backend and 1 device with drm-backend.
IMHO having the fbdev-backend in the default weston.ini would be better.

craftyguy:
The reason we originally decided to not ship a weston.ini for each device and the reason I implemented the original code for installing device-specific weston options into weston.ini is because this will overwrite an existing weston.ini on disk if it exists. One example would be if a user installs pmos, uses it for a while (incl. customizing weston.ini), then received a device package update which would then proceed to blow away their changes.

PureTryOut:
@drebrez I believe the goal is to get as many devices as possible to run on DRM. Because of this I'd set that to be the default, and change it per device if it doesn't support it yet. Yes, *right now* most devices run on fbdev, but (hopefully) in the future this will not be the case.

@craftyguy This discussion started in #499, please continue there.

drebrez:
@PureTryOut thanks for the clarification about the goal, didn't know it 👍. Now I'm interested in this DRM and I will investigate what steps I have to do in order to enable it on my devices.

aports/main/weston/APKBUILD
```diff
@@ -92,6 +92,8 @@ xwayland() {
 	mkdir -p "$subpkgdir"/usr/lib/$_libdir || return 1
 	mv "$pkgdir"/usr/lib/$_libdir/xwayland.so \
 		"$subpkgdir"/usr/lib/$_libdir
+
+	echo "xwayland=true" >> "$srcdir"/weston.ini
```
ollieparanoid: Why is this necessary?

PureTryOut: It isn't, it's left over from me missing around. I removed it

Issue #513, opened 2017-09-04
Upstream linux-firmware package contains outdated and vulnerable Wi-Fi drivers


zhuowei:
Alpine's [linux-firmware](https://pkgs.alpinelinux.org/package/edge/main/x86/linux-firmware) package pulls from the upstream linux-firmware repo: the Wi-Fi drivers available there are [very old](https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/log/brcm). For example, the BCM4358 firmware available there was built in September 2016 and is vulnerable to [Broadpwn](https://blog.exodusintel.com/2017/07/26/broadpwn/). This is a problem as [four devices](https://github.com/postmarketOS/pmbootstrap/search?utf8=✓&q=linux-firmware&type=) use the outdated linux-firmware blobs.

Should we use manufacturer Wi-Fi firmware or [AOSP firmware](https://android.googlesource.com/platform/hardware/broadcom/wlan/+log/android-8.0.0_r4/bcmdhd/firmware) if they are newer, or somehow get Broadcom to release updated Wi-Fi drivers to the linux-firmware repo?

ollieparanoid:
###### +CC @ncopa (maintainer of `linux-firmware` - *is it OK when I do it like that, or would you prefer to be notified in another way?*)


@ncopa: How does Alpine handle this issue, would you guys consider "patching" (using another blob) in the `linux-firmware` package?

@zhuowei: For postmarketOS, we can use any source (e.g. AOSP like you have linked) for firmware instead in case we need Broadcom firmware.

Of course this should also be fixed upstream in the kernel. I recommend to do some research first - has this been discussed on the Linux kernel mailing list already? Who is the responsible person for the drivers in the kernel (as the kernel has maintainers for everything)? - Then either joining the discussion on the LKML or starting a new one/writing the maintainer. I am not involved in kernel development though, this is how I think it should be done with my limited knowledge from an outside view.

Thank you for pointing this out! I would appreciate it very much if you tried to get this fixed in the kernel (and of course also in pmOS depending on how it is handled in Alpine, ideally we can continue using the Alpine package or even help our upstream friends resolve it!).

zhuowei:
[...]

Issue #182, opened 2017-07-18
Default WL_CALIBRATION


MartijnBraam:
It looks like most phones are very close to being calibrated already and I think the calibrations are already done in the touchscreen controller themselves.

For example the calibration for the htc ace:
```
ENV{WL_CALIBRATION}="0.963488 0.004947 14.422623 -0.076271 1.026566 -7.303467"
```
Is very close to a empty calibration:
```
ENV{WL_CALIBRATION}="1 0 0 0 1 0"
```

The formula from the calibration values is:

```
x' = Ax + By + C
y' = Dx + Ey + F
```
The relative big change in the third and last parameter is because those are in pixel values and those are quite close together on phones.

Maybe it's better to use the emtpy calibration for most phones instead the one copied from the mako while porting.

craftyguy:
Just an idea, but what if weston-calibrator is called once on 'first boot' to generate the necessary values for WL_CALIBRATION? This way users have a chance to adjust their screen without having to rely on someone who may not be tapping completely straight :P

MartijnBraam:
Yeah but the point is that most devices should work perfectly fine without a calibration, just by using the empty calibration which maps the pixels from the touchscreen exactly to the display pixels. The only devices that should require user calibration are the ones with a resistive touchscreen like the n900

drebrez:
@MartijnBraam The last device I ported, huawei y530, I didn't added the WL_CALIBRATION and worked fine. I will try to remove it also from my other device and see if they work fine 👍

montvid:
All the devices need empty callibration because if you rotate the screen you need to recalibrate the touchscreen. It is not done automaticaly. At least not in Mate and Hildon.

ollieparanoid:
@montvid: I am not 100% sure, but I think this config is only being used by Weston, and not by X11 (maybe through libinput? but nobody reported that being the case).

craftyguy:
libinput uses a different calibration variable: `LIBINPUT_CALIBRATION_MATRIX`

I configure it on the nokia-rx51 device in a udev rule so that touch works correctly under X11 (when using the libinput driver)

montvid:
Weston and X11 Libinput uses the udev values in the file to calibrate the ts. I use this for X11 and it works fine:
https://github.com/postmarketOS/pmbootstrap/blob/master/aports/device/device-asus-flo/90-android-touch-dev.rules

montvid:
[...]

PR #555, opened 2017-09-11
angler: support Wi-Fi and graphics

zhuowei:
With these patches, the Nexus 6P supports Wi-Fi and ~almost~ actually supports Weston.

Issues/questions:

- ~Graphics still don't work properly: I had to run `while [ ! ]; do cat /sys/class/graphics/fb0/modes > /sys/class/graphics/fb0/mode; done`. Not sure [how the Nexus 5 fixed this](https://github.com/postmarketOS/pmbootstrap/pull/66); I guess I can add a patch in Weston to automatically run this (or the equivalent ioctl) every frame. (With that command running, graphics work well.)~ msm-fb-refresher fixes this
- ~(Note that the Android recovery does send this ioctl every frame for double buffering)~
- ~KGSL is disabled in the kernel, so freedreno won't work, and everything's software rendered.~ KGSL is not disabled in the kernel anymore, although I haven't enabled the KGSL_DRM compat layer in the kernel yet, nor have I packaged the a440 blobs, so freedreno still won't run. That's for later.
- ~Should a kernel have NFS enabled? I had it enabled for testing.~ - based on review, yes.
- ~Wi-Fi uses the blobs from linux-firmware, which are old; I can package the AOSP blobs (used on this device's stock firmware) that are more up to date if you want. (see the discussion on #513)~ newer blobs packaged.
- The Wi-Fi calibration file is different for various revisions of this hardware; I've added the one that was for my device, but it may not work for earlier/later revisions. Android chooses the correct calibration file at boot in the init.angler.rc.

Tested over an NFS mounted root, but it should work on regular devices.


zhuowei:
@ollieparanoid I ended up making a Weston patch (#558) as I'm not sure how to modify the kernel. Do you have any advice on fixing the kernel instead?

Re. AOSP blobs: should I package all of them or just the bcm4358 needed for angler?

ollieparanoid:
I think the weston patch is a good solution, if I understand correctly the double-buffering isn't a bug, but valid usage of the framebuffer API.
In my opinion, it is enough to package just the blob needed for angler.

aports/device/device-huawei-angler/90-android.rules
```diff
@@ -1,2 +1,7 @@
 # udev rules file
 # All device names can be read from weston's logfile (/tmp/weston.log in postmarketOS)
+# Not calibrated for now.
```
drebrez: @zhuowei is the "Not calibrated for now" comment still valid?

aports/device/device-huawei-angler/deviceinfo
```diff
@@ -29,4 +29,4 @@ deviceinfo_flash_offset_tags="0x01e00000"
 deviceinfo_kernel_cmdline="androidboot.hardware=angler androidboot.console=ttyHSL0 msm_rtb.filter=0x37 ehci-hcd.park=3 lpm_levels.sleep_disabled=1 boot_cpus=0-3"
 deviceinfo_flash_pagesize="4096"
 
-deviceinfo_weston_core_modules="xwayland.so"
+deviceinfo_weston_core_modules=""
```
drebrez: @zhuowei you removed the "xwayland.so" by purpose? It doesn't work for this device?

zhuowei: It does work; I removed it for debugging; I'll add it back

aports/device/device-huawei-angler/90-android.rules
zhuowei: It seems that the default `1 0 0 0 1 0` calibration works the best, actually; I'll switch it over

zhuowei:
Rebased, squashed into two commits, updated to latest AOSP Wi-Fi firmware blobs, enabled msm-fb-refresher.

With these changes, Nexus 6P supports screen output and Wi-Fi (yay!)

This is ready for review and merge.

Note that I only tested this with NFS: I'll probably try flashing the built image later. Some newer devices require a sparse image for flashing, but I haven't tested this for Nexus 6P yet.

aports/device/device-huawei-angler/90-android-touch-dev.rules
```diff
@@ -1,2 +1,7 @@
 # udev rules file
 # All device names can be read from weston's logfile (/tmp/weston.log in postmarketOS)
+# Use default calibration for now.
+SUBSYSTEM=="input", ATTRS{name}=="synaptics_dsx", \
+ENV{ID_INPUT}="1", ENV{ID_INPUT_TOUCHSCREEN}="1" \
+ENV{ID_INPUT_TOUCHPAD}="" ENV{ID_INPUT_KEY}="" \
+ENV{LIBINPUT_CALIBRATION_MATRIX}="1 0 0 0 1 0"
```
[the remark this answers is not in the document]
zhuowei: it's to explicitly indicate that we are using default config; it's not actually needed.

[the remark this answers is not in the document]
ollieparanoid: This is fine for me. we could discuss in #182 whether we prefer the explicit notation or leaving it out and then apply it to all devices at once.

ollieparanoid:
In my opinion, this PR is ready to be shipped! :ship: Would it be OK for you if we squashed the two commits into one? That is how we usually do it. If not, here are some options: https://github.com/postmarketOS/pmbootstrap/pull/481#issuecomment-328563246.

zhuowei:
@ollieparanoid of course. I had a separate commit for the firmware to make review easier; feel free to squash into one.

Issue #35, opened 2017-06-01
[hammerhead] Make `pmbootstrap flasher flash_kernel` work


ollieparanoid:
Hi @dzervas and @MartijnBraam,

amazing progress with hammerhead! Wifi is working? Wow! :D

So there is `pmboostrap flasher boot`, which should do the same as the command specified in the wiki: `fastboot boot /mnt/rootfs_lg-hammerhead/boot/vmlinuz-lg-hammerhead /mnt/rootfs_lg-hammerhead/boot/initramfs-lg-hammerhead`

Does that work?

(Also, can you upload photos and maybe share them on /r/postmarketOS?)

MartijnBraam:
The fastboot boot command isn't what currently makes it boot, I shall update the wiki.

The workaround at the bottom of the wiki with generating a seperate boot.img and flashing that to the boot partition or starting it with fastboot boot boot.img does work. The automatic image creation from fastboot seems to have issues.

ollieparanoid:
To fix that, we could add a script, that builds the bootimg in a temporary folder with [mkbootimg](https://github.com/postmarketOS/pmbootstrap/tree/master/aports/mkbootimg), and then flashes that bootimg.

If someone wants to try it:
* modify the fastboot package and add such a shell script
* [change the config(https://github.com/postmarketOS/pmbootstrap/blob/master/pmb/config/__init__.py#L202), so the script gets called instead of fastboot directly (look at the heimdall part - there are also scripts getting called, instead of heimdall directly)

ollieparanoid:
Is the only problem here the kernel commandline, that does not get passed to fastboot yet (so this would be fixed in #50)?
Or is fastboot really creating broken bootimg files, and we must use `mkbootimg` instead, and pass the generated image to fastboot?

MartijnBraam:
I still use mkbootimg manually to flash my images.

ollieparanoid:
I know, but is it really necessary, or is the actual reason missing kernel commandlines in `pmbootstrap flasher`?

MartijnBraam:
It's necesairy because fastboot cannot live-create a bootable image for me for some reason, If I don't specify the cmdline in mkbootimg then I still get a flashable image, If I let fastboot create the boot.img then it will complain about the partition table.

lawl:
`flasher boot` works for me on my hammerhead

MartijnBraam:
[...]

PR #556, opened 2017-09-11
Added base offset to hammerhead deviceinfo

MartijnBraam:
This fixes flashing the kernel on my hammerhead

Closes: #35


ollieparanoid:
I assume @PabloCastellano clicked it by accident.

PR #557, opened 2017-09-11
Unify names of scripts and patches

PabloCastellano:
- 90-android.rules -> 90-android-touch-dev.rules
- reduce_lzo_compression.patch -> 02_reduce_lzo_compression.patch
- mdss_fb_refresh_rate.patch -> 02_mdss_fb_refresh_rate.patch
- aeea3592a13bf12861943e44fc48f1f270941f8d.patch -> 00_fix_return_address.patch
- gpu-msm-fix-gcc5-compile.patch -> 02_gpu-msm-fix-gcc5-compile.patch
- udev-hack.sh -> udev-v4l-hack.sh
- Separate udev-v4l-hack.sh from profile.sh


MartijnBraam:
should we also unify the underscores and dashes in the file names? 
Why do the touch dev rules need android in the name? I guess it can be called `90-wayland-touch-dev.rules`
Also sorry for that patch with a git hash as name.

ollieparanoid:
I've merged this PR directly, because it will conflict pretty soon if something gets changed in the repo, and then it might be a lot of work to get it right again. It already improves the situation as it is.

---

> should we also unify the underscores and dashes in the file names?

That would be sort of nice (but not necessary in my opinion), we could do that in a new PR (and also establish a convention, that we write down in the wiki?)

> Why do the touch dev rules need android in the name? I guess it can be called 90-wayland-touch-dev.rules

That sounds good!

PabloCastellano:
@ollieparanoid Thanks for merging quickly. I was afraid of how many conflicts I would cause

@MartijnBraam Yes, keeping android in the name makes no sense, but since I don't know where it comes from, I unified them and left it. I don't really mind about underscores/dashes as long as the files have the same name

Issue #375, opened 2017-08-13
[flame] Screen output flicker and inactivity


yuvadm:
The Flame device screen is essentially inactive. A very brief flicker appears upon boot, but immediately disappears and screen turns blank. Debugging this by entering a `loop_forever` right after `show_splash` yielded the same behavior.

In `dmesg` the following (truncated) output shows that framebuffer seems to load correctly:

```
[    0.000000] Node qcom,mdss_fb_primary memblock_reserve memory 3200000-3a00000
[    0.740375] mdss_dsi_panel_init: Panel Name = ili9806c wvga video mode dsi panel
[    0.740634] mdss_dsi_panel_init:1033 Continuous splash flag enabled.
[    0.740647] mdss_dsi_panel_init:1045 Partial update disabled.
[    0.741262] mdss_register_panel: adding framebuffer device fdd00000.qcom,mdss_dsi
[    0.768116] mdp3_alloc: allocating 3280896 bytes at c3200000 (3200000 phys) for fb 0
[    0.768510] mdss_fb_register: FrameBuffer[0] 480x854 size=3280896 registered successfully!
[    0.768728] Registered led device: lcd-backlight
```

There are also some errors coming from `msm_vidc` which do not appear on a normal B2G boot, but do appear on a B2G recovery boot and the screen works there, so they can possibly be ignored (might be related to more advanced screen functions irrelevant for the framebuffer driver)
```
[    3.390004] msm_vidc: 1: Failed to read qcom,load-freq-tbl from device tree
[    3.390017] msm_vidc: 1: no elements in frequency table
[    3.390027] msm_vidc: 1: Failed to read qcom,reg-presets from device tree
[    3.390106] msm_vidc: 1: Failed to read qcom,buffer-type-tz-usage-table from device tree
[    3.390127] msm_vidc: 1: Q6 hfi device probe called
```

All sysfs parameters seem to be set correctly when diffed against the B2G params, except for one param that is set manually to match:

```bash
echo "480,1708" > /sys/devices/virtual/graphics/fb0/virtual_size
```

`fbset` output:

```bash
mode "480x854-4"
	# D: 2.252 MHz, H: 3.263 kHz, V: 3.650 Hz
	geometry 480 854 480 1708 32
	timings 444139 100 100 8 20 10 12
	accel false
	rgba 8/24,8/16,8/8,8/0
endmode
```

This workaround didn't seem to help either:

```
[...]

PR #558, opened 2017-09-11
weston: add a patch to force FBIOPUT on Qualcomm devices

zhuowei:
Qualcomm's framebuffer is designed for double buffering, and only updates the screen
after an FBIOPUT ioctl is submitted to flip the buffer. Detect this to force Weston
to send this ioctl at the end of every frame even when single buffering.

This should be equivalent to the
`cat /sys/class/graphics/fb0/modes > /sys/class/graphics/fb0/mode`
hack.

This is tested on Nexus 6P (angler), and Weston is able to correctly draw the screen without the cat modes hack. On non-Qualcomm devices, this shouldn't activate, but I haven't tested there.

Edit: for testers: in /tmp/weston.log, if you have a device that needed the cat modes hack, applying this should say
`Force FBIOPUT workaround for Qualcomm framebuffer is enabled`
for other devices it should say
`Force FBIOPUT workaround for Qualcomm framebuffer is disabled`

### DIFF
--- a/aports/main/weston/0007-compsitor-fbdev-add-Force-FBIOPUT-workaround-for-MSM.patch
+++ b/aports/main/weston/0007-compsitor-fbdev-add-Force-FBIOPUT-workaround-for-MSM.patch
@@ -1,0 +1,102 @@
+From 45e92a17c7006d36a54bfbc2f4a35281fbd7b58c Mon Sep 17 00:00:00 2001
+From: Zhuowei Zhang <zhuoweizhang@yahoo.com>
+Date: Mon, 11 Sep 2017 11:13:18 -0400
+Subject: [PATCH] compsitor-fbdev: add Force FBIOPUT workaround for MSM
+ framebuffer
+
+Qualcomm MSM devices have a framebuffer that only updates the display after
+a FBIOPUT, as they are intended for double buffered operation only.
+Add a hack to do a FBIOGET/FBIOPUT pair every frame to force an update.
+---
+ libweston/compositor-fbdev.c | 40 +++++++++++++++++++++++++++++++++++++++-
+ 1 file changed, 39 insertions(+), 1 deletion(-)
+
+diff --git a/libweston/compositor-fbdev.c b/libweston/compositor-fbdev.c
+index b7d2c74..bce8d68 100644
+--- a/libweston/compositor-fbdev.c
++++ b/libweston/compositor-fbdev.c
+@@ -75,6 +75,8 @@ struct fbdev_screeninfo {
+ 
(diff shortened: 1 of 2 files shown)


ollieparanoid:
Thanks! Could someone test this? (QC and non-QC device)

@zhuowei: so as I understand, this *is* standard behavior, and weston just doesn't support it, right? Could you upstream this to weston then?

zhuowei:
@ollieparanoid This is _not_ standard behaviour; Qualcomm's framebuffer driver seems to be incomplete: Android Recovery always uses double buffering, so they didn't implement single buffering support. As discussed in #555 one should probably fix the kernel instead, but I think this is a reasonable stopgap solution while we figure out how to modify Qualcomm's framebuffer driver.

drebrez:
@zhuowei looking at the wiki search result we have a lot of Qualcomm devices (https://wiki.postmarketos.org/index.php?search=cpu+%3D+Qualcomm). 
Does this bug affects specific models only or it's a generic driver issue? 
Maybe you can compare the source code with the one of a working device.

MartijnBraam:
I think this only affects the equalcomm devices using mdss

ollieparanoid:
Cool, thanks for explaining @zhuowei! Yeah, using this fix as a solution until we figure out how to patch the kernel sounds acceptable to me, better than the while true loop. (We'll still need to test this though, I might get around to it at some point, but I'm pretty busy right now with answering all the fantastic input in the tracker 👍)

yuvadm:
After just now getting the Flame screen working (#375), I tested the screen with these two hacks disabled https://github.com/postmarketOS/pmbootstrap/blob/594ce64321d4f775a81c4b0c527282b4803bd8c0/aports/device/device-t2m-flame/initfs-hook.sh#L4-L5 and have not yet applied the patch provided here, but the screen still works properly.

zhuowei:
@yuvadm which kernel patches are you using? Can I try them?

I'm going to close this PR for now and try the kernel patches from the other Huawei device, the Moto G, and possibly the Flame to see if they help.

yuvadm:
@zhuowei sure, take a look at the current branch I'm testing https://github.com/postmarketOS/pmbootstrap/tree/device/t2m-flame

zhuowei:
@yuvadm I tried applying the mdss off_fnc patch, tried porting it to the mdp5 connection used in this phone, and tried applying the pixel clock patch, but none of those seemed to work. Any advice?

drebrez:
@zhuowei did you try enabling the mdp3 driver? `CONFIG_FB_MSM_MDSS_MDP3`

zhuowei:
@drebrez pretty sure this phone uses mdp5.

PR #559, opened 2017-09-11
Enable virtual terminal, update display/touchscreen settings

flacks:
At the moment, video output only works when the following script is run:
`while [ ! ]; do cat /sys/class/graphics/fb0/modes >
/sys/class/graphics/fb0/mode; done`.


zhuowei:
The Huawei Y530 and  Mozilla Flame ports seem to have fixed the `cat modes` thing: see #375

PR #560, opened 2017-09-11
T2M Flame updates

### LINKED ISSUE #375: [flame] Screen output flicker and inactivity

yuvadm:
- Fixes #375 
- Removes unneccesary hacks
- Fixes bug in firmware package after rename
- Fix touchscreen


aports/device/device-t2m-flame/initfs-hook.sh
```diff
@@ -1,5 +1,2 @@
 #!/bin/sh
```
drebrez: @yuvadm if the initfs-hook script is not necessary anymore, I suggest removing it completly

yuvadm: I thought I'd keep it since it might become useful again at some point, but I guess I can delete it.

yuvadm:
@drebrez done, thanks for the review!

yuvadm:
@ollieparanoid thanks for the feedback, done!

PR #561, opened 2017-09-11
Strip kernel modules

drebrez:
Adding INSTALL_MOD_STRIP will cause modules to be stripped after
they are installed (option --strip-debug will be used)


craftyguy:
Here are some *really* good reasons why this was never enabled by default: https://github.com/torvalds/linux/commit/2ea038917bbdd51a7ae4a898c6a04641324dd033

I rely on ccache & distcc because, as crazy as this sounds, I frequently have to compile on my device and using these tools is the only way I can get anywhere within my lifetime :)

drebrez:
@craftyguy I'm not sure what's your position about having the INSTALL_MOD_STRIP set to 1 during the module installation.

ollieparanoid:
As I understand it, @craftyguy  is against it, because of that commit he found. I wonder if this is still accurate though, because that commit is from *2009*.

1. Compatibility issues with a binutils `2.15.92.0.2` and lower is outdated, we're at `2.28` now.

2. Ccache and distcc issues - I wonder why that would happen in the first place. As I understand, that command affects the `make install` part only, so why would `ccache` or `distcc` be involved anyway? Too bad that the commit message doesn't specify this more clearly.

3. Installed modules increased in size: @drebrez checked that this is not the case (and it sounds like a bug that has been resolved - why would they get any *bigger* with stripped symbols?)

@drebrez even said that he would rebuild all kernels where he placed that option, just to make sure.

So from my understanding, we don't have a risk here and that message is outdated/too specific to that commit to be relevant for us (I guess the ccache/distcc part is messed up, because its wrappers caused the detection of the strip binary to fail or something). I could not find *anything* besides that commit message (and [this thread quoting it](https://superuser.com/a/705190)) on the net, that points out that there is a general problem with stripping kernel modules and ccache/distcc.

@craftyguy, could you run some tests, that prove that this fails with ccache/distcc, or find another source that says so? Otherwise I would like to merge this if you and everyone else are okay with it.

craftyguy:
Since there's a lot of burden of proof for me to prove that distcc and ccache don't break with this change (this will take some time to verify!), I'd like to understand why you feel this change is necessary in the first place. Having symbols in place for debug is hugely helpful, most especially in the case of pmOS where we are exercising modules that haven't really seen widespread attention or are very new to the kernel. I'm not aware of any major benefit in stripping modules, other than perhaps saving a handful of MB in rootfs.

drebrez:
@craftyguy @ollieparanoid I tested the compilation of all the involved kernels, I tested the linux-postmarketos kernel with qemu and it works, the linux-samsung-i9070 also works.
The INSTALL_MOD_STRIP option is only set in the `make modules_install`, it doesn't compile them again, so I think ccache/distcc are not affected.
The kernel modules seems to contain a lot of debugging stuff, look at the package size:
before  >>> `linux-postmarketos*: Package size: 293.7 MB`
after     >>> `linux-postmarketos*: Package size: 17.3 MB`
If you need to debug a problematic kernel module you can simply remove the INSTALL_MOD_STRIP option and compile it again.

craftyguy:
I guess my point is that it's probably too early to start optimizing out debug info. I'm still heavily involved with debugging things on the N900 that are related to the kernel and its modules, and it seems like the qemu camp is also trying to get DRM working (again, a kernel module). 

For me, this change would be an annoyance because I'd have to remember to manually edit the kernel APKBUILD after every update to it to remove this flag, and when I inevitably forget and install stripped modules when trying to debug, there will be much cursing :P

Once pmOS is "sufficiently stable", then this change would, as you pointed out, save a few MB in rootfs for folks using pmOS as a daily driver :)

ollieparanoid:
Yeah, the benefit would be saving 270 (!) MB here. But if it makes debugging harder for you, then I think we should *not* do it. Sorry @drebrez, I didn't think about this aspect enough.

drebrez:
@craftyguy is it ok for you if I revert the changes for the linux-postmarketos, which is used in qemu and like you say it's more useful to have it with the debug stuff, but strip the kernel modules of the devices which reduces the size and improves the flashing speed?

craftyguy:
@drebrez I'm fine with that proposed change because I only use linux-postmarketos and none of the other kernels, but you may want to get buy-in from other device maintainers who are working with those kernels 😄

ollieparanoid:
Merged. In case someone isn't happy with it, please open a PR that reverts it for your device, or start a discussion as a new issue. Thanks for the contribution @drebrez!

PR #565, opened 2017-09-12
flasher flash_system: add possibility to select partition

### LINKED ISSUE #484: Add command to flash system image to userdata partition
This would be nice for devices with a small system partition. Either `flash_userdata` (which might be taken by some other functionaly in the future if we implement LVM) or `flash_system --to-userdata`

drebrez:
Add argument `--partition` to `pmbootstrap flasher flash_system` to specify a partition to flash the system image

Closes #484


montvid:
How does it work? --partition=userdata and --partition userdata do not work for me.

drebrez:
@montvid the command should be `pmbootstrap flasher flash_system --partition userdata`, but you have to replace **userdata** with the name of the partition you want to use in your device (most devices have a partition called userdata, but might be different, for example in the [Samsung Galaxy S Advance](https://wiki.postmarketos.org/wiki/Samsung_Galaxy_S_Advance_(samsung-i9070)#Device_partitions) is called **datafs**)

montvid:
parted /dev/mmcblk0 print
30      2416MB  15.8GB  13.3GB               userdata
How can i find the name of my partition?

montvid:
@drebrez I used to have pmos system on /system partition but i want it in /userdata partition. The problem might be that the kernel is loading the /system partition instead of /userdata partition. How do I delete my /system partition from pmos? I want to see will it load from /userdata then.

ollieparanoid:
That sounds very plausible! when you have SSH or telnet access (e.g. via the usb-shell hook), you can use `dd` to overwrite the beginning of the `system` partition with zeros, that should do it. Be very careful in finding the right partition!

Alternatively, you could also flash an Android image on the `system` partition.

montvid:
Thanks guys, solved it by flashing twrp to /recovery and formating the /system partition with twrp.

PR #567, opened 2017-09-13
Feature/rx51 ofono

### LINKED ISSUE #520: [rx51] integrate ofono
ofonod v1.6 from Debian works with Nokia N900, given right kernel. It should be possible to integrate it to pmos, too.

@craftyguy

craftyguy:
Note: There are two changes to the Alpine upstream ofono package in this PR:

1) Apply two patches necessary for enabling the N900 modem. These patches are already merged into the master branch of ofono.

2) Install the ofono udev rules

I suggest we wait to try to upstream (to Alpine) these APKBUILD changes until ofono 1.21 is released, so then we're just upstreaming udev rules install and not the patches.

fixes #520


pavelmachek:
I have checked it on my N900, and it seems to work. I can power on modem, etc. Rest was not tested as this phone does not have SIM card.

MartijnBraam:
I'm getting a nice mini-sim for testing in a few days so I can test this.

craftyguy:
Thanks @ollieparanoid !

I should have mentioned this earlier, but you'll need to do some setup through ofonod's dbus interface to actually get grps and/or SMS up and running: https://wiki.postmarketos.org/wiki/Nokia_N900_(nokia-rx51)#GPRS_.2F_SMS

Ideally, this would be handled by something like connman or networkmanager. NM claims, as of 1.4, to support ofonod, but I haven't been able to find any documentation on how to actually set it up. Soo.... connman will probably be my initial focus.

Issue #551, opened 2017-09-10
New init configuration option for free space on the image file


BrianOtto:
I am currently running **pmbootstrap** in **qemu** and trying to compile a very large application after it boots up. The 20% extra space that #336 automatically adds is not enough. I have to hard code a larger size in the `get_subpartitions_size()` function at `pmb/install/install.py`.

e.g.

```
full *= 1.20

# my hard coded value
full += 2200 * 1024 * 1024

boot += 15 * 1024 * 1024
return (full, boot)
```

It would be nice if this was a configuration option that was asked during `/pmbootstrap.py init`.

ollieparanoid:
It is possible to resize the installation image, after it has been created as documented in the [QEMU](https://wiki.postmarketos.org/wiki/Qemu) wiki article. But I understands, that as we have it now, this is not really intuitive and you are not the first one who asked this so we should improve the situation.

For some context, the initramfs of pmbootstrap automatically adjusts the partition table and filesystem inside the image to take up the full space, in case it takes up less space already. That means, that we don't need to specify the size at the time of generating the image, but we could increase it at any later point as necessary.

I propose the following:
* Add a `--image-size` option, that can be used with qemu like this: `pmbootstrap qemu --image-size=1G`. It would run `truncate` on the image as described in the wiki article.
* Inform the user when starting `pmbootstrap qemu` without that parameter with a message printed to the screen - something like:
> NOTE: Run 'pmbootstrap qemu --image-size=...' to increase the system image size when you run out of space!

(Maybe someone can come up with a better name than `--image-size`?)

If you agree with this solution, would you take a shot at it @BrianOtto?
Thanks for opening this issue!

BrianOtto:
I didn't realize you could resize the image after it had been created. This solution sounds good to me!

I can take a shot at making the changes. The `--image-size` name make sense too, but I'll post here if any other ideas come up.

BrianOtto:
PR #568 has been created, tested and confirmed working on my setup.

PR #568, opened 2017-09-13
Added a new option that allows you to increase the system image size for qemu

BrianOtto:
I have added a new `--image-size` option for qemu, as described in #551.

I feel like `--increase-image-size` is a more accurate name for what the option does, but it seemed a little long to type. Let me know if you think it should be changed, or if you have any better ideas.

### DIFF
--- a/pmb/qemu/run.py
+++ b/pmb/qemu/run.py
@@ -19,6 +19,7 @@ along with pmbootstrap.  If not, see <http://www.gnu.org/licenses/>.
 import logging
 import os
 import shutil
+import re
 
 import pmb.build
 import pmb.chroot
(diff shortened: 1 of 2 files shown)


drebrez:
@BrianOtto Nice feature, I agree with you that `--image-size` might be confusing, since the value specified gets added to the actual image.
In theory we can have both arguments, `--increase-image-size` that increase the image by the amount specified and `--image-size` which sets the image to that specific size, it will only change the `truncate` command to have the `+` or not.
By setting the specific size we should add a check to not allow to specify a smaller value than the actual image size, this because the initramfs script which doesn't handle the shrink of the partitions, we can think about it in the future.
@BrianOtto and @ollieparanoid , what you think about having both arguments?

BrianOtto:
@drebrez I am okay with adding two arguments.

Another alternative is maybe we just make it behave as you would expect, but with the safety check @drebrez mentioned, i.e.
> `--image-size` which sets the image to that specific size, it will only change the truncate command to have the + or not.
By setting the specific size we should add a check to not allow to specify a smaller value than the actual image size, this because the initramfs script which doesn't handle the shrink of the partitions, we can think about it in the future.

If we do it this way, then you get the additional benefit of not accidentally increasing the image size when running it again from your command line history. I actually ran into that a few times when testing this update :)

Having said that, there is one issue we could run into. If I set the size to say 1GB on one run, but then decide to decrease it on the next run, how do I know the original image size now? On the first run, we know this by doing a os.path.getsize(img_path), but on the 2nd run this will be 1GB (i.e. the **increased size**). Is the original image size (base image + size of all packages) stored in cache somewhere?

ollieparanoid:
> If we do it this way, then you get the additional benefit of not accidentally increasing the image size when running it again from your command line history. I actually ran into that a few times when testing this :)

That's a very good argument! So If I understood you right, you're proposing that we only use one argument, and that sets the absolute size? That sounds like a great solution to me!

About the problem you mentioned, I would never allow to shrink it again. Because as soon as you boot it after resizing, you have no guarantee that there hasn't been any data written to the increased space.
After all, re-building the image is just less than a minute of zap and install away :)

BrianOtto:
> So If I understood you right, you're proposing that we only use one argument, and that sets the absolute size?

Yes, that's correct.

> About the problem you mentioned, I would never allow to shrink it again. Because as soon as you boot it after resizing, you have no guarantee that there hasn't been any data written to the increased space.
After all, re-building the image is just less than a minute of zap and install away :)

Good point, yes, you wouldn't want to shrink it... and it is easy to re-build too!

Great, thanks everyone! I will make this change.

BrianOtto:
The updates have been made. Let me know what you think.

pmb/qemu/run.py
```diff
@@ -154,6 +155,42 @@ def run(args):
     command = qemu_command(args, arch, device, img_path)
 
     logging.info("Command: " + " ".join(command))
+
+    if args.image_size:
+        # current image size in bytes
```
ollieparanoid: The new block is quite large, so I would move it to an extra function, like this:
```python
def resize_image(args, img_size_new):
    """
    (Description of the function)

    :param img_size_new: new image size in M or G
    """
    
    # ...


def run(args):
    # ...
    if args.image_size:
        resize_image(args, args.image_size)
    # ...

```

BrianOtto:
The `--image-size` logic has been moved to a new function now.

ollieparanoid:
I like how well commented everything is! Thank you very much, tested and working. :ship:

PR #569, opened 2017-09-13
Reorder entries in the menu, so that shutdown is always available,

pavelmachek:
even on N900.


pavelmachek:
Package updated.

... set the `pkgrel` to the next higher number, and `pkgrel` back to `0`...

I assumed first one should be pkgver. Let me know if it was other way around.

drebrez:
@pavelmachek your assumption is correct, FYI  https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#pkgrel

ollieparanoid:
You understood it perfectly, sorry for the confusing message.

PR #571, opened 2017-09-13
Support for Ed25519 SSH keys

flacks:
Check for an Ed25519 SSH key if an RSA SSH key isn't found.


pmb/install/install.py
```diff
@@ -125,13 +125,20 @@ def copy_ssh_key(args):
     """
     Offer to copy user's SSH public key to the device if it exists
     """
-    user_ssh_pubkey = os.path.expanduser("~/.ssh/id_rsa.pub")
+    user_ssh_rsa_pubkey = os.path.expanduser("~/.ssh/id_rsa.pub")
+    user_ssh_ed25519_pubkey = os.path.expanduser("~/.ssh/id_ed25519.pub")
     target = args.work + "/chroot_native/mnt/install/home/user/.ssh"
-    if os.path.exists(user_ssh_pubkey):
-        if pmb.helpers.cli.confirm(args, "Would you like to copy your SSH public key to the device?"):
+    if os.path.exists(user_ssh_rsa_pubkey):
```
ollieparanoid: You're duplicating quite a few lines here. How about looping over the possible pub keys?

Something like that:
```python
for key in ["rsa", "ed25519"]:
    user_ssh_pubkey = os.path.expanduser("~/.ssh/id_" + key + ".pub")
    if not os.path.exists(user_ssh_pubkey) or not pmb.helpers.cli.confirm(args,
        "Would you like to copy your " + key.upper() + " key to the device?"):
        continue
    # ... copy stuff
```

flacks: You're right. Let me know what you think of the follow-up patch.

pmb/install/install.py
```diff
@@ -125,16 +125,18 @@ def copy_ssh_key(args):
...
-            pmb.helpers.run.root(args, ["mkdir", target])
-            pmb.helpers.run.root(args, ["chmod", "700", target])
-            pmb.helpers.run.root(args, ["cp", user_ssh_pubkey, target + "/authorized_keys"])
-            pmb.helpers.run.root(args, ["chown", "-R", "12345:12345", target])
-    else:
-        logging.info("NOTE: No public SSH key found, you will only be able to use SSH password authentication!")
+    for key in ["RSA", "Ed25519"]:
+        user_ssh_pubkey = os.path.expanduser("~/.ssh/id_" + key.lower() + ".pub")
+        target = args.work + "/chroot_native/mnt/install/home/user/.ssh"
+        if not os.path.exists(user_ssh_pubkey) or not pmb.helpers.cli.confirm(
```
ollieparanoid: Sorry, I just realize that in my example code I should not have combined the two `if` conditions, because now the `NOTE:` even gets printed when the user doesn't want the key to be copied.

Could you split it again in two `if` and print the `NOTE:` only if the key was not found?

Pseudo-code to illustrate what I mean:
```python
if key not found:
	print note
	continue
if not confirmation:
	continue

copy keys here
```

Thanks again!

flacks: No problem; I didn't catch that, glad you did. Follow-up should do it.

pmb/install/install.py
```diff
@@ -125,16 +125,20 @@ def copy_ssh_key(args):
...
+        if not os.path.exists(user_ssh_pubkey):
+            logging.info("NOTE: Public " + key + " SSH key not found. If no SSH key " +
+                         "is copied, you will need to use SSH password authentication!")
+            continue
+        if not pmb.helpers.cli.confirm(
+                args, "Would you like to copy your " + key + " SSH public key to the device?"):
+            continue
+        pmb.helpers.run.root(args, ["mkdir", target])
+        pmb.helpers.run.root(args, ["chmod", "700", target])
+        pmb.helpers.run.root(args, ["cp", user_ssh_pubkey, target + "/authorized_keys"])
```
ollieparanoid: Thanks for making the changes! I have tested it, and unfortunately there is still one corner case, that leads to unexpected behavior. When the user has both the `rsa` and `ed25519` keys installed, and wishes to copy both of them, the `ed25519` key will overwrite the `rsa` key.

With that in mind, I propose:
* create a list of keys to be written
* fill it in the loop
* when it is empty after the loop, return
* write the keys to a in `args.work + "/chroot_native/tmp/"` with Python code
* then move that file inside the chroot to the right location and chown it


Some more pseudo code:
```python

keys = []
for key in ...:
	...
	# read the public key file, and add its contents to "keys"


if not len(keys):
	return

# write all keys to args.work + "/chroot_native/tmp/authorized_keys"
# use pmb.chroot.root to:
# - create the folder structure with right permissions
# - move the file to the right location
# - chown it
```

flacks: Good eye. I took your advice and gave it another go. What do you think?

Issue #572, opened 2017-09-13
Building linux-samsung-i9100 fails due to checksum mismatch issue


bhuvi8:
Checksum for linux kernel linux-samsung-i9100-349a3e91e76d17e67ef6213e1f6712e700695631.tar.gz in file aports/device/linux-samsung-i9100/APKBUILD is given as 
1cfaabb03669854037312d708731396ffd4592771fdb01d3b652c6ff7ebc05eebd7a93e7634595148d420e781262a369206008a9f1ed62bdb1a7738b2baa8fa2  linux-samsung-i9100-349a3e91e76d17e67ef6213e1f6712e700695631.tar.gz

whereas the correct checksum is 
285ffd2c82d0444b99da583573d19443b73de22d7382aae54413af229a28bebf4ea32ecbe248ae54ae64bbbc0af7d722b1ef86d5dd099e2990dcc6284fb195fa  linux-samsung-i9100-349a3e91e76d17e67ef6213e1f6712e700695631.tar.gz

bhuvi8:
This checksum value was modified in commit 6a5a57c4bc7aed8fb7ea19ead7d8cab0e2026056 when changing the linux kernel source format from .zip to .tar.gz 

Need someone to verify 285ffd2c82d0444b99da583573d19443b73de22d7382aae54413af229a28bebf4ea32ecbe248ae54ae64bbbc0af7d722b1ef86d5dd099e2990dcc6284fb195fa is indeed the correct checksum.

I will create a pull request for it in the mean time.

PR #573, opened 2017-09-13
Fix checksum linux-samsung-i9100 (#572)

bhuvi8:
Updated with correct checksum values for linux kernel


ollieparanoid:
Thank you!

PR #577, opened 2017-09-13
fix checksums

drebrez:
deleted all the download files in the `cache_distfiles` folder and executed:
```
ls aports/cross/ | xargs pmbootstrap checksum && \
ls aports/device/ | xargs pmbootstrap checksum && \
ls aports/firmware/ | xargs pmbootstrap checksum && \
ls aports/kde/ | xargs pmbootstrap checksum && \
ls aports/maemo/ | xargs pmbootstrap checksum && \
ls aports/main/ | xargs pmbootstrap checksum
```
a lot of packages resulted having wrong checksums.

Seems like something changed with the `.tar.gz`  files from github.


bhuvi8:
Yes It was working till last week and its showing checksum error. Faced the same with samsun-i9100.tar.gz kernel source

PR #578, opened 2017-09-14
[rx51] Fix wpa_supplicant initialization

craftyguy:
This commit instructs wpa_supplicant to use the wext & nl80211 kernel 
interface, and starts the service with the dbus interface enabled (useful for
things like networkmanager, etc). Prior to this, wpa_supplicant would
not manage the wifi device on the N900 since it was not using wext.


aports/device/device-nokia-rx51/device-nokia-rx51.post-install
```diff
@@ -50,4 +50,10 @@ EOF
 
 fi
 
+# Enable wext driver & dbus control for wpa_supplicant
+if [ -z "$(grep Dwext /etc/conf.d/wpa_supplicant)" ]; then
```
ollieparanoid: How about refactoring that to `if ! grep -q Dwext /etc/conf.d/wpa_supplicant; then`?

craftyguy:
@ollieparanoid thanks for the suggestion. You're quite the shell guru!

MartijnBraam:
If we change the interface to `-Dwext,nl80211` then it should work on all devices instead of only the n900. most kernel drivers are nl80211

craftyguy:
You bring up an interesting point..

Should this PR target postmarketos-base and not a specific device
package if it would be useful for other devices? I targeted the n900
with this PR, but since I think every device would want the capability
to connect to a secure wifi network maybe I should re-write this PR to
fix wpa_supplicant for all devices (and include the nl80211 driver
parameter to wpa_supplicant).

Thoughts?

ollieparanoid:
Configuring `-Dwext,nl80211` so it works for all devices in `postmarketos-base` would be better indeed!

craftyguy:
@ollieparanoid 
Changed the scope of this PR to target `postmarketos-base`, and added `nl80211` to the list of interface drivers.

Issue #373, opened 2017-08-13
wlan-prima (wireless module) and wcnss-service


PabloCastellano:
Continuation of https://github.com/postmarketOS/pmbootstrap/issues/143 and https://github.com/postmarketOS/pmbootstrap/issues/358

@yuvadm and I are working (see my [titan-wifi](https://github.com/PabloCastellano/pmbootstrap/tree/titan-wifi) branch and yuvadm's [device-mozilla-flame](https://github.com/yuvadm/pmbootstrap/tree/device-mozilla-flame)) on the dynamic firmware loading feature (https://github.com/postmarketOS/pmbootstrap/issues/147) which is required by the wlan-prima module to load blobs _(are these considered blobs?)_ in order to initialize the wireless interface later. This is the case for mozilla-flame and motorola-titan so far.

This is how it works:
- The [wlan-prima](https://github.com/t2m-foxfone/android_platform_vendor_qcom-opensource_wlan_prima) kernel module is loaded
- wlan-prima tries to load the firmware (multiple files) (will be solved in issue https://github.com/postmarketOS/pmbootstrap/issues/147)
- [wcnss-service](https://android.googlesource.com/platform/hardware/qcom/wlan/+/master/qcwcn/wcnss-service/) initializes the wireless interface

I made the wcnss-service (initially written for Android) some modifications (removed Android stuff, mocked Android functions) and added a Makefile and now it compiles and works in real linuxes.

This so-called service reads the following system variables:
- `ro.serialno`
- `ro.baseband` (the MODEM_BASEBAND_PROPERTY constant)

and writes `wlan.driver.config` and `wlan.driver.ath`.

Maybe writing the variables is not needed for postmarketOS. However `ro.serialno` is used to generate a serial number, used to initialize the interface. I have hardcoded this value for testing but it can be read from /proc/cmdline.

However my device gets frozen right after running wcnss-service. There must be something I've left back.

Things we should take the chance to discuss:
- Firmware path(s). Is it `/lib/firmware` or `/lib/firmware/postmarketos`?

yuvadm commented on Matrix:
>also keep in mind that at least for the flame I have firmware that is flashed on some special partitions and ideally i would want to mount them to /lib/firmware/$foo

This is a draft TODO:
- [x] udev dynamic firmware loading https://github.com/postmarketOS/pmbootstrap/issues/143 (pick firmware.sh and the udev rule from some of the branches and include them in postmarketos-base
[...]

PR #579, opened 2017-09-14
[huawei-y530] add kernel modules

drebrez:
To make the wifi work, the next step I think is the `wcnss-service` (#373), because loading the wlan module it fails with the following log:
```
[1166, modprobe] [ 1002.613395] wlan: loading driver v3.2.3.172
[1166, modprobe] [ 1007.657719] wlan: [1166:F :HDD] hdd_driver_init: WCNSS driver not ready
```


drebrez:
@ollieparanoid thanks for the info about the return 1, I remember there is an issue in the tracker about updating all the APKBUILD we have.

PR #585, opened 2017-09-15
Enable swap file support (#574)

### LINKED ISSUE #574: [nokia-rx51] implement encrypted SWAP
I've read through the `fstab` yesterday and noticed, that the N900 has an unencrypted swap.
In my opinion, we should have encrypted swap by default, and if someone has this use case, make it optional to turn it off (maybe along with the `--no-fde` option?)

craftyguy:
This implements swap file support (as discussed in #574), and also removes swap partition settings from the N900.

New deviceinfo variable `deviceinfo_swap_size_recommended` can be used to specify a swap size for specific devices.

`/etc/conf.d/swapfile` contains two variables:
- `swap_size` which will override `deviceinfo_swap_size_recommended`
- `swap_file` which defaults to `/swapfile`

If the config file doesn't exist, then the service defaults looks to `deviceinfo_swap_size_recommended` for size. If that is unset, then the service will not set up any swap file.

PR configures the `swapfile` service to start on bootup.

If a user wants to reconfigure the swap size while pmos is booted, they just have to edit the config file and to `sudo rc-service swapfile restart` to have it applied.

Also, as discussed previously, "encrypted swap" comes for "free" since this file exists on the rootfs, which would presumably be encrypted (assuming no `--no-fde` at install time)

Please test!

Fixes #574


aports/main/postmarketos-base/swapfile/swapfile
```diff
@@ -0,0 +1,92 @@
...
+# $2 = swap file size
+function make_swap() {
+	local sf=$1
+	local ss=$2
+
+	# Allocate space on disk for swap
+	fallocate -l ${ss}M ${sf}
+
+	#format swap file
+	mkswap ${swap_file} 1>/dev/null
```
ollieparanoid: I guess you meant to use `sf` there.

I am personally in favor of descriptive names btw, if you agree you could call the variable `_swap_file` instead (so it doesn't clash with the global variable name).

aports/main/postmarketos-base/swapfile/swapfile
```diff
@@ -0,0 +1,92 @@
...
+# This file is part of postmarketos-base					 #
+#										 #
+# postmarketos-base is free software: you can redistribute it and/or modify	 #
+# it under the terms of the GNU General Public License as published by		 #
+# the Free Software Foundation, either version 3 of the License, or		 #
+# (at your option) any later version.						 #
+#										 #
+# postmarketos-base is distributed in the hope that it will be useful,		 #
+# but WITHOUT ANY WARRANTY; without even the implied warranty of		 #
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the			 #
```
ollieparanoid: That hash looks out of line there. Maybe remove the hashes on the right side altogether?

aports/main/postmarketos-base/swapfile/swapfile
```diff
@@ -0,0 +1,92 @@
...
+	echo "Unable to find config at: ${config}"
+	exit 1
+fi
+
+# deviceinfo is a pmOS-specific thing, but it's not *required*
+source /etc/deviceinfo
+source /etc/conf.d/swapfile
+
+# Configure swap size. Size is in MB
+swap_size="${swap_size:-${deviceinfo_swap_size}}"
```
ollieparanoid: How about naming the variable `deviceinfo_swap_size_recommended`?

aports/main/postmarketos-base/swapfile/swapfile
```diff
@@ -0,0 +1,92 @@
...
+if [ -z "${swap_size}" ]; then
+	# Set safe default if nothing set swap_size
+	swap_size=768
```
ollieparanoid: How about defaulting to `0` instead (and adding extra code to not create a swap file in that case, I think that wouldn't work right now). Reason is, that newer devices usually come with enough RAM, so SWAP isn't even necessary.

aports/main/postmarketos-base/swapfile/swapfile.conf
```diff
@@ -0,0 +1,3 @@
+swap_file=/swapfile
+# Size is in MB
+swap_size=1024
```
ollieparanoid: Again, I would default to `0` here.

craftyguy:
@ollieparanoid 

I incorporated your changes and then added some logic to handle a 'bug' I identified.. So now the script will look at `swap_size` in the config file, if it is unset, then it looks at the `deviceinfo_swap_size_recommended` variable. It then writes the value of this deviceinfo variable to the config file (if deviceinfo var is not set, then it writes 0 to config).

The reason I did this is because it allows the value in `deviceinfo_swap_size_recommended` to actually be used by default. The user can then override this by editing the config file and then whatever is in `deviceinfo_swap_size_recommended`  would have no effect on the swap size at that point (including device package updates, those would not override user-specific configuration).

aports/main/postmarketos-base/swapfile/swapfile.conf
```diff
@@ -0,0 +1,3 @@
+swap_file=/swapfile
+# Size is in MB
+swap_size=
```
ollieparanoid: Surprisingly, OpenRC changes that into `swap_size=0` as soon as it runs `Caching service dependencies...` (-> as soon as you do anything with it). Reported upstream: https://github.com/OpenRC/openrc/issues/165

A workaround is changing that line into `# swap_size=1024` for example, that does not get replaced.
(I've also tried `swap_size=""`, which also gets replaced.)

_(When `swap_size=0`, the recommended value from the `deviceinfo` does not get used - so currently it never gets used.)_

aports/main/postmarketos-base/swapfile/swapfile
```diff
@@ -0,0 +1,103 @@
...
+
+swap_file="${swap_file:-/swapfile}"
+
+echo "Setting up swap file of size ${swap_size}MB at ${swap_file}..."
+
+# Clean up old swap file if new size is different than current size of file
+if [ -f ${swap_file} ]; then
+	swapoff ${swap_file} 2>/dev/null
+	# Get current swap file size
+	cur_size=$(du -h /swap |awk '{split($0, s, ".");print s[1]};')
```
ollieparanoid: That is error-prone, I've often got such lines:
```
localhost:/etc$ sudo service swapfile stop
 * Deactivating swap file ... [ ok ]
localhost:/etc$ sudo service swapfile start
 * Activating swap file ...
Setting up swap file of size 99MB at /swapfile...
du: /swap: No such file or directory
Unable to remove old swap file! [ ok ]
localhost:/etc$
```

I recommend always deleting the file and recreating it, it happens instantly with `fallocate` anyway.

aports/main/postmarketos-base/swapfile/swapfile
```diff
@@ -0,0 +1,103 @@
...
+	if [ "${cur_size}" != "${swap_size}" ]; then
+		remove_old_swap ${swap_file}
+		make_swap ${swap_file} ${swap_size}
+	fi
+else
+# No previous file, so make a new one
```
ollieparanoid: (Indentation)

aports/main/postmarketos-base/swapfile/swapfile
```diff
@@ -0,0 +1,103 @@
...
+	make_swap ${swap_file} ${swap_size}
+fi
+
+# Set correct file permissions for swap file
+chown root:root ${swap_file}
+chmod 600 ${swap_file}
```
ollieparanoid: Please move the `chown` and `chmod` lines into `make_swap`, before executing `mkswap`.
Otherwise we get warnings like the following:
```
localhost:/etc$ sudo service swapfile start
 * Activating swap file ...
Setting up swap file of size 99MB at /swapfile...
mkswap: /swapfile: insecure permissions 0644, 0600 suggested.
Successfully activated swap file! [ ok ]
```

craftyguy:
Does your last commit break the sed statement for the conf file?

ollieparanoid:
Yes it does, good catch! I'll fix it.

ollieparanoid:
Wait a second, could it be that the magical overwrite comes from the `/sbin/swapfile`?

craftyguy:
> Wait a second, could it be that the magical overwrite comes from the /sbin/swapfile?

I write the value from `deviceinfo_swap_size_recommended` to `swap_size` variable in the conf file [here.](https://github.com/postmarketOS/pmbootstrap/pull/585/files#diff-a0719f243701319a67222537ca42718cR73), but only if `swap_size` is unset (or missing) from the config file.. It (my script) shouldn't delete any existing values (e.g. if swap_size is set), or even replace the comment `# swap_size=` with a blank line..

Issue #584, opened 2017-09-15
Package msm framebuffer refresher


ollieparanoid:
@drebrez found this:
https://github.com/AsteroidOS/msm-fb-refresher

Some context: we have noticed, that some framebuffer drivers from Android kernels only implement double-buffering, but not single-buffering. weston, fbsplash and probably a lot of other Linux programs assume, that the single-buffering is working.

Right now we have a `while true` shell-script, that forces the refresh all the time. The `msm-fb-refresher` does basically the same thing, but more elegantly by only calling the necessary kernel syscall and - most importantly - with a sleep between the calls instead of running just all the time.

**TODO:**
* [x] package `msm-fb-refresher`
* [x] add a new deviceinfo option, that enables this hack
* [x] extend the mkinitfs code to put msm-fb-refresher into the initramfs, in case that option is set
* [x] extend the init script to actually run it, if it exists
* [ ] enable the option in all devices, that require the while true hack right now, and depend on the msm-fb-refresher package. also remove the existing while true hack.

craftyguy:
Unless I'm missing something, #587 implements the first 4 items on that list. I've updated the pmOS wiki to mention the new deviceinfo var for toggling this (`deviceinfo_msm_refresher`).

montvid:
There is another one here: [https://github.com/crondog/arch-flo/blob/master/refresher.c](https://github.com/crondog/arch-flo/blob/master/refresher.c) Is it any better?

ollieparanoid:
Thanks for the link @montvid! I've compared the code, and while it is different, it is written differently, it is essentially doing the same thing:

```c
struct fb_var_screeninfo var;
int fd = open("/dev/fb0", O_RDWR);
ioctl(fd, FBIOGET_VSCREENINFO, &var);
while(1) {
    ioctl(fd, FBIOPAN_DISPLAY, &var);
    usleep(16666);
}
```
So I assume that it behaves the same way (not tested though).

drebrez:
Seems that only `device-lg-lenok` still uses the `while echo '0,0' >  pan` workaround, @JamesHealdUK do you have time to test the `msm-fb-refreshed` on the LG Lenok?

ollieparanoid:
We have packaged it successfully, and quite a few devices are using it now. For some devices it does not work (causing reboots), but we have issue #604 for that. Closing.

PR #587, opened 2017-09-16
Add msm-fb-refresher package and initfs config

craftyguy:
For #584.

This is meant to be a starting point, I anticipate that I may have done more than one controversial thing below. My main motivation was to try and prevent any merging headaches with the OSK PR :smiley:  

In order to have `msm-fb-refresher` to run on boot in initramfs, set `deviceinfo_msm_refresher="true"` in the device `deviceinfo` file.

Note: I tested that packaging is successful, but didn't test this on my device because I don't have a device that is affected by the problem this is trying to solve, and I don't want to risk my device!


aports/main/msm-fb-refresher/APKBUILD
```diff
@@ -0,0 +1,22 @@
...
+pkgdesc="Display refresher for Qualcomm devices"
+arch="all"
+url="https://github.com/AsteroidOS/msm-fb-refresher"
+license="GPL2"
+makedepends="linux-headers"
+source="${pkgname}-${_hash}.tar.gz::https://github.com/AsteroidOS/msm-fb-refresher/archive/${_hash}.tar.gz"
+
+build() {
+    cd $srcdir/$pkgname-$_hash
+    gcc refresher.c -o msm-fb-refresher
```
ollieparanoid: To make use of the cross-compiler, we'd need to refactor that to:
```shell
gcc refresher.c -o refresher.o -c
gcc refresher.o -o msm-fb-refresher
```

aports/main/msm-fb-refresher/APKBUILD
```diff
@@ -0,0 +1,22 @@
+_hash=9560ceb1b0300d407fe10b47782cc2cf60acf379
```
ollieparanoid: They tagged a release for us, which we can use now 👍 
https://github.com/AsteroidOS/msm-fb-refresher/releases

aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -97,6 +97,9 @@ get_modules()
 get_binaries()
 {
 	BINARIES="/bin/busybox /bin/busybox-extras /usr/sbin/telnetd /sbin/kpartx"
+	if [ "${deviceinfo_msm_refresher}" == "true" ]; then
+		BINARIES="${BINARIES} /usr/sbin/msm-fb-refresher"
```
ollieparanoid: How about adding this here?
```shell
if ! [ -e /usr/sbin/msm-fb-refresher ]; then
    echo "ERROR: Please add msm-fb-refresher as dependency to your device package,"
    echo "or set msm_refresher to false in your deviceinfo!"
    exit 1
fi
```

aports/main/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -299,6 +299,13 @@ echo_connect_ssh_message() {
 	echo "ssh user@$IP"
 }
 
+start_msm_refresher() {
+	# shellcheck disable=SC2154,SC2086
+	if [ "${deviceinfo_msm_refresher}" = "true" ]; then
+		/usr/sbin/msm-fb-refresher &
```
ollieparanoid: I just realized, that we'll need to pass `--loop` as argument. Otherwise it will only refresh once:
https://github.com/AsteroidOS/msm-fb-refresher/blob/9560ceb1b0300d407fe10b47782cc2cf60acf379/refresher.c#L39-L50

craftyguy:
@ollieparanoid 

feedback incorporated and ready to test!

aports/main/postmarketos-mkinitfs/mkinitfs.sh
craftyguy: As pointed out by @drebrez  in chat, this doesn't actually work in this location. Here's output:

https://bpaste.net/show/fcf871866ccd

drebrez:
@craftyguy @ollieparanoid I'm testing it on the device and I found an issue. 
Without killing it at the end of the initramfs script, something goes wrong.
Before switching the chroot the `/dev` is unmounted, since `msm-fb-refresher` opened the `/dev/fb0` (see [here](https://github.com/AsteroidOS/msm-fb-refresher/blob/master/refresher.c#L33), it crashes

drebrez:
Ok, by killing `msm-fb-refresher` at the end of the initramfs script:
```
- killall telnetd mdev 2>/dev/null
+ killall telnetd mdev msm-fb-refresher 2>/dev/null
```
and adding a `profile.d` script with `/usr/sbin/msm-fb-refresher --loop &`, everything started correctly.

craftyguy:
I actually suggest adding it as a openrc script (in `/etc/init.d/msm-fb-refresher`), and enable with `rc-update add msm-fb-refresher`. This way it would always start on boot after initfs, since profile.d is only processed on login (and you probably don't want it starting multiple times on multiple VT logins in the future!)

drebrez:
@craftyguy that's probably the best way to go, the profile.d was only to test it

craftyguy:
I just pushed a commit that adds an openrc service for the refresher and starts it on the `boot` runlevel. I don't know if this actually works, but it can be adjusted by editing the `postmarketos-base.post-install` script to start it at earlier (or later) runlevels if necessary.

drebrez:
@craftyguy you also have to add `msm-fb-refresher` in the killall command here => https://github.com/postmarketOS/pmbootstrap/blob/3c7f1b7a57268d33a1c26601010a7af44cc9355a/aports/main/postmarketos-mkinitfs/init.sh.in#L44

craftyguy:
> you also have to add msm-fb-refresher in the killall command here =>

Done!

> about the runlevel, do you know when the /dev is mounted?

`devfs` is in the `sysinit` runlevel. the flow is: `sysinit->boot->default`

Issue #554, opened 2017-09-10
pmbootstrap should automatically check for required kernel config options


zhuowei:
After running `pmbootstrap menuconfig`, it may be helpful to check if the kernel configs (e.g. devfs, vt) required to boot postmarketOS are enabled correctly. We may be able to borrow [Ubuntu Touch's script](https://github.com/janimo/phablet-porting-scripts/blob/master/kernel/check-config).

ollieparanoid:
Great idea, I'm also in favor of doing that!

For the implementation however, I recommend doing it in pmbootstrap's Python code. Then it would be faster than with starting an extra process, and it would be better integrated. One example how it could be done:

* New function `check(args, details=False)` in `pmb/parse/kconfig.py`
* When that function gets called with `details=False`, it does all checks and prints out the following message in case something is wrong:

> WARNING: Your kernel isn't configured properly for postmarketOS, run `pmbootstrap kconfig_check` for details

(or maybe `kcheck`, that's shorter?)
* With details enabled, it tells you exactly which options are set wrong and what effect this has (such as the usual config check scripts).
* We could call the function with details right after menuconfig as you proposed
* We could call it without details right before flashing and before booting the QEMU VM

I see a lot of positive emojis there, so I wonder what everyone thinks about this - and maybe someone wants to give it a shot?

drebrez:
I'm also interested in this because I always forget about what to enable when starting with a new config.
I suggest we first start identifying all the required configs.
It would be good to have a dedicated page in the wiki also (here an example: https://wiki.gentoo.org/wiki/Systemd#Kernel).

zhuowei:
@drebrez There is a page for this: https://wiki.postmarketos.org/wiki/Kernel_configuration

drebrez:
@zhuowei lol 🤣, I missed it completely. Thanks for let me know 😉

ata2001:
Is it safe, to rely on `# CONFIG_XXXX is not set` comments for determining, which option is enabled?

zhuowei:
@ata2001 The Ubuntu script throws an error if one of the required options isn't set, so I guess not?

ata2001:
@zhuowei That means it is set to no I believe, the ubuntu script checks if it's either y or m.

ata2001:
@ollieparanoid I don't know, whether we should design it with non-boolean config options in mind. We don't have any yet, but we may need it in the future.

ata2001:
[...]

PR #589, opened 2017-09-16
kernel config checking

ata2001:
#554


ata2001:
Could easily be extended to support string or integer options.

ollieparanoid:
I have refactored some details to match the rest of the code-base better, and applied the following functional changes:
* Moved `is_set` to an own function, and return `True`/`False` if it was found or not (the regex function would return `None` or a regex object otherwise, which means we can't use it in simple `True`/`False` checks directly)
* Raise a meaningful exception, when the config lists something other than `True`/`False` as value
* Return `True`/`False` to indicate if the check was successful or not
* Use that information to raise an exception on failure, when called with "pmbootstrap kconfig_check"
* Always check the configs of all architectures, not just the current device architecture (otherwise, when `qemu-amd64` is the current device, you can't check the other device configs for example)
* Allow to specify multiple packages/kernels
* Allow to leave out the packages/kernels argument, check all kernels in that case (it is super fast anyway)
* Add the check for all kernels to the Travis CI config

I hope that the changes are okay with you - please tell me if you don't like any of this and I'll revert it. Thanks a lot for making this PR, I'm sure this will prevent a lot of frustration caused by wrong kernel configs.

ata2001:
It's much better now.

PR #597, opened 2017-09-19
Corrected small typo in ~/pmbootstrap/pmb/config/init.py

lyubod:
Instead of "pmbootstrap zap" to delete chroots, it needs the full filename pmbootstrap.py. Hence it's missing ".py" on line 143.

### DIFF
--- a/aports/device/device-samsung-ypg1/90-android-touch-dev.rules
+++ b/aports/device/device-samsung-ypg1/90-android-touch-dev.rules
@@ -1,0 +1,8 @@
+# udev rules file
+# All device names can be read from weston's logfile (/tmp/weston.log in postmarketOS)
+
+# Touchscreen (use 'weston-calibrator' to calibrate)
+SUBSYSTEM=="input", ATTRS{name}=="sec_touchscreen", \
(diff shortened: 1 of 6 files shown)


lyubod:
Please merge this if it doesn't break anything. It shouldn't.
It corrects a 3 character typo in the command: "pmbootstrap zap" - The updated text reads: "pmbootstrap.py zap"

ollieparanoid:
This is not a typo, but actually used in a lot of places in the code as shortform for `./pmbootstrap.py`. Some people also made it as alias in their shell. The idea is, that when we have `pmbootstrap` ready to be packaged for other distributions, then `pmbootstrap` would really be the command you would execute.

How about adding a notice to the end of `pmbootstrap init` about this short form instead (in a new PR)?

Oh and it seems tha you accidentally pushed another change into this branch. GitHub will add it automatically to the active pull-request from your branch, so you would need to make a new branch for everything that you do not want to have in the PR (`git checkout master`, `git checkout -b my-new-branch`).

Thanks for making the PR and for your interest in postmarketOS!

PR #599, opened 2017-09-20
Improve copy in mkinitfs

craftyguy:
'install' does not properly handle symlinks, it copies the entire file
instead of a symlink to it. This PR uses `cp -a` to preserve symlinks.

For example, with `install`, the files libc.musl-armhf.so.1 and
ld-musl-armhf.so.1 are the same size, despite the fact that in rootfs
libc.musl-armhf.so.1 is actually a symlink to ld-musl-armhf.so.1.

From @drebrez:
```
initramfs:
old size => 1 567 930
new size => 1 168 591
```

This might also help with #126.

This PR is from a commit in the 'add osk' branch that I made, but it's
useful enough to have it here since it reduces the initramfs size by
copying any symlinks over as symlinks. For osk-sdl's dependencies, this
change dramatically reduces the size of initramfs-extras since binaries
are not copied twice.


ollieparanoid:
Thanks for making this PR @craftyguy and thanks for testing the PR and especially that the archive parameter preserves the permissions, as it should @drebrez!

PR #607, opened 2017-09-21
Update wiki link

z3ntu:


ollieparanoid:
Thanks, the link with the subdomain looks better and is in line with the link from the Alpine wiki. Could you please also increase the `pkgrel`?

z3ntu:
@ollieparanoid Better? 😃

PR #610, opened 2017-09-21
Remove \ as it is not needed

montvid:
Forgot to remove the \ the kernel throws a warning if not removed.


montvid:
i am learning to use github... and i need to add the hash to apkbuild too...

Issue #411, opened 2017-08-18
Only create the initramfs when necessary


ollieparanoid:
Right now we create the initramfs whenever it could be necessary (simply to avoid running an old version, which is really annoying for development, and because it introduces bugs all the time).

It depends on a lot of factors, when one of those changes we should rebuild the initramfs:
* installed kernel (with the flavor for which we want to generate the initramfs)
* device package
* postmarketos-splash
* postmarketos-mkinitfs
* initramfs hooks changed/got updated

Maybe we should save a file with all versions of the last generated initramfs, and compare that file with the last one to decide if a rebuild is necessary. Note that generating the initramfs is done in the chroot, and must work completely independent of pmbootstrap (so we can update it once it is installed on the device with the package manager).

ollieparanoid:
FYI, if someone is annoyed by this, use `pmbootstrap -s` to skip initramfs generation during the current `pmbootstrap` run.

ollieparanoid:
@drebrez asked me to create a bunch of checkboxes for an initial version, that caches the splashes and assign him to this ticket.

**TODO (cache splashes):**
- [ ] Write new splash screens to this path first: `/var/cache/postmarketos-mkinitfs/splashes/splash_$hash.ppm.gz` (quickly made up file name, please reconsider if that really makes sense). $hash would be a hash of the postmarketos-splash version, the config file, the device resolution and the text to be displayed
- [ ] Only generate a new splash screen, when that output file does not exist.
- [ ] Put the splash into the initramfs as usually (but from that location instead of from some temp path)
- [ ] When a new splash has been generated, delete the whole cache folder. Usually we would need to rebuild all splashes in one go anyway, and that way we don't have left-overs.

**TODO (only build if necessary):**
- [ ] Create a list of files related to the initramfs (either the ones that get included, or the ones that help generating it) at the beginning of mkinitfs.sh, together with the last modified timestamp of those files (`ls` is good enough)
- [ ] Put that information into a temporary file and hash it
- [ ] Compare that hash against the last file list hash (stored in `/var/cache/postmarketos-mkinitfs/` somewhere)
- [ ] if there is no old file list hash, or no old initramfs file, or the hash differs, build a new initramfs. otherwise, do nothing.
- [ ] save the current file list hash in that cache folder

ollieparanoid:
[...]

PR #620, opened 2017-09-23
Implement cache for generated splash screens

drebrez:
This PR addresses the first part of the issue #411 (see TODO [here](https://github.com/postmarketOS/pmbootstrap/issues/411#issuecomment-329922545))

[Here](https://bpaste.net/show/fbcf6692eba3) you can see the improvement (first run without cache, second with generation of cached file, third with cache already present).

Most of the time is spent during the computation of the dependencies of the various executables, mostly for those in the initramfs-extra.


aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -198,34 +198,62 @@ create_bootimg()
...
+	[ "$1" != "false" ] && clean="true" || clean="false"
+
+	splash_version=$(apk info -v | grep postmarketos-splash)
+	splash_config="/etc/postmarketos/splash.ini"
+	splash_config_hash=$(md5sum "$splash_config")
+	splash_width=${deviceinfo_screen_width:-720}
+	splash_height=${deviceinfo_screen_height:-1280}
+	splash_cache_dir="/var/cache/postmarketos-splashes"
+
+	set -- "splash-telnet"           "On-screen keyboard is not implemented yet, plug in a USB cable and run on your PC:\\ntelnet 172.16.42.1" "" \
```
ollieparanoid: Nice hack! How about adding a comment above that line?
```shell
# Overwrite $@ to easily iterate over the splash screens. Format:
# $1: splash_name
# $2: text
# $3: arguments
```
Also we are overwriting a global variable `$@` here, which could lead to strange errors in case someone wants to access this later (when modifying the code in the future). How about putting everything from `set` until `done` into a subshell? Then it should not affect the real `$@` variable.

drebrez: @ollieparanoid I will add the comment like you suggest, that was also my idea but I just forgot it.

About the variable `$@`, I just created this script to test it:
```
#!/bin/sh

function test()
{
   set -- "overwrite global variable"
   echo "test: $@"
}
echo "main: $@"
test
echo "main: $@"
```
and the result is:
```
/ # sh test.sh test
main: test
test: overwrite global variable
main: test
```

this means that the global `$@` variable is safe, only the local one (in function) gets replaced.

Do you still want to change it to run in a subshell?

ollieparanoid: Ah I see, then we don't need it. Thanks for the explanation!

*When I think about it, $1, $2 etc. are also different in functions, so it is logical that $@ is not a global variable.*

Issue #478, opened 2017-08-28
Kernel updates


ata2001:
Right now you either need to use fastboot/odin or a recovery to flash a kernel update. Since we have a cool package manager we could avoid this.

The kernel package installs the kernel image in /boot, all we need is a shell script that generates and flashes the boot.img.

ollieparanoid:
Good idea! Some remarks:
* As the N900 boots with uboot from `/boot` on the sdcard, it is already working there (in theory, if you have a binary package for the kernel update)
* Our `mkbootimg` already generates the `boot.img` file (and `mkbootimg` gets called when you update the kernel, just like in Alpine!), so all that is left to do is flashing it to the boot partition afterwards.
* We would need to make sure that we are actually on the device, so it never gets flashed when running inside pmbootstrap (!)

craftyguy:
I've confirmed that kernel updates, using apk, install properly on the N900 now that /boot is mounted on boot.

ata2001:
@ollieparanoid where is mkbootimg called?

ollieparanoid:
@ata2001 here:
https://github.com/postmarketOS/pmbootstrap/blob/946ba82cd451baabc383cae367ac1e35d09f814c/aports/main/postmarketos-mkinitfs/mkinitfs.sh#L174-L197

ollieparanoid:
(We always call `mkbootimg` through `mkinitfs`)

ata2001:
I'd prefer a command like `update-kernel`, that you can run manually to flash the updated kernel.

ollieparanoid:
Yeah, that seems to be a lot safer right now, good idea!

pavelmachek:
Just word of warning -- if power fails during update. There should still be way to recover with fastboot.

ollieparanoid:
As long as we do not overwrite fastboot, this will still work 👍

z3ntu:
The best solution of course would be having a proper recovery for postmarketOS which gets flashed to the recovery partition which can fix stuff 😉

ollieparanoid:
[...]

PR #625, opened 2017-09-25
Add kernel updater script

ata2001:
#478


aports/main/postmarketos-update-kernel/update-kernel.sh
```diff
@@ -0,0 +1,33 @@
...
+		echo "Usage: update-kernel [flavor]"
+		exit 1
+		;;
+esac
+
+# shellcheck disable=SC1091
+. /etc/deviceinfo
+FLAVOR=$1
+mountpoint -q /boot/ || {
+	echo "Mounting boot partition..."
```
MartijnBraam: I guess this message should be "Remounting boot partition readonly..."?

ata2001: @MartijnBraam That only gets executed, if /boot/ is not mounted already.

ata2001:
It could assert, whether it is running on a supported device before mounting `boot`, but I'm not sure that is needed, since it will only get installed on the supported devices.

ata2001:
Perhaps we should mount the boot partition automatically with the fstab like on most distros, that way it won't cause a problem on the N900, and this script could be simpler.

ollieparanoid:
Yeah, I would prefer if `boot` was mounted automatically. Otherwise, kernel updates will go wrong (typical mistake in Arch installations). How about making this as a second PR, which we review+merge first?

ata2001:
@ollieparanoid Okay, I'll do that first.

drebrez:
@ollieparanoid @ata2001 I also agree to always mount the `/boot` partition automatically

ata2001:
Now we can continue to work on this. Made a commit, now it leaves mounting the boot partition to the fstab.

aports/main/postmarketos-update-kernel/update-kernel.sh
```diff
@@ -0,0 +1,39 @@
...
+	heimdall-isorec)
+		echo "Flashing kernel with isorec method..."
+		KERNEL_PARTITION=$(findfs PARTLABEL="${deviceinfo_heimdall_partition_kernel:?}")
+		INITFS_PARTITION=$(findfs PARTLABEL="${deviceinfo_heimdall_partition_initfs:?}")
+		dd if=/boot/vmlinuz-"$FLAVOR" of="$KERNEL_PARTITION"
+		gunzip -c /boot/initramfs-"$FLAVOR" | lzop > "$INITFS_PARTITION"
+		;;
+	0xFFFF)
+		echo "No need to use this utility, since uboot loads the kernel directly from \
+			the boot partition. Your kernel should be updated already."
```
ollieparanoid: This will cause three tabs `\t\t\t` to be displayed inside the message.
I recommend simply using two echo statements.

ata2001: @ollieparanoid I fixed it now, it isn't POSIX compliant, but we are using busybox' ash anyway.

drebrez:
@ata2001 should I test the `isorec` use case?

ata2001:
@drebrez Please do if you have time, I've only tested it on a bootimg device.

aports/main/postmarketos-update-kernel/update-kernel.sh
```diff
@@ -0,0 +1,40 @@
...
+case $METHOD in
+	fastboot|heimdall-bootimg)
+		echo "Flashing boot.img..."
+		BOOT_PARTITION=$(findfs PARTLABEL="boot")
+		dd if=/boot/boot.img-"$FLAVOR" of="$BOOT_PARTITION"
+		;;
+	heimdall-isorec)
+		echo "Flashing kernel with isorec method..."
+		KERNEL_PARTITION=$(findfs PARTLABEL="${deviceinfo_heimdall_partition_kernel:?}")
+		INITFS_PARTITION=$(findfs PARTLABEL="${deviceinfo_heimdall_partition_initfs:?}")
```
drebrez: @ata2001 the variables of the partitions the are wrong:
`deviceinfo_heimdall_partition_kernel` should be `deviceinfo_flash_heimdall_partition_kernel` and
`deviceinfo_heimdall_partition_initfs` should be `deviceinfo_flash_heimdall_partition_initfs`

aports/main/postmarketos-update-kernel/update-kernel.sh
```diff
@@ -0,0 +1,40 @@
...
+# shellcheck disable=SC1091
+. /etc/deviceinfo
+
+FLAVOR=$1
+METHOD=${deviceinfo_flash_methods:?}
+case $METHOD in
+	fastboot|heimdall-bootimg)
+		echo "Flashing boot.img..."
+		BOOT_PARTITION=$(findfs PARTLABEL="boot")
+		dd if=/boot/boot.img-"$FLAVOR" of="$BOOT_PARTITION"
```
drebrez: @ata2001 I suggest adding `bs=1M` to the various `dd` commands, it speed up a lot.

aports/main/postmarketos-update-kernel/update-kernel.sh
```diff
@@ -0,0 +1,40 @@
...
+		;;
+	heimdall-isorec)
+		echo "Flashing kernel with isorec method..."
```
drebrez: @ata2001 what you think about splitting the `echo` in `Flashing kernel...` and after the `dd` add `Flashing initramfs...`?

actual output:
```
localhost:~$ sudo /sbin/pmos-update-kernel samsung-i9070
Flashing kernel with isorec method...
22688+1 records in
22688+1 records out
Done.
```

new output:

```
localhost:~$ sudo /sbin/pmos-update-kernel samsung-i9070
Flashing kernel...
22688+1 records in
22688+1 records out
Flashing initramfs...
Done.
```

aports/main/postmarketos-update-kernel/update-kernel.sh
ata2001: @drebrez It indeed speeds a lot up.

```
~ # dd if=boot.img of=/dev/block/platform/msm_sdcc.1/by-name/boot
18512+0 records in
18512+0 records out
9478144 bytes (9.0MB) copied, 0.966787 seconds, 9.3MB/s
~ # dd if=boot.img of=/dev/block/platform/msm_sdcc.1/by-name/boot bs=1M
9+1 records in
9+1 records out
9478144 bytes (9.0MB) copied, 0.254191 seconds, 35.6MB/s
```

ata2001:
@drebrez Thank you for testing and reviewing, I believe I made the requested changes.

drebrez:
@ata2001 @ollieparanoid tested the `isorec` use case again and it works perfectly. 

I was thinking that since it's just a small script and doesn't use much space, would be good to always install it, so add it as a dependency of postmarketos-base, what you think?

ollieparanoid:
Thanks @ata2001 and @drebrez!

> I was thinking that since it's just a small script and doesn't use much space, would be good to always install it, so add it as a dependency of postmarketos-base, what you think?

As discussed in the chat, I recommend against adding this to `postmarketos-base`, as long as it depends on `util-linux` (for `findfs`). Because when we have `util-linux` in the base installation, it will lead us to write code that does not work with just busybox installed.

PR #627, opened 2017-09-25
asus-flo wifi & freedreno support + new firmware-asus-flo

montvid:
I created the firmware-asus-flo package to add all possible firmware a flo device might need. Tested wifi and freedreno as working.


montvid:
Thanks for the good vibrations! I am testing the device a lot. Still testing other things.

PR #634, opened 2017-09-26
Move to a stable kernel branch to fully fix freedreno.

montvid:
Tested with weston and hildon. Works without lockups, better performance in mesa demos. Drm enabled by default as it works flawlessly. Moved from flo-wip to flo-4.11.
Glxgears in weston show: 1366 frames in 5.0 seconds = 273.078 FPS


MartijnBraam:
Nice 👍

montvid:
Thanks. I am learning a lot.

Issue #633, opened 2017-09-26
postmarketos-demos not started from /etc/xdg/autostart, causing race condition


montvid:
I am running weston with drm-backend.so and I can't see the pmos demo menu anymore after boot. Changing to fbdev-backend.so the menu is there again. Running glxgears with drm-backend shows gears running on the screen so there should not be a problem with the output to weston. When I "sudo killall weston" weston respawns with the menu. I would like that the menu would be there at boot as not to ask people to run "sudo killall weston" to see the menu.

ollieparanoid:
This is probably a timing issue here:

https://github.com/postmarketOS/pmbootstrap/blob/e60eee7dfa80d2ae07f31c83d1bbb9c56e3d20e4/aports/main/postmarketos-ui-weston/start_weston.sh#L22-L23

You could try to increase that to 3 seconds as a workaround. A better fix would be using a proper autostart mechanism of weston (I couldn't find one yet), or retrying for 4 seconds or so to find the weston server (it should be possible to extend `postmarketos-demos` for that).

MartijnBraam:
I have read somewhere in the weston docs that it can launch .desktop files in the xdg autostart directory

montvid:
Yes! sleep 3 fixed it so far. Please update.

ollieparanoid:
Thanks for trying it out @montvid, and thanks for telling how to do it properly @MartijnBraam!

I'll fix it properly with the `/etc/xdg/autostart` solution.

montvid:
@ollieparanoid sleep 3 would work fine for now and it's a 1 minute fix.

PR #638, opened 2017-09-27
Workaround for #633: increase sleep before pmOS-demos shows up

ollieparanoid:
@montvid suggested we provide a fast fix here, so this is what I did. I'm working on starting postmarketos-demos with the xdg autostart properly, as discussed in #633.

@MartijnBraam, please do a review and merge afterwards if everything is fine.


montvid:
Thanks!

MartijnBraam:
Oops forgot the merging part

PR #639, opened 2017-09-27
[samsung-i9070] Add framebuffer mode in initfs-hook

drebrez:
After various tests we finally discovered that the framebuffer mode have to be set otherwise the `osk-sdl` initialized the framebuffer with a wrong memory size and doesn't work.

Even the order of the framebuffer settings matters, if the `cat ..modes > ..mode` is placed before the `bits_per_pixel` and the `virtual_size`, it doesn't work correctly.

([Here](https://bpaste.net/show/ae4ad4cccfe6) a small utility used to do the investigations)


craftyguy:
Just commenting here in case others have some ideas.. but I would *really* like to come up with a better way of doing this for all affected devices in osk-sdl, but if this causes osk-sdl to display 100% of the time then maybe it's fair to add this change to all devices with this display issue (i.e. all qualcomm mdss/msm devices it seems..). @ollieparanoid thoughts?

ollieparanoid:
Thanks for debugging this and finding a working solution!

@craftyguy: I wonder if `osk-sdl` would be the right place to apply this workaround. As I understand it, the problem is that the msm-driver does not specify a default framebuffer mode, but some applications expect it to exist.

So this affects more than osk-sdl. How about a new function in the initramfs, that we call after the hooks have been executed?


```shell
# Fix for framebuffer drivers, which do not have a default mode
set_framebuffer_mode()
{
	[ -e "/sys/class/graphics/fb0/modes" ] || return
	[ -z "$(cat /sys/class/graphics/fb0/mode)" ] || return

	_mode="$(cat /sys/class/graphics/fb0/modes)"
	echo "Setting framebuffer mode to: $_mode"
	echo "$_mode" > /sys/class/graphics/fb0/mode
}

```
(I have *not* tested this function, just written it down to illustrate what I mean.)

The best part is, that we can remove the `cat modes > mode` hook from all device packages, where we currently use it.

What do you guys think about it?

craftyguy:
That seems like a great solution, as long as it fixes the broken devices and doesn't break other devices.. maybe have some 'fix_stupid_msm' deviceinfo var? :P

ollieparanoid:
If necessary we could add a variable, but the code above checks if the `mode` exists and is empty, and only then sets it in that case - the idea was, that this hopefully does *not* break other devices.

craftyguy:
Ah, I missed that. @drebrez could you test out what @ollieparanoid has to see if it accomplishes the same thing on your device? If so, this seems like a better solution that would help all affected devices so we don't have to remember to put in what you have here into other device packages 😄

drebrez:
@ollieparanoid @craftyguy yes that solution seems the best one. I think I proposed it some time ago but I forgot about it. 
Looking at the packages now, a lot of devices already have that `cat modes > mode` setting, let's hope it doesn't break anything on the devices that don't have it right now.
If you want I can implement and test it, and if it works I also update all the other packages, ok?
I would suggest that we add the variable only when we get the first device that doesn't need it, maybe the variable can be to disable it, that one can already be added (ex: `skip_set_framebuffer_mode`), what you think?

ollieparanoid:
> If you want I can implement and test it, and if it works I also update all the other packages, ok?

That would be awesome \o/

> I would suggest that we add the variable only when we get the first device that doesn't need it, maybe the variable can be to disable it, that one can already be added (ex: skip_set_framebuffer_mode), what you think?

If you mean with "doesn't need it" that it breaks when executing the function, then yes, I also think we should add the deviceinfo variable only if there is a device that breaks with this change.

PR #640, opened 2017-09-28
Remove SDL2 package from aports

craftyguy:
Now that SDL2 2.0.6 is out and updated in Alpine, there's no reason for
us to keep this package around, at least from an osk-sdl perspective!


craftyguy:
This isn't ready to merge yet.. not until Alpine accepts directfb & tslib packages, and then I'll need to submit a patch to add directfb support to the Alpine SDL2 package :(

craftyguy:
Alpine merged SDL2 w/ directfb support, this PR is good to go!

PR #641, opened 2017-09-28
[osk-sdl] update to 0.49

craftyguy:


ollieparanoid:
Why the don't merge PR label here?

craftyguy:
This depends on SDL 2.0.6 w/ directfb support :(
See #640

craftyguy:
Alpine has support for this new version of osk-sdl now, so this PR should be good to go!

PR #643, opened 2017-09-28
Package unicsy_demo (basic hardware demo)

pavelmachek:
Create simple demo, that nevertheless demonstrates basic hardware functionality, including battery status, light sensor status, accelerometer status, LED control, control of backlight on keyboard and screen.


aports/main/unicsy_demo/APKBUILD
```diff
@@ -0,0 +1,24 @@
+pkgname=unicsy_demo
+pkgver=0.2
+pkgrel=0
+pkgdesc="Demo of phone functionality"
```
ollieparanoid: How about being a bit more specific about what it demonstrates? E.g.:
`"Phone hardware demo (battery status, light sensors, accelerometer, LEDs, backlights)"`

aports/main/unicsy_demo/APKBUILD
```diff
@@ -0,0 +1,24 @@
+pkgname=unicsy_demo
```
ollieparanoid: From our [package naming convention](https://wiki.postmarketos.org/wiki/Package_naming_convention):
> Please use dashes (-) only, no underscores (_).

aports/main/unicsy_demo/APKBUILD
```diff
@@ -0,0 +1,24 @@
...
+pkgver=0.2
+pkgrel=0
+pkgdesc="Demo of phone functionality"
+url="https://github.com/postmarketOS/osk-sdl"
```
ollieparanoid: Please change the URL.

aports/main/unicsy_demo/APKBUILD
```diff
@@ -0,0 +1,24 @@
...
+license="GPL3"
+depends="py-gtk-dev"
+makedepends=""
+subpackages=""
+
+source="${pkgname}-${pkgver}-${pkgrel}.tar.gz::https://github.com/pavelmachek/unicsy_demo/archive/${pkgver}.tar.gz"
+
+options="!strip"
+
+build() {
```
ollieparanoid: You can just remove the `build()` function if it does nothing :)

pavelmachek:
I added /usr/bin/unicsy_demo symlink for easy launch. Ultimately, it
would be nice to have a button in weston-demo, or launch it by
default.

Unicsy is a word play at Unix, which itself is a word play at Multics,
which means ... anyway what I wanted to say is that I plan to keep
traditional Unix philosophy where possible: simple programs that do
one thing and do it right and plain text configuration and user data
files are important to me. More here:
https://pavelmachek.livejournal.com/139660.html

Demo should work on non-N900 machines -- I test PC from time to
time. I expect hardware.py will need to be modified to support
all features on different phones. Help will probably be needed there.

I changed comment to be more specific, and changed to dash (not
underscore), fixed url, and removed empty bulid function.

ollieparanoid:
Thanks for the explanation and the fixes! I'm afraid the code isn't pushed yet though, because it does not show up here.

pavelmachek:
I did some testing and pushed fixed version.

aports/main/unicsy-demo/APKBUILD
```diff
@@ -0,0 +1,22 @@
...
+pkgrel=1
+pkgdesc="Phone hardware demo (battery status, light sensors, accelerometer, LEDs, backlights)"
+url="https://github.com/pavelmachek/unicsy_demo"
+arch="all"
+license="GPL3"
+depends="py-gtk-dev"
+makedepends=""
+subpackages=""
+
+source="${pkgname}-${pkgver}-${pkgrel}.tar.gz::https://github.com/pavelmachek/unicsy_demo/archive/${pkgver}.tar.gz"
```
ollieparanoid: Please remove the `pkgrel` from the source variable.
The `pkgrel` gets increased when something in the package gets changed, that is *not* related to the source, so the source should not get downloaded again.

pavelmachek:
Makes sense, I fixed it not to redownload. I pushed the changes.

drebrez:
@ollieparanoid @pavelmachek I would like to test it on my devices, but I noticed some hardcoded paths in the [hardware.py](https://github.com/pavelmachek/unicsy_demo/blob/master/lib/hardware.py) file (battery, leds,backlight,...).
We should consider adding those paths in some variables in the `deviceinfo` file of the device.

pavelmachek:
I re-tested on N900, still works fine.

pavelmachek:
@drebrez Which device is that? Can you install the package and see
what works?

Yes, we'll need to do better job on different hardware, but I don't
think we want to start mechanically moving strings to deviceinfo.

For example... backlight. We probably want to just use
/sys/class/backlight/*/ if there's just single entry there. No need to
add that to deviceinfo.

Situation might be different with accelerometer. There, we may want to
list known accelerometer devices, and just use one that is
available. Or laybe lis3lv02d is quite common and we'll not need to do
a list.

I did support on hardware I have working. More testing on different
hardware is needed, and then we can select best solutions.

drebrez:
@pavelmachek I agree that we don't need all the paths in variables, it's better if the tool tries to find it by itself.
I will try it on the various device I have (https://wiki.postmarketos.org/wiki/The-big-list-of-who-has-what-device#drebrez) and try to identify the paths that are different for each device.

ollieparanoid:
Autodetection where possible, and using the deviceinfo otherwise sounds good to me! I've merged this PR so we have an initial version, that works on the N900 and recommend that we discuss how to implement support for different devices in the program's own issue tracker:

https://github.com/pavelmachek/unicsy_demo/issues

Thanks again!

PR #647, opened 2017-09-28
postmarketos-base: Execute setup-udev, remove mosh from dependencies

ollieparanoid:
(`postmarketos-ui-hildon` package depends on this: #649)

---
* Run setup-udev `[0]`, because that is what `setup-xorg-base` does `[1]`.
It configures more than one OpenRC service and it also messes with the
runlevels. Only after running that, Hildon works (tested in Qemu).

`[0]`: https://github.com/alpinelinux/aports/blob/2fbea8b8dae00613de97862e509c6f76988ddeaf/main/eudev/setup-udev
`[1]`: https://github.com/alpinelinux/alpine-conf/blob/06300dd48343364aee4136dd6dd36e251188e8b8/setup-xorg-base.in#L25

* Remove `mosh` from `postmarketos-base` dependencies. I don't think anyone
uses it right now, and it's easy to add custom packages in `pmbootstrap init`
nowadays.

* Minor style changes in the `APKBUILD`.

PS: I think this should be outside of the `for` loop, I'll make an upstream PR soon:
https://github.com/alpinelinux/aports/blob/2fbea8b8dae00613de97862e509c6f76988ddeaf/main/eudev/setup-udev#L21-L26


ollieparanoid:
Thanks @craftyguy!

PR #648, opened 2017-09-28
Fix #633: start postmarketos-demos with weston

### LINKED ISSUE #633: postmarketos-demos not started from /etc/xdg/autostart, causing race condition

ollieparanoid:
We thought, that weston would honor `/etc/xdg/autostart` (see #633). However, that was *not* the case when I tested it, and also weston's source code does not have any trace of that folder :-\

So I did a different workaround. Try to run `postmarketos-demos` for 4 seconds, each half second.

~`dont_merge_pr` because of the `FIXME` in the code. After #647 is merged, we should test if we need the udev hacks at all - and if not, we can remove them with this PR.~ Tested on `i9100` and touch screen is still working after removing these lines. Yay, one less dirty hack! \o/


MartijnBraam:
I've looked through the weston mailinglist again and it looks like there was some talk about implementing xdg autostart in 2015 but they never implemented it :(

aports/main/postmarketos-ui-weston/start_weston.sh
```diff
@@ -17,10 +17,21 @@ if test -z "${XDG_RUNTIME_DIR}"; then
...
 		udevadm settle
-		# Workaround! See #633 for discussing a proper fix
-		(sleep 3; postmarketos-demos) &
+
+		# #633: Weston doesn't support autostarting applications (yet), so
+		# we try to run postmarketos-demos for 4 seconds, until it succeeds.
+		(
+			for i in $(seq 0 7); do
+				postmarketos-demos && break
+				sleep 0.5
```
MartijnBraam: I think we should add an initial sleep here since we start this the same time as weston itself so the first few invocations will fail guaranteed. Or just simply swapping the sleep and postmarketos-demos line

ollieparanoid: Good catch!

ollieparanoid:
It's ready now - would be cool if someone could test this. @montvid for example, because you needed the 3s sleep.

montvid:
Well, it seems this solution works better than sleep 3 as sleep 3 does not work sometimes. At first I got an empty screen with this code but after that I did more than 10 tests and the demos booted fine. Btw if you kill weston on weston reboot the demos load always! So this is a kind of boot issue. Thanks for the fix. Does it work on other devices too?

ollieparanoid:
Thanks for the exhaustive testing! I have increased it to trying for 10 seconds now, so it should really start every time now.

montvid:
I tried testing the demo menu again with the button "restart weston" and sometimes I get 2 menus and sometimes 1 or none at all. I will try to set different times and see how it goes...

Issue #199, opened 2017-07-19
Do not run weston as root


craftyguy:
We shouldn't be running weston as root. Supporting this will likely require re-enabling weston-launcher, which is currently disabled in the weston APKBUILD in aports.

craftyguy:
Since the APKBUILD upstream also has these configuration options for disabling weston-launch and suid install, I filed an upstream bug here: https://bugs.alpinelinux.org/issues/7556

Probably worth keeping this issue open in github though, since profile.d/postmarketos.sh will have to change if upstream APKBUILD is modified to build these required components.

ollieparanoid:
In my experience with Alpine's bug tracker it takes some time until you get a response. I've had better results with this approach:

* Compile the change you want to make yourself (it's easy with weston anyway, because it is in our own aports currently)
* Test it on your device
* Make an upstream PR
* Hang out in `#alpine-devel` and ask if this could be merged

Then the Alpine devs have less work, and our changes go through faster (and we actually tested them our selves) :)

(For complex changes I suggest asking in IRC first.)

PabloCastellano:
The reason seems to be that weston-launcher is not compatible with musl (it uses GNU's `error()`).

I found [a patch](https://lists.freedesktop.org/archives/wayland-devel/2016-September/031179.html) from September 2016 to fix it that never got merged.

Since we are having building issues (see https://github.com/postmarketOS/pmbootstrap/issues/546) with weston now that weston 3.0.0 is out and available for Alpine, I've merged this patch in https://github.com/postmarketOS/pmbootstrap/pull/495.

Maybe we can kill two birds with one stone by updating weston to 3.0.0 and launching it using weston-launcher from now on.

drebrez:
@craftyguy @PabloCastellano @ollieparanoid now that we have weston 3.0.0 we should try to make it work without root.

craftyguy:
Can you built Weston-launcher now? I think that suid utility is the only was to run it as a regular user.

drebrez:
@craftyguy I didn't built anything, weston-launcher was already present in 3.0.0, I think thanks to @PabloCastellano 's patch

craftyguy:
Ah, nice! In that case, probably just a matter of someone trying to use it to start weston as `user` :)

drebrez:
I can confirm that weston can be started as user.

Here what I've tried:
```
$ sudo groupadd weston-launch
$ sudo usermod -a -G weston-launch $USER
[...]

PR #649, opened 2017-09-28
Package postmarketos-ui-hildon

ollieparanoid:
This depends on #647.

When running `pmbootstrap init`, you will be able to choose `hildon` with this PR.
I was able to boot straight into Hildon with Qemu with this (and #647).


ollieparanoid:
(@PabloCastellano clicked you by accident.) Reviewing makes sense now, but testing really depends on #647.

aports/main/postmarketos-ui-hildon/APKBUILD
```diff
@@ -0,0 +1,34 @@
+pkgname=postmarketos-ui-hildon
+pkgver=1
+pkgrel=0
+pkgdesc="Lightweight desktop for resistive touchscreens"
```
drebrez: @ollieparanoid it's only for `resistive touchscreens`?

drebrez:
Code looks good, maybe the package description can be improved.

aports/main/postmarketos-ui-hildon/APKBUILD
MartijnBraam: Well it doesn't implement any multitouch features, but I guess most multitouch DEs also don't implement that a lot, only pinch to zoom on some places.

ollieparanoid: I've read in the channel, that it is suitable for resistive touchscreens, because it implements long presses for bringing up context menus or something like that. Ideas for better descriptions welcome. The one from Wikipedia for example is much too long :-\

ollieparanoid:
@PureTryOut wrote in the chat:
>  I see you made mesa-dri-swrast a dependency of postmarketos-ui-hildon. since it's more of a Qemu dependency to get 2d acceleration working, shouldn't it be in the Qemu deps instead? 

AFAIK, this is not only for qemu, but also for all devices where we have no proper GPU driver, such as the N900. For the Androids, we have 2D framebuffer drivers in the kernel at least. Yeah, we could also put it in the device packages instead, would you prefer that?

craftyguy:
@ollieparanoid not that you asked me or anything, but adding mesa-dri-swrast to the device package (e.g. N900) if it's a relatively low number of devices that would use it seems like a good idea to me. I just assumed that any device could fallback to using llvmpipe (swrast) if anything better was unavailable for some reason.

ollieparanoid:
That was worded poorly, I'm interested in all opinions 😉
Both directions make sense to me. I suggest we put it in the device package for now, and if it turns out, that it would have a major advantage in practice to have the fallback for all devices, we could add it to `pmbootstrap-base` - how about that?

PureTryOut:
Yeah, imo adding it to the device packages makes more sense.

drebrez:
Just tried it on samsung-i9070 without success, here is the error log => https://hastebin.com/qipojeviti.sql

AeternaMia:
I had the same issue. Try adding `xf86-input-evdev,xf86-video-fbdev` to your extra packages at `pmbootstrap init` time.

craftyguy:
@drebrez
what's your full Xorg.0.log file?

drebrez:
To rotate the screen using the `fbdev` driver:
```
cat > /etc/X11/xorg.conf.d/1-fbdev.conf << EOF
Section "Device"
  Identifier "LCD"
  Driver "fbdev"
  Option "Rotate" "CW"
EndSection
EOF
```
reference: https://www.x.org/archive//X11R6.8.0/doc/fbdev.4.html

drebrez:
@ollieparanoid, by adding the packages @AeternaMia suggested (`xf86-input-evdev`, `xf86-video-fbdev`), started working.

aports/main/postmarketos-ui-hildon/APKBUILD
craftyguy: "Lightweight desktop, optimized for single-touch touchscreen devices"

aports/device/device-qemu-amd64/APKBUILD
```diff
@@ -1,11 +1,11 @@
 pkgname=device-qemu-amd64
 pkgver=1
-pkgrel=4
+pkgrel=5
 pkgdesc="Simulated device in qemu with an x86 platform"
 url="https://github.com/postmarketOS"
 arch="noarch"
 license="MIT"
-depends="linux-postmarketos"
+depends="linux-postmarketos xf86-video-qxl"
```
craftyguy: What if someone is running wayland in qemu and not X11? Then this would be installed needlessly! This seems very much related to what we discussed in #499, and adding this package would best be handled by an `install_if`.. if only `install_if` worked!

AeternaMia:
I still needed to do `sudo ln -s /usr/share/themes/alpha /usr/share/themes/default` (as mentioned on https://wiki.postmarketos.org/wiki/Hildon) after installation (I found errors in `/tmp/x11.log` that it couldn't find the theme at that location)

AeternaMia:
I'm currently poking at this set of errors I get in `/tmp/x111.log`:
```
** (hildon-desktop:1382): CRITICAL **: DefaultBackgroundColor: unknown color
** (hildon-desktop:1382): WARNING **: No active views. Make first view active
** (hildon-desktop:1382): CRITICAL **: SelectionColor: unknown color
** (hildon-desktop:1382): CRITICAL **: LargeSystemFont: unknown font
** (hildon-desktop:1382): CRITICAL **: SystemFont: unknown font
** (hildon-desktop:1382): CRITICAL **: SmallSystemFont: unknown font
** (hildon-desktop:1382): CRITICAL **: DefaultTextColor: unknown color
** (hildon-desktop:1382): CRITICAL **: NotificationTextColor: unknown color
** (hildon-desktop:1382): CRITICAL **: NotificationSecondaryTextColor: unknown color
** (hildon-desktop:1382): CRITICAL **: TitleTextColor: unknown color
** (hildon-desktop:1382): CRITICAL **: SystemFont: unknown font
```
The definitions in the `gtkrc` in the theme files in `/usr/share/themes/default` look plausible, and `fc-match` doesn't fail to resolve a font by the same name

ollieparanoid:
Thanks for all the feedback everyone!

TODOs:
* [ ] change pkgdesc
* [ ] set the default gtk theme to alpha
* [ ] put `mesa-dri-swrast` in the relevant device packages, remove it from `postmarketos-ui-hildon`
* [ ] xf86-video-qxl should only get installed if both device-qemu-* and x11 are installed
* [ ] add the [screen rotation](https://github.com/postmarketOS/pmbootstrap/pull/649#issuecomment-333320142) to the relevant packages (also only makes sense, when x11 is installed)
* [ ] do the same for xf86-input-evdev,xf86-video-fbdev?

aports/main/postmarketos-ui-hildon/postmarketos-ui-hildon.post-install
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# Autologin on tty1, let busybox autoconfigure 2-6
+autologin="user"
+for i in 1 2 3 4 5 6; do
+	old="^tty$i::respawn:/sbin/getty 38400 tty$i"
+	new="# tty$i::respawn:/sbin/getty 38400 tty$i"
+	[ "$i" = "1" ] && new="tty1::respawn:/bin/login -f $autologin"
```
pavelmachek: Hmm. Weston needs this change, too, xfce4 will want it, too. Perhaps we should just do autologin on tty1 unconditionally? ... because this sed job is quite ugly. (For example, it will persist after package is uninstalled).

pavelmachek: ...and having to log-in makes little sense on phone, anyway...

ollieparanoid: We can't just provide the `/etc/inittab` file, because that is provided by a package from Alpine (unless we replace that package entirely, which we might do at some point).

However, we had the autologin in `postmarketos-base` instead of `postmarketos-ui-weston` until recently, but as `weston` requires root (and not "user" to run, bug #199), we moved it to the `weston` package.

In my opinion, the proper fix would be a display manager #656.

aports/main/postmarketos-ui-hildon/start_hildon.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# Hildon autostart on tty1 (Autologin on tty1 is enabled in
+# /etc/inittab by postmarketos-base post-install.hook).
+# This is a temporary solution, we'll need something like a
+# login manager in the long run.
+if [ "$(id -u)" = "12345" ] && [ "$(tty)" = "/dev/tty1" ]; then
+	# Start X11 with Hildon
+	startx /etc/postmarketos-ui/xinitrc_hildon.sh >/tmp/x11.log 2>&1
+
```
pavelmachek: Redirect to /tmp/x11.log is dangerous. Someone could have put symlink there. Better put it to ~ or something.

ollieparanoid: The right location for logs is probably `/var/log` (but then again this is not user-writeable). This could also be done properly with a display manager.
Could you explain that symlink attack scenario a bit more?

aports/device/device-qemu-amd64/APKBUILD
ollieparanoid: Actually, it does! Implemented as you recommended in #696.

pavelmachek:
/var/log is not user writable... but you could put log into
/home/user...

Attack scenario is... If I have login "pavel" on your phone, I can do
ln -s /home/user/your_precious_source.c /tmp/x11.log , and then wait
for you to restart Hildon. On next boot, your precious source file will be
gone.

Now... different users on a phone are maybe unusual, but pmos still
uses different uids for some daemons to enhance security.

ollieparanoid:
Thanks for the explanation! Yes, then we should put the log into the home folder, until we have a proper display manager. We should also fix this for Weston, I'll make a PR for that soon.

ollieparanoid:
Alright, I've fixed everything addressed in the feedback. Shipping it now!
Thanks again for all the reviewing and testing!

PR #650, opened 2017-09-28
Fix missing Hildon dependencies (found with `pmbootstrap build --strict`)

ollieparanoid:
Here's how one can find them:
```shell
cd aports/maemo
pmbootstrap build --force --strict *
```


ollieparanoid:
Thanks for reviewing!

PR #653, opened 2017-09-29
Add fstab

ata2001:


aports/main/postmarketos-base/postmarketos-base.post-install
```diff
@@ -19,6 +19,13 @@ if [ "${deviceinfo_msm_refresher}" = "true" ]; then
 	rc-update -q add msm-fb-refresher boot
 fi
 
+# Install /etc/fstab
+{
```
ollieparanoid: How about adding a `if ! apk audit /etc | grep -q etc/fstab; then` here, like we did below for `issue` and `motd`?

aports/main/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -182,7 +182,7 @@ resize_root_filesystem() {
 mount_root_partition() {
 	partition="$(find_root_partition)"
 	echo "Mount root partition ($partition)"
-	mount -w -t ext4 "$partition" /sysroot
+	mount -t ext4 -o ro "$partition" /sysroot
```
ollieparanoid: Why the `-o ro` here?

craftyguy: I don't want to mount rootfs as RO on the N900

ata2001: @ollieparanoid That's how it's done on most desktop systems, it gets remounted as rw later based on the fstab.

ollieparanoid: Makes sense, thanks for the explanation. @craftyguy, does this cover your use case, or do you mean something else?

craftyguy: I see. I will test this on my N900 to see if it does as advertised, but I don't see any obvious reasons why it would not :)

aports/main/postmarketos-base/postmarketos-base.post-install
ata2001: @ollieparanoid Sorry, forgot that, fixed it now.

ollieparanoid: Clever fix!

craftyguy: Wow, that's elegant as fuck! Nice work @ata2001

ollieparanoid:
@ata2001, thanks for making this!
@craftyguy, thanks for testing!

Issue #631, opened 2017-09-26
[BUG] TWRP/CWM flashable zip not work.


zhenyolka:
I make flashable zip. system large is 650mb. i try install with twrp 3.1. I get error on step "INSTALLING rootfs...", in log i see 
"tar -xz rootfs.tar.gz /system
KILLED" i see error about 10 second on start extracting rootfs. I get this error,because my RAM(1 GB)is very small(i see it in output of df command, i run it before start flashing and in time of flashing).

dakk:
How did you generate the flashable zip?

zhenyolka:
pmbootstrap install --android-recovery-zip

ata2001:
@zhenyolka I guess the issue is, the tar archive is extracted from the zip first, so it is present both in the zip, and extracted in /tmp, both in tmpfs. Could you please post the whole log file?

zhenyolka:
@ata2001 i flash success. need not touch phone and poweroff screen. i this case /tmp is empty and free space for flash

ata2001:
@zhenyolka Please try [this patch](https://github.com/postmarketOS/pmbootstrap/commit/fc5e52c5910875fdce78754eee060503f6259e47) out!

PR #654, opened 2017-09-29
Don't extract tar archive in tmpfs when installing from recovery

ata2001:
Closes: #631 (possibly)


drebrez:
I can test it, will let you know the result soon.

ata2001:
@ollieparanoid I've tested it of course, but I don't know, whether it solves the issue mentioned above, because @zhenyolka did not reply yet.

ollieparanoid:
My bad, I forgot about #631. So we wait with the merge until @zhenyolka respons whether this fixes the issue? Or would you prefer if we merged it already, because it is an improvement even if it does not fix that issue completely?

ata2001:
Does it really reduce memory usage?

drebrez:
@ata2001 @ollieparanoid I tested it and it works, now the percentage stops at 4%, and continues when on screen it shows `Installing rootfs...`. Also gives a better idea of the progress. No idea about the memory usage, probably it's better now

ollieparanoid:
I am pretty sure that the memory usage is better now. Without this PR, the archive would get extracted to `/tmp` (to RAM) first, and now it gets streamed into the tar extraction utility, one chunk at a time.

Thanks for testing @drebrez! Merging this, as it is an improvement ("gives a better idea of the progress" + memory usage). Great work, @ata2001! 🎉

PR #660, opened 2017-09-30
Start swclock on boot

### LINKED ISSUE #570: [rx51] real time clock problems
As soon as you remove battery, RTC loses time. Unfortunately, I remove batteries all the time during development. Wrong time and/or time going backwards does bad stuff to unix system: git clone fails to work due to expired certificates, make fails to rebuild binaries, etc.

I re-built pmOS image, and system booted with date in 1970, even through /etc/passwd had todays date. We can do better.

I'm using this:

#!/bin/bash
F=/etc/zaurus.time
if [ $(date +%s) -lt $(< $F) ];
    then
    date -s @$(< $F)
    fi
date +%s > $F

..runing it during boot and during shutdown. That way, I get approximately right time, good enough for https and make. I see bash might be an issue; if needed, I can rewrite it in sh, C or Python. Would something like this be acceptable?

craftyguy:
Fixes #570. @pavelmachek gets credit for finding this service


drebrez:
@craftyguy isn't it better to put this fix for all the devices? they probably have the same issue with the clock

craftyguy:
Do other devices have this problem? The RTC on the N900 relies on a
small battery on the board and is almost always going to be
non-functional at this point due to age. If other devices have a hard
time keeping track of time between reboots and/or poweroffs, then
I can change the PR to make it default for any device running pmOS.

drebrez:
@craftyguy dont know exactly, here is what I tried on the samsung-i9070:
```
localhost:~$ date
Sun Jan  1 02:00:02 GMT 2012
localhost:/home/user# rc-update add swclock boot
 * service swclock added to runlevel boot
localhost:/home/user# rc-service swclock start
 * Setting the local clock based on last shutdown time ...                                                                              [ ok ]
localhost:/home/user# date
Tue Mar 21 12:38:26 GMT 2017
```

craftyguy:
I see. Ok, I'll re-adjust this PR

craftyguy:
@drebrez 

PR has been changed to start swclock on all devices at boot. Can you please test that this works for your device too?

drebrez:
I tested it and it works 👍

steps done:
- boot pmos
- set specific date
- reboot
- check date in initramfs => year 1970
- boot pmos
- date restored correctly

Issue #659, opened 2017-09-30
Is it possible to get more packages to build with cross compilation in the native chroot?


zhuowei:
Currently pmbootstrap builds all packages with qemu user mode emulation except for the [Linux kernels](https://github.com/postmarketOS/pmbootstrap/blob/master/pmb/config/__init__.py#L135). This is slower than building with a cross compiler, as one has to emulate the build system using qemu even if the compiles themselves are done with distcc, and [distcc frequently fails](https://github.com/postmarketOS/pmbootstrap/issues/480).

abuild [supports cross building](http://lists.alpinelinux.org/alpine-devel/5427.html) if there is [a cross compiler and a sysroot](https://dev.alpinelinux.org/~tteras/bootstrap/abuild-crossbuild-aarch64.conf), and if the build system itself supports cross building (autotools and cmake are common build systems that do support cross compiles). Could we try enabling cross compile for normal packages? We already have a sysroot (our target chroot), and we already have the cross compilers for building the kernels. How can I help integrate this?

ollieparanoid:
### Terminology
There are `"native"` and `"distcc"` as cross-compile types available in pmbootstrap ([details](https://wiki.postmarketos.org/wiki/Build_internals#Cross-compile_types)), just to make sure that it is clear what I'm talking about.

### Why it was implemented this way
The idea was to make sure, that we really can cross-compile all packages, and this approach is [widely used in Arch Linux ARM](https://archlinuxarm.org/wiki/Distcc_Cross-Compiling) already.

### How enabling "native" for more packages works now
As you have seen already, the way to enable another package for "native" cross-compilation is extending this list with the package names:

https://github.com/postmarketOS/pmbootstrap/blob/cfbf06ee71eed6f9bc1ec40087c3a6c69ab629c6/pmb/config/__init__.py#L133-L135

One example is the WIP [openiboot](https://github.com/postmarketOS/pmbootstrap/issues/399#issuecomment-323116579) packaging.

### "native" for almost everything?

The tricky part is, when the package, that you are about to compile, depends on libraries of the same architecture. We don't handle this in `pmbootstrap` yet, and you raise an excellent point there. When thinking about it, it might be pretty feasible to do that actually.

So you asked how to integrate this. I recommend starting with a proof-of-concept shell script, something along the lines of the following (not tested, will need some adjusting to actually work).
If we get it working that way, we could brainstorm how to integrate this in pmbootstrap.
```shell
#!/bin/sh -e
[...]

Issue #546, opened 2017-09-10
weston build missing /usr/lib/libweston-desktop-2.so.0.0.0 library


zhuowei:
The weston package seems to be missing a library file:

```pmbootstrap/chroot_rootfs_huawei-angler/usr/lib$ ls -l libweston-desktop-2.so.0
lrwxrwxrwx 1 root root 28 Sep  4 01:53 libweston-desktop-2.so.0 -> libweston-desktop-2.so.0.0.0
```

And listing the package confirms that the library is missing: 

```pmbootstrap/packages/aarch64$ tar tf libweston-desktop-9999-r2.apk
.SIGN.RSA.user-59ac85cf.rsa.pub
.PKGINFO
usr/
usr/lib/
tar: Ignoring unknown extended header keyword 'APK-TOOLS.checksum.SHA1'
usr/lib/libweston-desktop-2.so.0
```

I'm using the latest master (e421313632676cf2e332e62a0fe039a8499a9021) with a few Nexus 6P patches (no changes outside of aports/device/device-huawei-angler, aports/device/linux-huawei-angler)

drebrez:
@zhuowei I'm getting the same issue, I think it might be due the recent update of libweston to the version 3.0.0

ollieparanoid:
Yep, weston just got updated to 3.0.0 in Alpine upstream. Does this issue go away when building weston 3.0.0 from #495? It is in branch `update-weston-3.0.0`.

drebrez:
I've found another solution, simply adding the `libweston` package to the weston dependencies solve it, creating the PR right now

drebrez:
@ollieparanoid I'm not sure anymore about my solution, adding `libweston` dependency to our weston APKBUILD is kind of strange.
What I've done was, add the depends and run `pmbootstrap build weston --arch=armhf`, this fixed it.
Maybe the build was enough without adding the dependency.
@zhuowei can you try to rebuild the package to see if it works?

zhuowei:
@drebrez I did a rebuild with `pmbootstrap build weston --force -- arch=aarch64` before filing the bug report; it didn't help. Should I try again?

zhuowei:
@ollieparanoid I'm getting a different error after cherry-picking the weston 3 branch:

```
localhost:/usr/lib$ cat /tmp/weston.log 
Error loading shared library libweston-3.so.0: No such file or directory (needed by /usr/bin/weston)
[...]

PR #665, opened 2017-10-01
pmbootstrap: support native cross compile for Autoconf packages

zhuowei:
Note: this PR depends on #664, so its commits are duplicated in the changes list. Only the last commit is specific to this PR.

This commit adds support for cross-compiling Autotools-based packages without distcc and qemu. This should be faster and more reliable.

See #659 for more information.

This installs dependencies in the foreign arch chroot, then bind mounts the foreign chroot into the native chroot for a cross compile sysroot. Finally, this sets env variables to trigger a cross compile based on https://dev.alpinelinux.org/~tteras/bootstrap/abuild-crossbuild-aarch64.conf

Currently chocolate-doom is the only Autotools package whitelisted to build using this configuration, and the executable it builds for aarch64 seems to at least start (inside the aarch64 build chroot, not tested on device; don't have a Doom WAD handy to test with)

====

Other changes I need to do:

- [ ] Do I need to install makedepends in the builder root as well as in the target root? Does the target root ever need the makedepends packages? 

Note: some packages in upstream aports have both a makedepends_host and makedepends_build variable, which is used by abuild to install the correct packages to each sysroot. However, the kernel sources do not.

- [ ] Make sure kernels still build
- [ ] Whitelist a few more packages and compare builds - can pmbootstrap's compare function work for this?
- [ ] Switch to a heuristic for whitelisting packages (maybe anything containing `./configure --host=$(HOST)` or something?) - should probably be controlled by a config var?
- [ ] Think of a way to warn when someone submits an APKBUILD that wouldn't work for cross compile?

Please suggest test cases!

### DIFF
--- a/pmb/build/package.py
+++ b/pmb/build/package.py
@@ -62,7 +62,9 @@ def package(args, pkgname, carch, force=False, buildinfo=False, strict=False):
     if native_cross_with_deps:
         pmb.build.init(args, "buildroot_" + carch)
     if len(apkbuild["makedepends"]):
+        # build: install in build chroot
         makedepends_build = apkbuild["makedepends"]
+        # host: install in foreign sysroot if cross compiling
         makedepends_host = []
         if native_cross_with_deps:
             if len(apkbuild["makedepends_build"]) != 0 and len(apkbuild["makedepends_host"]) != 0:
@@ -80,7 +82,7 @@ def package(args, pkgname, carch, force=False, buildinfo=False, strict=False):
         else:
             pmb.chroot.apk.install(args, makedepends_build, suffix)
             if len(makedepends_host) != 0:
-                pmb.chroot.apk.install(args, makedepends_build, "buildroot_" + carch)
+                pmb.chroot.apk.install(args, makedepends_host, "buildroot_" + carch)
 
     if cross:
         pmb.chroot.apk.install(args, ["gcc-" + carch_buildenv,
--- a/pmb/build/package.py
+++ b/pmb/build/package.py
@@ -59,13 +59,29 @@ def package(args, pkgname, carch, force=False, buildinfo=False, strict=False):
 
     # Initialize build environment, install/build makedepends
     pmb.build.init(args, suffix)
+    if native_cross_with_deps:
+        pmb.build.init(args, "buildroot_" + carch)
     if len(apkbuild["makedepends"]):
+        makedepends_build = apkbuild["makedepends"]
+        makedepends_host = []
+        if native_cross_with_deps:
+            if len(apkbuild["makedepends_build"]) != 0 and len(apkbuild["makedepends_host"]) != 0:
+                makedepends_build = apkbuild["makedepends_build"]
+                makedepends_host = apkbuild["makedepends_host"]
+            else:
+                makedepends_host = makedepends_build
+
         if strict:
-            for makedepend in apkbuild["makedepends"]:
+            carch_makedepends_build = args.arch_native if native_cross_with_deps else carch_buildenv
+            for makedepend in makedepends_build:
+                package(args, makedepend, carch_makedepends_build, strict=True)
+            for makedepend in makedepends_host:
                 package(args, makedepend, carch_buildenv, strict=True)
         else:
-            deps_sysroot_suffix = "buildroot_" + carch if native_cross_with_deps else suffix
-            pmb.chroot.apk.install(args, apkbuild["makedepends"], deps_sysroot_suffix)
+            pmb.chroot.apk.install(args, makedepends_build, suffix)
+            if len(makedepends_host) != 0:
+                pmb.chroot.apk.install(args, makedepends_build, "buildroot_" + carch)
+
     if cross:
         pmb.chroot.apk.install(args, ["gcc-" + carch_buildenv,
                                       "g++-" + carch_buildenv,
@@ -92,6 +108,8 @@ def package(args, pkgname, carch, force=False, buildinfo=False, strict=False):
         logging.info("WARNING: Option !tracedeps is not set, but we're"
                      " cross-compiling in the native chroot. This will probably"
                      " fail!")
+        if strict:
+            logging.info("Strict building is not supported by native cross compile yet!")
 
     # Run abuild
     pmb.build.copy_to_buildpath(args, pkgname, suffix)
@@ -112,9 +130,13 @@ def package(args, pkgname, carch, force=False, buildinfo=False, strict=False):
             env["CHOST"] = hostspec
             env["CROSS_CFLAGS"] = "--sysroot=" + cbuildroot
             env["CPPFLAGS"] = "--sysroot=" + cbuildroot
-            env["LDFLAGS"] = "\"--sysroot=" + cbuildroot + " -L" + cbuildroot + "/lib\""
-            # FIXME: this isn't enough; SDL's pkg-config still returns just /usr/include/SDL
-            # Need to make a wrapper script that sets these immediately before running real pkg-config
+            # the rpath-link is a workaround: proper solution is to add a ld.so.conf?
+            # https://sysprogs.com/w/fixing-rpath-link-issues-with-cross-compilers/
+            env["LDFLAGS"] = "\"--sysroot=" + cbuildroot + " -L" + cbuildroot + \
+                             "/lib -Wl,-rpath-link," + cbuildroot + "/lib," + \
+                             "-rpath-link," + cbuildroot + "/usr/lib\""
+            # FIXME: check if this is enough: SDL's pkg-config still returns just /usr/include/SDL?
+            # May need a wrapper script that sets these immediately before running real pkg-config
             env["PKG_CONFIG_PATH"] = cbuildroot + "/usr/lib/pkgconfig/:" + cbuildroot + "/usr/share/pkgconfig"
             env["PKG_CONFIG_SYSROOT_DIR"] = cbuildroot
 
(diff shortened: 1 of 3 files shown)


zhuowei:
Added Weston to the whitelist. Haven't had a chance to try the compiled binary yet.

Also, note that strict mode isn't supported yet.

ollieparanoid:
Awesome that you threw something together so quickly @zhuowei! \o/

> Note: some packages in upstream aports have both a makedepends_host and makedepends_build variable, which is used by abuild to install the correct packages to each sysroot. However, the kernel sources do not.

The kernel does not depend on anything in the build environment, if I'm not mistaken.


> compare builds - can pmbootstrap's compare function work for this?

Excellent idea! Yes, that should be possible:

```shell
git checkout master
pmbootstrap build hello-world --force --buildinfo
git checkout native-cross
pmbootstrap challenge $(pmbootstrap config work)/packages/x86_64/hello-world-1-r2.apk
```

>  Switch to a heuristic for whitelisting packages (maybe anything containing ./configure --host=$(HOST) or something?) - should probably be controlled by a config var?

> Think of a way to warn when someone submits an APKBUILD that wouldn't work for cross compile?

I would create a function, that checks one package for this heuristic, and expose that directly as pmbootstrap command to run on either one specific or all packages (just like `pmbootstrap kconfig_check`). We could keep a list of packages, that are known to require distcc compilation, and when there was a new package found, that is not in this list, pmbootstrap could fail (with a good explanation, possibly a wiki entry). Then we could put that in the TravisCI job.

> Please suggest test cases!

You suggested using `pmbootstrap challenge` already, which is perfect to test if the output of the packages is the same. Just run that on a bunch of packages :)

Please note, that I have not looked at the code diff yet, but I can do this at this early stage if you want me to - or wait until this PR is further progressed.

Thank you very much for this!

zhuowei:
@ollieparanoid Thanks: I would appreciate a code review, as the remaining component for this PR is mainly testing and supporting strict mode.

pmb/build/package.py
```diff
@@ -51,19 +51,37 @@ def package(args, pkgname, carch, force=False, buildinfo=False, strict=False):
     suffix = pmb.build.autodetect.suffix(args, apkbuild, carch_buildenv)
     cross = pmb.build.autodetect.crosscompile(args, apkbuild, carch_buildenv,
                                               suffix)
+    native_cross_with_deps = cross == "native" and "!tracedeps" not in apkbuild["options"]
```
ollieparanoid: I don't understand why you check for the `!tracedeps` flag here. The `tracedeps` is for automatically finding dependencies once the package is built, by scanning the resulting binaries for all libraries which they link against (with `scanelf`).

I think this is problematic in general, as this may not necessarily work for a `native` cross-compiled package. When I tested with `weston`, it *seems* to be working with `aarch64`, but not with `armhf` (`scanelf` fails there, causing the build to fail). So we would need to add `!tracedeps` for all packages that we want to compile with `native`, which means the `native_cross_with_deps` will always be `False`.

In fact, why do we need this condition anyway, how about executing the new code whenever a package would get compiled with the `native` method?

**Note:** Because `!tracedeps` is a hack here, that avoids `scanelf`, I would recommend that we set it dynamically for the `native` method, after copying the `APKBUILD` to `/home/user/build` (we could do this in a new PR to get this one in faster).

pmb/build/package.py
```diff
@@ -86,10 +104,12 @@ def package(args, pkgname, carch, force=False, buildinfo=False, strict=False):
     logging.info("(" + suffix + ") build " + output)
 
     # Sanity check
-    if cross == "native" and "!tracedeps" not in apkbuild["options"]:
+    if native_cross_with_deps:
         logging.info("WARNING: Option !tracedeps is not set, but we're"
                      " cross-compiling in the native chroot. This will probably"
                      " fail!")
+        if strict:
+            logging.info("Strict building is not supported by native cross compile yet!")
```
ollieparanoid: Because we don't have a way to let `abuild` install the dependencies here, and because `!tracedeps` is problematic, I recommend that when `!strict` is defined, the `native` method will never be used, so it falls back to the (slower) `distcc` method.

pmb/config/__init__.py
```diff
@@ -132,7 +132,7 @@
 
 # fnmatch for supported pkgnames, that can be directly compiled inside
 # the native chroot and a cross-compiler, without using distcc
-build_cross_native = ["linux-*"]
+build_cross_native = ["linux-*", "chocolate-doom", "weston"]
```
ollieparanoid: I recommend only adding `weston` here for now, so we get this PR in faster (less testing necessary) in order to workaround #546.

zhuowei:
PR updated.
- Only whitelist weston for now
- add a config option to disable native build
- add a list of packages to build in native build with no dependencies, like the old behaviour (only kernels should be in here); get rid of tracedeps
- disable native build in strict mode
- fix linking weston on armhf

zhuowei:
@ollieparanoid The Weston build aren't identical because the libgcc.a files differ (compare /usr/lib/gcc/armv6-alpine-linux-muslgnueabihf/6.4.0/libgcc.a in the two chroots); is there something we can change in our cross toolchain to make sure it builds the same libgcc.a?

aports/cross/cmake-cross-wrappers/APKBUILD
```diff
@@ -0,0 +1,31 @@
...
+source="template.toolchain.cmake cmake.sh"
+options="!check"
+
+package() {
+	local _archs="armhf aarch64"
+	for _arch in $_archs; do
+		_bindir="$pkgdir/usr/lib/cmake-cross-wrappers/$_arch/bin"
+		_hostspec="$(arch_to_hostspec $_arch)"
+		_processor="$_arch"
+		if [ "$_processor" = "armhf" ]
```
ollieparanoid: How about shortening that?
```shell
[  "$_arch" = "armhf" ] && _processor="arm"
```

aports/cross/cmake-cross-wrappers/cmake.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+if [ "z$1" = "z-E" ]
```
ollieparanoid: How about changing that to:
```shell
if [ "$1" = "-E" ]; then
    exec /usr/bin/cmake "$@"
else
    exec /usr/bin/cmake -DCMAKE_TOOLCHAIN_FILE="$(dirname "$0")/toolchain.cmake" "$@"
fi
```
* The "z" in the syntax is only necessary if you specify a flag to `sh`, which errors on empty variables as far as I know. So I would remove it for readability here.
* To pass all arguments, please use `"$@"` instead of `$@`, as the second one can cause problems ([see also](https://stackoverflow.com/a/18492575)).

ollieparanoid:
> The Weston builds aren't identical because the libgcc.a files differ (compare /usr/lib/gcc/armv6-alpine-linux-muslgnueabihf/6.4.0/libgcc.a in the two chroots), and same with crtbegin/crtend; is there something we can change in our cross toolchain to make sure it builds the same libgcc.a/crtbegin/end?

Actually, I would not look into this more. Does the built weston package work on the device?

zhuowei:
@ollieparanoid I haven't tested the built weston yet. (Been busy.)

I also pushed some initial CMake support into this PR for review, but I might take them out since I might want to get this merged for Weston first.

ollieparanoid:
> I haven't tested the built weston yet. (Been busy.)

I hope that I didn't sound unfriendly. This is a free-time project, so no need to explain yourself there. Thanks for making this in the first place 👍

zhuowei:
@ollieparanoid No worries - thank you for reviewing this.

ollieparanoid:
Could you revert the cmake commit?

```shell
git revert 8aaf85b3a0520a328d8efc033b12f60885173688
```

zhuowei:
@ollieparanoid Rebased and removed the cmake commit: I'll send that in a separate pull request eventually.

pmb/build/package.py
```diff
@@ -99,6 +113,27 @@ def package(args, pkgname, carch, force=False, buildinfo=False, strict=False):
         hostspec = pmb.parse.arch.alpine_to_hostspec(carch_buildenv)
         env["CROSS_COMPILE"] = hostspec + "-"
         env["CC"] = hostspec + "-gcc"
+        if native_cross_with_deps:
+            # bind mount sysroot into native and set env to point to it
+            foreign_sysroot_path = args.work + "/chroot_buildroot_" + carch
+            cbuildroot = "/home/user/cross_sysroot/chroot_buildroot_" + carch
```
PabloCastellano: Since #725 the user is configurable and so, `args.user` must be used

ollieparanoid:
@PabloCastellano: good catch!

@zhuowei: I'm sorry that I didn't spend much attention on this PR lately. I'll try to get it ready to be merged within the next week (you've allowed to write to your branch for whoever reviews your PR, so I'll just do that). Thanks again for all the work and I value it very much!

ollieparanoid:
Little update: I have realized, that this code is pretty hard to integrate. And that is not because of the changes, but because `package.py` is one long function, that does too many things. So I will refactor it, and then apply @zhuowei's great research from this PR on top of it!

ollieparanoid:
Unassigned myself, as I'm not working on this at the moment. We have the binary repository now, so reducing the compile time is not such a critical task for me anymore compared to what else there is to do. The `pmb/build/_package.py` code has been refactored a lot though, and gained lots of test coverage (almost 100%?), so this should be much easier to integrate if someone wants to pick it up.

ollieparanoid:
On second thought, we have an issue for this already, so I don't think keeping this outdated PR open is helping much. When taking another shot at this, it is probably easier to do it from scratch, using this one as reference only. Closing.

PR #670, opened 2017-10-01
[grouper] Add osk-sdl config for device

craftyguy:
@zhuowei please take a look at this!


ollieparanoid:
Looks like @zhuowei doesn't have time. @craftyguy, how did you create the config? If it comes directly from @zhuowei, I would suggest we merge it.

craftyguy:
Yea it came directly from them

PR #674, opened 2017-10-01
Add ntp client & timezone config

### LINKED ISSUE #588: Choose & set up a NTP client
From @ollieparanoid in https://github.com/postmarketOS/pmbootstrap/issues/570#issuecomment-329855915:

> A few notes regarding ntpd:
>
> * If we set the hardware clock to UTC/GMT instead of the localtime, we don't need to know the time zone beforehand (it will get calculated based on the UTC time then) ([more info](https://wiki.archlinux.org/index.php/Time#Time_standard))
> * Although that might not be obvious, we should take extra care choosing a suitable `ntpd`, *especially* when we want to have it running all the time. After all, it connects to the internet and executes code based on the input it gets, so it has a similar attack surface than curl or wget. Besides busybox' `ntpd`, there is [`openntpd`](http://www.openntpd.org/features.html) from OpenBSD for example, which does proper privilege separation among other features. It is packaged for Alpine.
> * Also running an `ntpd` creates characteristic traffic, which may be used to fingerprint your OS/device (I know that we'll need it for the N900, but we should keep that in mind at least).

craftyguy:
This adds chrony as the ntp client (starting on boot), and also adds a
prompt in the init step for configuring the timezone.

fixes #588


craftyguy:
For reference, here's what the prompting looks like. This is from the `setup-timezone` script in Alpine, which I think is quite nice for navigating to the right timezone.

```
[15:14:46] Would you like to configure the timezone? (y/n) [n]: y                         
Which timezone are you in? ('?' for list) [UTC] ?                                                                                                                                                                                                                                                                                                                       
Africa/      Atlantic/    Canada/      EST5EDT      Factory      GMT-0        Iceland      Japan        MST7MDT      PRC          ROC          US/          Zulu                                                                                                                                                                                                        
America/     Australia/   Chile/       Egypt        GB           GMT0         Indian/      Kwajalein    Mexico/      PST8PDT      ROK          UTC          posixrules
Antarctica/  Brazil/      Cuba         Eire         GB-Eire      Greenwich    Iran         Libya        NZ           Pacific/     Singapore    Universal    right/
Arctic/      CET          EET          Etc/         GMT          HST          Israel       MET          NZ-CHAT      Poland       Turkey       W-SU
Asia/        CST6CDT      EST          Europe/      GMT+0        Hongkong     Jamaica      MST          Navajo       Portugal     UCT          WET
Which timezone are you in? ('?' for list) [?] US                                          
What sub-timezone of 'US' are you in? ('?' for list) ?                                    
ls: US/: No such file or directory                                                                                                                                                  
Alaska          Arizona         East-Indiana    Hawaii          Michigan        Pacific         Samoa
Aleutian        Central         Eastern         Indiana-Starke  Mountain        Pacific-New
What sub-timezone of 'US' are you in? ('?' for list) Pacific                                                                                                                                                                                                                                                                                                            
[15:15:08] *** (3/5) PREPARE INSTALL BLOCKDEVICE ***                                                                 
```

craftyguy:
> So in case the user has set it in pmbootstrap init, we could call setup timezone -z TIMEZONE in the installation to install it without a prompt.

Ah! Yea that is a good idea.. I will move this to `init`

craftyguy:
@ollieparanoid 

So I'm trying to implement this now, but this:

> I would actually hardcode the available timezones,

.. is a little crazy because there are a lot of timezone options presented by setup-timezone. There are quite [a few UTC offsets](https://en.wikipedia.org/wiki/List_of_UTC_time_offsets), and I think some timezones have different settings for DST. This would amount to duplicating all of [the options in the tzdata package](https://pkgs.alpinelinux.org/contents?branch=edge&name=tzdata&arch=armhf&repo=main). That's not really maintainable..


I still don't know a good way to do the caching part since user input is directly into the setup-timezone script and not pmbootstrap. "Reaching" into the device chroot to pull out this config seems hack-ish and prone to breaking, but maybe I am wrong!


Some options I could think of:

1) Put the tz config late in the `pmbootstrap init`, initialize the device chroot, run `setup-timezone`, then cache the result into pmbootstrap's config. 

2) keep the tz config in `install` but make an option like `Using timezone GMT, is this OK? (y/n)`, this way if running `pmbootstrap -y install` it will just roll with GMT. This would avoid the issue of 'how to cache setting in pmbootstrap config?', it would not get stuck if going `pmbootstrap -y install`, but it would prompt the user every time to set up timezone when running `pmbootstrap install`.

ollieparanoid:
Agreed with duplicating the tzdata not being a good idea.

In case we *do* enter the chroot, it should be the native chroot. The device chroot requires additional setup with qemu (even building/repacking the qemu package).

Of the two above, I would prefer 1.).

However, I have another idea. What if we simply read the host timezone? Because that is what the user usually would like to have inside the device, right?
Alpine/OpenRC based systems save the config in [`/etc/zoneinfo/localtime`](https://github.com/alpinelinux/alpine-conf/blob/06300dd48343364aee4136dd6dd36e251188e8b8/setup-timezone.in#L82), while systemd-based systems save it in [`/etc/localtime`](https://www.freedesktop.org/software/systemd/man/localtime.html). In both cases, it is a symlink to the actual timezone. So as I see it, we could use the symlink target to find out what the host timezone is, and apply the same config to the device chroot during installation.

So how about this (in init)?
```
Your host timezone: Europe/London
Use this timezone instead of GMT? [Y/n] # defaults to previous selection or Y
```

craftyguy:
Oh, interesting idea! Are the names of timezones used by systemd, openrc (and their permutations across multiple distros) consistent with the names in tzdata?

ollieparanoid:
I'm pretty sure that they all package `tzdata` :)

craftyguy:
@ollieparanoid I have implemented this now so that it will try to detect the timezone set on the host (/etc/localtime) and default to GMT if unable to. 

I have some concerns though about this implementation:

1) it requires `readlink`, which is part of `coreutils` on Arch Linux and I would assume is on other distros by default, but I really don't know.

2) It expects `/etc/localtime` to be on the host system, and for it to be a symlink to the appropriate timezone file

3) It expects `tzdata` to be on the host, and for it to be under `/usr/share/zoneinfo/`

If you decide to merge this, please preserve my concerns above in the commit message.

pmb/config/init.py
```diff
@@ -79,6 +80,32 @@ def ask_for_keymaps(args, device):
                       " one from the list above.")
 
 
+def ask_for_timezone(args):
+    localtime = "/etc/localtime"
```
ollieparanoid: How about using an array here, and iterating over both known paths?

```python
localtimes = ["/etc/zoneinfo/localtime", "/etc/localtime"]
zoneinfo_path  ="/usr/share/zoneinfo"
for localtime in localtimes:
    if not os.path.exists(localtime):
        continue
    # ...
```

pmb/config/init.py
```diff
@@ -79,6 +80,32 @@ def ask_for_keymaps(args, device):
...
+    zoneinfoPath = "/usr/share/zoneinfo/"
+    tz = ""
+    if os.path.exists(localtime):
+        try:
+            tzpath = pmb.helpers.run.user(args, ["readlink", "-f", localtime],
```
ollieparanoid: Instead of `readlink`, we can use [`os.path.realpath`](https://docs.python.org/3/library/os.path.html#os.path.realpath) here - that should also make the code a bit simpler.

pmb/config/init.py
```diff
@@ -79,6 +80,32 @@ def ask_for_keymaps(args, device):
...
+                                          return_stdout=True)
+        except:
+            tzpath = ""
+        tzpath = tzpath.rstrip()
+        if os.path.exists(tzpath):
+            try:
+                _, tz = tzpath.split(zoneinfoPath)
+            except:
+                pass
+    if tz != "":
```
ollieparanoid: How about: `if tz:`?

pmb/config/init.py
```diff
@@ -79,6 +80,32 @@ def ask_for_keymaps(args, device):
...
+        logging.info("Your host timezone: " + tz)
+        if pmb.helpers.cli.confirm(args, "Use this timezone instead of GMT?",
+                                   default="y"):
+            return tz
+    else:
```
ollieparanoid: I would remove the `else:` here, because you return above anyway (and also the `return` at the end isn't part of the `else` block).

pmb/install/install.py
```diff
@@ -296,6 +296,10 @@ def install(args):
     # Set the keymap if the device requires it
     setup_keymap(args)
 
+    # Set timezone
+    if args.timezone != "":
+        pmb.chroot.root(args, ["setup-timezone", "-z", args.timezone], suffix, log=False)
```
ollieparanoid: * We default to GMT, so there's no need to check whether it is empty, right?
* Is the `log=False` still necessary? Looks like that was for the interactive prompt.

craftyguy: It would be empty if a user pulled this change but didn't re-run `pmbootstrap init`

ollieparanoid:
@z3ntu's comment from https://github.com/postmarketOS/pmbootstrap/issues/373#issuecomment-335620826 seems to be relevant here:
> But I needed `makestep 1 -1` in my `/etc/chrony/chrony.conf` otherwise chrony wouldn't synchronize the time (it thinks it's really 1970 and apparently only changes it back slowly)

craftyguy:
> @z3ntu's comment from #373 (comment) seems to be relevant here:

Hopefully that's not needed anymore since we now use swclock to set time to something reasonable, but I will look at adding that option to `/etc/chrony/chrony.conf` just in case!

craftyguy:
@ollieparanoid 
Requested changes have been made, and tested on my N900 (including tests w/o a 'localtime' symlink set and w/o running `init`). Thank you so much for the help!

PR #675, opened 2017-10-01
Only load loop module if it is not already loaded in the kernel

### LINKED ISSUE #673: [qemu] error when modprobing loop should not be fatal
In losetup.py, I suggest something like this: I do have loop built-in, so modprobe fails, but loop is still usable.

```
--- a/pmb/install/losetup.py
+++ b/pmb/install/losetup.py
@@ -27,7 +27,10 @@ import pmb.chroot
 
 
 def init(args):
-    pmb.helpers.run.root(args, ["modprobe", "loop"])
+    try:
+        pmb.helpers.run.root(args, ["modprobe", "loop"])
+    except:
+        print("Can't modprobe loop, perhaps it is built-in?")

```

craftyguy:
This checks for /sys/modules/loop before modprobing the loop module. My
understanding is that if this module is built into the kernel, that this
directory is still created. For a kernel without loop built in,
losetup.py will try to load the module using modprobe.

fixes #673


craftyguy:
@pavelmachek please test that this works for you!

pavelmachek:
I did quick test, and this seems to work for me. Thanks!

PR #680, opened 2017-10-02
Get rid of duplicated adduser's, move adduser late, so /etc/skel has chance of working

pavelmachek:


pavelmachek:
Hmm. Adding "-k /etc/skel" to the useradd would be nice, so we get working /etc/skel.

Another potential issue -- we keep chroots between builds, so useradd will be called with /home already existing....?

pmb/install/install.py
```diff
@@ -111,6 +111,13 @@ def set_user_password(args):
     """
     logging.info(" *** SET LOGIN PASSWORD FOR: 'user' ***")
     suffix = "rootfs_" + args.device
+
+    logging.debug("Add user")
```
ollieparanoid: This will cause problems, as you have written in your comment. If I understand correctly, the point was that `skel` could be used, when you generate the user here.

How about installing `skel` before adding the user in `pmb/chroot/init.py`, and then always specifying `-k /etc/skel`? That should work for all chroots and you don't have to worry about what happens if the installer runs multiple times.

pavelmachek:
We only have /etc/skel complete after all packages are installed.

Anyway, -k at adduser time is not enough, because final copy to
/mnt/install explicitely does _not_ copy /home.

I believe I can simply copy /etc/skel to /mnt/install/home/user.

And yes, it looks like adduser needs to be done for all chroots; but
that brings problem -- user is uid 12345 in final installation and is
not in group users (etc). It would be best to separate "user for final
installation" and "user for compiling".

Help welcome. I'm not familiar enough with pmbootstrap to do a good
job.

pavelmachek:
I have created new pull request. That one should address problems pointed out here.

ollieparanoid:
(Continued in #685)

PR #684, opened 2017-10-03
Add mkfs.ext2 and mkfs.ext4 binaries to the android-recovery-installer

drebrez:
Using the unofficial TWRP for the Huawey Y530 (no official support yet) I encountered this issue:
```
+ make_ext4fs -L pmOS_root /dev/mapper/pm_crypt
/tmp/postmarketos/bin/pmos_install: line 1: make_ext4fs: not found
```
I have no idea where the `make_ext4fs` command comes from but the mkfs.ext4 file was missing. 

This PR will ensure that both the necessary binaries to create the partitions are present.

All the other commands executed in the recovery script seems related to the busybox, if in the future there will be problems with that too, we may consider to extract all the files we need in a folder and chroot it (maybe reusing some functions from [init_functions.sh](https://github.com/postmarketOS/pmbootstrap/blob/d45c7d2d844be0da16d15c74c55a17b833d4f970/aports/main/postmarketos-mkinitfs/init_functions.sh)).

### DIFF
--- a/aports/main/postmarketos-android-recovery-installer/APKBUILD
+++ b/aports/main/postmarketos-android-recovery-installer/APKBUILD
@@ -1,10 +1,10 @@
 pkgname=postmarketos-android-recovery-installer
-pkgver=0.0.4
+pkgver=0.0.5
 pkgrel=0
 pkgdesc="TWRP compatible postmarketOS installer script"
 url="https://github.com/postmarketOS"
 # multipath-tools: kpartx
-depends="busybox-extras lddtree cryptsetup multipath-tools device-mapper parted util-linux zip"
+depends="busybox-extras lddtree cryptsetup multipath-tools device-mapper parted util-linux zip e2fsprogs"
 source="build_zip.sh
 	update-binary
 	pmos_install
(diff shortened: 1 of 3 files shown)


ata2001:
@drebrez, @ollieparanoid make_ext4fs is part of the aosp source code (https://android.googlesource.com/platform/system/extras/+/master/ext4_utils/make_ext4fs.c).

I've used it, because it was the only available tool for creating ext4 filesystems in the version of twrp on my device. (I've extracted some recovery installers, to read the installer script, and they were using this command.)

ata2001:
> All the other commands executed in the recovery script seems related to the busybox, if in the future there will be problems with that too, we may consider to extract all the files we need in a folder and chroot it (maybe reusing some functions from init_functions.sh).

The idea was to make the zip file smaller by using the busybox included in the recovery (it was present in all of the twrp versions I've encountered). I agree, if this is going to cause any problem, we'll add that too, and do everything inside a chroot.

drebrez:
@ata2001 considering that the whole zip file includes the system partition, adding 300-400 kb for some additional executables it shouldn't be a problem.
If you can test it on your device is better, I didn't tested in a device with the file already present.

ata2001:
@drebrez works great

ollieparanoid:
Excellent teamwork \o/

PR #685, opened 2017-10-03
Adduser 2

pavelmachek:
Remove the adduser that ends up being useless, add copying of /etc/skel directory.


pmb/install/install.py
```diff
@@ -100,8 +100,9 @@ def copy_files_other(args):
         pmb.helpers.run.root(args, ["cp", key, rootfs + "/etc/apk/keys/"])
 
     # Create /home/user
-    pmb.helpers.run.root(args, ["mkdir", "-p", rootfs + "/home/user"])
-    pmb.helpers.run.root(args, ["chown", pmb.config.chroot_uid_user,
+    pmb.helpers.run.root(args, ["mkdir", rootfs + "/home"])
+    pmb.helpers.run.root(args, ["cp", "-a", rootfs + "/etc/skel", rootfs + "/home/user"])
```
ollieparanoid: Why would you want the `/etc/skel` file in `/home/user`?

MartijnBraam: /etc/skel is not a file but a directory, it contains a template for new homedirs. In my case in arch it contains `.bashrc`, `.bash_profile` and `.bash_logout` but it can be used to provide default values for new users

ollieparanoid: ah I see, thanks!

PR #691, opened 2017-10-03
Removed a lot of hardcoded modules in mkinitramfs

MartijnBraam:
This removes a few globs from the mkinitramfs module copy part. This is an issue in linux-postmarketos since it will copy gpu drivers for all possible hardware to the initramfs.

This should be done in the deviceinfo variables for the modules instead and is already done for the current devices that use module loading. This changes the initramfs (uncompressed) from 16 MB to 11 MB just because the qemu graphics drivers aren't included.

I've testted the qemu machines and they still work the same as before this patch.


MartijnBraam:
I've tried it on the n900 and the graphics and storage still work fine with this patch, I think the omapfb module is built in into the kernel (its quite small)

PR #695, opened 2017-10-04
Add xfce4, including configuration that is suitable for touchscreen.

pavelmachek:
Nokia N900 does not have working hardware acceleration, so xfce4 with enlarged buttons is easiest way to usable user interface at the moment.

### DIFF
--- a/aports/main/postmarketos-ui-xfce4/APKBUILD
+++ b/aports/main/postmarketos-ui-xfce4/APKBUILD
@@ -1,0 +1,24 @@
+pkgname=postmarketos-ui-xfce4
+pkgver=0.0
+pkgrel=0
+pkgdesc="Meta package for xfce4"
+url="https://github.com/postmarketOS/xfce4-phone"
+arch="noarch"
+license="GPL3"
+depends="xfce4"
+makedepends=""
+install="$pkgname.post-install"
+subpackages=""
+source="start_xfce4.sh $install ${pkgname}-${pkgver}-${pkgrel}.tar.gz::https://github.com/postmarketOS/xfce4-phone/archive/${pkgver}.tar.gz"
+# g/tui/ofone/config.n900/xfce4/xfconf/xfce-perchannel-xml/
+options="!check"
+
+package() {
+	install -D -m644 "$srcdir"/start_xfce4.sh \
+		"$pkgdir"/etc/profile.d/start_xfce4.sh
+	install -d -m755 "$pkgdir"/etc/skel
+	cp -a "${srcdir}/xfce4-phone-${pkgver}"/config "$pkgdir"/etc/skel/.config
+}
+sha512sums="6aa496acf16a5348050ed7aca07203f20f4663b37acd63255c31561e0931f90886050aa3c221db2224e3dc357fe5329dc703ec06f5938e7cea8606ad6b741ee6  start_xfce4.sh
+86b0d3d83da393aafc5d17552457030d9dfae071a35dbe93125e101361962f1752575f17c7dd4140d2a5c6cd4e09367bb2226b532a0194c26c5a77899e80842f  postmarketos-ui-xfce4.post-install
+758562d2820a11b2c96e1805e00450f78203e6fd224115c373825f041b344985490a84b74d2880b423d5606140d5ffb41e79b0a8eb3e5b9a0ce7874c4ccfb666  postmarketos-ui-xfce4-0.0-0.tar.gz"
--- a/aports/main/postmarketos-ui-xfce4/APKBUILD
+++ b/aports/main/postmarketos-ui-xfce4/APKBUILD
@@ -10,7 +10,6 @@ makedepends=""
 install="$pkgname.post-install"
 subpackages=""
 source="start_xfce4.sh $install ${pkgname}-${pkgver}-${pkgrel}.tar.gz::https://github.com/postmarketOS/xfce4-phone/archive/${pkgver}.tar.gz"
-# g/tui/ofone/config.n900/xfce4/xfconf/xfce-perchannel-xml/
 options="!check"
 
 package() {
--- a/aports/main/postmarketos-ui-xfce4/start_xfce4.sh
+++ b/aports/main/postmarketos-ui-xfce4/start_xfce4.sh
@@ -1,0 +1,8 @@
+
+if [ "$(id -u)" = "12345" ] && [ "$(tty)" = "/dev/tty1" ]; then
+	startxfce4 > ~/x11.log 2>&1
+
+	# In case of failure, restart after 1s
+	sleep 1
+	exit
+fi
--- a/aports/main/postmarketos-ui-xfce4/postmarketos-ui-xfce4.post-install
+++ b/aports/main/postmarketos-ui-xfce4/postmarketos-ui-xfce4.post-install
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+. /etc/deviceinfo
+
+# Autologin on tty1, let busybox autoconfigure 2-6
+autologin="user"
+for i in 1 2 3 4 5 6; do
+	old="^tty$i::respawn:/sbin/getty 38400 tty$i"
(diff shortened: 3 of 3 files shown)


aports/main/postmarketos-ui-xfce4/start_xfce4.sh
```diff
@@ -0,0 +1,9 @@
+
+if [ "$(id -u)" = "12345" ] && [ "$(tty)" = "/dev/tty1" ]; then
+
+    startxfce4
+
+    # In case of failure, restart after 1s
+    sleep 1
+    exit
+    fi
```
z3ntu: Indentation is wrong here

aports/main/postmarketos-ui-xfce4/APKBUILD
```diff
@@ -0,0 +1,30 @@
+pkgname=postmarketos-ui-xfce4
+pkgver=0.0
+pkgrel=0
+pkgdesc="Meta package for xfce4"
+url="https://github.com/pavelmachek/xfce4-phone"
+arch="noarch"
+license="GPL3"
+depends="xfce4 xorg-server xf86-video-vesa xf86-input-evdev xf86-input-mouse xf86-input-keyboard udev xinput"
```
ollieparanoid: * Does `xorg-server` not get pulled in by `xfce4` (or one of its dependencies?)
* `udev` is a dependency of `postmarketos-base`, so we don't need it here
* `xf86-video-vesa xf86-input-evdev xf86-input-mouse xf86-input-keyboard` these will get pulled in automatically by `postmarketos-base-x11` (so we have them in one place) once #696 gets merged
* `xinput` is for testing, and not required for running the desktop normally, right?

With all this considered, we would have (once #696 gets merged, I recommend we wait until then): `depends="xfce4"`

aports/main/postmarketos-ui-xfce4/APKBUILD
```diff
@@ -0,0 +1,30 @@
+pkgname=postmarketos-ui-xfce4
+pkgver=0.0
+pkgrel=0
+pkgdesc="Meta package for xfce4"
+url="https://github.com/pavelmachek/xfce4-phone"
```
ollieparanoid: These are just a bunch of config files, so I suggest we put them inside this package instead of hosting them elsewhere. We could also make a subfolder for them (see `postmarketos-base` for an example). What do you think about that?

aports/main/postmarketos-ui-xfce4/APKBUILD
```diff
@@ -0,0 +1,30 @@
...
+makedepends=""
+install="$pkgname.post-install"
+subpackages=""
+source="start_xfce4.sh $install ${pkgname}-${pkgver}-${pkgrel}.tar.gz::https://github.com/pavelmachek/xfce4-phone/archive/${pkgver}.tar.gz"
+# g/tui/ofone/config.n900/xfce4/xfconf/xfce-perchannel-xml/
+options="!check"
+
+package() {
+	install -D -m644 "$srcdir"/start_xfce4.sh \
+		"$pkgdir"/etc/profile.d/start_xfce4.sh || return 1
```
ollieparanoid: The `|| return 1` is obsolete.

aports/main/postmarketos-ui-xfce4/APKBUILD
```diff
@@ -0,0 +1,30 @@
...
+	USER="$pkgdir"/home/user2
```
ollieparanoid: I don't understand why you write something to `/home/user2` at all. Shouldn't these config files go to `/etc/skel`?
You could also put them in `/usr/share/...` and put them in the current user's home folder (in case they don't exist yet) inside your `xinitrc` script, just before starting `xfce4`. This is done in the `postmarketos-ui-hildon` `xinitrc` script.

aports/main/postmarketos-ui-xfce4/APKBUILD
```diff
@@ -0,0 +1,30 @@
...
+	ASUSER="-D -m644 -o user"
```
ollieparanoid: This variable isn't used anywhere else, delete it?

aports/main/postmarketos-ui-xfce4/APKBUILD
```diff
@@ -0,0 +1,30 @@
...
+	install -d -m755 "$USER"
+	install -d -m755 "$pkgdir"/etc/skel
+	cp -a "${srcdir}/xfce4-phone-${pkgver}"/config "${USER}"/.config
+	cp -a "${srcdir}/xfce4-phone-${pkgver}"/config "${USER}"/config-xfce4-phone
+	#chown -R user $USER/.config
```
ollieparanoid: Please remove commented out code.

aports/main/postmarketos-ui-xfce4/APKBUILD
pavelmachek: Sorry about that, I forgot to clean up debugging code.

pavelmachek:
Ok, all but one comment. I'd like to keep separate repository for the files. It allows me to keep APKBUILD simple, may allow other projects to use it easily, and provides a place to put code such as top panel.

ollieparanoid:
Thanks for making the changes!

> may allow other projects to use it easily

That is a good point. The package is called `postmarketos-ui-xfce4` though, so I recommend that we put the `xfce4-phone` repository to the postmarketOS team on GitHub, would that make sense?

As always, we would set it up so you have all necessary rights (e.g. to write to the master branch).

> provides a place to put code such as top panel.

It might make sense to put that in an extra package if this becomes something bigger, and possibly upstreaming it to xfce.

ollieparanoid:
As discussed in the chat, the repo is now here (you should have write access once you click the invite links, I think you get them per mail):
https://github.com/postmarketOS/xfce4-phone

It would be nice if you could review and test #696, so we can get both merged ASAP.

drebrez:
```
/usr/bin/startxfce4: Starting X server
/etc/X11/xinit/xserverrc: exec: line 2: /usr/bin/X: not found
```

drebrez:
@pavelmachek I suggest adding the `xfce4-battery-plugin` package as a dependency, it recognize successfully the battery automatically and shows the percentage 👍

pavelmachek:
Well, problem is that on N900 it displays _three_ batteries (when we
only have one), and displays it in place that will end up hidden as
soon as user window is visible.

Yes, we'll want to display battery status, and thanks for pointer, but
I'm not adding it at the moment.

pavelmachek:
It seems to work on top of x11-install-if branch; moved repository as discussed on chat.

aports/main/postmarketos-ui-xfce4/start_xfce4.sh
```diff
@@ -0,0 +1,8 @@
+
+if [ "$(id -u)" = "12345" ] && [ "$(tty)" = "/dev/tty1" ]; then
+	startxfce4
```
drebrez: @pavelmachek can you add a redirect of the output to a log file? like it's done for other UIs (see [example](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/main/postmarketos-ui-weston/start_weston.sh#L30))

pavelmachek:
I don't think logging to a file is good idea.

X already has its own log, and tells user on console what is going on
and where to find that log.

ollieparanoid:
@drebrez: Awesome photo!

@pavelmachek: At least on Android phones, it is usually very hard to see the console, but reading the log can really be useful, as it contains messages that are not in X11's log ([example](https://github.com/postmarketOS/pmbootstrap/pull/649#issuecomment-333379569)).

I have redirected the logging to `~/x11.log` in `postmarketos-ui-hildon` now, how about we do the same here? It would be consistent then, and this should only be a workaround until we can properly redirect the logging to `/var/log` with a display manager anyway.

Other than that and rebasing this PR, I'd say let's ship it! 🚢

pavelmachek:
Okay, here you go. This should address review comments, logging now goes to a file.

PR #696, opened 2017-10-04
Automatically install X11 and mesa dependencies (using install_if)

ollieparanoid:
This creates a new `postmarketos-base-x11`, that gets automatically installed when `postmarketos-base` and `xorg-server` are installed. It depends on the `xf86` packages, which we need for (almost) every device, so we don't need to specify them over and over again in the `postmarketos-ui-*` packages (e.g. hildon #649 and xfce4 #695).

For `device-qemu-amd64`, it adds a `x11` and `mesa` subpackage, which pull in the appropriate packages once `xorg-server` / `mesa` get installed.


opendata26:
You should probably add xf86-input-libinput and libinput

pavelmachek:
Seems to work for me.

aports/main/postmarketos-base/APKBUILD
```diff
@@ -51,6 +51,18 @@ package() {
...
 }
+
+x11() {
+	install_if="$pkgname xorg-server"
+	depends="
+		xf86-input-evdev
+		xf86-input-keyboard
+		xf86-input-mouse
+		xf86-video-fbdev
+		xf86-video-vesa"
```
drebrez: Just found [this ](https://wiki.archlinux.org/index.php/xorg) which says:

> Xorg searches for installed drivers automatically:
If it cannot find the specific driver installed for the hardware (listed below), it first searches for fbdev (xf86-video-fbdev).
If that is not found, it searches for vesa (xf86-video-vesa), the generic driver, which handles a large number of chipsets but does not include any 2D or 3D acceleration.
If vesa is not found, Xorg will fall back to kernel mode setting, which includes GLAMOR acceleration (see modesetting(4)).

Does this means that `xf86-video-vesa` will never be used because it will always find the `xf86-video-fbdev`?

ollieparanoid: I guess there could be devices, which need `vesa` and for which `fbdev` does not work. But then we could add it in that device package. Thanks for pointing this out, I'll remove it from this list!

ollieparanoid:
@drebrez, @opendata26: thanks for reviewing!
@pavelmachek: thanks for testing!

aports/main/postmarketos-base/APKBUILD
drebrez: @ollieparanoid I don't know how Xorg works, maybe it falls back to `vesa` also if it can find `fbdev` but it doesn't work. The `xf86-video-vesa` package is only 27 kB according to the [alpine package page](https://pkgs.alpinelinux.org/package/edge/main/armhf/xf86-video-vesa), so shouldn't be a problem to have it there.

ollieparanoid: You're right with the small file size. But in general I would rather avoid carrying around packages we don't need, even if they are very small. Would it be OK for you if we wait until we find a device, that needs `vesa`, and add it to the device package then?

Sorry for closing this PR so quickly, I did this because two other PRs were depending on it (xfce4 and Hildon) and because I thought that removing it would be the obvious choice here. But I can give it more time in a similar situation in the future, if you prefer that.

drebrez: @ollieparanoid no problem, I also prefer to have minimal packages, I thought you wanted to have it, since you added it at the beginning.

PR #698, opened 2017-10-04
Fix #681: gnome-menus2 didn't compile for aarch64

### LINKED ISSUE #681: gnome-menu2 doesn't build on aarch64
Reported by rn3_pmOS in IRC:
> i'm trying to build hildon for aarch64 device. gnome-menus2 fails. 
> i get "./config.guess: unable to guess system type" 
> tried copying latest config.guess to build root but, still the timestamp won't change 
> https://paste.ubuntu.com/25653428/ 

The solution is probably using a similar prepare function like for chocolate-doom in #664.

ollieparanoid:
### How to test
`pmbootstrap build --arch=aarch64 gnome-menus2`

It should run through. Without this PR it does not.

### DIFF
--- a/aports/maemo/gnome-menus2/APKBUILD
+++ b/aports/maemo/gnome-menus2/APKBUILD
@@ -1,6 +1,6 @@
 pkgname=gnome-menus2
 pkgver=3.0.1
-pkgrel=2
+pkgrel=3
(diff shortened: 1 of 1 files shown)


ollieparanoid:
Thanks for testing!

PR #702, opened 2017-10-05
Provide default mixer configuration, so sound works by default.

pavelmachek:
Configuring mixers is not easy on system as complex as Nokia N900. Provide a default config file, so at least playback through loudspeaker works.


ollieparanoid:
Thanks for making this! I have read about speakers getting destroyed on the N900 easily, has this been fixed in the kernel or does this config mute the speakers by default?

MartijnBraam:
The n900 needs a highpass filter in the audio pipeline to protect the speakers since it doesn't have analog filtering for that. I'm not sure alsa has something for that, in maemo its handled by a patched pulseaudio.

pavelmachek:
Not sure about "easily". No, I'm not handling that problem.I'm not even sure I believe such problem exists :-). Link would be useful. This simply sets default configuration for N900 speakers.

As you need to be root to use alsaplay, anyway, this does not add any new problem.

ollieparanoid:
I just want to make sure, that no one destroys these speakers by accident with postmarketOS. Here is [one source](http://wiki.maemo.org/Gentoo/N900):
> This basically "just works", but be careful when unmuting - it's possible to play audio that destroys the internal speakers! If you're experienced in writing highpass filters, please contact [...]

So is it muted by default?

pavelmachek:
No, its not muted. If I make it muted by default, it will be a)
useless, and b) first thing anyone does is unmute it anyway.

Unfortunately, gentoo pointer is useless. Nokia N900 has two
"loudspeakers" and speaker for a call. Actually, I searched a bit
more, and came up with:
http://talk.maemo.org/showthread.php?t=58993&page=2 . More FUD.

We could assume it is the loudspeaker and set the output to 80% by
default. Maybe. I'll set output to 100% on mine, anyway. I don't
believe the FUD. Phone without audio is useless. Phone that can't ring
at useful level is useless, too.

ollieparanoid:
80% by default sounds good to me until we know more. Could you do that and increase the pkgrel?

MartijnBraam:
Looks like the n900 _does_ have some highpass filtering in the audio hardware.

screenshot

I don't have specs on the codec or amplifier in the schematic so I don't know at what frequency its filtering but the hardware probably has enough protection from bass frequencies on the speakers.

pavelmachek:
I updated the version and mixer is now 80%. Should be ready.

NikkSaan:
I have been using my N900 as a DAP quite successfully for the past four months. So, I'm uploading my configurations in case someone finds them useful, or wants to include them in the main package. I also include a script that switches between speakers and jack output, and is used to control the volume of the designated mixer - "PCM" for speakers, "TPA6130A2 Headphone" for headphone and Radio Tx.

I'm interested in hearing comments about whether or not I should set the headphone amp to a fixed value, and use PCM as master control for both outputs. Please let me know!

Note that these settings affect only the audio out thru speakers and jack, everything else is muted, and that these are prior to audio breaking #1404 in last update! Also, before you apply any of the settings, I recommend creating a diff between the included _asound.state.muted_ and _asound.state.<headset, speakers>_ to see the changes being made in order to enable the output.

[n900-audio-package.zip](https://github.com/postmarketOS/pmbootstrap/files/1905557/n900-audio-package.zip)

PR #705, opened 2017-10-05
[ofono] update to 1.21

craftyguy:
This release (1.21) includes 3 patches upstreamed earlier. The 4th patch didn't make the cutoff, but has just been merged so it'll be in 1.22


ollieparanoid:
Awesome, thanks for upstreaming and updating \o/

PR #706, opened 2017-10-05
[postmarketos-base] enable udev services to start at boot 

### LINKED ISSUE #700: udevadm trigger is needed before ofonod works
Current init scripts do not do udevadm trigger, so ofonod starts and immediately exits. Running udevadm trigger fixes that.

```
localhost:~$ sudo udevadm trigger
localhost:~$ ps aux | grep ofonod
 1639 user       0:00 grep ofonod
localhost:~$ sudo ofonod
localhost:~$ ps aux | grep ofonod
 1642 root       0:00 ofonod
 1644 user       0:00 grep ofonod
localhost:~$ 

```

craftyguy:
Fixes #700


craftyguy:
@pavelmachek 

Please test that this change fixes ofono for you on the n900

ollieparanoid:
Thanks for the PR! I'm pretty sure, that this fixes it. We should fix this for all devices though, so either in `postmarketos-base` or upstream in the `setup-udev` script. I'll ask upstream whether `setup-udev` should enable the services or not.

craftyguy:
@ollieparanoid 
Oh my. Yea that's no good.

I agree, I submitted the PR thinking that this was a N900-specific quirk and not that the setup-udev was a noop! I'll re-adjust the PR to start these udev services on boot and add that to postmarketos-base's post-install script!

craftyguy:
@ollieparanoid 

I have added `udev` and `udev-postmount` to sysvinit runlevel, and `udev-trigger` to default runlevel to give things (like the `hwdrivers` service) time to load all modules, etc before the `trigger` is executed.

ollieparanoid:
Thanks for making this! The thing is, we need to call `setup-udev` anyway, it is not entirely a NOP as of now. It also changes the inittab, which is required to get X11 starting (or at least Hildon).

So I've asked in upstream, and `setup-udev` *should* enable the services:
```
2017-10-05 20:53:56 <ollieparanoid[m]> hi guys, I have a question about udev-setup: https://github.com/alpinelinux/aports/blob/master/main/eudev/setup-udev
2017-10-05 20:54:42 <ollieparanoid[m]> 1. I'm pretty sure the if [ -z "$dryrun" ] block should be outside of the loop, I'll make a patch for that soon unless I'm confusing this
2017-10-05 20:54:54 <ollieparanoid[m]> 2. should this script enable the services, or only start them once?
2017-10-05 20:55:25 <ollieparanoid[m]> because when you run them once, you have udev set up for the current session, until you reboot. but as I understand, the setup-* script should make persistent changes, right?
2017-10-05 21:09:09 <kaniini> i believe it does both
2017-10-05 21:09:12 <kaniini> if it does not, its a bug
```

I suggest that we keep the `setup-udev` call for now (together with a comment that this is a workaround, and a link to this thread would be nice!), then we merge this, and then I make a PR to fix this upstream. Once it is fixed upstream, we can remove the manual service starting from the postinstall script again.

craftyguy:
This PR seems largely unnecessary given recent comentary (not reflected here) by @ollieparanoid , but I still see ofonod starting successfully on boot with this PR. I also saw it starting successfully on boot before this PR.. 

@pavelmachek please test this and report back. If it doesn't work, then we should start discussing any differences in your config vs mine!

craftyguy:
@pavelmachek any chance you can test this to see if it resolves the udev issue you experienced?

ollieparanoid:
@craftyguy: I recommend rebasing this one, because we've changed a lot in-between. Also if it fixes the issue for you, I'd consider it to be good enough to be merged, the code looks good. (Actual testing by other people with the same hardware is of course preferred.)

craftyguy:
@ollieparanoid well, I never had the problem @pavelmachek reported.. udev has always started on boot correctly for me. So I'm not sure my testing will help if he had some differences in configuration.

ollieparanoid:
Oh, I see, then we really need him to test it.

PureTryOut:
@pavelmachek did you manage to test this yet?

pavelmachek:
I'm sorry, I'm busy with other stuff, and likely will not be able to
test this.

ollieparanoid:
@pavelmachek: actually you wrote in https://github.com/postmarketOS/pmbootstrap/issues/700#issuecomment-343576758 that you already tested it:
> Thanks a lot. I updated to 03f3833...I re-ran install and tested, but no change; still needed to do udevadm trigger before ofonod would start.

So unfortunately it does not seem to be fixing the issue. Thanks for rebasing though @PureTryOut (and for going through all the open PRs, that's very helpful!).

PR #707, opened 2017-10-05
Run weston with user instead of root

drebrez:
I tried to set the `weston-launch` suid bit directly in APKBUILD like suggested but it's not allowed ([reference](https://github.com/alpinelinux/abuild/blob/master/abuild.in#L689-L697))

Closed: #199


drebrez:
@ollieparanoid oh nice, now looking at the reference link I added there is `if ! options_has "suid"; the`, how can I have missed that, lol.
Now I can add the `chmod +s /usr/bin/weston-launch` directly in the APKBUILD package function of weston.
Should I also move the group creation and user assignation somewhere else?

ollieparanoid:
> Should I also move the group creation and user assignation somewhere else?

I think it is good for now, in the long run it would make sense to put this in the user creation in the Python code (see #709).

drebrez:
@ollieparanoid changes done, rebased on master, conflicts solved, re-tested on device, everything should be fine now.

PureTryOut:
Ouch, setting suid while it's not allowed. Not sure if I'm a fan of that. Are we sure this can't be handled by some Polkit rules or something instead?

drebrez:
@PureTryOut I followed this guide in order to run weston as user => https://wayland.freedesktop.org/building.html

PureTryOut:
So you added the user to the `weston-launch` group? If so, why is `suid` still in `$options` in the Weston APKBUILD?

ollieparanoid:
@PureTryOut: `weston-launch` first gets executed as `root` through the `suid` bit. It sets up everything, and [drops rights afterwards](https://github.com/wayland-project/weston/blob/be1090b5234a583ce82edfc3f562066717c41d73/libweston/weston-launch.c#L625-L634). That is pretty standard procedure AFAIK, and not that bad from a security perspective (considering that there are no security holes in the short code, that gets used to set up everything, where root rights are needed).

Polkit is packaged for Alpine Linux, so it might be possible to use it in someway to avoid the `suid` binary. However, Weston does not seem to have support to do it via `polkit`. From [building weston](https://wayland.freedesktop.org/building.html):
> Otherwise (outside of X) it will run on the KMS/DRM framebuffer and take input from evdev devices. Use weston-launch, which as a setuid-root program does privileged operations on Weston's behalf. It also requires that you either enable systemd session support for weston-launch (by using systemd and having the systemd-login devel headers at configure time), or add yourself to the "weston-launch" group:

In my opinion, this PR is a big security improvement with no downsides, and that's why I'll merge it. We could still do further research on better privilege separation for Weston (through Polkit or other programs, maybe even SELinux), but that would be work for another PR.

If someone thinks, that this should have been discussed more before merging, please let me know and I'll do better in the future.

PR #713, opened 2017-10-06
X11 screen rotation config for i9070

ollieparanoid:
@drebrez, I simply took your config from https://github.com/postmarketOS/pmbootstrap/pull/649#issuecomment-333320142 and put it with `install_if` to `device-samsung-i9070`.

I am not sure if you need to have the screen rotated only for Hildon, or for all X11 desktops. So I'm not sure if this is the right way to handle it. What do you think?

Also feel free to put any commits on top of this branch, if you have a better solution 👍

### DIFF
--- a/aports/device/device-samsung-i9070/APKBUILD
+++ b/aports/device/device-samsung-i9070/APKBUILD
@@ -1,6 +1,6 @@
 pkgname=device-samsung-i9070
 pkgver=1
-pkgrel=12
+pkgrel=13
 pkgdesc="Samsung Galaxy S Advance"
 url="https://github.com/postmarketOS"
 arch="noarch"
@@ -8,12 +8,13 @@ license="MIT"
 depends="linux-samsung-i9070 firmware-samsung-i9070"
 makedepends=""
 install=""
-subpackages=""
+subpackages="$pkgname-x11"
 source="
 	deviceinfo
 	initfs-hook.sh
 	90-android-touch-dev.rules
 	modules-load.conf
+	x11-fbdev-rotate.conf
 "
 options="!check"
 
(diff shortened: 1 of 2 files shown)


drebrez:
@ollieparanoid for Hildon was necessary otherwise half screen was black. With xfce4 works pretty good in portrait mode, I tried rotating it with this same config file, but the touchscreen wasn't re-calibrated automatically.

montvid:
Yes screen rotation and touchscreen are not rotated together. For screen rotation I use xrandr and calibration libinput settings..

ollieparanoid:
Okay, that sounds logical. How about we abandon this PR then, and point people who want to rotate the screen for now for Hildon to [the wiki](https://wiki.postmarketos.org/wiki/Hildon)? I've seen that @drebrez documented it there already 👍

montvid:
Yes, screen rotation is only needed for hildon. Mate desktop also runs fine without it.

drebrez:
@ollieparanoid don't close this PR, I will try to add a configuration to calibrate the touchscreen correctly

montvid:
Weston and X11 Libinput uses the udev values in the file to calibrate the ts. I use this for X11 and it works fine:
https://github.com/postmarketOS/pmbootstrap/blob/master/aports/device/device-asus-flo/90-android-touch-dev.rules

@drebrez try this - works for me. Would be good to atomate screen rotation for Hildon as this is a probem on all devices.

drebrez:
@montvid thanks, I will try that one.

@ollieparanoid should we consider to add a config to rotate x11 and touchscreen for all devices with an install_if="hildon"?

ollieparanoid:
So as I understand, we need to both rotate the screen (Xrandr or xorg config), as discussed in https://github.com/postmarketOS/pmbootstrap/issues/749#issuecomment-336411321 and to rotate the touchscreen input matrix with udev.

I guess my preferred solution would be adjusting the udev config on the fly in the `postmarketos-ui-hildon` xinitrc, if that is possible, so we don't need to know the touch input name in advance - just like it is possible with xrandr.

Does anyone know if this is possible?

montvid:
xrandr -o left
xinput set-prop 'Your touchscreen name' 'Coordinate Transformation Matrix' 0 -1 1 1 0 0 0 0 1
exec hildon-desktop

There should be something in libinput too. Have to check.

montvid:
Ok libinput can't change settings on the fly we need to use xinput. I propose adding xrandr and xinput apks to hildon and adding lines to sh:
xrandr -o right
xinput set-prop '*Your touchscreen name' 'Coordinate Transformation Matrix' 0 1 0 -1 0 1 0 0 1
*if someone knows how to parse the name of the touch device please share.

ollieparanoid:
Without having checked it, I guess we can parse it from `xinput`'s output?

drebrez:
From [Hildon page](https://wiki.postmarketos.org/wiki/Hildon#Running) in wiki: `hildon-desktop can be run in a horizontal view only.`
I think we can abandon this PR and find a way to rotate screen/touchscreen coordinates directly in `postmarketos-ui-hildon` package, instead of doing it for every device.
Xfce4 works great in portrait mode and shouldn't be enforced to landscape IMO.

montvid:
@drebez, yes we are discussing this in this pr too. You can join - so far noone has found an answer how to automate that in hildon i.e. how to rotate the screen and touchscreen at once.

ollieparanoid:
Yeah, this PR doesn't really work. Let's move the discussion to #819, I've quoted @montvid there.

Issue #716, opened 2017-10-07
Is ccache broken in distcc compilation mode?


zhuowei:
Ccache doesn't seem to work inside a chroot with distcc enabled.

To reproduce:  Build kwin, weston, or any other large package twice for aarch64.

Expected: Second build is blazingly fast, with no distcc invocations as all the built files should be cached

Actual: Second build same speed as the first build; looking at the distcc log shows that it's still invoked on all the files

[This post](http://blog.kfish.org/2010/06/speeding-up-cross-compiling-with-ccache.html) suggests that, to allow distcc and ccache to work together, one should use the CCACHE_PREFIX environmental variable instead of just adding them together in the PATH, which is what we do right now. I'm not sure if that information is accurate for our ccache and distcc wrapper setups, though.

Edit: no, CCACHE_PREFIX doesn't work; it compiles x86 executables instead for some reason.

Edit 2: https://wiki.gentoo.org/wiki/Distcc#With_automake seems to have a working solution: ccache must go before distcc in the path. This seems to work.

Edit 3: Adding ccache to the front of the path does make ccache cache distcc outputs: before the change, my distccd.log shows that all files are recompiled at each build; after the change, distccd.log only shows files being compiled on the first compile. 

This however doesn't seem to have an effect on compile time: building `weston` took 11 minutes without ccache and 10 minutes with ccache on my computer. I suspect the effect would be stronger for C++ packages such as kwin.

I have not compared with the old code, so I'm not sure if my change actually makes a difference though.

ollieparanoid:
The idea was, to have it working that way:
```
distcc-wrapper (foreign arch) -> distcc -> ccache (native) -> gcc (or cached result) (native)
```

...because `ccache` also needs some CPU time to look up if there is a cached result, which would be about 15 times as slow when done via QEMU.

So having entries in the `distcc` log for every compilation, even if cached, is expected. You could check the `ccache stats` for cache hits (or misses) though, `pmbootstrap stats` opens it.

ollieparanoid:
It could be, that `ccache` always misses for kde stuff, in case a header file contains a timestamp or something, which prevents it from caching. I suggest the following:
* check if `ccache` is used properly (does the cache size increase when compiling?)
* check if `ccache`'s cache has hits when compiling again (probably not?)
* check if there are options for cmake, which make the build reproducible to allow the caching to work

ollieparanoid:
I can reproduce this, but I need to do further investigation of why this happens.

PR #717, opened 2017-10-08
pmbootstrap: enable ccache when using distcc to cross compile

zhuowei:
Based on https://wiki.gentoo.org/wiki/Distcc#With_automake.

Compile time of weston is shortened from 11 minutes on first compile to 10 minutes on second.

This doesn't work for kwin, though - using this configuration cmake only builds with ccache and not distcc. Sigh. I don't known enough about cmake, distcc, and ccache: how would I start fixing this?

Note that I haven't checked if this actually speeds things up relative to the original code, and I'll rather not spend another 30 minutes building Weston, so I would really appreciate if someone can see if this actually has an effect.

This fixes issue #716.


ollieparanoid:
See the discussion in #716.

ollieparanoid:
Obsoleted by #1026 (which got merged!).

Issue #656, opened 2017-09-29
Display manager


ollieparanoid:
Right now we configure the tty0 to automatically log in as `user` (Hildon, plasma?) or `root` (weston) on boot (by editing the `inittab` in the `postmarketos-ui-*` package's postinstall script). Then we start the X11 server (for Hildon) or Wayland compositor (kwin for plasma-mobile, Weston) in an `/etc/profile.d/` script, that checks if the current tty is the one where the autologin happened.

While that works for installing one device, it is not possible to cleanly switch out user interfaces after the installation (because replacing the entry in the `inittab` does probably not work when switching), and it is not possible to install multiple user interfaces in parallel and switch dynamically.

Our requirements for a display manager, as I see it:
* Needs to be able to execute X11 UIs (desktop environments) as well as Wayland compositors (!)
* It should set up the login session (consolekit?) and udev stuff (as @MartijnBraam suggested)
* It should be lightweight, so we don't blow up the image with something that could be done in a few MB (like the `osk-sdl`)
* A config file, that selects the current UI would be nice as first step
* If only one UI is installed, it should boot that.
* Super fancy version with UI (future!)
  * user can select the UI with touch buttons
  * display date, time and battery
  * reboot and power-off buttons

In my opinion, we don't need to allow multiple users for now, or typing in a password (as we focus on security and have full disk encryption anyway, and getting multi-user secured is hard).

PureTryOut:
It would be nice to at least leave an option open to allow multiple users in the future. It might not be useful on phones and smartwatches, but it is on tablets and smart TV's. Although I guess the focus should still be on DE selection at that point. Or we could even make it so that the DE choice is remembered per user.

All future stuff though, let's focus on DE switching first.

MartijnBraam:
The simplest implementation I know of is nodm which is X only and single user/single DE (built for mobile devices), its just a shell script so maybe we can extend that to support what we want.

PureTryOut:
[This one?](https://github.com/spanezz/nodm) Well, it's not just a script, but it seems pretty minimal which is good. Implementing Wayland support should be our first priority then I guess.

drebrez:
What about Qingy?

> qingy is a replacement for getty. Written in C, it uses DirectFB to provide a fast, nice GUI without the overhead of the X Windows System. It allows the user to log in and start the session of his choice (text console, gnome, kde, wmaker, ...).


 http://qingy.sourceforge.net/

[...]

PR #718, opened 2017-10-08
plasma-mobile: fix checksums, start ConsoleKit before kwin, support RGB32 with BGR

zhuowei:
This allows kwin to start on Nexus 6P.

If I apply the QtQuick renderer change at https://wiki.postmarketos.org/wiki/Plasma_Mobile#Running I can get a gray bar at the bottom of the screen, two white rectangles with "W" on them in the middle of the screen and a top gray bar covering the "plasma won't work with OpenGL disabled" screen. Does that match qemu?

@PureTryOut


aports/kde/kwin/APKBUILD
```diff
@@ -17,14 +17,22 @@ makedepends="extra-cmake-modules qt5-qttools-dev kdoctools-dev kdecoration-dev k
 			 sonnet-dev kitemviews-dev solid-dev kdeclarative-dev libdrm-dev eudev-dev
 			 mesa-gbm mesa-libwayland-egl"
 source="https://download.kde.org/stable/plasma/${pkgver}/${pkgname}-${pkgver}.tar.xz
-		D6291.patch"
+		D6291.patch
+		0001-fbdev-support-RGB32-framebuffer-with-BGR-order.patch"
 subpackages="$pkgname-dev $pkgname-doc $pkgname-lang"
-
+options="!strip"
```
zhuowei: I don't think this works? I'm still getting stripped binaries.

aports/main/postmarketos-ui-plasma-mobile/start_plasma.sh
```diff
@@ -15,6 +15,7 @@ if test -z "${XDG_RUNTIME_DIR}"; then
 		udevadm settle
 
 		chown user:user /dev/fb0
+		console-kit-daemon
```
zhuowei: What's the proper place to start ConsoleKit2? I'm guessing this isn't it...

aports/kde/kwin/APKBUILD
ollieparanoid: I guess they don't get stripped by Alpine's `abuild` (which `pmbootstrap` wraps), but by `cmake`. Maybe you can set it to build-type: debug instead of release or something.

aports/main/postmarketos-ui-plasma-mobile/start_plasma.sh
ollieparanoid: You can let OpenRC start it:

https://github.com/postmarketOS/pmbootstrap/blob/4db9a1253e5248a97b385de39387e832c0bd406a/aports/main/postmarketos-base/postmarketos-base.post-install#L16-L18

In `postmarketos-ui-plasma-mobile`'s `.post-install` file would be a good place in my opinion.

zhuowei: I don't think consolekit2 ships with an openrc compatible service.

ollieparanoid:
Awesome work! It would be cool if you could upstream that `kwin` patch.
CC: @bhush9, don't miss this ;)

zhuowei:
@ollieparanoid On chat, @ata2001 mentioned that he also is working on getting kwin to work and has a similar patch, so I probably will wait for that version of the patch instead.

aports/main/postmarketos-ui-plasma-mobile/start_plasma.sh
MartijnBraam: It should be started by the display manager which we don't have yet.

ollieparanoid: Right, my comment is invalid. Where you have put it now is the best we can do until we have the display manager #656 as @MartijnBraam points out.

aports/kde/kwin/APKBUILD
PureTryOut: Yup, you need to add both `!strip` top `$options`, *and* set CMake's release type to Debug.

PureTryOut:
This is amazing work! Thank you! I wonder, isn't the Nexus 6P an aarch64 device? I hadn't yet managed to successfully build the Plasma Framework for aarch64, so I wonder why you did without many problems.

I did realize there were checksum errors, but running `./pmbootstrap.py checksum <package>` locally didn't change any. I'm glad you managed to get them though.

The interesting part is that on x86_64, I didn't have to start the ConsoleKit2 daemon. The log reported a succesful running kwin and ConsoleKit2. I guess it doesn't hurt having it though. Once we have a simple display manager, we can remove it from here.

zhuowei:
@PureTryOut I have to use the abuild-tar patch in #546 - I'll probably package my patch since I'm not sure when QEMU would fix that issue.

aports/kde/kwin/APKBUILD
```diff
@@ -17,14 +17,22 @@ makedepends="extra-cmake-modules qt5-qttools-dev kdoctools-dev kdecoration-dev k
...
 	cd $srcdir/$pkgname-$pkgver
-	patch -p1 < $srcdir/D6291.patch
+	for i in $source; do
+		case ${i%::*} in
+			*.patch)
+				msg "${i%::*}"
+				patch ${patch_args:--p1} -i "$srcdir/${i%::*}" || return 1
+				;;
+		esac
+	done
```
ollieparanoid: you can call `default_prepare` instead of applying that block. In fact, you could move the `mkdir -p $srcdir/build` to the `build()` step, and remove `prepare()` entirely.

PureTryOut:
Well if we're going to have `options=!strip`, then also set CMake to the Debug build (`-DCMAKE_BUILD_TYPE=Debug`). Aside from that, I think this is good to merge.

zhuowei:
@PureTryOut I removed the !strip line instead.

aports/kde/kwin/APKBUILD
ollieparanoid: On second thought, it is not that important to fix this right now. Merging this branch seems to be very beneficial for the plasma mobile work, and it is not merging to `master` anyway. So I'll dismiss my review and merge it.

PR #720, opened 2017-10-08
postmarketos-base-x11: Less dependencies

ollieparanoid:
Thanks to @montvid for [pointing this out](https://github.com/postmarketOS/pmbootstrap/commit/c10dc03eb8f618c3bcec339a41118a304916013c#commitcomment-24832397)!

From [Arch Wiki](https://wiki.archlinux.org/index.php/Libinput#Installation):
> If you wish to use libinput with Xorg, install the `xf86-input-libinput` package, which is "a thin wrapper around libinput and allows for libinput to be used for input devices in X. This driver can be used as as drop-in replacement for evdev and synaptics." **In other words, other packages used for input with X (i.e., those prefixed with `xf86-input-`) can be replaced with this driver.**


montvid:
Any news on this? Xfce, Hildon, Mate run fine without these apk for me.

drebrez:
I will check on my device, I remember I had to install `xf86-input-evdev` at some point when I tested the UI, will see if it's still necessary or can be removed

drebrez:
@ollieparanoid tested it with xfce UI and both, display+touchscreen, works correctly without the packages you removed.

PR #721, opened 2017-10-08
postmarketos-ui-xfce4: depend on xorg-server

ollieparanoid:
I thought `xfce4` depends on `xorg-server` already, so I recommended this wrong. This fixes it.
Thanks to AeternaMia for reporting in the channel!


montvid:
Should there be a dependency on postmarketos-base-x11 too?

ollieparanoid:
That gets pulled in automatically with the `install_if="postmarketos-base xorg-server"` it has.
https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#install_if

ollieparanoid:
Thanks for the review @drebrez!

montvid:
@ollieparanoid Xfce does not load for me without mesa-egl and touchscreen does not work without libinput, xf86-input-libinput. Where should these dependencies be added? I thought everyone needed them.
Touchscreen support:
https://en.wikipedia.org/wiki/Evdev
Linux kernel → libevdev → libinput → xf86-input-libinput → X server → X client

drebrez:
@montvid I think that when the PR #740 will be closed, the correct packages will be automatically installed. I tested yesterday the PR and with a fresh install everything was installed correctly and working

PR #725, opened 2017-10-08
Improve user creation

### LINKED ISSUE #709: Improved "user" creation code
Follow-up to #666 and what has been discussed in the channel.
We would like to have the following changes:
- [x] user used in the build chroots should be called "build", not "user" (to avoid confusion)
- [x] ask in `pmbootstrap init` for the install-user, default to `user`
- [x] during `pmbootstrap install`: the install-user should be created inside the chroot, with a UID of `1000`
- [x] give the install-user the normal set of "desktop" groups 

Keep in mind, that a lot of `--bind` mounts mount stuff to `/home/user`, so we would need to replace the username in there.

PabloCastellano:
Fixes #709

### Important note

After merging or changing branch for testing, you need to run the following commands to set your pmbootstrap folder in a consistent state.
These are the steps I had to run after changing branch from master:

```
pmb init
pmb -y zap -d -hc -p
rm -fr ~/.local/var/pmbootstrap/config_*
```

Or you will find these kind of errors:

```
(010391) [22:31:44] (native) % cd /home/build/packages/build/aarch64 && su build -c 'abuild-sign APKINDEX.tar.gz_'
Error opening key file /home/user/.abuild/user-59da88ae.rsa
139767470046092:error:02FFF002:system library:func(4095):No such file or directory:bio/bss_file.c:255:fopen('/home/user/.abuild/user-59da88ae.rsa', 'r')
139767470046092:error:20FFF002:BIO routines:CRYPTO_internal:system lib:bio/bss_file.c:257:
unable to load key file
```

```
ERROR: /home/build/packages/build/x86_64/postmarketos-base-3-r8.apk: UNTRUSTED signature
```

@ollieparanoid Please remember to add the needed commands to the commit message.


pmb/helpers/user.py
```diff
@@ -0,0 +1,29 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with pmbootstrap.  If not, see <http://www.gnu.org/licenses/>.
+"""
+import pmb.chroot
+
+
+def exists(args, username, suffix="native"):
```
ollieparanoid: That should go to somehwere in `pmb/chroot/` in my opinion. How about mapping it to `pmb.chroot.user_exists()` and putting it in `pmb/chroot/user.py`?

pmb/install/install.py
```diff
@@ -92,29 +93,43 @@ def copy_files_from_chroot(args):
...
+    homedir = rootfs + "/home/" + args.user
     pmb.helpers.run.root(args, ["mkdir", rootfs + "/home"])
-    pmb.helpers.run.root(args, ["cp", "-a", rootfs + "/etc/skel", rootfs + "/home/user"])
-    pmb.helpers.run.root(args, ["chown", "-R", pmb.config.chroot_uid_user,
-                                rootfs + "/home/user"])
+    pmb.helpers.run.root(args, ["cp", "-a", rootfs + "/etc/skel", homedir])
+    pmb.helpers.run.root(args, ["chown", "-R", "1000", homedir])
+
+
+def set_users(args):
```
ollieparanoid: We only support one user right now, so I would call that `set_user()`.

pmb/install/install.py
```diff
@@ -92,29 +93,43 @@ def copy_files_from_chroot(args):
...
+    """
+    Create user with UID 1000 if it doesn't exist
+    Remove 'build' user
+    """
+    suffix = "rootfs_" + args.device
+    if not pmb.helpers.user.exists(args, args.user, suffix):
+        pmb.chroot.root(args, ["adduser", "-D", "-u", "1000", args.user],
+                        suffix)
+        pmb.chroot.root(args, ["addgroup", args.user, "wheel"], suffix)
+    if pmb.helpers.user.exists(args, "build", suffix):
```
ollieparanoid: Why should the build user exist in the device chroot? We never use that chroot for building.

pmb/helpers/user.py
PabloCastellano: Makes sense :+1:

pmb/install/install.py
PabloCastellano: The thing is that the user existed once the image was created. I will remove this part and then we can debug why the user is created there.

PabloCastellano:
The safe upgrade path is a very good idea!

I also found strange the `/home/build/build` and makes sense that we change it. While I find `/home/bootstrap` more descriptive I'm still not sure about it.

What do you think about `/home/postmarketos` or `/home/pmos` ?

kskarthik:
I vote for `/home/pmos`

ollieparanoid:
`/home/pmos` would also be fine with me. it's short 👍 

@PabloCastellano: how do we move forward with this? If you prefer, I could also change the username and rebase it. I'd really not like to carry this longer around than one week because of the huge rebasing that needs to be done otherwise.

PabloCastellano:
@ollieparanoid I also prefer `pmos` so I will choose this option finally. 
Working on it at the moment so hopefully we can have it merged before midnight !

ollieparanoid:
Thanks! Not sure if I can merge it tonight though (need my time to test ;)), but I would merge this one ASAP and before merging any conflicting PRs!

PabloCastellano:
Tests are not passing because of the safe upgrade path. Not even after using `pmbootstrap -y init`.

Any idea?

craftyguy:
I think the build user should be named 'ollie', in honor of its creator!

ollieparanoid:
Did a lot of testing, fixed what was still broken and shipped it! ⛵️

PR #729, opened 2017-10-09
Samsung i9003 device support

dakk:
I started the port of samsung-galaxy-scl (i9003, or S1 SCL). At the moment the kernel (the one used by cyanogen, lineage and other roms) compiles fine (with few patches).

- [x] I'm stuck trying to figure out which are the correct partition for kernel; I tried various partition but got no luck, so any help is welcome.
- [x] The smartphone boots to the "boot partition not found" screen (with pm logo)
- [x] Fix boot partition


MartijnBraam:
My guess is partition 6 (normalboot) but I've not messed with samsung devices before.

dakk:
Ok I found out the problem (was a missing mkbootimg package in the APK of the device); the device boots the kernel, the screen works; I got the "boot partition not found" error, but I'm in the right way: https://imgur.com/a/ONUaE

ollieparanoid:
Nice progress! Have you flashed the system image? The postmarketOS boot partition is actually part of that. See also: [partition-layout](https://wiki.postmarketos.org/wiki/Partition-layout)

dakk:
@ollieparanoid yep, I flashed it in the system partition but it doesn't work. I'll try in the next days using the mmc drive as system partition; hope it works, I really want a linux phone :)

ollieparanoid:
What you're having already sounds pretty promising. Even if it does not work, we should be able to find out why it can't find our partition when debugging the initramfs.

dakk:
Update: using the sdcard, the kernel finds the system partition and start weston correctly. The touch works fine, I updated the calibration parameters.

ollieparanoid:
Awesome! Should we merge it then? It fulfills the requirements already (actually, booting the kernel and having usb networking is enough) to get merged into master, and you can make a new PR for wifi.

dakk:
@ollieparanoid ok, you can merge it

aports/device/linux-samsung-i9003/APKBUILD
```diff
@@ -0,0 +1,136 @@
...
+# - add findutils makedepend (necessary for the initramfs)
+# - it builds an initramfs - do not use ramdisk.cpio and
+#	ramdisk-recovery-device.cpio, in there, always use the
+#	initramfs from the recovery partition ("isorec")
+# - directly boot to that initramfs, not only when the recovery
+#	key combination was pressed.
+# - use Alpine's busybox-static instead of the pre-compiled
+#	binary, that comes with the source
+#
+# Kernel config changes, based on: arch/arm/configs/i9100_defconfig
```
ollieparanoid: Please adjust this line, so it points to the real defconfig that has been used.

aports/device/device-samsung-i9003/APKBUILD
```diff
@@ -0,0 +1,28 @@
+pkgname=device-samsung-i9003
+pkgver=1
+pkgrel=13
+pkgdesc="Samsung Galaxy SI SLC"
+url="https://github.com/postmarketOS"
+arch="noarch"
+license="MIT"
+depends="linux-samsung-i9003 firmware-samsung-i9003 libsamsung-ipc mkbootimg"
```
ollieparanoid: Please remove `firmware-samsung-i9003` for this PR, and also remove the `firmware-samsung-i9003` APKBUILD (because it doesn't work yet and is just an empty package).

dakk:
I think it's in sight with your reviews now

Issue #737, opened 2017-10-10
ANDROID_PARANOID_NETWORK restricts network access to users who are member of specific groups


bhuvi8:
when non root user tries to access network they get permission denied errors

bhuvi8:
did a strace and found that permission denied error occurs when a socket is created using socket() system call

bhuvi8:
on further investigation found that it is an android security feature

verified it by creating groups with ids 3003 and 3004 and assigned user to group and user is now able to access network
# apk add shadow
# groupadd -g 3003 aid_inet
# groupadd -g 3004 aid_inet_raw
# usermod -G aid_inet user
# usermod -G aid_inet_raw user

$ ping 8.8.8.8
now works as user

bhuvi8:
Additional reference https://groups.google.com/forum/#!topic/sluggers/g0XY-Qk5C3w

Most other devices having android kernels have this option set...This needs to be disabled for network features in non android OS'es to work properly

bhuvi8:
created a pull request https://github.com/postmarketOS/pmbootstrap/pull/738 to fix this in samsung-i9100

z3ntu:
Better reference: https://elinux.org/Android_Security#Paranoid_network-ing 😉

bhuvi8:
Modified the title to better reflect this issue

z3ntu:
I would edit it to something like "ANDROID_PARANOID_NETWORK restricts network access to root user"

bhuvi8:
Done

ollieparanoid:
We should add this to our `pmbootstrap kconfig_check` code, so we can fix it for all devices (and warn the user when it is enabled). Thanks for reporting! (If someone wants to take a shot at this, please go ahead 👍 )

z3ntu:
Am doing it right now.

bhuvi8:
@z3ntu nice job fixing it for all devices and nice work with the script :-)

PR #738, opened 2017-10-10
[samsung-i9100] Fix for non root users not able to access network #737

bhuvi8:


PabloCastellano:
Submitting PR to missing bump of pkgrel

Issue #731, opened 2017-10-09
postmarketos-base-x11 not pulled in automatically for armhf


drebrez:
If you zap the device arm chroot and you select `xfc4` as UI, when running `pmbootstrap install` the `postmarketos-base-x11` subpackage doesn't get installed.
The problem is not the `install_if` condition, but that the `postmarketos-base` APKBUILD has `noarch`, and after building the package it gets symlinked in the other arch folders ([here](https://github.com/postmarketOS/pmbootstrap/blob/bed1eacbb5b0a668032fc13d32698d93bc5e8b84/pmb/build/package.py#L127-L129)).

The problem is that the subpackages are not symlinked, so they don't exist in the other repositories.

I've tested manually symlinking the `postmarketos-base-x11` subpackage in the armhf and with `pmbootstrap install` was triggered correctly.

Don't know how to change the `symlink_noarch_package` function to automatically symlink the subpackages.

craftyguy:
Would this be fixed by changing `postmarketos-base` to `arch="all"` instead?

ollieparanoid:
Thanks for your detailed bug report 👍
This should be fixed in #740.

PR #739, opened 2017-10-10
Add support for device-specific weston config (#499)

### LINKED ISSUE #499: Weston config should be loaded on start

craftyguy:
Fixes #499. See this issue for more details


craftyguy:
This currently doesn't work due to #731

montvid:
nano /tmp/weston.log

> Old Xwayland module loading detected: Please use --xwayland command line option or set xwayland=true in the [core] section in weston.ini

Please change from modules=xwayland.so to xwayland=true It would be nice if it was changed in the default weston.ini too.

craftyguy:
There is no longer going to be a 'default weston.ini' file, that's the whole point of this PR and the discussion in #499.

montvid:
Well I don't think it is practical to have a weston.ini for every device as not many people will be using weston anyway as they need a more proper gui.

craftyguy:
Go read the discussion in #499, and add any additional comments you might have in that issue.

drebrez:
@craftyguy closing this PR will break weston in all the devices except N900? 
Or weston may work correctly without the config?

aports/device/device-nokia-rx51/APKBUILD
```diff
@@ -1,14 +1,14 @@
...
+pkgrel=24
 pkgdesc="Nokia N900"
 url="https://github.com/postmarketOS"
 arch="noarch"
 license="MIT"
 depends="linux-postmarketos uboot-tools linux-firmware kbd kbd-bkeymaps ofono"
 makedepends="uboot-tools kbd kbd-bkeymaps"
 install="$pkgname.post-install"
-subpackages=""
+subpackages="$pkgname-weston:weston:noarch"
```
ollieparanoid: The `:noarch` is unnecessary when the parent package has the same architecture.

craftyguy:
Oh shit, what did I volunteer for? 😛 

I would appreciate some help adding this to device packages.. since there are a lot.

ollieparanoid:
Here we go! I have no idea why one Travis check is failing though, I didn't even touch `firmware-huawei-y530` in this PR, and also the checksum are correct in master.

Reviewing this PR and testing it on real devices would be greatly appreciated. I've quickly tested it in `qemu-amd64` and it worked for me.

craftyguy:
wow, nice work @ollieparanoid !

aports/main/postmarketos-ui-weston/weston.ini.default
```diff
@@ -0,0 +1,6 @@
+# In postmarketOS, this file gets used, when no /etc/xdg/weston.ini exists. That 
+# is, so we can install a device specific config file (which may disable 
+# xwayland for example) from the device specific package to /etc/xdg/weston.ini.
+
+[core]
+modules=xwayland.so
```
drebrez:
```
[core]
xwayland=true
```
is not necessary here?

aports/device/device-oneplus-bacon/weston.ini
```diff
@@ -0,0 +1,8 @@
+# XWayland seems to be broken for this device.
```
drebrez: missing changes to the `aports/device/device-oneplus-bacon/APKBUILD`?

aports/main/postmarketos-ui-weston/weston.ini.default
ollieparanoid: Thanks! `modules=xwayland.so` is the obsolete version, `xwayland=true` is the proper one, see https://github.com/postmarketOS/pmbootstrap/pull/739#issuecomment-336144375 -- I'll adjust this.

aports/device/device-oneplus-bacon/weston.ini
ollieparanoid: Good catch, fixed.

ollieparanoid:
@drebrez wrote:
> `/etc/xdg/weston.ini` is not the default location that weston looks, if no `-c ...weston.ini` parameter is specified, it uses `/etc/xdg/weston/weston.ini` 

- [ ] use `/etc/xdg/weston/` for the configs

drebrez:
After a lot of testing and git bisecting, I decided to rebuild all the packages from scratch and finally only by adding `backend=fbdev-backend.so` to the `weston.ini.default` everything started working again

ollieparanoid:
Wow, that you so much for figuring that out! I'll adjust the PR accordingly.

drebrez:
@ollieparanoid i'm doing it right now

ollieparanoid:
Even better, thanks \o/

PR #740, opened 2017-10-10
Fix #731: Create symlinks for noarch-subpackages

### LINKED ISSUE #731: postmarketos-base-x11 not pulled in automatically for armhf

ollieparanoid:
* apkindex:
  * Also parse the architecture field
* symlink_noarch_package:
  * Renamed to symlink_noarch_packages
  * Always work on all packages (so we don't need to guess which
    subpackages have been generated after a certain build)
  * Get invoked when running 'pmbootstrap index'
  * Use 'apk index' to generate one index, where the architecture
    does not get rewritten (abuild does that by default, due to
    Alpine's repos not having a 'noarch' folder and diverging from
    that doesn't make things easier for us). That goes super fast,
    and then we know which packages are noarch packages and can
    create the symlinks.
* Made output less verbose:
  * Use -q for 'apk index' when calling it directly (when it gets
    called by abuild we can't control that)
  * Output that the APKINDEXes get reindexed only to the 'pmbootstrap
    log'.


drebrez:
@ollieparanoid maybe is better if you bump the `postmarketos-base` package version in order to rebuild it and automatically create the new symlinks, otherwise the user have to run `pmbootstrap index` to fix it. What you think?

ollieparanoid:
Great idea, done that. Thanks for testing and for that suggestion!

PR #746, opened 2017-10-11
Delete broken symbolic links during `pmbootstrap zap --mismatch-bins`

drebrez:


drebrez:
@ollieparanoid probably when you keep changing branches around or when you delete old apk versions, I usually run `pmbootstrap zap -m`, and lot's of time it says something that packages are not in aport folder, something like this, don't remember

PR #747, opened 2017-10-11
[android-recovery-zip] Add missing apk keys

drebrez:
I noticed that installing pmos with android-recovery, when running `sudo apk update` doesn't trust the alpine repository:
```
ERROR: https://nl.alpinelinux.org/alpine/edge/main: UNTRUSTED signature
```
With this PR the keys in `/etc/apk/keys` are also added. 

It also compress the rootfs archive with `gzip -f1` which reduces the time by ~30/40 seconds and the resulting zip grows only by 1-2 MB, for me it seems a good trade-off.


montvid:
You could try xz it could be even faster and the file smaller as it supports multithreading. :)

drebrez:
@montvid not sure busybox has `xz` compression, and that would mean to also add additional dependencies into the recovery zip. 
I'm not saying it's not possible, I suggest you create a new Issue with that idea, to keep track of it.

montvid:
Sorry, yes, busybox tar supports only:
Archive format selection:

        z       Filter the archive through gzip
        j       Filter the archive through bzip2
        a       Filter the archive through lzma
        Z       Filter the archive through compress

montvid:
Oh, I found busybox supports lzop so that should be faster than gzip. Anyway I don't know if it fits in this situation.
http://www.lzop.org/

> lzop is a file compressor which is very similar to gzip. lzop uses the LZO data compression library for compression services, and its main advantages over gzip are much higher compression and decompression speed (at the cost of some compression ratio).

drebrez:
@montvid please open a new issue with the `enhancement` label for changing the compression of the rootfs archive inside the android recovery zip.

ollieparanoid:
Code looks good to me, although I did not test it. What do you think, @ata2001?

drebrez:
@ollieparanoid conflict resolved

ata2001:
Could you please explain, why the keys are present in the native chroot, but not in the device chroot?

ollieparanoid:
They get mounted inside the chroot, so they are always the same in all chroots (we add the building key, once it gets generated, and that must be in all chroots afterwards).
In the normal installation, we umount everything, copy the rootfs, and then copy the keys afterwards.

ata2001:
@ollieparanoid Thanks for the explanation!

Looks Ok.

PR #753, opened 2017-10-12
[device-ouya-ouya] Adding device folder, osk-sdl setup

rendeko:
postmarketOS/pmbootstrap version is missing the essential device-ouya-ouya folder. Additionally, an fb.modes file has been added that adds support for osk-sdl when it is implemented. Checksums also updated.

I also deleted all the other devices from the device folder for this branch. Is this proper? I feel it would help avoid keeping outdated versions.

### DIFF
--- a/aports/device/device-ouya-ouya/APKBUILD
+++ b/aports/device/device-ouya-ouya/APKBUILD
@@ -1,0 +1,23 @@
+pkgname=device-ouya-ouya
+pkgver=1
+pkgdesc="Ouya"
+pkgrel=5
+url="https://github.com/postmarketOS"
+arch="noarch"
+license="MIT"
+depends="linux-ouya-ouya mkbootimg"
+makedepends=""
+install=""
+subpackages=""
+source="deviceinfo fb.modes"
+options="!check"
+
+package() {
+	install -D -m644 "$srcdir/deviceinfo" \
+		"$pkgdir/etc/deviceinfo"
+	install -D -m644 "$srcdir"/fb.modes \
+		"$pkgdir/etc/fb.modes"
+}
+
+sha512sums="60c9aa95a15b4c8e48737636eb0f2a2e3a942e0c5ebd88beb3946262fc79b1a5e388cf31e12b253015fc59252234c4082fd63c57a92a1e65fef8eb448ce64e3f  deviceinfo
+bf442fac4dc0594e055ed7a7d8232b5c884e2e77543273e8c4f32b5fe3c199561c86d8a5f665e17406057ca7863fabb93e789fe5e168fcd89d2982e0046232ad  fb.modes"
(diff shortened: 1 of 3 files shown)


ollieparanoid:
> I also deleted all the other devices from the device folder for this branch. Is this proper?

No, please do not do that. Please keep the other devices, only your changes to the `device-ouya-ouya` branch will get merged into master, the other device packages will not get downgraded or anything similar.

rendeko:
My mistake. Fixed commit. Should be good to pull now.

aports/device/device-ouya-ouya/APKBUILD
```diff
@@ -0,0 +1,27 @@
...
+arch="noarch"
+license="MIT"
+depends="linux-ouya-ouya mkbootimg"
+makedepends=""
+install=""
+subpackages=""
+source="deviceinfo fb.modes"
+options="!check"
+
+build() {
```
ollieparanoid: Please remove this empty `build()` function.

rendeko: Removed.

ollieparanoid:
Thanks!

drebrez:
@ollieparanoid the wiki shows this device as supported, but it's not present in master branch. What to do?

ollieparanoid:
Good point! @rendeko if you are around and have time, do you want to make a new PR against master?

(Also how's the ouya going in general? If you have time, it would be nice to have photos to show for the next blog post.)

rendeko:
@ollieparanoid I was avoiding merging into master because the boot method means it can't be installed by pmbootstrap's regular flashing. As for the device itself, my main issue has been getting the wifi/bluetooth modules to work. The controller uses bluetooth, so I wouldn't want to show off the device without that major part working. I'll come by the matrix room later today to ask for assistance with it.

ollieparanoid:
That makes sense, and we faced a similar discussion with other devices where system flashing did not work (https://github.com/postmarketOS/wiki/issues/12). Whether we would call them supported or not.

The TL;DR is, that we don't call the devices supported anymore (because it's hard to draw a line!) and instead we list the devices where booting works. The Ouya falls into that category now.

As it says on the [devices](https://wiki.postmarketos.org/wiki/Devices) page now:
> These devices are included in the postmarketOS source tree, and it is at least possible to boot them. To confirm that, either the screen or networking/serial output must be working. If your device fulfills these requirements and is not yet on the list, please make a pull-request and edit the list after it has been accepted. 

So if you don't mind, I would recommend that you make a new PR against master with your work and move it from the "not booting" table up to the first one.

drebrez:
@ollieparanoid @rendeko nothing have to be changed in the wiki, as I mentioned in my last post, the device it's already listed in the main table

PR #758, opened 2017-10-13
Generating rootfs image is now 110% less scary

### LINKED ISSUE #708: Less scary warnings in ./pmbootstrap.py install
If we are building qemu image, we probably do not need to ask y/n if that's ok. Qemu images are always safe.

If we are doing sdcard install, it makes sense to print less scary warning if the sdcard already contains pmos.

craftyguy:
This adds a check to make sure there is enough disk space for creating the device image.

fixes #708


craftyguy:
@ollieparanoid I took the liberty of adding a check here to make sure there is enough disk space for creating the device disk image before it is created. hopefully this makes it more robust!

When it fails because there is not enough disk space, you see this message during `install`:
> [19:23:01] *** (3/5) PREPARE INSTALL BLOCKDEVICE *** 
> [19:23:01] ERROR: Not enough free space to create rootfs image! (free: 695M, required: 696M)

PR #345, opened 2017-08-07
Fix various build issues (fix #189, and probably #341)

ollieparanoid:
Even if this does not fix #341, it should turn a crash into a warning.

Changes:
* Removed the apkindex_files cache. That particular cache caused
  bug #189 and didn't bring any real-world performance improvements
  (tested 3x with that cache and 3x without, no significant speed
  difference). I decided to remove it instead of keeping it/adding
  even more code to resolve the bug.
* Fix the check for already built packages: always use the architecture,
  that the package should be built for instead of the architecture of
  the build environment (e.g. use armhf, even when building a noarch
  package in the x86_64 chroot). This partially resolves #341.
* Make pmb.chroot.apk.install_is_necessary() more robust: If the binary
  package is missing, although it should be there, print a warning and
  build it with force.

@antonok-edm: It would be awesome if you could try out this PR and confirm whether it fixes #189 for you. In my testing it did.
@PureTryOut: I'd like to know if this fixes #341 for you.

PR #760, opened 2017-10-13
Run fast testcases in Travis CI, remove obsolete testcase

ollieparanoid:
* This adds `test/testcases_fast.sh`, which also gets executed by Travis CI now.
* The ones which require cross-compilers or traffic-intense setup (cloning Alpine's aports repo) are disabled for now.
* The apkindex_files cache has been removed in #345, this commit finally deletes the obsolete testcase, which has been failing since.


ollieparanoid:
Interesting, two fail on Travis, but do not fail locally. I'll need some time to investigate and fix this (help welcome if anyone wants to).

ollieparanoid:
@ata2001 @craftyguy @drebrez @kskarthik @opendata26 @PabloCastellano @pavelmachek @PureTryOut  @MartijnBraam @yuvadm @z3ntu @zhuowei:

It would be nice if you could take a look at this. The following function adds an offset to the folder size, although it should only return the size on disk as I understand it:

https://github.com/postmarketOS/pmbootstrap/blob/5e7596d3c525aa663311208d09606c8981c3604b/pmb/helpers/other.py#L26-L38

I assumed, that the offset is constant with `140` bytes, but that is not the case, as it [fails on Travis](https://travis-ci.org/postmarketOS/pmbootstrap/builds/288754759#L667):
```python
    def test_get_folder_size(args, tmpdir):
        # Write five 2 KB files to tmpdir
        tmpdir = str(tmpdir)
        files = 5
        for i in range(files):
            pmb.helpers.run.user(args, ["dd", "if=/dev/zero", "of=" +
                                        tmpdir + "/" + str(i), "bs=1K",
                                        "count=2", "conv=notrunc"])
    
        # Check if the size is correct
        # 140 bytes seem to get added by the filesystem for the directory
        size = 2048 * files + 140
>       assert pmb.helpers.other.folder_size(args, tmpdir) == size
E       AssertionError: assert 14336 == 10380
```

(This is the only failing testcase in the PR.)

zhuowei:
@ollieparanoid https://www.gnu.org/software/coreutils/manual/html_node/du-invocation.html the --apparent-size option may help here?

ollieparanoid:
Yeah, I also think it makes it better, I've enabled it in the last commit (updated my previous comment to show the updated code snippet). But there are still some bytes getting added to the folder for some reason :\

drebrez:
I did some investigation and found this:
```
$ find <folder> -type f -print0 | du --files0-from=- --summarize --apparent-size --block-size=1
2048    ./111
2048    ./333
2048    ./222
2048    ./444
2048    ./555
10240   total
```
what you think?

ollieparanoid:
@drebrez: Thanks for the idea! I've tried it, and I still get a different result on my PC compared to what you have.... I'm starting to doubt that it is possible to use `du` (which comes from "disk usage" apparently) to display the actual folder size, as if you would put it into an image file (although it says in the manpage that you can do just that with `--apparent-size` - it clearly does not work).

However, I think the roughly estimated size we get from `du` is good enough for our use case. So I've adjusted the testcase to be more tolerant with the output.

(I've commited and pushed a change, but somehow it won't show up here in GitHub...?)

craftyguy:
I haven't actually tried this, but instead of calling du (which may not be the same version in travis vs your system vs @drebrez's system), would something like `os.walk`ing a directory in python, calling `os.path.getsize` on each file in the directory's tree to add up their sizes lead to a deterministic total directory size across environments?

craftyguy:
lol, my suggestion might be what you are trying to change from. Sorry.

ollieparanoid:
We also did that at some point (I think @drebrez implemented it? or I've played with it) and it turned out to be really slow compared to `du`.

ollieparanoid:
Thanks for having a look at it!

PR #762, opened 2017-10-13
Weston: log to ~/weston.log instead of /tmp/weston.log

ollieparanoid:
Placing such files in /tmp/ is an unnecessary attack vector, as
@pavelmachek showed. All udev rules, which mentioned /tmp/weston.log
do not contain any path anymore, because that path is very likely
to change again (and then we would need to change all device packages
again!).


z3ntu:
I wouldn't want unnecessary files in my home directory in the future... Right now I don't really care, as it's just for development.

ollieparanoid:
Yep, we'll move them to `/var/log` or another sane location once we have the display manager #656. This PR makes it consistent with Hildon and XFCE4.

drebrez:
why not using something like `/var/log/postmarketos/`?

In theory we can start something like `syslogd -O /var/log/postmarketos.log` as root, and redirect the output of weston with `weston-launch -- ${WESTON_OPTS} 2>&1 | logger -t weston`

craftyguy:
@drebrez there is already a syslog in pmOS/Alpine, and we already start it on boot!

I don't know what security-related thing @pavelmachek found with withing to /tmp, but using the existing syslogd might not be a bad idea. The logger tag should include something like "username:application", this way we are ready for multi-user support!

ollieparanoid:
Excellent solution @drebrez and @craftyguy! I'll refactor this to use syslogd instead.

ollieparanoid:
Ready for review/testing! The logs can be read with `logread -f /var/log/messages | grep weston` now.

PR #763, opened 2017-10-13
pmbootstrap init: Show pkgdesc for each postmarketos-ui package

ollieparanoid:
This way we could give the user a rough idea what will be installed, and also use this to display a short warning about long compile times (e.g. until the plasma mobile stuff is upstreamed).

With this PR, it looks like:
```
[01:14:35] Available user interfaces (3):
[01:14:35] * none: Do everything via SSH (like a server PC)
[01:14:35] * hildon: (X11) Lightweight GTK+2 UI optimized for single-touch touchscreens
[01:14:35] * weston: (Wayland) Reference compositor (demo, not a phone interface)
[01:14:35] * xfce4: (X11) Lightweight GTK+3 desktop (stylus recommended)
[01:14:35] User interface [weston]:
```


pmb/helpers/ui.py
```diff
@@ -18,16 +18,18 @@
...
     """
-    Get all UIs, for which aports are available
-    :returns: ["postmarketos-ui-one", "postmarketos-ui-two", ..., "none"]
+    Get all UIs, for which aports are available with their description.
+
+    :returns: {"none": "Do everything...", "weston": "Wayland reference ..."}
     """
-    ret = []
-    for path in glob.glob(args.aports + "/main/postmarketos-ui-*"):
+    ret = {"none": "Do everything via SSH (like a server PC)"}
```
MartijnBraam: The description for `none` doesn't apply to everything, for devices with hardware keyboards or usb ports and a working VT you don't need to ssh, you can just use text mode.

ollieparanoid: Right... any chance you have a better description?

MartijnBraam: Hmm maybe something like "No graphical environment (use only when you have network access or a physical keyboard)" but that might be too long.

drebrez: `No graphical environment` is enough IMO, between parenthesis there is a description for the other UIs, this doesn't need a description. And `(use only when you have network access or a physical keyboard)` sounds more like a suggestion instead of a description

aports/main/postmarketos-ui-hildon/APKBUILD
```diff
@@ -1,7 +1,7 @@
 pkgname=postmarketos-ui-hildon
 pkgver=1
-pkgrel=1
-pkgdesc="Lightweight desktop, optimized for single-touch touchscreen devices"
+pkgrel=2
+pkgdesc="(X11) Lightweight GTK+2 UI optimized for single-touch touchscreens"
```
drebrez: maybe you can add parenthesis around `optimized for single-touch touchscreens`, more similar to the xfce description

pavelmachek:
ret = {"none": "No graphical user interface (use SSH or console)"}

?

ollieparanoid:
Thanks for the helpful feedback! This is how it looks now. Do you think it is ready for merge?
```
Available user interfaces (3):
* none: No graphical environment
* hildon: (X11) Lightweight GTK+2 UI (optimized for single-touch touchscreens)
* weston: (Wayland) Reference compositor (demo, not a phone interface)
* xfce4: (X11) Lightweight GTK+3 desktop (stylus recommended)
User interface [weston]:
```

PR #766, opened 2017-10-14
Check block dev for previous pmOS install

### LINKED ISSUE #708: Less scary warnings in ./pmbootstrap.py install

craftyguy:
fixes #708

The method of 'install detection' used here is to look for partitions with pmOS_ in the partition label. It's not a guarantee, but it works when FDE is enabled without having to unlock the partition to read files within it.

Here's what you see during installation if you try to install to a sd card with a previous installation of pmOS:
>[16:43:59] *** (3/5) PREPARE INSTALL BLOCKDEVICE ***
>[16:43:59] WARNING: This device may have a previous installation of pmOS. CONTINUE? (y/n) [n]: y

Note: in the implementation, we loop through the glob of sdcard partitions in a new loop since we only want to prompt once. The previous loop looking for mounted partitions would result in multiple prompts since we wouldn't want to break out of that loop (we would want to check every partition to see if it is mounted)

Note: this depends on 'blkid', which is on every Linux system I've ever come across, AFAIK, so that shouldn't be an issue.. right?


pmb/install/blockdevice.py
```diff
@@ -25,6 +25,17 @@
...
 
+def previous_install(args, blk_dev):
+    """
+    Search a block device for possible existence of a previous installation of pmOS
+
+    :param blk_dev: path to block device
+    """
+    label = pmb.helpers.run.user(args, ["blkid", "-s", "LABEL", "-o", "value",
+                                 blk_dev], return_stdout=True)
+    return "pmOS_" in label
```
ollieparanoid: Is there any reason for not checking for `pmOS_boot` directly?

pmb/install/blockdevice.py
```diff
@@ -37,6 +48,13 @@ def mount_sdcard(args):
         if pmb.helpers.mount.ismount(path):
             raise RuntimeError(path + " is mounted! We will not attempt"
                                " to format this!")
+    for path in glob.glob(args.sdcard + "*"):
+        if previous_install(args, path):
+            if pmb.helpers.cli.confirm(args, "WARNING: This device may have"
```
ollieparanoid: If we have such a `pmOS_` or even `pmOS_boot` partition, I'd say it's pretty sure that this is a postmarketOS installation. So I would reword that to: "This device has ...".

pmb/install/blockdevice.py
```diff
@@ -37,6 +48,13 @@ def mount_sdcard(args):
         if pmb.helpers.mount.ismount(path):
             raise RuntimeError(path + " is mounted! We will not attempt"
                                " to format this!")
+    for path in glob.glob(args.sdcard + "*"):
```
ollieparanoid: How about only checking the boot partition?
```python
for partition in ["1", "p1"]:
    path = args.sdcard + partition
    if not os.path.exists(path):
        continue
    # ...
```
(This handles `/dev/sdx1` and `/dev/mmcblkxp1`.)

craftyguy:
requested modifications made. 

I also removed the 'scary' message, but my only concern would be that this removes any chance that someone has to abort an install if they specified the wrong device (assuming they don't realize it ahead of this step and ctrl+c out).

If a pmOS install is detected (by looking for 'pmOS_boot' label), this is what the output looks like:
```
[18:49:44] *** (3/5) PREPARE INSTALL BLOCKDEVICE ***
[18:49:45] WARNING: This device has a previous installation of pmOS. CONTINUE? (y/n) [n]: y
[18:49:45] (native) mount /dev/install (host: /dev/sdc)
[18:49:45] (native) partition /dev/install (boot: 42M, root: the rest)
```

pavelmachek:
This is opposite of what I had in mind.

I propose:

Note: This device has previous installation of pmOS. Overwrite it? (y/n) [y]

(because updating pmOS to new version is pretty normal).

and

WARNING: About to install pmOS on new device. Everything on the device
will be completely lost. ERASE EVERYTHING? (y/n) [n]

(becuase typo between sdd and sde is easy to make and results would
not be nice).

Another possible check would be.. if device is bigger than 32GiB or
so (because uSD cards we want to overwrite are usually small, and
harddrives we really don't want to overwrite are usually big).

WARNING: About to erase big block device. Are you sure? (y/n) [n]


-- 

craftyguy:
So not the opposite, just what I had originally committed: 1) warn if pmos is detected and 2) warn before erasing disk.

ollieparanoid:
Sorry, I guess what I wrote was confusing. What I meant:

* When the target devices *does not* contain postmarketOS, ask for confirmation like we are already doing (`"EVERYTHING ON " + args.sdcard + " WILL BE ERASED! CONTINUE?"`)
* When the target device *does* contain postmarketOS, also ask for confirmation **once**, but replace the message with something like: "This device has a previous installation of pmOS. CONTINUE WITH OVERWRITE?"

I hope this makes sense now and that this also is what @pavel means (the crucial part being, that we *only ask once* in both scenarios).

> Note: this depends on 'blkid', which is on every Linux system I've ever come across, AFAIK, so that shouldn't be an issue.. right?

For Alpine it is provided by busybox, which does not seem to support the options you are using. That means, `pmbootstrap` would depend on `util-linux` from now on (I think right now we do not, but I am not 100% sure). Is it possible to avoid that dependency?

Thanks for picking this issue in the first place and working on it. Let me know if I can help with anything.

craftyguy:
> For Alpine it is provided by busybox, which does not seem to support the options you are using. That means, pmbootstrap would depend on util-linux from now on (I think right now we do not, but I am not 100% sure). Is it possible to avoid that dependency?

I don't know of any other way to read partition label from a disk (unless of course we resort to tearing apart the MBR using only tools in python)... Do you have any ideas? I'll continue to investigate!

craftyguy:
@ollieparanoid 

Ha, so apparently [there's a way using python](https://github.com/hamptus/pyMBR/blob/master/mbr.py) (external library someone wrote), but I *really* don't want to go there...

Another idea: can we do this check somehow from within the chroot, where we have blkid provided by busybox? It *seems* like I could move the check to [after this line](https://github.com/postmarketOS/pmbootstrap/blob/8aede0dea0cb32a3ca33e423c8cbb444e918e58f/pmb/install/blockdevice.py#L62), and call blkid with the command to run in chroot. That seems like it would work, right?

ollieparanoid:
> Ha, so apparently [there's a way using python](https://github.com/hamptus/pyMBR/blob/master/mbr.py) (external library someone wrote), but I *really* don't want to go there...

Agreed.

> Another idea: can we do this check somehow from within the chroot, where we have blkid provided by busybox? ...

That's a great idea! \o/

craftyguy:
@ollieparanoid 

Question about this: https://github.com/postmarketOS/pmbootstrap/blob/master/pmb/install/blockdevice.py#L45

Does this also expose the partition devices too in the chroot? For example, would this mount bind include the `p1`, so that in the chroot there would be a `/dev/installp1` after this command? If not, I'll need to rethink how I detect partition labels in the chroot because (presumably) blkid would not know about that partition in the chroot.

ollieparanoid:
Uh... right, it does not mount the partitions at that point. We could search for the boot partition (is it `args.device + "1"` or `args.device + "p1"`?) and mount it afterwards if it exists with another `pmb.helpers.mount.bind_blockdevice()` command, and directly umount it after we checked the label.

I would put all that in an extra function.

craftyguy:
Is there some 'unbind_blockdevice()' function in pmb that I could use for unmounting it?

ollieparanoid:
`pmb.helpers.run.root(args, ["umount", mountpoint])` does the job. Other than that, you can umount all mountpoints in a given folder with:

https://github.com/postmarketOS/pmbootstrap/blob/caf64719022bdf8170caf5f0179f56b8a84c1add/pmb/helpers/mount.py#L102-L109

craftyguy:
@ollieparanoid changes were made to show the "EVERYTHING ON " + args.sdcard + " WILL BE ERASED! CONTINUE?" message if pmOS is NOT detected, and show the "A previous install blah blah" message if the first partition on the sdcard has the label `pmOS_boot`. Tested on a blank sdcard and on an sdcard with a previous install of pmOS.

craftyguy:
@ollieparanoid 
Looks good to me. 🚢 it!

PR #770, opened 2017-10-14
Package firmware-sony-amami (wifi firmware)

ata2001:
tested it on sony-amami


z3ntu:
Please don't merge. Will explain why later (in the evening)

ata2001:
@z3ntu Added label. Of course it wouldn't have got merged before we tested and discussed it.

ata2001:
I've used the files from the sony rhine-common repo, not sure about the difference between the devices, we might need different WCNSS_qcom_cfg.ini files.

aports/firmware/firmware-qualcomm-prima-wlan/APKBUILD
```diff
@@ -0,0 +1,45 @@
+_files="wcnss.b00 wcnss.b01 wcnss.b02 wcnss.b04 wcnss.b06 wcnss.b07 wcnss.b08 wcnss.b09 wcnss.mdt"
+_files1="WCNSS_cfg.dat WCNSS_qcom_cfg.ini WCNSS_qcom_wlan_nv.bin"
+
+pkgname=firmware-qualcomm-prima-wlan
+pkgver=1
+pkgrel=0
+pkgdesc="Firmware for qualcomm prima wifi chips"
+url="https://github.com/TheMuppets/proprietary_vendor_sony"
+arch="armhf"
```
ollieparanoid: It might make more sense to put `noarch` here, although [`linux-firmware` from Alpine has `all`](https://github.com/alpinelinux/aports/blob/097cf28fcb19221db7d7e87d18040439d63c0241/main/linux-firmware/APKBUILD#L9). I'm asking for the reason for that in `#alpine-devel` and report the answer here.

aports/firmware/firmware-qualcomm-prima-wlan/wcnss-wlan.initd
```diff
@@ -0,0 +1,14 @@
+#!/sbin/openrc-run
+
+description="Initialize wcnss wlan driver"
+
+depend()
+{
+	need udev
+}
+
+start()
```
ollieparanoid: Nice! I guess it would also be possible to add a stop() function with `echo 0`?

z3ntu:
Okay so as far as I have seen, all devices have 1. different firmware files (`wcnss.*`) and also different configuration files (`WCNSS_*`). 
I think the configuration files also differ in the version of the kernel driver.
So probably we'll have to use different firmware packages for each device and a common "prima/pronto" package that provides the rc script.

ollieparanoid:
@ata2001: do you want to split the firmware package in two, one with shared files and one with amami related files? It could also be done in one APKBUILD with subpackages, if that makes the code cleaner and easier to maintain.

aports/firmware/firmware-qualcomm-prima-wlan/APKBUILD
ata2001: @ollieparanoid apk complained when I was trying to build it with noarch.

yuvadm:
The `t2m-flame` firmware, for example, already has devices-specific files packaged in `firmware-t2m-flame`.

ata2001:
I split up the packages to `firmware-sony-amami` and `wcnss-wlan`.

aports/main/wcnss-wlan/APKBUILD
```diff
@@ -0,0 +1,19 @@
+pkgname=wcnss-wlan
+pkgver=1
+pkgrel=0
+pkgdesc="Init script for enabling wcnss wlan driver"
+url="https://github.com/postmarketOS/pmbootstrap"
+arch="noarch"
+license="GPL"
```
ollieparanoid: GPLv1? If you want to put it under the GPLv3 and later, use `GPL3+`. [More information](https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package#license)

aports/main/wcnss-wlan/wcnss-wlan.initd
```diff
@@ -0,0 +1,21 @@
...
+start()
+{
+	ebegin "Enabling wlan driver"
+	echo 1 > /dev/wcnss_wlan 2>/dev/null
+	echo sta > /sys/module/wlan/parameters/fwpath
+}
+
+stop()
+{
+	return 0
```
ollieparanoid: I recommend not having a stop function instead of having an empty stop function that does nothing. Maybe stopping it is as easy as doing `echo 0 > /dev/wcnss_wlan`?

aports/main/wcnss-wlan/APKBUILD
```diff
@@ -0,0 +1,19 @@
+pkgname=wcnss-wlan
+pkgver=1
+pkgrel=0
+pkgdesc="Init script for enabling wcnss wlan driver"
```
ollieparanoid: This is for Qualcomm prima chips on non-mainlined Android kernels only, right? maybe we could mention it here. How about: `"Init script for Qualcomm prima wifi chips (non-mainline Android kernels)"`

aports/main/wcnss-wlan/wcnss-wlan.initd
ata2001: @ollieparanoid Sorry, I misunderstood your previous review. I was not able to turn it off this way when tried, but will check again, what that exactly does.

ata2001:
@ollieparanoid pushed to the branch

Issue #592, opened 2017-09-17
Package nexmon -- Broadcom firmware patches to fix broadpwn and enable monitor mode


Pneumaticat:
As mentioned in #513.

Repository is [here](https://github.com/seemoo-lab/nexmon).

Build instructions are only for Ubuntu/Raspberry Pi, and appear to require some 32-bit libraries that Alpine does not package.

I can think of three options, none of which are particularly fun:

1. Package and maintain the required 32-bit libraries.
2. Build firmware binaries and package them as blobs. (Not ideal for transparency.)
3. Create ubuntu buildroot with docker and build them that way? ;) Just kidding, probably.

Does anyone have any other ideas on how to package them?

ollieparanoid:
From looking at the source: When using Nexmon as described in their `README.md`, you would use a few programs/libraries, that are not built from source, but contained as binaries in the git repository. That's where the 32bit library dependencies come from.

So to package it properly, we would build everything, that we don't already have, from source and use our own cross-compiler.

Here are my recommended first steps:
- [ ] Set up a VM where you use it as expected by the developers, so it's easier to understand how the build system works
- [ ] Disable the statistics reporting
- [ ] Build the firmware files we want (at least one)
- [ ] Figure out all dependencies, that we need to build the firmware files (there is a lot in the `buildtools` folder, e.g. the cross-compiler and the `nexmon-gcc-plugin`)

One we have that, I would recommend further brainstorming on how to do the actual packaging of the dependencies and firmware files, and how to handle the cross-compiling (see also: [build internals](https://wiki.postmarketos.org/wiki/Build_internals)).

*I'm really excited about this topic \o/*

PR #771, opened 2017-10-14
Motorola Titan wifi firmware

PabloCastellano:
I have symlinked all files available in the firmware partitions. I could not find only one file which is downloaded from the lineageos repository.

Side note: the wcnss.* files are not the same packaged in #770.

There are the md5sums of the files used in titan:

```
0f3d4a65a08bf0eda8c3a23e48e438ad  wcnss.b00
f08f46da25e790ce436f6b31e8dc348f  wcnss.b01
bc5fc4f6405e9d3d9a4ff633e5cdfcd4  wcnss.b02
4355bc68a61c942125781266d7a89a9d  wcnss.b04
60c898e99adc6cf9a5c1cb0f6c2c3a52  wcnss.b06
4ae71336e44bf9bf79d2752e234818a5  wcnss.b07
cb18a5d28e77522dfec6a6255bc3847e  wcnss.b08
88049c23cdf003738bb0cefb4dc49dfb  wcnss.b09
8bf067bc8a30b9277a9b228e8a2a2252  wcnss.mdt

2922cdca251f38d5128b3d158ba510ac  WCNSS_cfg.dat
5d0e9d67ef4e1887ff31298f602e0ee2  WCNSS_qcom_cfg.ini
898d38a57c2e2d2bef7f0f1720e73df0  WCNSS_qcom_wlan_factory_nv.bin
d5071632bf2764908c7189e87c4d323a  WCNSS_qcom_wlan_nv.bin
```


ollieparanoid:
For consistency with how we handle it for other devices, we would download the firmware files from TheMuppets or similar, and use those instead of creating symlinks.

This has the following advantages:
* We know exactly which files will be used, no matter what ROM was installed before
* We can update the firmware files (even patch binary firmware #592) without having to modify that partition
* In multiboot scenarios (nfs boot, ...) it is not possible, that other ROMs overwrite the firmware files we provide

montvid:
@PabloCastellano I guess everything is working for you I just would extract all the firmware from the original device rom and double check the sha sums if I was paranoid. :)

montvid:
Sorry - you have a seperate firmware partition? Is it from the original device? Is it just your device?

ollieparanoid:
Recommended TODOs:
* [ ] download the firmware files with the APKBUILD (the firmware APKBUILD style has been refactored in master in #788, please try to stick to that so we have it consistent)
* [ ] test on device

PabloCastellano:
Hey!

I'm back to this issue.

Unfortunately I could not find any of the wcnss.b?? files available in any repository.
I have found the same files but with different hashes but I guess these will not work.

The only option here would be hosting these myself in some repository as @yuvadm has done for [t2m-flame](https://github.com/yuvadm/mozilla-flame-firmware).

I will soon update the branch accordingly

drebrez:
@PabloCastellano I had to do the same for the huawei-y530; extracted from firmware partition and uploaded to github. 
Then packaged like this => https://github.com/postmarketOS/pmbootstrap/blob/master/aports/firmware/firmware-huawei-y530/APKBUILD#L13

PR #775, opened 2017-10-14
Do not create device-mapper mapping if no subpartitions

### LINKED ISSUE #774: Mounting of /boot fails when mounting by disk label
The new /etc/fstab will try to mount /boot by label, and it fails on the N900 (sdcard install):

```
localhost:/home/user# mount -v LABEL="pmOS_boot" /boot    
mount: /boot: /dev/mmcblk0p1 already mounted or mount point busy.
```

@ollieparanoid pointed me to [this section in the initramfs](https://github.com/postmarketOS/pmbootstrap/blob/de84ec61a401be284df02c554cb53acbaf32b8c0/aports/main/postmarketos-mkinitfs/init_functions.sh#L66-L69), kpartx is being used to create mappings to handle subpartitions on some devices (Android-only?).
This results in a real partition and a dm mapping with the same labels (and UUIDs):
```
localhost:/home/user# blkid 
/dev/dm-0: LABEL="pmOS_boot" UUID="70e8a3f2-e295-4765-a6bc-1427911bde01" TYPE="ext2"
/dev/dm-1: LABEL="pmOS_root" UUID="8fd44248-b745-4642-8bd1-67e18f5c5634" TYPE="ext4" 
/dev/mmcblk0p2: LABEL="pmOS_root" UUID="8fd44248-b745-4642-8bd1-67e18f5c5634" TYPE="ext4" 
/dev/mmcblk0p1: LABEL="pmOS_boot" UUID="70e8a3f2-e295-4765-a6bc-1427911bde01" TYPE="ext2"      
```

craftyguy:
This causes initramfs to skip calling 'kpartx' when pmOS has been
installed on an external disk (e.g. sdcard).

Tested on my N900 and works great. 

Note that this will cause the 'resize' stuff to be skipped over, but it's not needed because the call to parted for creating the root partition already instructs parted to use 100% of the available space, and since it's working on a real device, there's no need to resize in initramfs :)

fixes #774


aports/main/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -39,6 +39,11 @@ setup_mdev() {
 }
 
 mount_subpartitions() {
+	# Do not create subpartition mappings if install was on external
+	# disk (e.g. an sdcard)
+	# shellcheck disable=SC2154
+	[ "${deviceinfo_external_disk_install}" = "true" ] && return
```
ollieparanoid: As written in the chat: there are Android devices, which could have it both installed to sdcard or to the internal flash memory.
I think the key is running kpartx only on partitions which really have subpartitions. If there's no good way to check for that, we could write a tiny C program for that purpose.

craftyguy:
This checks that BOTH `pmOS_root` AND `pmOS_boot` partition labels are present, if both of them are, then it returns immediately. If either one or both are not present, then it will proceed to call kpartx to set up subpartition mapping.

ollieparanoid:
I'm afraid that this still does not work as intended in all cases. Just to make sure I have looked up where we set the label when the root partition is encrypted. Turns out, we set it inside the cryptsetup container, which means that it is only accessible once the cryptdevice has been unlocked (which happens later in the initramfs).

https://github.com/postmarketOS/pmbootstrap/blob/9e99ff6f23fc8b932a858688b729c756045de053/pmb/install/format.py#L51-L59

So in order to get this right, **we can only check for `pmOS_boot`, *not* for `pmOS_root`**. I know how much work you put in that single blkid grep line to check for both. Sorry :-\

The good news is, that resizing the image for Qemu is still working with this PR, just tested that (with `--no-fde`) because I thought that it could have been broken.

Recommendations before merging:
- [ ] only check for pmOS_boot
- [ ] update the long comment in `find_root_partition()` (the two paragraphs at the bottom are unnecessary now as I see it, we could add a link to [the wiki](https://wiki.postmarketos.org/wiki/Partition-layout))
- [ ] test on an android device

craftyguy:
@ollieparanoid I've fixed this per your suggestion, including the long comment. I still need to test on my N900 (will do it later when I am around it), and would appreciate if someone could test this on an Android device as well!

craftyguy:
Ok I think this is ready to be reviewed. Works as intended on my N900, is anyone able to test on an android device?

drebrez:
It boots correctly but probably it doesn't resize the partition because it
will probably find the /dev/dm-1
I will also test it later

drebrez:
Ok, did some testing and discovered that `/dev/dm-0` and `/dev/dm-1` are created/triggered by `kpartx`,
this means that checking `pmOS_boot` with `blkid` like you did, works perfectly fine, it doesn't find it and mount the subpartitions.

Here is the log => https://hastebin.com/wumuguciqo.coffeescript

Ps: I also added a sleep between `setup_mdev` and `mount_subpartitions` just to exclude that mdev was creating the `/dev/dm-X` files

ollieparanoid:
Well, resizing is not necessary for the sdcard as we always take up the entire space when formatting. The only edge case would be QEMU, but it appeared to be still working when I tested here: https://github.com/postmarketOS/pmbootstrap/pull/775#issuecomment-337011163. I am pretty sure that the code changes since then do not influence the test outcome, but testing again won't hurt!

Thanks for thinking along!

drebrez:
@ollieparanoid I tested only because I didn't know that `kpartx` was creating the `/dev/dm-0`, I thought the kernel or busybox was mounting them. 
If that was the case, the `blkid |grep -q "pmOS_boot"` would have returned and skipped the mount of subpartitions, boot was still possible because the `find_root_partition` now also searches for `/dev/dm-X`, but the resize function doesn't work with `/dev/dm-X`.
Hope it's more clear what was my concern.

ollieparanoid:
Ah, now I understand. Thanks for clarifying!

Issue #750, opened 2017-10-12
No boot image generated


fyfyone:
Hello, I am having noticing issues during `pmbootstrap install`. First of all,
 [this is my Log for both `pmbootstrap build` and `pmboostrap install` from line 4006 onward as well as APKBUILD](https://gist.github.com/fyfyone/c0ee1da32abea1c6f2202b033ec5cfb8).

I think there is no boot image generated because
* of [these](https://gist.github.com/fyfyone/c0ee1da32abea1c6f2202b033ec5cfb8#file-log-txt-L4238-L4247) lines where it says `/boot/dt.img` was not found in `mkbootimg`.
* a generated recovery zip file fails because `boot.img` could not be found.

What can I do to fix this issue and generate `boot/dt.img`?

drebrez:
@fyfyone seems like you mixed some stuff, the `/boot/dt.img` file is necessary if you have to append the DTB to the `boot.img` instead of the kernel.
Should should try to understand where you have to put the DTB for your device, appended to the kernel? or appended to the boot.img?
The `deviceinfo_bootimg_qcdt="true"` variable in the `deviceinfo` file is to append the `/boot/dt.img` to the `boot.img`. 
And from your APKBUILD you are using the kernel with the DTB file appended (`$srcdir/build/arch/arm64/boot/Image.gz-dtb`).
Using the DTB in boot.img you have to generate that `/boot/dt.img` as in [this APKBUILD](https://github.com/postmarketOS/pmbootstrap/blob/0cc7869576dc340a4b42c9789fdef705bc02b92b/aports/device/linux-huawei-y530/APKBUILD#L116)

ollieparanoid:
@fyfyone: are you sure, that you need `deviceinfo_bootimg_qcdt="true"` at all? If you don't know, you probably don't need it and it should work when you remove the line.

drebrez:
@fyfyone I compared the offset you specified in the deviceinfo with those of TWRP and they are different. 
Can you try with the following offests?
```
deviceinfo_flash_offset_base="0x80000000"
deviceinfo_flash_offset_kernel="0x00008000"
deviceinfo_flash_offset_ramdisk="0x02000000"
deviceinfo_flash_offset_second="0x00f00000"
deviceinfo_flash_offset_tags="0x01e00000"
```
edit the `aports/device-huawei-kiwi/deviceinfo` and then the following commands:
`pmbootstrap checksum device-huawei-kiwi`
[...]

PR #780, opened 2017-10-16
Initial commit of files for Huawei Honor 5X

fyfyone:
I am merge-requesting for fellow folks to find my work because I noticed working on this is time-consuming and wanted to take a break.

What works:
* The kernel compiles

What does not work:
* Unidentified problem and subject of #750 

I already created a [wiki page](https://wiki.postmarketos.org/wiki/Huawei_Honor_5X_(huawei-kiwi)) for my device.


ollieparanoid:
I'm merging this into the new `device-huawei-kiwi` branch, because the device doesn't boot yet and adding a link to the wiki page.

Thank you for the work you have already done on the device! This wouldn't be the first time, that one person starts the postmarketOS device port, and another person picks it up later and continues the port. So it's really good that you're making this PR and that you have created the device page in the wiki 👍

PR #783, opened 2017-10-17
Install dbus by default

craftyguy:
Since we are starting it on boot, and wpa_supplicant is being run in 'dbus mode', it makes sense to have it installed by default

### DIFF
--- a/aports/main/postmarketos-base/APKBUILD
+++ b/aports/main/postmarketos-base/APKBUILD
@@ -1,12 +1,13 @@
 pkgname=postmarketos-base
 pkgver=3
-pkgrel=10
+pkgrel=11
 pkgdesc="Meta package for minimal postmarketOS base"
 url="https://github.com/postmarketOS"
 arch="noarch"
 license="GPL3+"
 depends="
 	alpine-base
+	dbus
 	chrony
 	cryptsetup
 	eudev


ollieparanoid:
`wpa_supplicant` already depends on `dbus`, but we can add it as explicit dependency if you think that makes sense. Do you prefer that?

craftyguy:
@ollieparanoid, yea I see that wpa_supplicant depends on dbus, but @montvid reported that dbus wasn't being installed for him automatically.. maybe he removed it? In any case, since we try to start the service in the post-install script, maybe it's best to be explicit about installing it? In many cases (except @montvid's I guess) this would be a noop for apk if dbus is already installed.

ollieparanoid:
@craftyguy: I agree that this wouldn't hurt other devices where it works. But I'm afraid, that we might cover up a bug here in the worst case, so I'd rather find out what is going wrong.

@montvid: do you have any details on how this happened, or could you provide instructions on how to reproduce this (or how you can reproduce it)? Thanks!

montvid:
While choosing "none" gui in the installer menu you get a bare system without dbus. Try yourself and see.

ollieparanoid:
I will do that and report back, thanks!

montvid:
> localhost:~$ sudo wpa_supplicant -u -Dnl80211 -B -c/etc/wpa_supplicant/wpa_supplicant.conf -iwlan0
dbus: Could not acquire the system bus: org.freedesktop.DBus.Error.FileNotFound - Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
Failed to initialize wpa_supplicant

> localhost:~$ sudo rc-update add dbus
rc-update: service `dbus' does not exist

craftyguy:
@montvid that is very strange. As @ollieparanoid pointed, out (and I confirmed in the APKBUILD), dbus is a dependency for wpa_supplicant.

If you `apk add wpa_supplicant` does it pull in dbus?

montvid:
It is not strange at all if you read the link - no it does not because it can run without it:
https://pkgs.alpinelinux.org/package/edge/main/aarch64/wpa_supplicant

craftyguy:
Ah, you're right. Man, I could have sworn that I saw `depends=dbus` in that APKBUILD earlier but I must have imagined it. This wasn't a problem on the N900 since I also install ofono, which does have a hard dependency on dbus.

ergo.. this PR is necessary then!

montvid:
@craftyguy we need not just the package but also run/enable the service at boot with openrc.

craftyguy:
@montvid that already happens in postmarketos-base today if it is installed: https://github.com/postmarketOS/pmbootstrap/blob/master/aports/main/postmarketos-base/postmarketos-base.post-install#L16

Also see #628

ollieparanoid:
@montvid: My bad! I must have looked at the dependencies of [3.6](https://pkgs.alpinelinux.org/package/v3.5/main/aarch64/wpa_supplicant) instead of [edge](https://pkgs.alpinelinux.org/package/edge/main/aarch64/wpa_supplicant) (which is what we use) and it depends on `dbus` there. In `edge` it only depends on `dbus-libs`.

Thanks for pointing us to the link, where it's obvious, merging this PR.

PR #786, opened 2017-10-17
[recovery-installer] Don't fail when reflashing after failure

ata2001:


ollieparanoid:
I can't test it any time soon. I assume you have tested it already and it is ready to merge?

ata2001:
@ollieparanoid worked for me. I disconnected the device, while it was extracting the rootfs, then started flashing again and it succeeded.

PR #773, opened 2017-10-14
Add firmware for the Fairphone 2

z3ntu:
Currently includes firmware and configuration files for the prima/pronto
wlan driver

Depends on #770

PR #788, opened 2017-10-18
Fix firmware packages

drebrez:
Added package name and version to source files in order to avoid conflicts in cache folder


ollieparanoid:
Could you change that to only include the `pkgver`, not the `pkgrel`? Because when we only change the APKBUILD, the files should not get downloaded again.

Thank you very much for making this @drebrez, that explains why Travis failed in recent PRs, which merged master!

ollieparanoid:
As discussed in the chat, I'll refactor the firmware packages to be more consistent and to use the commit hash instead of the pkgver/pkgrel instead.

ollieparanoid:
All firmware-packages are using the commit hash everywhere in the download file output (as specified in `source=`) and the APKBUILDs are refactored to be more readable (e.g. all source related variables are in one block now).

Because I've changed a lot, I wrote the following script to check for regressions. It builds each firmware package from master, and from this PR's branch, then checks if the file names in the tar (apk) archive are the same. It caught one regression, and now the APKBUILDs produce the same output again - so I'm pretty confident with this PR.

Would be cool if someone could review it though, before we ship it.

```shell
#!/bin/sh -e

PMBDIR=~/code/pmbootstrap/
BRANCH="fix/fix-firmware-packages"
TEMP="/tmp/test-APKBUILD-refactoring/"
WORK="$(pmbootstrap config work)"
ARCH="x86_64"
DIFF="colordiff"

# $1: name of package to be removed
remove_existing() {
	echo "remove existing $1 package"
	sudo rm -rv "$WORK/packages/$ARCH/$1"-* || true
	pmbootstrap index
}

# $1: package name
temp_copy_package() {
	echo "copy package $1 to temp folder"
	mkdir -p "$TEMP"
	cp -v "$WORK/packages/$ARCH/$1"-* "$TEMP/"
}

temp_clear() {
	echo "clear temp folder"
	[ -e "$TEMP" ] && rm -rf "$TEMP" || true
}

# $1: package
# $2: branch
build_on_branch() {
	cd "$PMBDIR"
	echo "building $1 on branch $2"
	git checkout "$2"
	remove_existing "$1"
	pmbootstrap build "$1"
}

# List package contents from package folder and from temp folder, then
# compare contents
# $1: package name
compare() {
	echo "Compare builds"
	tar --list -f "$TEMP/$1"-* | sort > "$TEMP/_old"
	tar --list -f "$WORK/packages/$ARCH/$1"-* | sort > "$TEMP/_new"
	"$DIFF" "$TEMP/_old" "$TEMP/_new"
}

cd "$PMBDIR/aports/firmware"
for i in *; do
	echo ":: $i"
	# Build on master, copy to temp folder
	build_on_branch "$i" "master"
	temp_clear
	temp_copy_package "$i"

	# Build on other branch, compare
	build_on_branch "$i" "$BRANCH"
	compare "$i"
done
```

drebrez:
Nice job @ollieparanoid, now they look really better.
I reviewed the code and didn't find anything else to change.

drebrez:
The same refactor/cleanup should be reflected also to PR #773 and #771

ollieparanoid:
Thanks for the review! \o/

PR #792, opened 2017-10-18
Various small xfce4/Hildon fixes

ollieparanoid:
* xfce4/Hildon: Log to syslog instead of `~/x11.log` (see also: <https://github.com/postmarketOS/pmbootstrap/pull/762>)
* Fix xfce4 didn't start in qemu-amd64, because it required
  libEGL.so.1, which is provided by mesa-egl. This is a dependency
  of postmarketos-base-x11 now. (It worked for Hildon and Weston,
  because they pull mesa-egl in with other packages.)


pavelmachek:
Most of the changes look good to me. Not sure about adding the mesa-egl stuff ... That was not required for me. Why is it required now?

craftyguy:
Adding Mesa-egl actually breaks X on the N900, since it requires the GL blob to run

ollieparanoid:
Thanks for the feedback, the new dependency is now in `device-qemu-amd64-x11`.

montvid:
nexus 7 needs mesa-egl to run xfce. It is added to the device already by default.

ollieparanoid:
Reviews welcome! But we should wait with merging this until #808 gets merged (otherwise it will break for Android devices without that kernel config option enabled).

ollieparanoid:
Thanks for testing! 👍

PR #795, opened 2017-10-19
Fix flashing offsets

fyfyone:
Thanks to @drebrez for finding this issue on #750! Flashing finishes successfully now.


drebrez:
@fyfyone Merged, thanks for the update.

PR #796, opened 2017-10-20
Implement cache for initramfs-extra

drebrez:
Here the steps in few words:

- list and sort all the files in BINARIES_EXTRA and osk_conf (stored in `initfs_extra_files`)
- compute the md5 of each file (stored in `initfs_extra_files_hashes`)
- compute md5 of the previous list of md5 (stored in `initfs_extra_hash`)
- check if the cached file with the name `initramfs-device-extra_hash` exist
- in case it dosn't exist, clear any existing cache and generate it again (same process as before)
- copy cached file to final location

Initial idea was to compute md5 of all the files that will be saved in the archive, but the `lddtree` command is very time consuming.

Closes: #411


aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -287,6 +287,47 @@ append_device_tree()
...
+generate_initramfs_extra()
+{
+	echo "==> initramfs: creating $1"
+
+	osk_conf="$(get_osk_config)"
+	if [ $? -eq 1 ]; then
+		echo "ERROR: Font specified in /etc/osk.conf does not exist!"
+		exit 1
+	fi
+
```
ollieparanoid: A comment, that describes the hashing in the following block would be nice, something like:
`# Generate cache output filename (initfs_extra_cache) by hashing all input files`

aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -287,6 +287,47 @@ append_device_tree()
...
+	# Ensure cache folder exists
+	mkdir -p "$mkinitfs_cache_dir"
+
+	if ! [ -e "$initfs_extra_cache" ]; then
+		# If a cached file is missing, clear the whole cache and create it
+		rm -f ${mkinitfs_cache_dir}/*
+
+		tmpdir_extra=$(mktemp -d /tmp/mkinitfs.XXXXXX)
+
+		# set up initfs-extra in temp folder
```
ollieparanoid: Using capital letters at the beginning of comments would be nice for consistency.

aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -287,6 +287,47 @@ append_device_tree()
...
+	initfs_extra_cache="$mkinitfs_cache_dir/$(basename $1)_${initfs_extra_hash}"
+
+	# Ensure cache folder exists
+	mkdir -p "$mkinitfs_cache_dir"
+
+	if ! [ -e "$initfs_extra_cache" ]; then
+		# If a cached file is missing, clear the whole cache and create it
+		rm -f ${mkinitfs_cache_dir}/*
+
+		tmpdir_extra=$(mktemp -d /tmp/mkinitfs.XXXXXX)
```
ollieparanoid: How about moving that line after the `# Set up initfs-extra` comment below?

aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -287,6 +287,47 @@ append_device_tree()
...
+
+		# set up initfs-extra in temp folder
+		mkdir -p "$tmpdir_extra"
+		copy_files "$(get_binaries_extra)" "$tmpdir_extra"
+		copy_files "$osk_conf" "$tmpdir_extra"
+
+		# finish up
+		create_cpio_image "$tmpdir_extra" "$initfs_extra_cache"
+
```
ollieparanoid: How about removing that empty line?

aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -287,6 +287,47 @@ append_device_tree()
...
+	echo "==> initramfs: creating $1"
+
+	osk_conf="$(get_osk_config)"
+	if [ $? -eq 1 ]; then
+		echo "ERROR: Font specified in /etc/osk.conf does not exist!"
+		exit 1
+	fi
+
+	mkinitfs_cache_dir="/var/cache/postmarketos-mkinitfs"
+	initfs_extra_files=$(echo "$BINARIES_EXTRA$osk_conf" | xargs -0 -I{} sh -c 'ls $1 2>/dev/null' -- {} | sort -u)
```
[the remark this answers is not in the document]
drebrez: @ollieparanoid the `ls` is to resolve the path like `/usr/lib/ts/*` defined in `BINARIES_EXTRA`, I didn't find a better way to do it, any idea/suggestion to simplify it?

aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -287,6 +287,47 @@ append_device_tree()
...
+	initfs_extra_files_hashes=$(echo "$initramfs_extra_files" | xargs md5sum)
```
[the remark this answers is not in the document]
drebrez: @ollieparanoid It works but not like you suggested, need to remove the double quote of the variable

aports/main/postmarketos-mkinitfs/mkinitfs.sh
drebrez: 👍

aports/main/postmarketos-mkinitfs/mkinitfs.sh
drebrez: 👍

aports/main/postmarketos-mkinitfs/mkinitfs.sh
drebrez: 👍

aports/main/postmarketos-mkinitfs/mkinitfs.sh
drebrez: 👍

drebrez:
@ollieparanoid thanks for the review, I addressed the comments.
I thought that wasn't necessary to cache also the non-extra archive since it's really fast. But for consistency and also save some extra seconds it might be a good idea to do it. 
I will look into it in the next days.

ps: here you can see a verbose output of what's going on => https://hastebin.com/uyewawulen.go

ollieparanoid:
Awesome! \o/

drebrez:
@ollieparanoid doesn't make sense to do it in another PR? 
I'm not going to work on it right now, the non-extra package depends on a lot of external files (modules, deviceinfo, hooks, busybox link?) and it's probably more difficult to have an hash of everything.

craftyguy:
Wow, this is really nice @drebrez!

I really want to test it with osk-sdl, but won't be able to for a few days... 

If y'all don't want to wait for me, can someone (with a device that works with osk-sdl) perform the following tests and report back here:

* [ ] Enable fde (don't do `--no-fde`), install with this PR, test that you can unlock

* [ ] Rebuild pmOS (don't zap or disable fde) so that initramfs caching (this PR) is relied on to restore all osk-sdl deps. Test that you can boot/unlock through osk-sdl

ollieparanoid:
@drebrez: sure, then let's make it in another PR!
@craftyguy: thanks for listing the osk-sdl related testing steps!

I'll do the testing soon, then ship it. 🛥

ollieparanoid:
Alright, I have tested the following way on the `i9100`:
```shell
pmbootstrap zap
pmbootstrap install --sdcard=...
# insert sd card into target device, enter flashing mode
pmbootstrap flasher flash_kernel
```
And it worked as expected, I was able to use the OSK to unlock the device and boot straight into weston.

@craftyguy: I did not do the whole procedure again, because I realized (again) that we already build the initramfs multiple times during one `pmbootstrap install` call. It gets built in the post-install hook of `postmarketos-mkinitfs` the first time, and also by Python code when you run install and flasher. This was implemented to make sure, that we never flash the old initramfs if something did not work (and that is why the cache is so useful, so we can avoid the additional building steps). So what I have tested was already built with the cache being used.

It is good to go in my opinion. Terrific work again, @drebrez!

Issue #688, opened 2017-10-03
Create a kernel package template


ollieparanoid:
It would be nice if we had modern template somewhere in the code tree for new kernels with the following:
* using abuild's `default_prepare` for applying the patches (like @drebrez did)
* include commonly used patches by default (fix return address, timeconst fix, msm driver fix). it's much easier to remove patches that do not apply than finding which patch is necessary to fix a specific compilation error.
* using "_commit" instead of "_hash", because "_hash" is confusing
* use .tar.gz (not zip!) sources from LineageOS
* add useful comments everywhere
* in general: throw out everything that we do not need
* patching out -Werror* by default in `prepare()` (as I did [for linux-ouya-ouya](https://github.com/postmarketOS/pmbootstrap/blob/dafd1b8d526dc7fef31c2b4f7560b3b9d7e468ec/aports/device/linux-ouya-ouya/APKBUILD#L64-L66)):
```shell
# Remove all -Werror statements
find . -type f -name Makefile -print0 | xargs -0 sed -i 's/-Werror-/-W/g'
find . -type f -name Makefile -print0 | xargs -0 sed -i 's/-Werror//g'
```

MartijnBraam:
I think a small improvement could to remove `$_flavor` for the default source url and use a `$_repository` variable containing the github identifier (like LineageOS/android_kernel_lge_bullhead), with the `$_repository` variable and the `$_commit` variable together you can generate the correct source .tar.gz url and the correct `$kbuilddir` variable every time

ollieparanoid:
Alright, I have given this a lot of thought and came up with a whole new porting workflow around this template:
1. `pmbootstrap init`: user types in new device
2. Do you want to start a new port to $vendor-$name? [Y/n] y
  * To prevent typos in the device name causing the wizard to run by accident
3. OK, copying default files
  *  Creates the `device-{device}` and `linux-{device}` package (see template below, `{device}` gets automatically replaced in the template) with extended `aportgen` code
  *  Displays steps on how to continue with the port, and a link to the porting guide.

Furthermore, I would create an empty file as kernel config file, and, as soon as the user runs `pmbootstrap menuconfig` for the first time, copy the appropriate default config. That way we don't need to store this information in a comment, but in an actually used variable, and we save the user from manually pulling out that file and putting it in the right place.

The guide in the wiki would need to be adjusted, and the shortlinks would need to be created.

[...]

PR #799, opened 2017-10-21
Add linux-postmarketos-mainline aport

ata2001:
Aport based on `linux-postmarketos`. Work in progress.

Configs are from `linux-postmarketos` (ran olddefconfig) except armhf, that one is for `sony-amami`, we will probably need to change that.

Tested it on `sony-amami`.


ollieparanoid:
If I recall correctly, @opendata26 and @craftyguy were discussing whether it makes sense to ship a combined `linux-next` kernel, and came to the conclusion that it is too much effort to keep up with the release cycle: [They often have a lot of releases in a row, with the distance of one day .](https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/refs/)

So my recommendation would be putting the amami related patches on top of our `linux-postmarketos` package, if that is possible, or temporarily package an extra `linux-sony-amami-next` kernel, which you could update whenever something relevant changes in `linux-next` (so not necessarily every day).

Thanks for the fantastic work on mainlining this device \o/

craftyguy:
We could follow the RC kernels, which are weekly, and apply any patches from 'next' to help some devices out.

Following the next kernel would be a bit crazy due to the frequent changes :P

ata2001:
@craftyguy that would make more sense :)

Do you mean, we should add a new package and keep `linux-postmarketos` at stable, or the other way around? I think creating a linux-next based kernel package for each device is not good either.

craftyguy:
We could create a new package like 'linux-postmarketos-rc' or even call it '-next' (because it is technically next) but track mainline RCs instead of the real linux-next.

craftyguy:
In Matrix, I proposed the following to keep our kernel names in line with [what is used on kernel.org](https://www.kernel.org/), but totally open to feedback! (paging Dr @ollieparanoid )

Rename `linux-postmarketos` --> `linux-stable-postmarketos` to track [the latest stable kernel](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/log/?h=v4.13.9)

Add `linux-mainline-postmarketos` to track [Linus' tree](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/log/?h=v4.14-rc5)

ollieparanoid:
Oh, there we are talking about "mainline" kernel all the time, and what we mean is the "stable" branch, thanks for pointing that out! I think renaming the packages accordingly makes sense, but I would switch the order, so they are displayed nicely below each other in directory listings:
```
linux-postmarketos-mainline
linux-postmarketos-stable # renamed from linux-postmarketos
```

I recommend to adjust the pkgdesc to make it clear, that mainline is not the stable branch and to rename linux-postmarketos to linux-postmarketos-stable (and adjust the dependency in qemu and rx51) in a new PR.

[Relevant literature](https://www.kernel.org/category/releases.html):
> **Mainline**
>    Mainline tree is maintained by Linus Torvalds. It's the tree where all new features are introduced and where all the exciting new development happens. New mainline kernels are released every 2-3 months.
> **Stable**
>    After each mainline kernel is released, it is considered "stable." Any bug fixes for a stable kernel are backported from the mainline tree and applied by a designated stable kernel maintainer. There are usually only a few bugfix kernel releases until next mainline kernel becomes available -- unless it is designated a "longterm maintenance kernel." Stable kernel updates are released on as-needed basis, usually 2-3 a month.

craftyguy:
That naming is nice too! @ata2001 if you need help implementing this, let me know!

This is not huge at all just a classic @craftyguy ramble. No changes required beyond here folks. 
The ordering seems to be a product of their (arbitrary) names, it works now but if they decide on a `kernel-<any letter less than 'm'>` then maybe we'll be back here again!

aports/main/linux-mainline/APKBUILD
```diff
@@ -0,0 +1,179 @@
...
+pkgver=4.14-rc5
+case $pkgver in
+	*.*-rc*) _kernver=${pkgver%-rc*};;
+	*.*.*)	_kernver=${pkgver%.*};;
+	*.*) _kernver=$pkgver;;
+esac
+pkgrel=0
+
+arch="all"
+pkgdesc="Linux for pmOS supported chipsets (mainline)"
```
ollieparanoid: How about:
```shell
pkgdesc="pmOS kernel based on mainline (more bleeding-edge than stable)"
```

aports/main/linux-mainline/APKBUILD
```diff
@@ -0,0 +1,179 @@
+# APKBUILD based on linux-postmarketos aport. Changes:
+#
+# - use mainline branch instead of stable
+
+_flavor=mainline
```
ollieparanoid: As discussed in the ticket, `postmarketos-mainline` seems to be good.

aports/main/linux-mainline/APKBUILD
```diff
@@ -0,0 +1,179 @@
+# APKBUILD based on linux-postmarketos aport. Changes:
```
ollieparanoid: I am not sure how useful those three comment lines are anymore (I'm also leaning towards removing them in all the `device/linux-*` APKBUILDs in the long run, see https://github.com/postmarketOS/pmbootstrap/issues/688#issuecomment-334953621). What do you think about removing them?

aports/main/linux-mainline/APKBUILD
```diff
@@ -0,0 +1,179 @@
...
+#
+# - use mainline branch instead of stable
+
+_flavor=mainline
+_config="config-${_flavor}.${CARCH}"
+
+pkgname=linux-${_flavor}
+
+pkgver=4.14-rc5
+case $pkgver in
```
ollieparanoid: The `_kernver` variable isn't used anywhere, so the whole `case` block can be removed.

(`linux-vanilla` from Alpine and `linux-postmarketos` use it to download the `.0` release and then apply the patchset on top to save the building machines from downloading and storing the entire kernel source for each patchlevel increase.)

aports/main/linux-mainline/APKBUILD
```diff
@@ -0,0 +1,179 @@
+# APKBUILD based on linux-postmarketos aport. Changes:
+#
+# - use mainline branch instead of stable
+
+_flavor=mainline
+_config="config-${_flavor}.${CARCH}"
+
+pkgname=linux-${_flavor}
+
+pkgver=4.14-rc5
```
ollieparanoid: *NOTE: This pkgver is invalid from Alpine's standards (`4.14_rc5` would be valid). But `pmbootstrap` will build it fine unless you use `--strict`.  As this gets directly used in the download path and therefore makes the code simpler, I would keep it like it is now and only change it if we run into problems.*

aports/main/linux-mainline/APKBUILD
ata2001: @ollieparanoid couldn't we do the same with mainline release candidates?

ollieparanoid: Yes sure! We would just need to add code to prepare() to handle that.

ata2001:
@ollieparanoid now it downloads patches when updating, not the whole archive every time

ollieparanoid:
@ata2001, thanks for making all the changes! Code looks good to me now (although I didn't compile it). You've even renamed linux-postmarketos-stable and adjusted all devices using it.

The only thing I would change is increasing the pkgrels for the device-* packages you have modified.
What's your opinion, is it ready to get merged?

ata2001:
@ollieparanoid The latest rc patch should be optional, it's only needed, if the rc version is greater than 1. I'll fix that.

ata2001:
Hmm, actually requesting a diff of the same version returns an empty document, maybe it doesn't even need change.

ata2001:
@ollieparanoid The config included is for, and only tested on sony-amami. If you're okay with merging that, we can make it work at least on castor_windy too in another PR.

ollieparanoid:
Yep, that sounds good! Merging. Travis is still failing for the same unrelated reason.

PR #804, opened 2017-10-21
package nanopb with protocol definitions for generator

tssk:


tssk:
This commit should fix:

* nanopb is not packaged properly (we'll need to package the whole
  generator folder and run make in it first or something)

aports/android/nanopb/APKBUILD
```diff
@@ -33,6 +33,18 @@ package() {
 	for i in nanopb_generator.py protoc-gen-nanopb; do
 		install -Dm755 "generator/$i" "$pkgdir/usr/bin/$i"
 	done
+
+	# Protocol definitions for generator
+	cd generator/proto
```
ollieparanoid: `make -C generator/proto` should be equivalent to these three lines.

ollieparanoid:
NOTE: This is *not* for the `master` branch, and the code I've provided earlier for the same branch isn't high quality either. So if you prefer if we make it work first, and then make it pretty when merging it into `master`, I'd also be fine with that. Just let me know and I'll merge it.

tssk:
I agree the change you suggest is much more elegant.

And it also show that I am not very skilled about compliling things. It took me two evenings to prepare this. But I hope to get better over time :) One small step forward after another.

tssk:
For the next step - it seems that utils/Log.h is missing - probably part of cutils. And I am not sure that includes from libcutils and cutils is the same?

ollieparanoid:
I've rebased the `feature/android-rild` branch on master (so it has the new work folder format), and I'm currently writing a ticket with what is left to do to get this working 👍

ollieparanoid:
New issue: #806, I'll try to answer your log.h question there.

PR #814, opened 2017-10-23
Support multiple architectures in `pmbootstrap menuconfig` command

drebrez:
This adds the `--arch` argument to the `menuconfig` command, it's only necessary when the kernel package have `arch="all"`, which lets the user choose which config to edit. 
In case the package have a specific architecture and the `--arch` argument is passed, it should match the architecture of the package


drebrez:
@ollieparanoid looks good 👍

Issue #806, opened 2017-10-22
Package RIL from Android for compatibility


ollieparanoid:
_NOTE: Packaging can be done without running postmarketOS on a real device._

There's at least for quite a few Samsung phones an open source component, that talks to RIL instead of ofono. If we had RIL, then we should be able to use it. See #598 for details.

@tssk  and I have started work in the [feature/android-rild](https://github.com/postmarketOS/pmbootstrap/tree/feature/android-rild/aports/android) branch.

In contrary to other software, the RIL daemon was not meant to be packaged for normal Linux distributions. To get it working, we write our own Makefiles and patch stuff out, that is Android specific and not useful for us (e.g. [wakelocks](https://lwn.net/Articles/318611/) are already stubbed out).

Dependencies (may be incomplete, in brackets is the package name in postmarketOS):
```
rild (android-ril)
- nanopb (nanopb)
- librilutils (android-ril)
- libril (android-ril)
    - libhardware_legacy (patched out)
    - libcutils (android-libcutils)
        - liblog (android-liblog)
```
Status:
* **nanopb**: compiles and appears to be working
* **android-liblog**: compiles (will probably not work as expected once running? patch it to log to syslog would be nice)
* **android-libcutils**: compiles, but the resulting shared library isn't linked right against liblog (check with ldd)
* libhardware_legacy: would be necessary for [wakelocks](https://lwn.net/Articles/318611/) only, which isn't in mainline Linux anyway so we can use a stub instead
* **android-rild**: Makefiles are partially wrong, needs fixing, doesn't compile

How to help:
* clone the pmbootstrap source code
* `pmbootstrap init` (defaults are fine)
* `git checkout feature/android-rild`
* `pmbootstrap build android-ril`
* see the logs in `pmbootstrap log`
* try to fix stuff by `editing aports/android/*`
* make pull-requests with fixes

ollieparanoid:
@tssk wrote in https://github.com/postmarketOS/pmbootstrap/pull/804#issuecomment-338461706:
> For the next step - it seems that utils/Log.h is missing - probably part of cutils. And I am not sure that includes from libcutils and cutils is the same?

[...]

Issue #793, opened 2017-10-18
Package "timekeep" to use Qualcomm's realtime clock (non-mainline kernels)?


ollieparanoid:
@opendata26 found this repository:
https://github.com/sonyxperiadev/timekeep

From what I understand, this is a workaround for non-standard behavior of the realtime clock driver in the Android kernels for Qualcomm chipsets.

We could package this with a custom OpenRC service, that loads the realtime clock at boot time and stores it on poweroff.

z3ntu:
```
/ # timekeep
I/TimeKeep(17953): usage: timekeep store|restore
/ # timekeep store
I/TimeKeep(17977): Failed to read epoch while storing
/ # timekeep restore
I/TimeKeep(17978): Failed to read from /sys/class/rtc/rtc0/since_epoch (-2), bailing out
```
I have currently completely patched out the `property_get()` and `property_set()` methods, so it's not functional. Should I write the value to some file somewhere?

z3ntu:
Found more information in the TWRP source: https://github.com/omnirom/android_bootable_recovery/blob/android-8.0/twrp-functions.cpp#L914-L921

PR #815, opened 2017-10-23
Packaging timekeep

z3ntu:
Contains partly as the branch for #806 but I like these Makefiles (which
are from Debian) more. :)

For issue #793


opendata26:
You would need to change the ATC file path to somewhere else, the property_functions should probably  save to a file and the time would have to be set before first run (probably using openntpd)

ollieparanoid:
## Makefiles

* In general it would be good if the individual C files were compiled to objects, otherwise distcc will not be used, so everything will be compiled in Qemu which is a lot slower.

* I'd be very interested in the source of the Makefiles, I didn't know that someone already packaged them for Debian!

* It's great that you got it to compile, that means that the Android dependencies are packaged in a way, that they are usable and this is probably what we'll need for the android-rild.

## Do we need to package it that way?

[`timekeep.c`](https://github.com/sonyxperiadev/timekeep/blob/master/timekeep.c) seems to be really simple, so I wonder if it is worth depending on all the android specific code for that. All it seems to do is reading the time from `/sys/class/rtc/rtc0/since_epoch`, write to `/data/time/ats_2` and use property_get/property_set to sync the value to `persist.sys.timeadjust` (which is even patched out).


## Alternative approach

### Reading and writing with hwclock

Usually one would use `hwclock` to set the hardware time in a Linux distribution, and pass the real time clock device with the `-f` parameter. So I recommend doing that, and I expect that only reading works, but not writing. As it says in the [README.md](https://github.com/sonyxperiadev/timekeep/blob/master/README.md):

> TimeKeep is a small utility to keep track of time & date since the **RTC driver** on Qualcomm chipset is **read-only.**

It would be something like `hwclock -r -f /sys/class/rtc/rtc0/device` to read, and `hwclock -w -f /sys/class/rtc/rtc0/device`.

### Wrting manually, like timekeep does it

So to se the time, we need to write to `/data/time/ats_2` as I understand it, and the realtime clock will pick it up then. Right now I don't think we mount the `/data` partition, it would be great if someone could boot into TWRP/Android and check which partition it belongs to.

With that information, we should also be able to mount it, and then I would try to write the time to the file, and checking if it worked by reading it with hwclock again.


### Packaging

If all the manual steps are working, we could provide a shell-script (or C program) as wrapper for hwclock, that calls busybox' hwclock for reading and does the non-standard way of writing manually. That, together with an appropriate fstab entry and with enabling the hwclock service at boot (in a post install script) could go into an extra package, `qualcomm-hwclock-workaround`, on which the Qualcomm devices could then depend on.


## Conclusion
As it is right now, I would rather not merge it, because timekeep doesn't really do anything with the patched out code. But with some further brainstorming and testing we should be able to figure out a good way to integrate the functionallity.

Thank you very much for packaging this, I'm sure we can use the efforts somehow, especially your work on the dependencies, which are shard with `android-rild`!

ollieparanoid:
No progress in more than 2 months. Feel free to reopen when you're working on this again.

bhush9:
Btw, for next person reading this: there is https://github.com/ubports/timekeeper which doesn't use android_properties and standard libc and it keeps the timestmap in file in /data/time/timekeep (which can be compile time selected).

PR #817, opened 2017-10-24
 [linux-postmarketos-mainline] Avoid name conflict when downloading patches

ata2001:
Sorry, I only realized this issue, after it got merged.


ollieparanoid:
Could you please explain the issue? We're using `$pkgname` in the source file, which is different for `postmarketos-linux-stable` and `postmarketos-linux-mainline`, so I'd expect it to be working (not tested).

ata2001:
@ollieparanoid The conflict happens, when we're on rc-1, the old naming only contains the target version, so there will be two file named `postmarketos-linux-mainline-4.15-rc1.patch`: one that updates the kernel from the latest stable, and one empty.

We could make the rc-1-->rc-x patch conditional, but even then I think this naming makes more sense.

PR #820, opened 2017-10-24
Close #761: Integrate coveralls.io to show test coverage

### LINKED ISSUE #761: Use coveralls.io to display testing coverage?
PR #760 will integrate automatic testing on Travis CI, at least for the fast testcases. I would like to display how much testing coverage we have for pmbootstrap in the README.md (with a badge and with details when you click on it).

**Proprietary** service [coveralls.io](https://coveralls.io/) looks like a nice interface and is free to use for open source projects. It is compatible to GitHub and gitlab [in case we decide to move there](https://github.com/postmarketOS/postmarketos.org/issues/37).

I have not found a similar service, where the source code is available (is there any?).

Worst case is, that for some reason we suddenly can't use that service anymore and lose the coverage reports from the past. We can always generate coverage reports ourselves with [pytest-cov](https://github.com/pytest-dev/pytest-cov) (which is the output that gets used for coveralls.)

And of course the upside is, that this motivates (at least me) to make more testcases and have good testing coverage.

So what do you think about this proposal, fellow hackers?

ollieparanoid:
Whether we do this or not has been discussed here, and no one spoke out against it:
https://github.com/postmarketOS/pmbootstrap/issues/761

Preview of the README.md:
https://github.com/postmarketOS/pmbootstrap/blob/feature/coveralls/README.md

Submitting the coverage is working:
https://coveralls.io/github/postmarketOS/pmbootstrap

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,14 @@ addons:
     - debian-sid
     packages:
     - shellcheck
-install: "pip install flake8 pytest-cov"
+install: "pip install flake8 pytest-cov python-coveralls"
 script:
   - test/static_code_analysis.sh
   - yes "" | ./pmbootstrap.py init
   - ./pmbootstrap.py kconfig_check
   - test/testcases_fast.sh
   - test/check_checksums.py
+after_success:
+  - coveralls
 notifications:
   - email: false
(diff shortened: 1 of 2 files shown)


drebrez:
@ollieparanoid branch `feature/coveralls` is still present, I think it can be removed (https://github.com/postmarketOS/pmbootstrap/tree/feature/coveralls)

ollieparanoid:
Thanks for noticing, done!

Issue #811, opened 2017-10-23
`pmbootstrap shutdown` only shuts down the binfmt of the currently selected device arch


drebrez:
I tried to enter the chroot for the aarch64 but it fails with:
```
drebrez@drebrez-laptop:~/pmbootstrap$ pmbootstrap chroot --suffix buildroot_aarch64
[20:31:43] (buildroot_aarch64) % sh
/usr/sbin/chroot: failed to run command 'sh': No such file or directory
[20:31:43] ERROR: Command failed: (buildroot_aarch64) % sh
```
here the failing command:
```
subprocess.CalledProcessError: Command '['sudo', '/bin/sh', '-c', "env -i CHARSET=UTF-8 
PATH=/usr/lib/ccache/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin 
SHELL=/bin/ash HISTFILE=~/.ash_history /usr/sbin/chroot 
/home/drebrez/.local/var/pmbootstrap/chroot_buildroot_aarch64 sh -c 'cd /;sh'"]' 
returned non-zero exit status 127
```

What I discovered until now:
```
drebrez@drebrez-laptop:~/.local/var/pmbootstrap/chroot_buildroot_aarch64$ lddtree bin/busybox
busybox => bin/busybox (interpreter => /lib/ld-musl-aarch64.so.1)
    libc.musl-aarch64.so.1 => not found
```

ollieparanoid:
I think this might happen, if you kill `pmbootstrap` while it is setting up the chroot (e.g. with `^C`).
What happens, when you zap your chroots and try it again?

drebrez:
Already tried running all possible combinations of `pmbootstrap shutdown` and `pmbootstrap zap`

ollieparanoid:
Could you please zap your chroots again, clear your log file, run `pmbootstrap chroot --suffix buildroot_aarch64` and then upload the entire logfile?

drebrez:
https://hastebin.com/vuceneyaje.swift

ollieparanoid:
This might be an upstream issue, there's some discussion in `#alpine-devel` about /bin/sh missing or something. I need to look into this more to understand it.

kaniini:
it’s failing because you’re not running the initial apk process inside qemu, so the scripts can’t run.  apk cannot set up the /bin/sh symlink because it can’t run busybox, since it’s a different architecture.

[...]

PR #822, opened 2017-10-24
Unregister all binfmt architectures during `pmbootstrap shutdown`

drebrez:
also removed the `-r, --rootfs` argument from `pmbootstrap build_init` command

Close: #811


drebrez:
Issue found and fixed, hope that everything works now 👍

PR #823, opened 2017-10-24
KFHDX 3rd Gen (thor, possibly apollo) basic support

MoreRobustThanYou:
It compiles. Boots into weston successfully!
When flashed using this script in TWRP,
```
dd if=amazon-thor.img of=/dev/block/platform/msm_sdcc.1/by-name/system bs=1M
dd if=boot.img of=/dev/block/platform/msm_sdcc.1/by-name/boot bs=1M
```


zhuowei:
@MoreRobustThanYou It looks like the kernel config is missing a line to enable the amazon ram console: Did you base your kernel config on the LineageOS defconfig? I think it's https://github.com/LineageOS/android_kernel_amazon_hdx-common/blob/cm-14.1/arch/arm/configs/msm8974-hdx_defconfig

aports/device/device-amazon-thor/90-android-touch-dev.rules
```diff
@@ -0,0 +1,8 @@
+# udev rules file
+# All device names can be read from weston's logfile
+
+# Touchscreen (use 'weston-calibrator' to calibrate)
+SUBSYSTEM=="input", ATTRS{name}=="sec_touchscreen", \
+ENV{ID_INPUT}="1", ENV{ID_INPUT_TOUCHSCREEN}="1" \
+ENV{WL_CALIBRATION}="0.954174 0.016888 -2.626083 -0.006031 1.011271 14.175903"
```
drebrez: Could you try if it still works after omitting the WL_CALIBRATION? It should not be necessary for capacitive touchscreens (yes, we still have it in lots of other aports, see #182):

Example: https://wiki.postmarketos.org/wiki/Screen_Calibration#weston

aports/device/device-amazon-thor/APKBUILD
```diff
@@ -0,0 +1,22 @@
+pkgname=device-amazon-thor
+pkgver=1
+pkgrel=0
+pkgdesc="Amazon KFHDX 7inch 2013"
+url="https://github.com/postmarketOS"
+arch="armhf"
```
drebrez: Please set `arch="noarch"` here, as we have for all the other device packages

aports/device/device-amazon-thor/APKBUILD
```diff
@@ -0,0 +1,22 @@
...
+license="MIT"
+depends="linux-amazon-thor"
+makedepends=""
+install="$pkgname.post-install"
+subpackages=""
+source="
+	deviceinfo
```
drebrez: please add `$install` to the source variable

aports/device/device-amazon-thor/deviceinfo
```diff
@@ -0,0 +1,34 @@
...
+deviceinfo_dev_touchscreen="/dev/input/event2"
+
+# Kernel related
+deviceinfo_kernel_cmdline="console=ttyHSL0,115200,n8 androidboot.hardware=qcom user_debug=31 msm_rtb.filter=0x37 ehci-hcd.park=3 buildvariant=userdebug"
+deviceinfo_flash_offset_kernel="00000000"
+deviceinfo_flash_pagesize="2048"
+deviceinfo_flash_offset_ramdisk="02000000"
+deviceinfo_flash_offset_second="00f00000"
+deviceinfo_flash_offset_tags="01e00000"
+deviceinfo_flash_offset_base="00000000"
```
drebrez: does it work without the `0x` prefix in the flash offset variables?

aports/device/linux-amazon-thor/APKBUILD
```diff
@@ -0,0 +1,107 @@
+# APKBUILD based on lg-mako aport. Changes:
+# - disabled module installation
+# - add !check !tracedeps
+# - package: just install zimage and kernel.release, because the kernel config
+#	does not generate modules or dtb files
+# - do not create -dev subpackage (makes no sense without module support)
+#
+# Kernel config based on: arch/arm/configs/lineageos_mako_defconfig
```
drebrez: please update this comment with the actual config you started

aports/device/linux-amazon-thor/APKBUILD
```diff
@@ -0,0 +1,107 @@
...
+# Changes:
+# - enable devtmpfs (needed for udev -> touch support in weston)
+# - disable ANDROID_PARANOID_NETWORK (removes network restrictions)
+
+_vendor=amazon
+_flavor=amazon-thor
+_hash="cm-14.1"
```
drebrez: please use an actual commit instead of the branch name, otherwise everytime they commit on that branch the package fails to compile

aports/device/linux-amazon-thor/APKBUILD
```diff
@@ -0,0 +1,107 @@
...
+_config="config-${_flavor}.armhf"
+
+pkgname=linux-${_flavor}
+pkgver=3.4.0
+case $pkgver in
+	*.*.*)  _kernver=${pkgver%.*};;
+	*.*) _kernver=$pkgver;;
+esac
+pkgrel=8
```
drebrez: Please change the pkgrel to version **0** since it will be closed in master and it will be the first version

MoreRobustThanYou:
Progress.... kinda https://pastebin.com/fEiTkszJ

zhuowei:
@MoreRobustThanYou would applying https://github.com/postmarketOS/pmbootstrap/blob/master/aports/device/linux-huawei-angler/kernel-use-the-gnu89-standard-explicitly.patch help?

MoreRobustThanYou:
@zhuowei thanks! that worked.

zhuowei:
@MoreRobustThanYou Does it flash/boot? Amy other error messages?

If it boots but doesn't work properly, look at deviceinfo of other SD800 devices; in particular, turn on device_msm_refresher if the screen doesn't work.

MoreRobustThanYou:
@zhuowei actually it's still compiling (fucking AMD APU)

aports/device/device-amazon-thor/APKBUILD
MoreRobustThanYou: done

MoreRobustThanYou:
REEEEEEEEEEEE https://pastebin.com/trxXEdX6

zhuowei:
@MoreRobustThanYou You could try removing the inline patch and using the other patch I linked.

MoreRobustThanYou:
@zhuowei see op for new details/patch logs

MoreRobustThanYou:
IT COMPILES NOW!

ollieparanoid:
Awesome work! Before we put it into `master`, we would need to know if it actually boots on your device (let us know if it does not an we'll help you, you know the pattern 😉 ). So what is the status regarding boot on the device?

MoreRobustThanYou:
wtf mkbootimg doesnt run

aports/device/device-amazon-thor/APKBUILD
PureTryOut: Actually, abuild complains if you do that, so don't. The files being in $install is enough

drebrez: @PureTryOut didn't know that, thanks for the info 👍

drebrez: @PureTryOut are you sure that it complains? also without adding it to the source variable, it doesn't get checksummed, see [here](https://github.com/postmarketOS/pmbootstrap/pull/823/commits/e6e3f6599350a2a6194b95657211caf6fa2f8a77#diff-09609db7cb2104c362f7a8c5571bdfe0R22)

ollieparanoid:
What exactly does not work? (Also could you please use more descriptive commit messages?)

MoreRobustThanYou:
@ollieparanoid mkbootimg simply doesn't run, and ok

MoreRobustThanYou:
boot.img is now generated. However, it's still not packaged into the recovery zip. And, recovery zip still returns TWRP Error 255

aports/device/device-amazon-thor/APKBUILD
PureTryOut: Yes it complains. It doesn't break, it just warns you that it shouldn't be there. I've tried it before, it works fine if you don't specify it in `$source`. It clearly doesn't need to be checksummed.

Check the log carefully when building a package which *does* have `$install` inside `$source`, it'll mention it.

drebrez: @PureTryOut what about adding `$pkgname.post-install` in `source`? it shouldn't complain then, but at least we have the checksum. no?

PureTryOut: Why do you want a checksum in the first place? Abuild doesn't need it, why would we?

drebrez: the `$pkgname.post-install` it's a script added as a separated file, in every APKBUILD we have the checksum of all the other files except the APKBUILD itself, why this file shouldn't be checksummed?

PureTryOut: I don't know, you'll have to ask the Alpine devs. I just know it isn't needed so I'm not sure why we shouldn't do it.

I guess it's something that doesn't actually get installed, unlike other stuff put in `$source`. It would be a reasonable explanation to me.

MoreRobustThanYou:
@ollieparanoid this is mostly merge ready at this point, but i'd like to try to get wifi and maybe hildon working.

PureTryOut:
If it boots and shows Weston, I'd say merge it. You can always get other stuff working later on.

MoreRobustThanYou:
@PureTryOut well yeah, it boots and shows weston

aports/device/device-amazon-thor/APKBUILD
ollieparanoid: @MoreRobustThanYou: sorry if this was confusing. Alpine's abuild, which we use to build the packages, clearly states: ["You should not have $install in source"](https://github.com/alpinelinux/abuild/blob/800e4a818764a30c4bd39822ada63b358ba9bb31/abuild.in#L181-L183). I think this changed over time, and that is why they added the warning in the first place. But that is what `abuild` says now, so I'm pretty sure that this is the right way to do it.

PR #821, opened 2017-10-24
pmbootstrap init: Wizard for new port device- and linux-packages

ollieparanoid:
First PR for #688 (new kernel porting workflow)
* adds `pmbootstrap aportgen device-*` and  `pmbootstrap aportgen linux-*`
* ask for confirmation when selecting a non-existing device
* generate the packages directly from init
* refactor aportgen code
* fixed some easy things in the linux- APKBUILD (more to come in follow-up PRs!):
  * `_repository` variable as @MartijnBraam suggested
  * removed some unused variables
  * restructured it to be more clear
  * automatically fill out `pkgname`, `pkgdesc`, `arch`, `_carch`, `_flavor`
  * use `zImage-dtb` if present
  * renamed `_hash` to `_commit`
  * reduced long based on ... comment at the top
  * don't require the user to accept defaults anymore (the `yes "" |` thing)

### How to test
* `pmbootstrap init`, confirm that you want a new device and look at the resulting device- and linux- packages
* `pmbootstrap aportgen device-...`
* `pmbootstrap aportgen linnux-...`

### How it looks like
```
[01:01:03] Target device (either an existing one, or a new one for porting).
[01:01:03] Available (31): asus-flo, asus-grouper, fairphone-fp2, htc-ace, htc-bravo, huawei-angler, huawei-y530, lg-d285, lg-d855, lg-dory, lg-hammerhead, lg-lenok, lg-mako, motorola-osprey, motorola-titan, nokia-rx51, oneplus-bacon, qemu-aarch64, qemu-amd64, qemu-vexpress, samsung-i9003, samsung-i9070, samsung-i9100, samsung-i9305, samsung-maguro, samsung-n7100, sony-amami, sony-aries, sony-castor-windy, sony-yuga, t2m-flame
[01:01:03] Device [samsung-i9100]: purism-librem5
[01:01:10] You are about to do a new device port for 'purism-librem5'.
[01:01:10] Continue? (y/n) [y]: y
[01:01:12] Device architecture (armhf/aarch64) [armhf]: aarch64
[01:01:16] Who produced the device (e.g. LG)?
[01:01:16] Manufacturer: Purism
[01:01:19] What is the official name (e.g. Google Nexus 5)?
[01:01:19] Name: Purism Librem 5
[01:01:24] Username [user]:
...
```

Resulting kernel APKBUILD (more changes to come in follow-up PRs, see #688 ):
```shell
# Kernel config based on: arch/arm64/configs/(CHANGEME!)

[...]

PR #825, opened 2017-10-25
Cleanup linux-lg-mako APKBUILD

drebrez:
Changed APKBUILD to be like the template in #821 with the difference to use `builddir` variable to build directly in the source folder and call `default_prepare` to apply patches


ollieparanoid:
I'll build this and test it on the device before merging (unless another `lg-mako` user could test it, that would be highly appreciated).
Thank you very much @drebrez !

PR #830, opened 2017-10-26
kconfig_check: disable CONFIG_PFT

ollieparanoid:
From https://wiki.postmarketos.org/wiki/Kernel_configuration#CONFIG_PFT:
> This caused builds to fail with a modern GCC at least once, and while it originally was meant as security feature by Qualcomm, "it is unsupported by Qualcomm, and opens up to a wide range of potential attack surfaces that has not been audited by anyone." [(source)](http://www.merproject.org/logs/%23sailfishos-porters/%23sailfishos-porters.2016-08-26.log.html#t2016-08-26T17:43:06)

It failed building with that option enabled for `amazon-thor`, see #823.

I have not rebuilt all kernels yet to test that compilation is still working, doing that would be highly appreciated. Otherwise I'll do it the following days.


zhuowei:
@ollieparanoid Seems Google agrees: see this AOSP commit disabling this for the Nexus 5X kernel: https://android.googlesource.com/kernel/msm/+/7872047076626eed9a897a2fca6466d8fc023c1b

ollieparanoid:
Thanks for compiling it!

ollieparanoid:
This caused builds to fail with a modern GCC at least once, and while it originally was meant as
security feature by Qualcomm, "it is unsupported by Qualcomm, and opens up to a wide range
of potential attack surfaces that has not been audited by anyone."

PR #832, opened 2017-10-26
Add architecture argument for the buildroot chroot

drebrez:
This allows to run the following commands:

`pmbootstrap chroot --buildroot x86_64`
`pmbootstrap chroot --buildroot aarch64`
`pmbootstrap chroot --buildroot armhf`
and 
`pmbootstrap chroot --buildroot device` which is also the default if no argument is passed. 

The `device` gets replace with the architecture of the device selected in `pmbootstrap init`, which is the same behavior as before.

Also added a simple test case that runs `uname -m` in every chroot and checks the output.


drebrez:
Why it fails in one builder and not in the other?

ollieparanoid:
Thanks making this and for adding a testcase \o/ I'm surprised that the coverage catches when we open another `./pmbootstrap.py` process. Nice!

Could you try to run this at the beginning in the Travis script section?
`git symbolic-ref refs/remotes/origin/HEAD refs/remotes/origin/master`
...and if that does not work, also `git fetch` afterwards.

ollieparanoid:
@drebrez found this related issue: https://github.com/travis-ci/travis-ci/issues/6069

PR #833, opened 2017-10-27
Some kernel APKBUILD cleanup

drebrez:
The `pkgver` of the `linux-samsun-tuna` package was wrong, so now it changes from `3.4.0` to `3.0.31`.
This might be a problem because now the version is lower than before, or this is already handled somehow by pmbootstrap?


aports/device/linux-samsung-tuna/APKBUILD
```diff
@@ -1,104 +1,59 @@
...
-_config="config-${_flavor}.armhf"
-
-pkgname=linux-samsung-tuna
-pkgver=3.4.0
-case $pkgver in
-	*.*.*)  _kernver=${pkgver%.*};;
-	*.*) _kernver=$pkgver;;
-esac
-pkgrel=7
+pkgname="linux-samsung-tuna"
```
ollieparanoid: The `pkgver` downgrade will not be handled by `pmbootstrap` (if we really need this, we could migrate the work folder like done for that username PR).

What would you think about renaming the package to `linux-samsung-maguro`? Then we would have all device-specific linux packages with the name of the device and also the pkgver downgrade would work without problems.

If we port another device in the future, that also needs the samsung_tuna repository, it will have another kernel config anyway, and I think it would be easier to make an extra package than extending this one to support multiple devices.

drebrez: @ollieparanoid I agree with the rename, I also keep doing confusion about that maguro-tuna naming. I will rename and test it again 👍

drebrez:
Renamed `linux-samsung-tuna` package to `linux-samsung-maguro`, tested on device again and everything works

PR #840, opened 2017-10-28
[WIP] bq Aquaris U (chaozu) support

JBBgameich:
The device and kernel package are currently build-tested only.
I did not yet flash the images, but I'll do as soon as I can. I'll remove the WIP label as soon as it boots and I have SSH access.
A code review would already be nice because I have no experience with APKGBUILDS yet.


aports/device/linux-bq-chaozu/APKBUILD
```diff
@@ -0,0 +1,95 @@
...
+	cd "$srcdir"/build
+	unset LDFLAGS
+	make ARCH="$_carch" CC="${CC:-gcc}" \
+		KBUILD_BUILD_VERSION="$((pkgrel + 1 ))-Alpine" CONFIG_NO_ERROR_ON_MISMATCH=y
+	echo "--[ Build modules ]--"
+	make ARCH="$_carch" CC="${CC:-gcc}" \
+		KBUILD_BUILD_VERSION="$((pkgrel + 1 ))-Alpine" CONFIG_NO_ERROR_ON_MISMATCH=y \
+		modules
+
+	dtbTool -s 2048 -p "scripts/dtc/" -o "arch/arm/boot/dt.img" "arch/arm/boot/"
```
ollieparanoid: You're not using the `dt.img` you create here in the `package()` step. (probably copy pasted from another linux device APKBUILD, where it was necessary, which is totally fine of course - we'll add a wizard providing a good template soon).

So in case the `boot.img` you have created is working, and you don't need the `dt.img` file, you could remove this line and `dtbtool` from `makedepends`.

JBBgameich: Okay, I'll look into that tomorrow, thanks for the review!

JBBgameich:
The kernel I'm using is for all bq msm8938 devices (Aquaris U, U Plus, U Lite, V, V Plus etc), should I name the package linux-bq-chaozu or linux-bq-msm8937? Also can someone build the same kernel source with different defconfigs for this different devices if needed?

ollieparanoid:
`linux-bq-chaozu` would be the right name, and if another device needs the same sources, we would make another package. it's not worth adding support for multiple configs etc. for these kernel forks.

We want to make multiple devices able to use the same (mainline) kernel in the `linux-postmarketos-mainline` (bleeding edge) and `linux-postmarketos-stable` though, and have already started that (right now N900 and qemu only, the first Android with castor is close to be added.)

PureTryOut:
@JBBgameich how far did you get with the port? Did you manage to get SSH access?

JBBgameich:
I rebased it on the LineageOS kernel source locally using the new porting-wizard because the device is now officially supported by LineageOS. Unfortunately the kernel doesn't compile anymore (I will probably be able to work on that tomorrow). Sorry for the long inactivity on this PR :/

JBBgameich:
Currently the kernel doesn't compile:
```
In file included from drivers/net/ethernet/msm/rndis_ipa_trace.h:81:0,
                 from drivers/net/ethernet/msm/rndis_ipa.c:32:
include/trace/define_trace.h:83:43: fatal error: ./rndis_ipa_trace.h: No such file or directory
 #include TRACE_INCLUDE(TRACE_INCLUDE_FILE)
                                           ^
compilation terminated.
```
Does anyone know this error?

ollieparanoid:
Yes, there are some patches in `aports` containing `rndis_ipa_trace`. See `aports/device/linux-motorola-potter/03_fix_rndis_ipa_trace_header_not_found.patch` for example. :+1:

JBBgameich:
Thank you! I'll try compiling it with the patch tomorrow again.

ollieparanoid:
You're welcome, thanks for updating it! From reading the thread again, it looks like the only thing missing is confirming that the newer kernel boots as well, right?

And of course Travis CI should run through, currently it is failing [here](https://travis-ci.org/postmarketOS/pmbootstrap/builds/369289539#L653):
```
bq-chaozu: not in the wiki yet.
*** Wiki check failed!
Thank you for porting postmarketOS to a new device! \o/
Now it's time to add some documentation:
1) Create a device specific wiki page as described here:
   <https://wiki.postmarketos.org/wiki/Help:Device_Page>
2) Add your device to the overview matrix:
   <https://wiki.postmarketos.org/wiki/Devices>
3) Run these tests again with an empty commit in your PR:
   $ git commit --allow-empty -m 'run tests again'
Please take the time to do these steps. It will make your
precious porting efforts visible for others, and allow them
not only to use what you have created, but also to build upon
it more easily. Many times one person did a port with basic
functionallity, and then someone else jumped in and
contributed major new features.
```

JBBgameich:
I already created a wiki page, do I need to rename it so it's found by the travis tests?

ollieparanoid:
The full codename (`bq-chaozu`) needs to be in the devices table's codename column, and it needs to be moved to the upper table for booting devices to make it pass Travis. But that only makes sense after making sure that it boots.

JBBgameich:
Unfortunately the device currently just bootloops...

ollieparanoid:
Nice that the maximum attention hook is working! :+1: 
The IP of the device should be `172.16.42.1`, does SSH (or telnet after installing [debug-shell](https://postmarketos.org/debug-shell)) work, or which message do you get on failure?

See also: https://wiki.postmarketos.org/wiki/Troubleshooting:boot#Kernel_boots_without_network_interface

PR #841, opened 2017-10-28
Add argument to zap older versions of packages from all chroots

drebrez:
This adds argument `-o`, `--old-bins` to the `pmbootstrap zap` command which will simply call `apk -v cache clean` in all the chroots.
It's very useful to run it once in a while to free up some disk space.

Also moved the mismatch clean before zapping the chroots.
This avoids re-creating the chroot just to perform the repo index after the clean.


drebrez:
It deletes the old version of packages in folders: `$work/cache_apk_aarch64`, `$work/cache_apk_armhf` and `$work/cache_apk_x86_64`

ollieparanoid:
I couldn't make it delete anything so far, could you provide me with testing steps? This is how I have tested:
```shell
$ pmbootstrap build hello-world
(edit aports/main/hello-world/APKBUILD and increase pkgrel)
$ pmbootstrap build hello-world
$ pmbootstrap zap -o
$ ls ~/.local/var/pmbootstrap/packages/x86_64/hello-world-1-r*
/home/user/.local/var/pmbootstrap/packages/x86_64/hello-world-1-r3.apk
/home/user/.local/var/pmbootstrap/packages/x86_64/hello-world-1-r4.apk
```

Also it seems like the `armhf` and `aarch64` chroot gets set up whenever calling `pmbootstrap zap -o`, which should not be necessary as I understand it.

ollieparanoid:
Alright, I have tested it on another PC, which actually had old packages installed, and it is working! :+1:

pmb/chroot/zap.py
```diff
@@ -33,11 +34,26 @@ def zap(args, confirm=True, packages=False, http=False, mismatch_bins=False, dis
...
+    :arg old_bins: Clean out older versions of packages from all chroots
     :arg distfiles: Clear the downloaded files cache
 
     NOTE: This function gets called in pmb/config/init.py, with only args.work
     and args.device set!
     """
+
+    # Delete packages with a different version compared to aports, then re-index
+    if mismatch_bins and os.path.exists(args.work + "/packages/"):
+        binaries(args)
```
ollieparanoid: I know, that this PR didn't add the `-m` option, but what do you think about adding a confirmation for both `-m` and `-o`? That would be more consistent in my opinion, as we also ask for confirmation with `-p` and `-hc`.

pmb/parse/arguments.py
```diff
@@ -174,6 +174,8 @@ def arguments():
     zap.add_argument("-m", "--mismatch-bins", action="store_true", help="also delete"
                      " binary packages that are newer than the corresponding"
                      " package in aports")
+    zap.add_argument("-o", "--old-bins", action="store_true", help="also delete"
```
ollieparanoid: How about making clear, that this operates on the downloaded packages? (see my "testing comment")

drebrez:
Steps to reproduce:

- `pmbootstrap chroot`
- `echo 'http://dl-cdn.alpinelinux.org/alpine/v3.2/main' >> /etc/apk/repositories`
- `apk update`
- `apk add vim=7.4.712-r2`
- `apk del vim`
- `apk add vim`

at this point if you look in the `$work/cache_apk_x86_64` folder you should have both vim packages, version 7.4.x and latest version (8.x) 

running `pmbootstrap zap --old-bins` you should see in the logs:
```
(006027) [15:04:06] (native) % apk -v cache clean
deleting vim-7.4.712-r2.e04bc396.apk
```

this is the simple way I found to replicate it, by adding an old repository, since alpine only keep the last version of the package. But after some time you keep creating chroots and each time it downloads newer versions of packages, the cache starts growing and use various MB.

I will add the confirmation question as you suggested 👍

pmb/parse/arguments.py
drebrez: @ollieparanoid any suggestion of a better description?

pmb/chroot/zap.py
```diff
@@ -33,11 +34,26 @@ def zap(args, confirm=True, packages=False, http=False, mismatch_bins=False, dis
...
+        pmb.build.other.index_repo(args)
+
+    # Delete old packages
+    if old_bins:
+        pmb.chroot.root(args, ["apk", "-v", "cache", "clean"])
```
drebrez: add a check for the existence of `$work/cache_apk_<architecture>` folder, if it doesn't exist there is no need to create the chroot.

pmb/chroot/zap.py
```diff
@@ -33,11 +34,26 @@ def zap(args, confirm=True, packages=False, http=False, mismatch_bins=False, dis
...
     and args.device set!
     """
+
+    # Delete packages with a different version compared to aports, then re-index
+    if mismatch_bins and os.path.exists(args.work + "/packages/"):
+        binaries(args)
+        pmb.build.other.index_repo(args)
+
+    # Delete old packages
+    if old_bins:
```
drebrez: move in separate function like for mismatch_bins

pmb/parse/arguments.py
ollieparanoid: How about:
`-o`, `--old-upstream-packages` `"also delete outdated binary packages downloaded from mirrors (e.g. from Alpine)"`

pmb/chroot/zap.py
drebrez: @ollieparanoid do you think it's enough one single question or it should ask confirmation for each architecture?

drebrez:
@ollieparanoid I've addressed all the comment, can you please take a look again?

PR #843, opened 2017-10-29
make motorola-osprey use msm-fb-refresher

MoreRobustThanYou:
it was still using a loop of ```cat /sys/class/graphics/fb0/modes > /sys/class/graphics/fb0/mode```


ata2001:
@MoreRobustThanYou osprey is not using a loop to refresh the screen. That command is only issued once after the graphical environment launched.

drebrez:
@ata2001, @MoreRobustThanYou, if the command have to be executed only once, we may consider running `msm-fb-refresher` only once (without `--loop` argument)

ata2001:
@drebrez It would make sense. If we are going to do it that way, sony-amami also needs fixing.

PR #844, opened 2017-10-29
qemu-user-static-repack: fix download url

### LINKED ISSUE #842: pmbootstrap unusable due to https://ftp.us.debian.org problems
Hello everyone...
[10:33:39] (native) build x86_64/qemu-user-static-repack-2.8-r8.apk
...now fails, because
https://ftp.us.debian.org/debian/pool/main/q/qemu/qemu-user-static_2.8+dfsg-6+
deb9u2_amd64.deb is not available.
...making pmbootstrap unusable. Should happen to everyone...?

It seems there's wrong certificate on  https://ftp.us.debian.org ?

JBBgameich:
fixes #842 (hopefully)

btw, can someone tell me why PostmarketOS repacks this from a debian package and doesn't build it from source?


ollieparanoid:
> can someone tell me why PostmarketOS repacks this from a debian package and doesn't build it from source?

Because it is faster (not everyone has a fast CPU for quick compilations) and no one bothered with doing this properly instead. In my opinion, the proper way would be upstreaming `qemu-user-static` to Alpine, and installing the binary package from Alpine in `pmbootstrap` afterwards.

Thank you very much for the fix, tested and working!

PR #845, opened 2017-10-29
Adds amazon-thor firmware

MoreRobustThanYou:
https://github.com/LineageOS/android_device_amazon_thor/blob/cm-14.1/proprietary-files.txt#L380-L409


MoreRobustThanYou:
I used the same branch for this as the initial thor porting, so that explains all the commits

PureTryOut:
As follow up from your previous PR, could you make a [wiki page](https://wiki.postmarketos.org) for your device? Look at the other device pages for some examples.

MoreRobustThanYou:
@PureTryOut https://wiki.postmarketos.org/wiki/Amazon_Kindle_Fire_HDX

MoreRobustThanYou:
I need help on getting the kernel to use the firmware. What would I need to patch?

PureTryOut:
Good work on the wiki page! Please also add it to the supported devices list [here](https://wiki.postmarketos.org/wiki/Devices) so people can actually *find* the device 😉

ollieparanoid:
@MoreRobustThanYou as written in the chat (I haven't read the backlog yet, but maybe someone finds this question here):
* Change the path to the firmware files in the kernel, so it loads the right ones (if that is necessary) ([Example](https://github.com/postmarketOS/pmbootstrap/blob/6627599cf031c11a9df44dc93704be015d0484fb/aports/device/linux-samsung-i9070/04_change_dhd_firmware_dir.patch))
* You can grep the kernel source dir for possibly hardcoded paths (which should appear in dmesg, when the firmware loading fails afaik).
* [Here is a patching guide](https://wiki.postmarketos.org/wiki/How_to_create_a_patch_when_packaging_software)

ollieparanoid:
@MoreRobustThanYou: why did you close this one?

MoreRobustThanYou:
@ollieparanoid i'll bundle it in with my mainline port

PR #849, opened 2017-11-01
[samsung-maguro] Add firmware-samsung-maguro package

drebrez:
This PR adds the `firmware-samsung-maguro` package and remove the alpine `linux-firmware` package.

`firmware-samsung-maguro` installed size: 240 KiB
`linux-firmware` installed size: 202.14 MiB

It also changes the location of the installed firmware files, now it uses the `/lib/firmware/postmarketos/` folder.

I tested it on the device and wifi works. Also updated the [device wiki page](https://wiki.postmarketos.org/wiki/Google_Galaxy_Nexus_GSM_(samsung-maguro)#Connect_to_WiFi) and added the steps to connect to a wifi network.


drebrez:
@ollieparanoid I checked the AOSP github mirror that zhuowei created and it actually contains the `brcmfmac4330-sdio.bin` file, but comparing the files they are completely different. 
I can try if they work on the device.
In case it works, how do you suggest to package it? leave the `firmware-samsung-maguro` package with the calibration file (from what I understood it's different for each device) or add an `install_if="device-samsung-maguro"` condition in the `firmware-aosp-broadcom-wlan` package?
Btw, the `firmware-aosp-broadcom-wlan` installed size is already 11,05 MiB, does it make sense to install it? If you look at the repo [here](https://github.com/zhuowei/android_hardware_broadcom_wlan/tree/master/bcmdhd/firmware) you notice various folders for each version, does it make sense to have all those in sub-packages maybe?

ollieparanoid:
> but comparing the files they are completely different.

The one, that Alpine packages, is an inofficial build for the Raspberry Pi community (https://github.com/alpinelinux/aports/commit/46edfde10d8e5baec6bfaef2fedc311108465d82), which fixes broadpwn: https://github.com/postmarketOS/pmbootstrap/issues/513#issuecomment-327266718. That explains why it is completely different. We actually use the latest official one right now, which is *not* patched with the plan to patch it with nexmon #592.

Actually it would make a lot of sense, if we also used the inofficial Pi version until we got the nexmon thing properly going (do you want to make a PR for that, if you see it the same way?).

I would go with leaving the calibration file in the device specific firmware package. Otherwise we would have to update the shared package (-> all devices need to re-download it) whenever we add a new device.

Splitting the package up in subpackages sounds like a good idea to me!

drebrez:
@ollieparanoid I didn't understand what you suggest to package.

1. from alpine's `linux-firmware` package?
2. from `https://dev.alpinelinux.org/archive/linux-firmware/` directly?
3. from `https://github.com/RPi-Distro/firmware-nonfree/tree/master/brcm80211/brcm`?
4. from `https://android.googlesource.com/platform/hardware/broadcom/wlan/+/android-8.0.0_r12/bcmdhd/firmware/`?
5. other?

ollieparanoid:
Okay, sorry for the confusion. I suggest we use `firmware-aosp-broadcom-wlan` for this PR.

Optionally, for a new PR: update `firmware-aosp-broadcom-wlan` to use `https://github.com/RPi-Distro/firmware-nonfree/tree/master/brcm80211/brcm`.

drebrez:
I've updated the package to depend on `firmware-aosp-broadcom-wlan` instead of downloading `linux-firmware` and extract files from there.
Tested on device and it works.

PR #852, opened 2017-11-02
Fix #831: Disallow work dir being inside the pmbootstrap dir

### LINKED ISSUE #831: Check if work folder is inside pmbootstrap folder (not supported!)
I'm making efforts to port pmOS over to my old Droid 4, however when pmbootstrap is preparing to build the kernel it fails after building gcc-armhf with this message:
```
ERROR: [Errno 13] Permission denied: '/home/dominic/postmarketOS_experiments/pmbootstrap/ls/chroot_native/proc/1/cwd'
```

upon investigation of the log I see this:
```
ls/packages/x86_64/g++-armhf-6.4.0-r5.apk
ls/packages/x86_64/gcc-armhf-6.4.0-r5.apk
ls/packages/x86_64/musl-armhf-1.1.17-r1.apk
ls/packages/x86_64/musl-dev-armhf-1.1.17-r1.apk
ls/version
(005356) [17:27:02] ERROR: [Errno 13] Permission denied: '/home/dominic/postmarketOS_experiments/pmbootstrap/ls/chroot_native/proc/1/cwd'
```

Interesting... let's have a look at that file
```
ls: cannot read symbolic link ls/chroot_native/proc/1/cwd: Permission denied
lrwxrwxrwx 1 root root 0 Oct 26 17:26 ls/chroot_native/proc/1/cwd
```
You can't see it, but the file path was red, indicating a permission denied error.
I listed out the rest of the directory and saw that cwd, exe, and root were all broken.

I've done a zap, same problem. I've reinitialized, same problem. I'm actually unable to delete these files until I reboot my system and change the ownership of them, even running directly as root.

I've tried just about everything I can think of, this is a huge blocker and I haven't seen anything about it in the issues here, so my first assumption is downlevel packages used by debian, but I have no idea where to start looking.

Has anyone successfully completed the build process on Debian 8?

I've gone ahead and attached the log to this issue, in case anyone wants to try and dig through there to glean any information.
[log.txt](https://github.com/postmarketOS/pmbootstrap/files/1420148/log.txt)

ollieparanoid:
The timestamp based rebuilds folder tries to scan the `pmbootstrap` folder, and it runs into problems such as trying to scan the mounted `proc` folder with user rights, so we better disallow it altogether.

### How to test
* Run `pmbootstrap init`
* Try to put the work folder inside the folder where your `pmbootstrap` source is in (or any subfolder)


pmb/config/init.py
```diff
@@ -42,6 +42,17 @@ def ask_for_work_path(args):
         try:
             ret = os.path.expanduser(pmb.helpers.cli.ask(
                 args, "Work path", None, args.work, False))
+            ret = os.path.realpath(ret)
+
+            # Work must not be inside the pmbootstrap path
+            pmbootstrap_path = os.path.realpath(os.path.dirname(
+                                                os.path.abspath(__file__)) +
+                                                "/../../") + "/"
```
drebrez: seems like in `pmb/config/__init__.py` there is `pmb_src` which might already be the `pmbootstrap_path`, it can be reused here

ollieparanoid:
Good catch @drebrez, you're right, it makes more sense to use `pmb.config.pmb_src` there!
It's adjusted now. And it also disallows specifying the exact pmbootstrap folder, which would have passed before. Could you take another look at it?

ollieparanoid:
Thanks for testing! \o/

PR #853, opened 2017-11-02
libsamsung-ipc, linux-postmarketos-mainline: Fix invalid pkgrels

ollieparanoid:
We have more invalid `pkgver`s, but they are all in the `kde` folder and @PureTryOut fixed them already in the upcoming PR #440.

Both packages still compile after the change.


ollieparanoid:
Thanks for the review!

PR #854, opened 2017-11-02
Fix #839: Check pkgver after parsing APKBUILD

### LINKED ISSUE #839: pmbootstrap should verify APKBUILD versions when parsing APKBUILDs
We already have a validation function, which we would only need to call after parsing an APKBUILD:

https://github.com/postmarketOS/pmbootstrap/blob/11e44fe44a2c7c7aad6a66acd74138c70ec8c94a/pmb/parse/version.py#L198-L213

To keep it userfriendly, I suggest extending [APKBUILD Reference#pkgver](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#pkgver) in Alpine's wiki to describe valid version strings (especially which suffixes are allowed) and linking to that page in the error message.

The reason this is marked as bug is, that we assume valid pkgvers at other points in the code, and when the version currently contains `-r`, it will confuse pmbootstrap. @ata2001 reported:
> it did not crash, but rebuilt the package every time, then installed the old version 

We should also add a Travis CI testcase, that parses all APKBUILDs (in order to find broken pkgvers and problems with the parser). @MartijnBraam's PR #351  already does that afaik (so it might be worth getting that in first).

ollieparanoid:
Depends on #440 and #853, because they fix the existing wrong `pkgver`s. Reviews can be done already if someone wants to.

Also changed (made testing easier and is useful in general):
* allow to specify multiple packages to `pmbootstrap parse_apkbuild`
* specifying no package will parse all packages (like kconfig_check)
* JSON output is sorted in `parse_apkbuild`


ollieparanoid:
This PR has been rebased and is ready to be merged. A review would be highly appreciated.

drebrez:
@ollieparanoid how much complicated is to rename `parse_apkbuild` to `apkbuild_parse` to follow the same naming as `kconfig_check` and `bootimg_analyze`?

ollieparanoid:
Not at all. If you prefer, we could also do it as part of this PR - do you want to add a commit that renames them?

ollieparanoid:
@drebrez: rebased and implemented `parse_apk*` -> `apk*_parse`. Could you review it when you have time?

ollieparanoid:
Thanks for testing! The `*` gets interpreted in the `glob` when looking for the `APKBUILD` file. I've added a commit that detects this as invalid `pkgname` now.

Issue #848, opened 2017-10-31
OnePlus X (oneplus-onyx) kernel won't boot with postmarketOS compiler


magmastonealex:
I'm well into porting postmarketOS to the OnePlus X, but have run into a blocking issue: kernels built with the postmarketOS toolchain won't boot.

I thought I'd create this issue to solicit some ideas from the community,  and record my work in case this issue comes up with another device.

I diagnosed this as follows:

> 
> My initial symptom was that after flashing or `fastboot boot`, the device would reboot after a few seconds, never leaving the manufacturer splash screen, and return back to fastboot mode.
> 
> _(Note that this is different than a kernel panic - that results in a black screen and white LED, and the device switching to some kind of mass storage mode)._
> 
> I found that the LineageOS kernel did not exhibit this behaviour, even with the postmarketOS initramfs.
> 
> I narrowed it down:
> 
> \
> My first thought was device tree files. Both the LineageOS & postmarketOS build processes produce the same list of 10 dtb files, all are bit-for-bit identical.
> 
> I concatenated them on to the end of the lineageOS zImage file (the one made without the dtbs already added), and it boots!
> 
> I then concatenated them (same order) on to the end of the postmarketOS zImage, and that one doesn't boot, just waits a couple of seconds then reboots.
> 
> \
> Next, kernel configuration. I copied the configuration being used for postmarketOS to the lineage source tree and compiled it myself. The resulting kernel also booted (in fact, I was able to get weston & touchscreen working well).
> 
> \
> Both of these kernels are being built on the exact same git commit.
> 
> \
> At this point, the only differences between the two kernels are the compilers used to build them.
> 
> To confirm this, I built the LineageOS kernel out of the full lineage build environment, with the `arm-linux-androideabi` toolchain. This kernel boots.
> 
> I have seen a kernel which claims to be built using gcc 6 and work on the OnePlus X (blu_spark), and confirmed that it boots fine. But, when building from source, using the postmarketOS toolchain, I again run into the rebooting issue

At this point, I highly suspect the issue is something to do with the postmarketOS build tools.

In an attempt to resolve this, I tried to package the prebuilt `arm-linux-androideabi` toolchain for postmarketos, but unfortunately it was built with `glibc`, and it can't run in the alpine environment.

My next step is to try and convince the source for that toolchain to compile under the chroot, though I don't expect to have much success.

magmastonealex:
Interesting development: I finally got usable ram_console logs, and found this:
```
[...]

PR #856, opened 2017-11-02
Add new device oneplus-onyx (Oneplus X)

magmastonealex:
The kernel compiles & boots, and USB networking is functional but has issues with the frame buffer.

More details about the various kernel issues are [here](https://github.com/postmarketOS/pmbootstrap/issues/848).


aports/device/device-oneplus-onyx/APKBUILD
```diff
@@ -0,0 +1,34 @@
+pkgname=device-oneplus-onyx
+pkgver=1
+pkgrel=8
```
drebrez: Please change the pkgrel to version **0** since it will be closed in master and it will be the first version

aports/device/device-oneplus-onyx/weston.ini
```diff
@@ -0,0 +1,9 @@
+# XWayland seems to be broken for this device.
+# NOTE: This information might be outdated! simply delete /etc/xdg/weston/weston.ini
+# after installation to check if it is still broken, and please report if it
+# isn't broken anymore (or even better: make a pull request, that removes this
+# config file).
+
+[core]
+xwayland=true
```
drebrez: the custom `weston.ini` is actually the same as the default (see [here](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/main/postmarketos-ui-weston/weston.ini.default)).
You can actually delete the file and also remove the subpackage in the APKBUILD.

magmastonealex:
@drebrez Thanks for the comments. I've addressed what you've brought up.

Issue #855, opened 2017-11-02
Add support for custom fastboot args / vendor ID


MoreRobustThanYou:
For example, for fastboot to work on ```amazon-thor```, it requires that you add ```-i 0x1949``` after fastboot for it to work

perhaps add ```deviceinfo_fastboot_args```?

drebrez:
Maybe something more specific like `deviceinfo_flash_fastboot_vendor_id`

MoreRobustThanYou:
yeah

PR #857, opened 2017-11-03
Add "flash_fastboot_vendor_id" deviceinfo variable for fastboot flash method

drebrez:
This adds the possibility to specify the `-i` vendor id parameter of the `fastboot` command used by the flasher.
I added the `-i` argument for every command even if it's an empty string, becase `fastboot` ignores it and works like if it's not specified.

Close: #855


drebrez:
@ollieparanoid I'm fine with your change, you already did something similar in another PR of mine, next time I'll try to remember to use that notation 👍 
I asked @MoreRobustThanYou to test it, since he reported the issue with his device, but I also tested with my device at it worked both with and w/o that variable defined.
So for me it can be merged 😉

PR #859, opened 2017-11-03
Adds mkdtimg package

MoreRobustThanYou:


aports/main/mkdtimg/APKBUILD
```diff
@@ -0,0 +1,21 @@
+pkgname=mkdtimg
+pkgver=1
+pkgrel=0
+pkgdesc="dt.img creation tool for QCDT devices"
+url="https://github.com/LineageOS/android_system_core/blob/cm-14.1/mkbootimg/mkbootimg"
```
ollieparanoid: Looks like the real URL where the script was pulled out from is:
https://github.com/96boards-hikey/tools-images-hikey960

aports/main/mkdtimg/APKBUILD
```diff
@@ -0,0 +1,21 @@
+pkgname=mkdtimg
+pkgver=1
```
ollieparanoid: Since this program does not have proper releases, I recommend setting the `pkgver` to the day where the commit was made:
`pkgver=20170406`

(Much better would be an official repository with proper releases, but I've also searched and there seems to be none - this repository seems to be the next best thing.)

aports/main/mkdtimg/APKBUILD
```diff
@@ -0,0 +1,21 @@
...
+pkgrel=0
+pkgdesc="dt.img creation tool for QCDT devices"
+url="https://github.com/LineageOS/android_system_core/blob/cm-14.1/mkbootimg/mkbootimg"
+arch="noarch"
+license="APACHE2"
+depends="python3"
+subpackages=""
+source="$pkgname::https://raw.githubusercontent.com/96boards-hikey/tools-images-hikey960/6494008425d10193b037ea353fbdadfa3581299a/build-from-source/mkdtimg"
```
ollieparanoid: Please put the commit hash in an extra variable and use it inside the source URL and inside the destination file name. That way, when we update it, we won't have a conflict in the cached `distfiles` (two files with the same name but different content):
```shell
_commit=6494008425d10193b037ea353fbdadfa3581299a
source="$pkgname-$_commit.py::https://raw.githubusercontent.com/96boards-hikey/tools-images-hikey960/${_commit}/build-from-source/mkdtimg"
```

MoreRobustThanYou:
@ollieparanoid handled

ollieparanoid:
Thanks for the quick response. But please really test if it works - and not just if Travis runs through (it does not because of the checksums), but if the built program is still working.

MoreRobustThanYou:
@ollieparanoid well, tbh, it should work if the mkbootimg APKBUILD works, since this is just a modified mkbootimg APKBUILD

ollieparanoid:
Here's what happens when building `dtimg` now:
```shell
>>> mkdtimg: abuild 3.1.0-r1
>>> mkdtimg: Checking sanity of /home/pmos/build/APKBUILD...
>>> WARNING: mkdtimg: No maintainer
>>> mkdtimg: Cleaning temporary build dirs...
>>> mkdtimg: Fetching mkdtimg-6494008425d10193b037ea353fbdadfa3581299a.py::https://raw.githubusercontent.com/96boards-hikey/tools-images-hikey960/6494008425d10193b037ea3$
3fbdadfa3581299a/build-from-source/mkdtimg
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  4569  100  4569    0     0   4569      0  0:00:01 --:--:--  0:00:01 10479
>>> mkdtimg: Fetching mkdtimg-6494008425d10193b037ea353fbdadfa3581299a.py::https://raw.githubusercontent.com/96boards-hikey/tools-images-hikey960/6494008425d10193b037ea3$
3fbdadfa3581299a/build-from-source/mkdtimg
>>> mkdtimg: Checking sha512sums...
mkdtimg-6494008425d10193b037ea353fbdadfa3581299a.py: OK
sed: can't read /home/pmos/build/src/mkdtimg: No such file or directory
>>> ERROR: mkdtimg: all failed
```

I didn't expect this either, but it comes from the (necesssary!) change in the output file name.

So I know that all the testing seems unnecessary sometimes and it's annoying. But I think it is important nevertheless. We can really save ourselves from shipping broken stuff if we test properly, and in the end, fixing all the stuff that broke from one PR that we didn't really test takes even longer than if we did it properly in the first place and it's a bad user experience.

So could you please fix it, and test it (i.e. run the program after installing it)?

MoreRobustThanYou:
@z3ntu dtbTool takes a folder full of DTBs, mkdtimg uses a single DTB file. I had to use mkdtimg at one point, so I want to package it for those who also need to use it at some point.

PR #860, opened 2017-11-03
Update linux-postmarketos-mainline to rc7

MoreRobustThanYou:


ollieparanoid:
Travis fails, because the checksum have not been updated. Does the updated kernel compile for you?
Thanks!

MoreRobustThanYou:
@ollieparanoid actually can you include this in your pkgrel fix?

ollieparanoid:
I can do that, but #853 will only get merged after a review. If you have the time, could you review it? Reviewing would involve looking at the code changes, and building the packages.

PR #861, opened 2017-11-03
Close #453: Support mesa-dri-virtio in Qemu

### LINKED ISSUE #453: Improve Qemu video performance

ollieparanoid:
The mesa driver, which ends up in the installation image, needs to be known
before the installation is done (in other words: when running the qemu action,
it is to late as the image has already been generated). That's why one can
choose the Qemu mesa driver in `pmbootstrap init` now:

```
Device [qemu-amd64]:
Which mesa driver do you prefer for your Qemu device? Only select something other
than the default if you are having graphical problems (such as glitches).
Mesa driver (dri-swrast/dri-virtio) [dri-virtio]:
```

It is still possible to select `dri-swrast`, because `dri-virtio` may not work
in all cases, and that way we could easily debug it or experiment with other
mesa drivers (e.g. the "vmware" one, which is supported by mesa and Qemu).

Other changes:
* `pmbootstrap qemu` accepts a `--display` variable now, which passes the value
  directly to `qemu`'s `display` option. It defaults to `sdl,gl=on` (@PureTryOut
  reported that to work best with plasma mobile on his PC). `--display` and
  `--spice` (which is still working) are mutually exclusive.
* Removed obsolete telnet port pass-through: We only use the debug telnet port
  since osk-sdl has been merged.
* Add show-cursor to the Qemu command line, so it shows a cursor in X11
* Refactored the spice code (`command_spice` only returns the spice command,
  because it has all necessary information already) and the spice port can be
  specified on the commandline now (previously it was hardcoded in one place and
  then always looked up from there).
* Start comments with capital letters.
* Keep the log on the screen a bit shorter (e.g. Qemu command is written to the
  "pmbootstrap log" anyway, so there's no need to display it again).


PureTryOut:
Huh, don't you need to enable Virtio support in the kernel? I've always set `CONFIG_DRM_VIRTIO_GPU=m`, `CONFIG_VIRTIO_PCI=m`, and `CONFIG_VIRTIO_BALLOON=m` .

ollieparanoid:
You are right, I've missed that. It is enabled now and the kernels still compile. So, when you have time, could you test if the virtio stuff works with this PR and make plasma usable?

PureTryOut:
Interestingly enough, Plasma reports it's still using the llvmpipe driver, but peformance suggests otherwise. However this might be because the Qemu launch command is missing `-vga virtio`.

ollieparanoid:
I am not sure if you mean with that, that we should add it or not. So should we add `-vga virtio` to the `qemu` command line, in case the driver is the `dri-virtio`?

PureTryOut:
Yes, we should add it.

ollieparanoid:
OK, rebased and setting that option now, when `dri-virtio` is selected.

PR #862, opened 2017-11-03
[osk-sdl] Update to 0.51

craftyguy:
Update for osk-sdl. Changelog is in the osk-sdl repo under the relevant tagged release.

Please test this on devices that currently work with osk-sdl!!!


ollieparanoid:
Unlocking works the same on the `i9100`, no regression there 👍

ollieparanoid:
It is also working the same on the `lg-mako`. I think that and you testing it on the N900 makes it good enough for merging, what do you think?

craftyguy:
Sounds good to me. :ship: it!

PR #867, opened 2017-11-04
Clean up updated repo list after zapping chroots

### LINKED ISSUE #818: pmbootstrap build --strict sometimes fails to initialize the chroot
A similar error message to:
> `ERROR: Command failed: % sudo /home/kyle/PmOS/Work/apk.static -U --root /home/kyle/PmOS/Work/chroot_native --cache-dir /home/kyle/PmOS/Work/cache_apk_x86_64 --initdb --arch x86_64 add alpine-base`

@PureTryOut wrote in https://github.com/postmarketOS/pmbootstrap/issues/816#issuecomment-338899329:
> I have this error quite often, for me it's always resolved by just re-running the command.
>
> I'm not exactly sure how to reproduce it, but I think for example after building a package (`build --force --strict`) right after having run an install (`install --no-fde`).

I wrote in response:
> @PureTryOut: that sounds like a different issue related to `--strict`, I've also had that once during development of that feature and I thought I fixed it.

craftyguy:
Fixes #818


ollieparanoid:
Awesome that you made a fix for this! 🎉 
Does anybody know how to reproduce the original error message, so we can test if it went away?

craftyguy:
yea, I was easily reproducing it by calling this twice: `pmbootstrap build --arch armhf --force --strict osk-sdl`

ollieparanoid:
Thanks for the instructions! I've tested with:
```
pmbootstrap build --arch armhf --force --strict hello-world
pmbootstrap build --arch armhf --force --strict hello-world
```

(Without this PR it fails for the second one, and with this PR it does not.)

Issue #869, opened 2017-11-04
Root partition not resized automatically on non-x86_64 Qemu


PureTryOut:
Even though pmbootstrap reports otherwise, `--image-size <whatever>` doesn't actually do anything when running Qemu on a non-x86_64 architecture (in my testing, armhf).

If I manually resize the image using `truncate +1G qemu-vexpress.img`, pmbootstrap does however complain when not setting `--image-size` to the correct value.

```
ERROR: The system image size must be 3072M or greater
```

But when booting the image, the root partition doesn't get resized to use up all the new space.

drebrez:
The `--image-size` argument works as expected. The real issue is that the partition is not resized. 
This is due the way we use to detect it, see [here](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/main/postmarketos-mkinitfs/init_functions.sh#L146).
In the `qemu-vexpress` the root partition is `/dev/mmcblk0p2`, but if we change the resize function to detect that one, we will also start resizing the device partitions or the partitions in the actual sd-cards. 

I see 2 possible solutions to force the partition resize: 

1. using a deviceinfo variable (ex: `deviceinfo_initfs_force_partition_resize="true"`)
2. using a cmdline parameter (ex: `PMOS_FORCE_PARTITION_RESIZE`)

@ollieparanoid what you think about it?

ollieparanoid:
I would prefer the 2nd option, because we only need this in the `initramfs` and we do it the same way with the logging already.

PR #874, opened 2017-11-04
Add PMOS_FORCE_PARTITION_RESIZE kernel boot parameter

drebrez:
This kernel parameter will force the resize of the root partition during the initramfs boot script.

It will use the regex `'s/p?2$//'` to get the physical path of the device to be resized:

on x86_64 qemu: root partition `/dev/hda2`  => physical path `/dev/hda`
on armhf qemu: root partition `/dev/mmcblk0p2`  => physical path `/dev/mmcblk0`

The parameter has been added by default to all the cmdline variables of qemu packages.

I managed to test it successfully on x86_64 and armhf, but I didn't managed to test on aarch64, qemu wasn't starting at all.

Close: #869


drebrez:
Tested also on aarch64 and works fine.

Just for info, root partition on qemu aarch64 is `/dev/vda2`

PR #877, opened 2017-11-05
Add new device: samsung-i747m (Galaxy S3 USA/Canada version)

magmastonealex:
This is a different, but similar device to the existing samsung-i9305. It has a Snapdragon S4 instead of the Exynos processor in the i9305, and an Adreno 225 GPU

Supported so far:
- Boot from internal flash
- Touchscreen
- Weston/Xwayland
- WiFi (with firmware)


aports/device/linux-samsung-i747m/APKBUILD
```diff
@@ -0,0 +1,93 @@
+# Kernel config based on: arch/arm/configs/lineageos_d2_defconfig
+# Changes:
+# - Added DEVTEMPFS
+# - Enabled SYSVIPC
+# - Disabled ANDROID_PARANOID_NETWORK
+# Patches:
+# - All simply to allow it to compile under GCC6.
+
+pkgname="linux-samsung-i747m"
+pkgver=3.x.x
```
drebrez: Please add the correct kernel version, you can read it in the `Makefile` (see [here](https://github.com/LineageOS/android_kernel_samsung_d2/blob/716f00ee2e86adfd507842c5f193a738b57d3c49/Makefile#L1-L3))
In this case should be `3.4.112`

magmastonealex:
So close :) 

I've fixed the kernel version now.

drebrez:
@magmastonealex Please remember to create/update the wiki page of the device and add a new entry in the supported devices table [here](https://wiki.postmarketos.org/wiki/Supported_devices)

Issue #871, opened 2017-11-04
Binary Repository #2


ollieparanoid:
### Reproducible builds
The first binary repository approach involved reproducible builds, and it turned out not to be feasible and it wouldn't provide any real security benefit. Read https://github.com/postmarketOS/pmbootstrap/issues/64#issuecomment-334252443 for details. But we need a binary repository, it makes development much easier (compiling KDE takes forever, and we have some developers for whom compiling kernels also takes hours).

Please note, that I chose the reproducible build route to provide maximum security. We do not only run the binaries from this repository on the target devices (phones, tablets, ...), but also on the PCs where we develop (e.g. the cross-compilers `gcc-armhf`, ...), and sometimes even with root access (e.g. the `heimdall` flasher application).

What we have left from this approach and can use, are mirrors for hosting the packages (thanks to @MartijnBraam, see #365).

### Alpine building postmarketOS packages
So the next best thing would be, if Alpine built our packages along with the official packages. Then we wouldn't need to trust another person for providing the packages. Two Alpine developers had the idea of providing a PPA-like system at some point in the future, and therefore I've asked on the mailing list.

The short answer is, that they would need a new build system first, for which there is no time plan and that Alpine focuses on its upcoming 3.7 release so they don't have time for this.

[Looks like](https://lists.alpinelinux.org/alpine-infra/0198.html) we can talk about this again after 3.7, but realistically even if it worked out that way, it won't happen in the next few months, which means for me that we better have a short-term solution.

## Short-term solution
It seems to me, that the fastest way to get this going is if I build all the packages myself, but not reproducible. This might be the next best deal regarding trust, because you need to sort-of trust me already for the `pmbootstrap` code, which also runs on your computer and while you can just read it, people rarely read all code before executing it.

I would use an i3 CPU with cross-compilation (which I'm using right now to do bigger builds), which is slower than what @PureTryOut uses. But then again, we use ccache and we only need to build changed packages. As long as the packages compile regularly, the delay shouldn't be more than one day or so, which is a lot better than what we have now (no binary packages).

For the first iteration, I would trigger the builds manually and I might be able to set this up in a week or so. I would write a script, that does the following:
* Clone `pmbootstrap`
* Build some (later all) packages
* Upload the whole packages folder to the official mirror

Future ideas:
* Delete outdated packages from the repo (use `zap -m`)
* GitHub trigger to build automatically whenever a commit lands in master
[...]

PR #881, opened 2017-11-07
Fix: package gets rebuilt, even if it exists, when using default arch

ollieparanoid:
When not specifying an architecture for `pmbootstrap build`, and the `APKBUILD`
says that it can't be built for the native arch, it gets built for the right
foreign arch. `pmbootstrap` did not properly detect if packages were already
built in that case, and tried to build them again.

(I've noticed that while building packages for the binary repo #871.)

### How to test

Use any `linux-` package, that is not available for `x86_64` on a `x86_64` PC
and build it twice. It should get properly detected now:

```shell
pmbootstrap build linux-amazon-thor
pmbootstrap build linux-amazon-thor # should not get built again
```


pmb/build/package.py
```diff
@@ -53,7 +53,7 @@ def package(args, pkgname, carch, force=False, buildinfo=False, strict=False):
                                               suffix)
 
     # Skip already built versions
-    if not force and not pmb.build.is_necessary(args, carch, apkbuild):
+    if not force and not pmb.build.is_necessary(args, carch_buildenv, apkbuild):
         return
```
craftyguy: I think it would be very helpful to have a message printed here along the lines of "No updates detected, package is already the latest, not rebuilding". Currently it just outputs this immediately after hitting return:

```
❯ pmbootstrap fix/build-is-necessary-default-arch* ./pmbootstrap.py build linux-postmarketos-stable
[14:35:09] Done
```

That seems *ok*, but it also might seem like to the user that something is wrong since a kernel compile is never *that* fast :)

ollieparanoid: `make` prints out the following in such a case: `make: Nothing to be done for 'all'.`

So we could do something similar. But to be honest, I would do this on demand, meaning, when we really have someone come up and saying that this message not being there is confusing. Because we would need to figure out a good way to handle this, it should not come for every package that we check to be built (that happens a lot during `pmbootstrap install`!).

And when we implement it, it should really go into another PR, because this is a fix PR for another issue. If you have the time, could you review this PR as it is (i.e. read the code, verify whether the bug is fixed and comment on that fix) and if you consider necessary, open a new issue with the "nothing to do" message?

Thanks for looking at it!

craftyguy: ok, it was just a suggestion since I am always skeptical of commands that return instantly without any indication as to *why*. As I mentioned previously, I tested this out and it seems to do what you advertise.

ollieparanoid: This is a bit random, but I thought about this a few times and changed my mind. But I've forgotten to reply here again.

Anyway, @z3ntu asked for the same thing in https://github.com/postmarketOS/pmbootstrap/pull/1174#issuecomment-361055959 so I'm implementing this right now as part of that pull-request. :+1:

PR #882, opened 2017-11-07
README.md: Remove obsolete telnet unlocking text

ollieparanoid:
Preview:
https://github.com/postmarketOS/pmbootstrap/blob/feature/readme-remove-telnet-unlock/README.md


craftyguy:
I wonder if we should call out the osk porting guide, and a note about not enabling fde if the device hasn't been configured to work with osk-sdl?

drebrez:
@craftyguy what you suggest should be a warning message somewhere during the install, in the readme I think is better to show the happy flow only

craftyguy:
I'm just trying to reduce the number of folks who have a bad time
because they enable it without having configured it properly to work.

drebrez:
@craftyguy I know, I agree to show that warning message, but the Readme file of the github should try to get the attention to the project itself, if you already see warnings that it might not work, that's not encouraging IMO

ollieparanoid:
Thanks for the reviews!

PR #884, opened 2017-11-08
linux-samsung-i9003: fix compilation with PARANOID_NETWORK disabled

ollieparanoid:
Looks like, there are actually kernels where disabling a kernel option breaks the build, because they assume it is always enabled ._.

Anyway, this PR fixes it for `linux-samsung-i9003` and in the future we might be able to extend the automatic checksum check by Travis to an automatic package building check by Travis, just like Alpine does it (as soon as the binary repo is working, we could look into that, then Travis can use the pre-compiled cross-compiler packages).

The good news are, that there was already a patch on the net.


z3ntu:
Gotta love downstream kernels ^^

z3ntu:
Actually in the FP2 kernel repo, the original commit for that is https://github.com/LineageOS/android_kernel_fairphone_msm8974/commit/e9a85c71afa38ee304e71c86ca7d74ed4658318c , idk how it is in the other kernel repos (just do a "blame" of security/commoncap.c in the github interface to find out)

zhuowei:
@z3ntu Angler has the same patch (https://github.com/LineageOS/android_kernel_huawei_angler/commits/cm-14.1/security/commoncap.c), so it seems at least modern Qualcomm devices have it.

z3ntu:
@zhuowei Thanks for checking. I'll replace the patch with that one on this branch.

ollieparanoid:
Tested and the other patch also applies cleanly. Any opinions on adding the error message it fixes during compilation? (My intention was to make it easier to find the patch with that information.)

zhuowei:
@ollieparanoid could we autodetect kernels that needs this patch somehow?

ollieparanoid:
I have proposed `pmbootstrap autopatch $kernel`, that automatically tries out all known patches 😄 

See https://github.com/postmarketOS/pmbootstrap/issues/688#issuecomment-338407540, *3rd paragraph*. That issue is actually full of ideas to make porting easier, with a big TODO list on top. Help is appreciated if someone is interested.

z3ntu:
The easiest way of doing this would be to create a wiki page called "Common kernel patches" or so and linking them there. This would require no code changes and only slight modifications of existing wiki pages to link to that new site.
I would put every patch there which is needed for more than 2 or 3 kernels, possibly sorted by number of usages in this repo.

ollieparanoid:
@z3ntu: the extra wiki page would also work, but then again it [is already explained](https://wiki.postmarketos.org/wiki/Porting_guide#Get_a_patch) how to use grep the search through the existing patches (which works for all patches and we don't need the extra work of listing them in a wiki page). We could discuss this further in #688 if you want to.

Back to the topic, are you / is anyone else *against* including the actual error message from the compilation on top of the patch?

PR #887, opened 2017-11-09
Close #871: Enable binary repository

### LINKED ISSUE #871: Binary Repository #2

ollieparanoid:
It should be basically working already. ~~Not all packages have been compiled, some got stuck while building and need further investigation~~. Binary repository is located here: http://postmarketos.brixit.nl

The builder does not automatically compile packages yet, I need to trigger it manually as of now. But that is planned (not in scope of this PR).

### How to test
* Check out the branch
* Whenever you try to build something, that exists in the binary repo, it should get skipped
* Whenever you install something, that exists in the binary repo, it should get taken from there
* See the packages here (almost everything, except firmware packages, tell me if I've missed anything): http://postmarketos.brixit.nl/

### TODO
- [x] update the README.html of the mirrors
- [x] ~~use HTTPS mirror if possible (see below, right now it uses HTTP)~~ nope
- [x] update "keys" testcase to pull Alpine's keys out of the `alpine-keys` package we install anyway when setting up the chroot (so we can enable the testcase in Travis, it is currently disabled as it downloads the whole aports git repo) and to add the postmarketOS key
- [x] check upgrade path from pmbootstrap without the mirror configured to this version
- [x] fix regression in config loading/saving for `timestamp_based_rebuild`


MartijnBraam:
ssl is not activated because its not very useful for package mirrors, the packages are signed themselves to guarantee that its the right package. Enableing https for the mirrors will make running upstream caches for the packages way harder (like I do now for debian, ubuntu and arch with a transparent proxy).

ollieparanoid:
Good argument, no HTTPS it is then.

ollieparanoid:
Thanks for the review @drebrez! This should be resolved now, could you test again? Tests from other people are appreciated as well!

ollieparanoid:
Travis fails with:
> RuntimeError: Your aports repository does not have the 'origin/HEAD' reference. Please add it by running the following commands inside /home/travis/build/postmarketOS/pmbootstrap: git symbolic-ref refs/remotes/origin/HEAD refs/remotes/origin/master; git fetch

...which we also had in #832 and fixed that by adding to `.travis.yml`:
```shell
./pmbootstrap.py config timestamp_based_rebuild False
```

This means, that the reworked config code doesn't quite work yet - it doesn't save/load the option correctly!

ollieparanoid:
I'll try it again tomorrow. @drebrez found the correct way to patch it, but my patch was invalid:
> you changed it in the wrong place 
> it's few lines below, where it says `# Use defaults from pmb.config.defaults`

drebrez:
Another thing I noticed, in my config I have a specific `mirror_alpine`, now it's ignored in the load config function. But I think is the new normal behavior, right?
Also `jobs` seems ignored (always uses default), and that shouldn't be the case

drebrez:
Possible solution:
```
    # Use defaults from the user's config file
    args = parser.parse_args()
    cfg = pmb.config.load(args)
    for varname in cfg["pmbootstrap"]:
        if varname not in args or getattr(args, varname, "") == "" or getattr(args, varname) == None:
            value = cfg["pmbootstrap"][varname]
            if varname in pmb.config.defaults:
                default = pmb.config.defaults[varname]
                if isinstance(default, bool):
                    value = (value.lower() == "true")
            setattr(args, varname, value)

    # Use defaults from pmb.config.defaults
    for key, value in pmb.config.defaults.items():
        if key in args and getattr(args, key, "") != "" and getattr(args, key, "") != None:
            continue
        setattr(args, key, value)
```
In the first loop, it sets all the variables loaded from the config (only those configurable) and it sets them in the `args` only if are not present and not `empty/none`
In the second loop it's similar, only sets the default if in the `args` if the key is not present and not `empty/none`

ollieparanoid:
Thanks a lot for your input @drebrez, it helped me a lot to fix this.

First I have added a testcase, which covers the following:
* Generate a default config (like Travis does, but in a temporary folder)
* Check that `timestamp_based_rebuild` is True at first, then set it to False and then to True again, and verify each step by building the `args` variable with that config and checking the value
* Check if loading the `job` count from the config is still working, and check if it can be overridden with the `-j` parameter (that's considered more important than what is in the user's config).

Then I've adjusted the code to pass that case.

---

Alright, I consider this PR **ready to be reviewed and tested**. If someone did any of that, it would be greatly appreciated 😃

pmb/config/merge_with_args.py
```diff
@@ -0,0 +1,51 @@
...
+    pmb.config.config_keys, we look into the value set in the the user config.
+
+    When that is empty as well (e.g. just before pmbootstrap init), or the key
+    is not in pmb.config_keys, we use the default value from the internal
+    config.
+    """
+    # Use defaults from the user's config file
+    cfg = pmb.config.load(args)
+    for varname in cfg["pmbootstrap"]:
+        if varname not in args or not getattr(args, varname):
```
drebrez: @ollieparanoid did you check what happens with `not getattr(args, varname)` if the value returned by `getattr` is `False`?
From my tests I remember that is actually evaluating it as `not False`, and this means that the attribute will be changed here with the default value.

```
class TestClass(object):
    testBoolTrue = True
    testBoolFalse = False
    testStringEmpty = ""
    testStringNotEmpty = "this is a test"
    testNone = None

args = TestClass()

print("not getattr(args, \"testBoolTrue\")       => " + str(not getattr(args, "testBoolTrue")))
print("not getattr(args, \"testBoolFalse\")      => " + str(not getattr(args, "testBoolFalse")))
print("not getattr(args, \"testStringEmpty\")    => " + str(not getattr(args, "testStringEmpty")))
print("not getattr(args, \"testStringNotEmpty\") => " + str(not getattr(args, "testStringNotEmpty")))
print("not getattr(args, \"testNone\")           => " + str(not getattr(args, "testNone")))
print("")
print("getattr(args, \"testBoolTrue\") is None       => " + str(getattr(args, "testBoolTrue") is None))
print("getattr(args, \"testBoolFalse\") is None      => " + str(getattr(args, "testBoolFalse") is None))
print("getattr(args, \"testStringEmpty\") is None    => " + str(getattr(args, "testStringEmpty") is None))
print("getattr(args, \"testStringNotEmpty\") is None => " + str(getattr(args, "testStringNotEmpty") is None))
print("getattr(args, \"testNone\") is None           => " + str(getattr(args, "testNone") is None))
```
output:
> not getattr(args, "testBoolTrue")       => False
not getattr(args, "testBoolFalse")      => True
not getattr(args, "testStringEmpty")    => True
not getattr(args, "testStringNotEmpty") => False
not getattr(args, "testNone")           => True

> getattr(args, "testBoolTrue") is None       => False
getattr(args, "testBoolFalse") is None      => False
getattr(args, "testStringEmpty") is None    => False
getattr(args, "testStringNotEmpty") is None => False
getattr(args, "testNone") is None           => True


the check below for setting the default when `if key not in args or getattr(args, key) is None`, seems fine.

I'm not yet familiar with the test framework in python, so I don't know if this use cases is  covered.
But from what I understand you save a config with the default values (no variables with False), and you test by changin the `timestamp_based_rebuild ` only in the `args` variable, not in the actually `cfg["pmbootstrap"]["timestamp_based_rebuild "]`.

ollieparanoid: Thanks for spotting this, indeed this is not covered by the testsuite and would yield the wrong results if you passed something to the commandline, for which `not getattr(args, key)` results in `True`.

While we can't overwrite a Boolean (yet), this does make it impossible to unset the postmarketOS binary mirror (by setting it to an empty string, which `not ""` also returns `True`).

I will add a this to the testcase accordingly and fix the code. Thanks for your great reviewing work! \o/

ollieparanoid:
@MoreRobustThanYou: thanks for testing!
@drebrez: I've fixed what you've found (good catch!) - do you (or anyone else) see another reason why this shouldn't be merged?

drebrez:
@ollieparanoid code looks good, I didn't managed to test it again. 
PS: I'm trying to install some IDE with a debugger for python.

drebrez:
#### tested cases (with config key present in `pmb.config.config_keys`)

- config not defined, default with string => default value is used :heavy_check_mark:
- config not defined, default with True/False => default value is used :heavy_check_mark:
- config with string, default with string => config value is used :heavy_check_mark:
- config with string, default with empty string => config value is used :heavy_check_mark:
- config with True, default with False => config value is used :heavy_check_mark:
- config with False, default with True => config value is used :heavy_check_mark:
- override with argument --mirror-pmOS => value from argument is used :heavy_check_mark:
- override with argument --mirror-alpine => value from argument is used :heavy_check_mark:
- override with argument --jobs => value from argument is used :heavy_check_mark:
- override with argument --port-distccd => value from argument is used :heavy_check_mark:
- config with empty string, default with value => config value is used :x: WRONG?? (see below)

#### all the previous test cases with a config key **not** present in `pmb.config.config_keys` results in the default value is used :heavy_check_mark: (this is the new expected behavior)

The case with the empty string which doesn't use the default value but it actually keep the empty string, it might break something (for example with empty `user` ?).

My suggestion is to add a check for the user defined config key and if it's an empty string, use the default value. 
The only key with an empty string is `keymap`, because others key (`extra_packages`, `ui`) are handled with `none`, so right now is not a problem.

What you think? @ollieparanoid 

```
diff --git a/pmb/config/merge_with_args.py b/pmb/config/merge_with_args.py
index 2ff61ee..2411036 100644
--- a/pmb/config/merge_with_args.py
+++ b/pmb/config/merge_with_args.py
@@ -41,6 +41,8 @@ def merge_with_args(args):
             value = cfg["pmbootstrap"][key]
             if key in pmb.config.defaults:
                 default = pmb.config.defaults[key]
+                if value is "":
+                    value = default
                 if isinstance(default, bool):
                     value = (value.lower() == "true")
             setattr(args, key, value)
```

ollieparanoid:
(Rebased to master.)

Whoa, thanks for the extensive testing!

Regarding overwriting empty string settings with the default value: As you wrote, we don't have an use-case for this yet. For future options it might make sense to allow saving an empty string, so I think it makes sense if we don't overwrite it with the default unless we know that it is necessary at some point. In case someone sees this differently or simply wants to discuss this more, I also recommend doing that in a new issue.

I'll go with your recommendation and merge this now. Thank you so much, we finally have a binary repository \o/ 🎉  😀

PR #894, opened 2017-11-11
Close #836: Wrap Alpine's newapkbuild

### LINKED ISSUE #836: Wrap abuild's newapkbuild
Alpine's abuild provides a nice [newapkbuild](https://github.com/alpinelinux/abuild/blob/master/newapkbuild.in) tool, for which you can specify the source, and it will generate a state-of-the-art APKBUILD file accordingly by analyzing it.

We could simply wrap it with something like:
```shell
pmbootstrap newapkbuild [--folder=main] [everything passed through]
```

The `--folder` parameter will not collide, as it only uses single letter parameters.

When called, it would do the following:
* install `abuild` in the native chroot
* run `newapkbuild` with all parameters
* create a new folder in aports/$folder/$pkgname (we can parse the pkgname from the resulting APKBUILD)
* copy the generated APKBUILD to that new folder

ollieparanoid:
`newapkbuild` creates `APKBUILD`s and other files (e.g. OpenRC service file templates). It is quite powerful and always installed, when installing the build environment anyway. This PR adds a simple wrapper, to make it conveniently usable with `pmbootstrap newapkbuild`.

#### `pmbootstrap newapkbuild -h`
```
usage: pmbootstrap newapkbuild [-h] folder ...

positional arguments:
  folder       aports subfolder, where the new aport will be located (main,
               cross, device, ...)
  args_passed  arguments directly passed to Alpine's newapkbuild, more
               information: 'pmbootstrap newapkbuild main -h'

optional arguments:
```

#### `pmbootstrap newapkbuild main -h`
```
newapkbuild 3.1.0-r1 - generate a new APKBUILD
Usage: newapkbuild [-n NAME] [-d DESC] [-l LICENSE] [-u URL]
       [-aCpy] [-s] [-cfh]
       PKGNAME[-PKGVER]|SRCURL
Options:
  -n  Set package name to NAME
  -d  Set package description (pkgdesc) to DESC
  -l  Set package license to LICENSE
  -u  Set package URL
  -a  Create autotools package (use ./configure ...)
  -C  Create CMake pakckage (Assume cmake/ is there)
  -m  Create meson package (Assume meson.build is there)
  -p  Create perl package (Assume Makefile.PL is there)
  -y  Create python package (Assume setup.py is there)
  -s  Use sourceforge source URL
  -c  Copy a sample init.d, conf.d, and install script to new directory
  -f  Force even if directory already exist
  -h  Show this help
```

#### `pmbootstrap newapkbuild main -d "testdescription" -c testaport`
```
>>> testaport: Updating the sha512sums in APKBUILD...
[00:40:37] Create /home/user/code/pmbootstrap/aports/main/testaport
[00:40:37] Done
```


ollieparanoid:
Actually, `apkbuild` generates a `src` folder inside the `aport`, just like `abuild` does it when building packages. We don't want that folder to be versioned along with the `pmbootstrap` git repo, so it should not get copied to the `aports` folder. I think that should be changed before merging it, but it can already be tested.

ollieparanoid:
Thanks for looking at it!
Testing the normal flow would be enough - using `newapkbuild` to generate a new package and (what the testsuite doesn't cover), using `newapkbuild` to download sources and analyze them automatically.

ollieparanoid:
Thanks for testing! I didn't delete it, because it was easier to code and we're doing all that in the chroots we zap all the time anyway. But on second thought it would be cleaner if we removed the folder afterwards. Also this PR needs to be rebased, adding `dont_merge_pr`.

ollieparanoid:
Instead of using `/tmp/newapkbuild` (which is then not removed), I've updated it to use `/home/pmos/build`. This has the advantage, that you can directly look at extracted source code, which it puts in a src subfolder, if you specify a source download URL.

PR #899, opened 2017-11-11
Add max size option for fastboot

### LINKED ISSUE #897: Ability to set max fastboot flash size in deviceinfo
```amazon-thor``` can't flash files via fastboot that are over 500 MB. I'd like to be able to set the max flash byte size in deviceinfo.

### PR DESCRIPTION
**Pull request by craftyguy**
This adds a new deviceinfo 'flash_fastboot_max_size' used for preventing fastboot from flashing a system partition that is too large. Some devices do not support flashing over a certain size (e.g. 500MB).

fixes #897

### CODE REVIEW

File: pmb/flasher/frontend.py [hunk 1]
```diff
@@ -67,6 +67,15 @@ def system(args):
         raise RuntimeError("The system image has not been generated yet,"
                            " please run 'pmbootstrap install' first.")
 
+    # Do not flash if using fastboot & image is too large
+    method = args.flash_method or args.deviceinfo["flash_methods"]
+    if method == "fastboot" and args.deviceinfo["flash_fastboot_max_size"] != '':
```
ollieparanoid: How about `and not args.deviceinfo["flash_fastboot_max_size"]`?

craftyguy: Done! But it's actually `and args.device....` since we want that to be set before trying to use it :)

ollieparanoid: Right 😄  thank you!

ollieparanoid:
@craftyguy: So, I've written in the review above, that I would recommend to change the deviceinfo for the `device-amazon-thor`, too and edit the wiki. You've written in the chat, that you'd edit the wiki after this is merged (thanks!).

And it would also be possible to adjust the deviceinfo after merging this PR, so we could ship it now. Would you prefer that? 🚢

craftyguy:
Sure, I've updated the deviceinfo for the amazon-thor (I set it as `510`)

ollieparanoid:
Thanks again!

Issue #208, opened 2017-07-20
[rx51] Sometimes the phone boots without usable display


MartijnBraam:
Sometimes the phone doesn't boot right and you only see scrolling lines on the display:

![lines](https://user-images.githubusercontent.com/6928199/28429082-bbac110a-6d7b-11e7-8418-7e89168f1afa.jpg)

This has been reproduced by @craftyguy

craftyguy:
This is almost certainly related to the issue reported here: https://lkml.org/lkml/2017/6/14/901

Unfortunately it doesn't look like the 'fix' (reverting one patch and applying a new one from drm-next) have been staged for 4.13. I may produce a patch that we can apply to the kernel in the linux-nokia-rx51 APKBUILD..

ollieparanoid:
If you can't get it fixed, I suggest reaching out to the kernel devs working on the N900 patches.

craftyguy:
I agree. I am building 4.13-rc1 right now to test it to see if it's still there. They'll also want me to try to reproduce this with the drm-next branch before they'll talk to me (I highly suspect!).

craftyguy:
I sent a response to that thread on the linux-kernel mailing list asking if there was any update. None of the patches in 4.13 for omap-drm seem to mention this problem specifically in the rather short commit messages, and 4.13-rc1 showed the exact same issue so I suspect it hasn't been resolved yet.

craftyguy:
I noticed that if I compile gpio_keys into the kernel instead of as a LKM, this problem seems to happen 100% of the time on every boot.

craftyguy:
#309, via d6bf000, works around this issue by applying a patch to linux-postmarketos.

craftyguy:
Looks like the patch to fix this issue ([seen here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/drivers/gpu/drm/omapdrm/omap_drv.c?id=fc5cc9678e130196012c17b37e555d53d3d3476b)) has been merged into 4.14!

ollieparanoid:
Fixed in kernel 4.14 (#903).

PR #903, opened 2017-11-12
[linux-postmarketos-stable] Update to 4.14

craftyguy:
Note, I have removed the rx51 DRM patch that worked around the boot display issue (#208), this issue should be resolved in this kernel.


craftyguy:
yea I tested it on my N900, and it seems to work. I rebooted a few times and haven't been about to reproduce #208.

@opendata26 mentioned that the `linux-postmarketos-mainline` config + patches should be merged with this package now. Maybe it's trivial for this round, but I am concerned that folks will modify the config in 'stable', others will add/modify config in 'mainline', and merging the two will be quite tedious every time.

PR #905, opened 2017-11-13
Add `pmbootstrap bootimg_analyze` command and prompt for boot.img during new device wizard

drebrez:
This PR adds the command `pmbootstrap analyze_bootimg` which extract all the information from an existing boot.img and print out the variables for the deviceinfo.

```
drebrez@drebrez-laptop:~/pmbootstrap$ pmbootstrap analyze_bootimg ~/boot.img
[00:26:38] (native) install unpackbootimg
[00:26:39] Put these variables in the deviceinfo file of your device:

deviceinfo_kernel_cmdline="androidboot.hardware=qcom user_debug=31 msm_rtb.filter=0x37"
deviceinfo_generate_bootimg="true"
deviceinfo_bootimg_qcdt="true"
deviceinfo_flash_offset_base="0x00000000"
deviceinfo_flash_offset_kernel="0x00008000"
deviceinfo_flash_offset_ramdisk="0x02000000"
deviceinfo_flash_offset_second="0x00f00000"
deviceinfo_flash_offset_tags="0x01e00000"
deviceinfo_flash_pagesize="2048"

[00:26:39] Done
```

And it also prompt it during the initialization wizard for a new device.

This is a working draft implementation, please review the code and suggest how can I improve it.


zhuowei:
I haven't tested this, so I'm wondering what the result would be if one ran this on a non-standard boot.img, such as the [Pixel C's Chrome OS format boot.img](https://forum.xda-developers.com/pixel-c/help/unpack-repack-boot-img-info-to-build-t3367835) or a MediaTek/Sony ELF boot.img.

fyfyone:
I tested it with an Android boot image and it seems to have worked. 

I tried throwing a file that is not a boot image at it. The error message was just the generic one ("failed, see log"). In the log, the actual error message was "`Android boot magic not found.`" Could we please hand that over to the log so users don't have to open their log files?

TL;DR: "`Android boot magic not found.`" or other errors should be passed to the log.

drebrez:
@fyfyone I also noticed that the error is not propagated, I know that `pmb.chroot.user` call can have a flag to return the output of the command, I will investigate how to print it out in case of error. 
@zhuowei I tried with a lot of different boot.img and twrp from various devices (samsung, lg, mediatek, ...) and they all worked.
Thank you both for testing and suggestions

pmb/aportgen/device.py
```diff
@@ -79,8 +81,49 @@ def ask_for_flash_method(args):
...
 
+def ask_for_bootimg(args):
+    if not pmb.helpers.cli.confirm(args, "Do you want to extract fastboot information from an "
+                                   "existing boot.img?"):
+        return None
+
+    logging.info("Location of the 'boot.img' (stock boot.img, stock recovery.img, twrp.img, ...)")
+    while True:
+        try:
+            path = os.path.expanduser(pmb.helpers.cli.ask(
```
ollieparanoid: I would only wrap the `pmb.parse.bootimg` line with `try...catch`, because that is what it is for, right? (It could end in an infinite loop if `pmb.helpers.cli.ask` didn't work and threw an exception for example.)

pmb/aportgen/device.py
```diff
@@ -79,8 +81,49 @@ def ask_for_flash_method(args):
...
+            return pmb.parse.bootimg(args, path)
+        except KeyboardInterrupt:
+            return None
+        except Exception as e:
+            logging.fatal("ERROR: " + str(e) + ", Please try again.")
+
+
+def generate_deviceinfo_fastboot_content(args, bootimg):
+    if bootimg is None:
+        bootimg = {}
```
ollieparanoid: How about writing that as
```python
bootimg = {"cmdline": "",
    "qcdt": "false",
    ...}
````

File: pmb/aportgen/device.py [hunk 3]
```diff
@@ -190,7 +221,11 @@ def generate(args, pkgname):
     has_keyboard = ask_for_keyboard(args)
     has_external_storage = ask_for_external_storage(args)
     flash_method = ask_for_flash_method(args)
+    if flash_method in ["fastboot", "heimdall-bootimg"]:
+        bootimg = ask_for_bootimg(args)
+    else:
+        bootimg = None
```
ollieparanoid: This is shorter and in my opinion just as readable:
```python
bootimg = None
if flash_method in ["fastboot", "heimdall-bootimg"]:
    bootimg = ask_for_bootimg(args)
```

pmb/parse/arguments.py
```diff
@@ -316,6 +316,11 @@ def arguments():
     config.add_argument("name", nargs="?", help="variable name")
     config.add_argument("value", nargs="?", help="set variable to value")
 
+    # Action: analyze_bootimg
+    analyze_bootimg = sub.add_parser("analyze_bootimg", help="Extract all the"
```
ollieparanoid: I would have named it `bootimg_analyze` instead, because I thought we always put the verb at the end of actions (`kconfig_check`, `build_init`). But I realized that this is not always the case, e.g. `parse_apkbuild` and `parse_apkindex` so we don't have a real convention yet. However, the `parse_*` commands aren't used much so I guess no one would complain if we renamed them.

What do you think?

pmb/parse/bootimg.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+import os
+import pmb
+
+
+def bootimg(args, path):
+    if not os.path.exists(path):
+        raise RuntimeError("Could not find file '" + path + "'")
+
+    pmb.chroot.apk.install(args, ["unpackbootimg"])
```
ollieparanoid: How about also installing `file` and checking what `file` thinks the `boot.img` is? It would be relatively easy to see if it is:
* a real `boot.img` file
* just the linux kernel (-> show a message, that the user needs to use `heimdall-isorec`)
* none of the above (-> throw an error which contains the output of `file`)

pmb/aportgen/device.py
drebrez: @ollieparanoid actually there is a reason i've put the `pmb.helpers.cli.ask` in try/catch, to catch the `KeyboardInterrupt`, I forgot to write it somewhere. 
Let's consider the use case where a user answer `y` to provide the boot.img but all the boot.img he provides are not working, I thought about something like `(ctrl+c to skip)` and be able to continue without the boot.img.
Maybe split in 2 different try/catch blocks?

pmb/parse/arguments.py
drebrez: @ollieparanoid are you suggesting to start the convention now? 😄 
For me it's fine with the verb at the end, so I will change it. 
Should I change the parse commands also? Or it's better leave it for another PR?

drebrez: @ollieparanoid or should I change it to `parse_bootimg` to be aligned with the python namespace and the other commands?
Or should I create a new folder `analyze`?

pmb/aportgen/device.py
ollieparanoid: Good idea, but we don't use/handle `^C` in any of the other questions. How about skipping with an empty string (or "none") as input instead? In fact, then we could merge that with the confirmation at the beginning. Example of this idea:

> You can analyze a known working boot.img file to automatically fill out the flasher information for your deviceinfo file. Either specify the path to an image or press return to skip this step (you can do it later with pmbootstrap analyze_bootimg).
> Path:

pmb/parse/arguments.py
ollieparanoid: I think its good if we set up a convention now, yeah. I'll answer these questions to my own taste, but it's really a opinion thing and as always I'm happy to discuss all points.

How I would do it:
* rename it to `bootimg_analyze`
* rename the `parse_*` commands to `*_parse` (in a new PR)
* do not change the folder (because we also have `kconfig_check` in the `parse` folder, which would fit more in the `analyze` folder then, which means we would have to also change that to `analyze_kconfig` to be consistent. someone suggested having multiple verbs for `kconfig` in the future, so this idea would even be incompatible then).

And yeah, if we set up a convention here, I'll put it in the wiki for future reference 👍

pmb/parse/bootimg.py
drebrez: seems a good approach, I will try it.
For the real `boot.img` you think it's enough to check if there is `Android bootimg` in the whole output?

ollieparanoid: Yes, that should be enough.

drebrez:
@ollieparanoid I've addressed all the comments, can you please review it again and also test it? Thanks

ollieparanoid:
Unfortunately I can't do a full review/test today, but here are some thing I've noticed while testing:
* Use case: user uses pmbootstrap for the first time and wants to do a device port with analyzing a boot.img file
  * the user gets asked for the sudo-password after specifying a path in `pmbootstrap init` (new with this PR, previously that wasn't required in init)
  * I think it would be good to write *why* that is necessary (e.g. "NOTE: You will be prompted for your sudo password, so we can set up a chroot to extract and analyze your `boot.img` file")
  * So how about adding that right before `pmb.chroot.apk.install(args, ["file", "unpackbootimg"])`?

* Testcases would be great. We could add sample `bootimg` files (real and vmlinuz) in `test/testdata/bootimg/`, run the analyzer script on it and compare if it delivers the output it should. Also we could run it on a python file with the output we know it should generate.

Thanks for making this PR!

pmb/parse/bootimg.py
```diff
@@ -0,0 +1,65 @@
...
+    bootimg_path = args.work + "/chroot_native" + temp_path + "/boot.img"
+
+    # Copy the boot.img into the chroot temporary folder
+    pmb.helpers.run.root(args, ["cp", path, bootimg_path])
+
+    file_output = pmb.chroot.user(args, ["file", "-b", "boot.img"], working_dir=temp_path,
+                                  return_stdout=True).rstrip()
+    if "android bootimg" not in file_output.lower():
+        raise RuntimeError("File is not an Android bootimg. (" + file_output + ")")
+
```
ollieparanoid: How about also checking for `zImage`s, and telling the user that `heimdall-isorec` is needed in that case? I think the substring to search for would be `Linux kernel` (tested it with a `x86_64` kernel from my host Linux system).

pmb/aportgen/device.py
```diff
@@ -79,8 +81,47 @@ def ask_for_flash_method(args):
...
+        path = os.path.expanduser(pmb.helpers.cli.ask(args, "Path", None, "", False))
+        if not len(path):
+            return None
+        try:
+            return pmb.parse.bootimg(args, path)
+        except Exception as e:
+            logging.fatal("ERROR: " + str(e) + ". Please try again.")
+
+
+def generate_deviceinfo_fastboot_content(args, bootimg):
```
ollieparanoid: How about indicating, that filling out `bootimg` is optional by writing the following?
```python
def generate_deviceinfo_fastboot_content(args, bootimg=None):
```

pmb/aportgen/device.py
```diff
@@ -79,8 +81,47 @@ def ask_for_flash_method(args):
...
+        deviceinfo_flash_offset_ramdisk=\"""" + bootimg["ramdisk_offset"] + """\"
+        deviceinfo_flash_offset_second=\"""" + bootimg["second_offset"] + """\"
+        deviceinfo_flash_offset_tags=\"""" + bootimg["tags_offset"] + """\"
+        deviceinfo_flash_pagesize=\"""" + bootimg["pagesize"] + """\"
+        """
+
+
 def generate_deviceinfo(args, pkgname, name, manufacturer, arch, has_keyboard,
-                        has_external_storage, flash_method):
+                        has_external_storage, flash_method, bootimg):
```
ollieparanoid: I'd also put `bootimg=None` here.

PabloCastellano:
Also remember to add the command name to https://github.com/postmarketOS/pmbootstrap/blob/d9e7a3d7deeaab48a1be778af75c2aa850f05a6b/pmb/parse/arguments.py#L349 so that we can execute it no matter if the deviceinfo file doesn't exist yet

pmb/parse/bootimg.py
```diff
@@ -0,0 +1,65 @@
...
+        output["ramdisk_offset"] = ("0x%08x" % int(f.read().replace('\n', ''), 16))
+    with open(bootimg_path + "-second_offset", 'r') as f:
+        output["second_offset"] = ("0x%08x" % int(f.read().replace('\n', ''), 16))
+    with open(bootimg_path + "-tags_offset", 'r') as f:
+        output["tags_offset"] = ("0x%08x" % int(f.read().replace('\n', ''), 16))
+    with open(bootimg_path + "-pagesize", 'r') as f:
+        output["pagesize"] = f.read().replace('\n', '')
+    with open(bootimg_path + "-cmdline", 'r') as f:
+        output["cmdline"] = f.read().replace('\n', '')
+    output["qcdt"] = ("true" if os.path.isfile(bootimg_path + "-dt") and
```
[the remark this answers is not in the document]
ollieparanoid: Maybe it makes more sense to change it in `mkinitfs.sh`, let's discuss here: #914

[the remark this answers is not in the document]
PabloCastellano: As discussed, this is not a problem anymore. We will handle the variable properly (true/false/empty)

ollieparanoid:
Added testcases (as discussed in #postmarketOS) and rebased to master.

Thank you so much for making this @drebrez, it's already obvious how helpful this is \o/
Thanks to @PabloCastellano for reviewing and testing!

@drebrez: do you want to adjust the porting guide accordingly?

PR #907, opened 2017-11-14
Workaround for Qemu aarch64 abuild-tar bug

ollieparanoid:
Qemu has a bug in aarch64 emulation, that causes `abuild-tar` to omit files from the archives it generates in some cases (#546). This workaround ~~copies the `abuild-tar` binary from the native chroot into the buildroot_aarch64 and installs a wrapper to use it. That way we bypass emulation and it works as expected again.~~ builds a patched `abuild-tar`, which avoids the bug.

### DIFF
--- a/pmb/build/qemu_workaround_aarch64.py
+++ b/pmb/build/qemu_workaround_aarch64.py
@@ -1,0 +1,42 @@
+"""
+Copyright 2017 Oliver Smith
+
+This file is part of pmbootstrap.
+
+pmbootstrap is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+pmbootstrap is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with pmbootstrap.  If not, see <http://www.gnu.org/licenses/>.
+"""
+import pmb.chroot
+import pmb.chroot.apk_static
+import pmb.config
+import pmb.helpers.repo
+import pmb.helpers.run
+import pmb.parse.arch
+
+
+def qemu_workaround_aarch64(args, suffix="buildroot_aarch64"):
+    """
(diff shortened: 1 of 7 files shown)


ollieparanoid:
@zhuowei: I've assembled the patch from what you wrote in https://github.com/postmarketOS/pmbootstrap/issues/546#issuecomment-328380771. Is it the same as yours?

pmb/build/qemu_workaround_aarch64.py
```diff
@@ -0,0 +1,42 @@
...
+import pmb.config
+import pmb.helpers.repo
+import pmb.helpers.run
+import pmb.parse.arch
+
+
+def qemu_workaround_aarch64(args, suffix="buildroot_aarch64"):
+    """
+    Qemu has a bug in aarch64 emulation, that causes abuild-tar to omit files
+    from the archives it generates in some cases. This workaround copies the
```
zhuowei: is this comment still accurate? It seems we are building a patched aarch64 binary, not a x86_64 binary.

zhuowei:
I think this will work: the comment seems to be misleading since this seems to build an aarch64 abuild-tar binary, not an x86_64 one, but it should work around the issue.

I currently don't have time to test this on my device, unfortunately. Testing in qemu-aarch64 sould be enough if anyone wants to do that.

pmb/build/qemu_workaround_aarch64.py
ollieparanoid: Right, that comment is outdated. I've tried to copy it first (and using various hacks to make it find the .so files), but it all failed because `abuild-tar` appears to be running in `fakeroot`.

ollieparanoid: I've even copied it into this PRs submit comment 😆

ollieparanoid:
Thanks for the review! I'll delete the `aarch64` folder from the binary repository and build these packages again to see if it works. I've also had errors there with files missing, causing packages depending on other packages not to compile correctly.

zhuowei:
@ollieparanoid When I did the port to Nexus 6P, only Weston had issues since it was the only package built from source that had symlinks. The kernels and device packages should be fine.

ollieparanoid:
Thanks for the information. I think I've also had issues with building `hildon` because of that, and maybe another program (chocolate-doom because sdl-net wasn't built right or something).

ollieparanoid:
Building all `aarch64` packages works now. I've also compared the contents of files in all `weston` packages from `armhf` against the ones in `aarch64` and they match. 🚢

PR #908, opened 2017-11-14
Unset restrictive umask before creating files

jspricke:
If a user has a restrictive umask (for example 0077) set, it will be
inherited to sudo and thus files created with sudo where not readable by
the normal user. In that case, when setting up a new chroot, the
etc/apk/repositories file would have umask 600 and a os.path.exists()
(in update_repository_list) would return false. Setting the umask to 0
first, results in world readable files and directories, so the user
running `./pmbootstrap.py install` can read them.


pmb/__init__.py
```diff
@@ -35,6 +35,7 @@ def main():
     # Parse arguments, set up logging
     args = parse.arguments()
     pmb_logging.init(args)
+    os.umask(0)
```
ollieparanoid: Why did you chose `0`? The default `umask` on my system seems to be `022`, so I would have used that.
Thanks for making the PR!

jspricke: Fixed and explained :)

PR #909, opened 2017-11-15
Packages xweston

MoreRobustThanYou:
xweston is a script that allows you to use weston as a host, in order to run XWayland easily.

this will likely require rewriting of xfce4/hildon scripts to use the  ```~/.xinitrc``` file, as xweston runs that after it starts


ollieparanoid:
Cool! Is it working on your device already (with xfce maybe?), and do you want a review?
Thank you for working on this!

PR #914, opened 2017-11-18
Handle deviceinfo_bootimg_qcdt boolean variable properly

PabloCastellano:
Reference: https://github.com/postmarketOS/pmbootstrap/blob/master/aports/main/postmarketos-mkinitfs/mkinitfs.sh#L210

### DIFF
--- a/aports/main/postmarketos-mkinitfs/APKBUILD
+++ b/aports/main/postmarketos-mkinitfs/APKBUILD
@@ -10,6 +10,7 @@ source="init.sh.in init_functions.sh mkinitfs.sh"
 arch="noarch"
 license="GPL2"
 provides="mkinitfs=0.0.1"
+options="!check"
 
 package() {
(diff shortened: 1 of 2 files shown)


ollieparanoid:
Great find! I recommend, that we change `postmarketos-mkinitfs`, so it accepts "true" and "false". Because that is also what we allow for the other booleans (see [deviceinfo reference](https://wiki.postmarketos.org/wiki/Deviceinfo_reference)).

Opinions on this?

drebrez:
Nice catch @PabloCastellano I also prefer to handle a boolean in the correct way. But also an empty string should be fine.
I think in general they should be handle like this:
variable not defined => false
variable defined with empty string => false
variable with "false" or any other wrong value => false
variable with "true" => true

in other words, always check if variable is present and value is `true` for condition to match. 
What you think ?

PabloCastellano:
Yeah, sounds more reasonable. I'm working on it

PR #916, opened 2017-11-19
Add support for Sony Xperia Z1 (sony-honami)

limiter121:
Working:
- Display
- Touchscreen
- Network through USB
- Telnet in initramfs
- SSH
- Wifi

Not working:
- FDE (WIP)


aports/device/device-sony-honami/90-android-touch-dev.rules
```diff
@@ -0,0 +1,7 @@
+# udev rules file
+# All device names can be read from weston's logfile
+
+# Touchscreen (use 'weston-calibrator' to calibrate)
+SUBSYSTEM=="input", ATTRS{name}=="clearpad", \
+ENV{ID_INPUT}="1", ENV{ID_INPUT_TOUCHSCREEN}="1" \
+ENV{WL_CALIBRATION}="1.013786 0.035310 -24.920929 -0.008370 0.990421 12.094299"
```
MartijnBraam: Have you tried this without the calibration value? the calibration is not needed for most devices since the touchscreen controller is factory calibrated.

aports/device/device-sony-honami/APKBUILD
```diff
@@ -0,0 +1,23 @@
+pkgname="device-sony-honami"
+pkgdesc="Sony Xperia Z1"
+pkgver=1.0
+pkgrel=10
```
MartijnBraam: For merging it should be pkgrel=0 since this is the first release of this package in the master branch

aports/device/device-sony-honami/deviceinfo
```diff
@@ -0,0 +1,28 @@
+# Reference: <https://postmarketos.org/deviceinfo>
+# Please use double quotes only. You can source this file in shell scripts.
+
+deviceinfo_format_version="0"
+deviceinfo_name="Xperia Z1"
```
MartijnBraam: `deviceinfo_name` should include the manufacturer

aports/device/device-sony-honami/initfs-hook.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+# start msm-fb-refresher (to show splash)
+msm-fb-refresher
```
MartijnBraam: does this really work? since it only refreshes a single frame this way, its probably handled by the msm_refresher setting in the deviceinfo

aports/device/linux-sony-honami/APKBUILD
```diff
@@ -0,0 +1,79 @@
+# Kernel config based on: arch/arm/configs/(CHANGEME!)
```
MartijnBraam: You forgot to update the comment in the header

aports/firmware/firmware-sony-honami/APKBUILD
```diff
@@ -0,0 +1,52 @@
+pkgname=firmware-sony-honami
+pkgver=1
+pkgrel=1
```
MartijnBraam: pkgrel should also be 0 here.

aports/device/device-sony-honami/initfs-hook.sh
limiter121: Nope, it does not, the display works even without this. I will remove it.

aports/firmware/firmware-sony-honami/APKBUILD
```diff
@@ -0,0 +1,52 @@
+pkgname=firmware-sony-honami
```
ollieparanoid: This looks like a copy of `firmware-sony-amami`. If you need exactly the same files, how about depending on that package instead?

*I think longterm it would be good to have firmware packages for the specific components (e.g. prima wifi chip) instead of firmware packages per device - and for some components we already have this, but it is not yet consistent.*

limiter121: Yeah, that sounds like a better approach.

Issue #480, opened 2017-08-29
Distcc fails sporadically, causing single files to be compiled with qemu (slow!)


ollieparanoid:
I suspect, that `distcc` has a timeout for the compiler, but more research needs to be done on this. This does not block compiling, but slows it down unnecessarily.
```
distcc[19959] ERROR: compile generic.c on 127.0.0.1:33632 failed with exit code 118
distcc[19959] (dcc_build_somewhere) Warning: remote compilation of 'generic.c' failed, retrying locally
distcc[19959] Warning: failed to distribute generic.c to 127.0.0.1:33632, running locally instead
```

craftyguy:
Just FYI, I found this response which says that the timeout we would be interested in is not configurable: https://github.com/distcc/distcc/issues/212

exit code 118 is definitely a "timeout", according to `error_codes.h`

ollieparanoid:
Great research work and thanks for asking upstream how to handle this!

PR #921, opened 2017-11-20
Allow configuring ccache size and skipping build options in init

ollieparanoid:
This is very useful for the binary repository, because I'm almost hitting the default cache size of 5 GB from ccache already (4.7 or so is what I have right now) and I have not even started with compiling KDE. With this PR, one can set it as big/as small, as desired.

### Output
Here's how `pmbootstrap init` looks like with this PR. First the usual stuff...
```
[18:15:49] Target device (either an existing one, or a new one for porting).
[18:15:49] Available (35): amazon-thor, asus-flo, asus-grouper, ...
[18:15:49] Device [samsung-i9100]:
[18:15:49] Username [user]:
[18:15:49] Available user interfaces (3):
[18:15:49] * none: No graphical environment
[18:15:49] * hildon: (X11) Lightweight GTK+2 UI (optimized for single-touch touchscreens)
[18:15:49] * weston: (Wayland) Reference compositor (demo, not a phone interface)
[18:15:49] * xfce4: (X11) Lightweight GTK+2 desktop (stylus recommended)
[18:15:49] User interface [weston]:
[18:15:50] Location of the 'work' path. Multiple chroots (native, device arch, device rootfs) will be created in there.
[18:15:50] Work path [/home/user/.local/var/pmbootstrap]:
```

Then the grouped build options:
```
[18:15:51] Build options: Parallel jobs: 3, ccache size: 5G, timestamp based rebuilds: False
[18:15:51] Change? (y/n) [n]: y
```

When answering with `y`:
```
[18:15:57] How many jobs should run parallel on this machine, when compiling?
[18:15:57] Jobs [3]:
[18:15:58] We use ccache to speed up building the same code multiple times. How much space should the ccache folder take up? (After init is through, you can check the current usage with 'pmbootstrap stats'.)
[18:15:58] 0: infinite
[18:15:58] Ccache size [5G]:
[18:16:00] Rebuild packages, when the last modified timestamp changed, even if the version did not change? This makes pmbootstrap behave more like 'make'.
[18:16:00] Timestamp based rebuilds (y/n) [y]:
```

When finished with the above or pressed `n` there, then you'll continue as usually:
```
[18:16:03] Additional packages that will be installed to rootfs. Specify them in a comma separated list (e.g.: vim,file) or "none"
[18:16:03] Extra packages [none]:
...
```

### Detailed changes
* allow configuring the ccache size
* allow to skip changing build related options in `pmbootstrap init`
  (so we stop making the list of questions longer and longer)
* sort config keys alphabetically
* remove the arch parameter from `pmbootstrap stats` (which shows
  the ccache stats). The reason is, that whenever you are
  cross-compiling with distcc, the ccache layer is used in the
  native chroot (so the ccache code runs without emulation and is
  therefore faster). That means, it makes little sense to choose
  another architecture to get the statistics, and removing it
  makes the code a bit simpler.
* move ccache_stats out of `pmb/build/__init__.py again`, directly
  putting the two lines of logic into `frontend.py` makes it simpler
* fix: make sure, that ccache is installed when running
  `pmbootstrap stats`

### DIFF
--- a/pmb/build/__init__.py
+++ b/pmb/build/__init__.py
@@ -20,7 +20,7 @@ along with pmbootstrap.  If not, see <http://www.gnu.org/licenses/>.
 from pmb.build.init import init
 from pmb.build.checksum import checksum
 from pmb.build.other import copy_to_buildpath, is_necessary, \
-    symlink_noarch_packages, find_aport, ccache_stats, index_repo
+    symlink_noarch_packages, find_aport, index_repo
 from pmb.build.package import package
 from pmb.build.menuconfig import menuconfig
 from pmb.build.qemu_workaround_aarch64 import qemu_workaround_aarch64
--- a/pmb/build/other.py
+++ b/pmb/build/other.py
@@ -302,15 +302,12 @@ def symlink_noarch_packages(args):
         index_repo(args, arch)
 
 
-def ccache_stats(args, arch):
-    suffix = "native"
-    if args.arch:
-        suffix = "buildroot_" + arch
-    pmb.chroot.user(args, ["ccache", "-s"], suffix, log=False)
-
-
-# set the correct JOBS count in abuild.conf
 def configure_abuild(args, suffix, verify=False):
+    """
+    Set the correct JOBS count in abuild.conf
+
+    :param verify: internally used to test if changing the config has worked.
+    """
     path = args.work + "/chroot_" + suffix + "/etc/abuild.conf"
     prefix = "export JOBS="
     with open(path, encoding="utf-8") as handle:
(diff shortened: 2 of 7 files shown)


drebrez:
@ollieparanoid I don't understand the change related to the `arch` option. 
Right now I see these folders in my work dir:
- cache_ccache_aarch64 (200KB)
- cache_ccache_armhf (6MB)
- cache_ccache_x86 (empty)
- cache_ccache_x86_64 (4.1GB)
With this PR if I delete the other architecture folders and leave only the native, the others they shouldn't be created anymore because ccache will run only in the native?

ollieparanoid:
**Good catch!** The thing is, that `ccache` only runs in the native chroot normally and does all caching there, even when cross-compiling with distcc.

But there are exceptions:
* `pmbootstrap --no-cross build` (which will disable cross-compiling with `distcc` altogether and only use `qemu` - very slow, might be useful for debugging though)
* Compiling a file takes too long via `distcc`, so it runs into a timeout and uses the native compiler instead (running with qemu and taking even longer, see #480)
* What I forgot about: when no emulation is necessary, then the ccache folder of that architecture also gets used. Example: compiling for `armhf` on a `aarch64` machine, or for `x86` on a `x86_64` machine.

So while the first two ones are debatable, the third bullet point makes it clear that we need to have full support for ccache related stuff for every architecture. This means, I need to change this PR to do the following:
- [ ] allow displaying the stats for all architectures again
- [ ] the function, that sets the `ccache` size, should also do it for foreign architectures when building there (right now it only supports native)
- [ ] the explanation should make clear, that you might end up with *multiple* ccache folders, that have this size.

ollieparanoid:
I've rebased the PR, fixed what you've commented and also fixed ccache usage for non-native chroots in #1026. Could you test/review it again?

PR #924, opened 2017-11-20
[hammerhead] Added redscreen kernel patch and xorg config

MartijnBraam:
This makes xorg and xfce4 work on the hammerhead thanks to the changes found by @zhenyolka


aports/device/device-lg-hammerhead/APKBUILD
```diff
@@ -30,6 +31,8 @@ package() {
 		"$pkgdir"/etc/profile.d/00-udev-v4l-hack.sh
 	install -Dm644 "$srcdir"/bcmdhd.cal \
 		"$pkgdir"/etc/wifi/bcmdhd.cal
+	install -Dm644 "$srcdir"/xorg.conf \
```
drebrez: @MartijnBraam what about installing the xorg configuration with an `install_if` condition? (similar to [this](https://github.com/postmarketOS/pmbootstrap/blob/7ab85bc4a5ae748b84759d4e12c367e4979db96d/aports/main/postmarketos-base/APKBUILD#L56))

MartijnBraam: Ohh yes, good idea

aports/device/device-lg-hammerhead/APKBUILD
```diff
@@ -32,8 +33,20 @@ package() {
 		"$pkgdir"/etc/wifi/bcmdhd.cal
 }
 
+x11() {
+	install_if="$pkgname xorg-server"
+	depends="
+		libinput
+		xf86-input-libinput
+		xf86-video-fbdev"
+	mkdir "$subpkgdir"
```
ollieparanoid: `abuild` requires the `$subpkgdir` to exist, that's why `postmarketos-base` has this in the subpkg functions. But here the subfolder gets created already with `install -D`, so the line is redundant.

aports/device/device-lg-hammerhead/APKBUILD
```diff
@@ -32,8 +33,20 @@ package() {
...
+	install -Dm644 "$srcdir"/xorg.conf \
+		"$subpkgdir"/etc/X11/xorg.conf.d/10-hammerhead.conf
+}
+
 weston() {
 	install_if="$pkgname weston"
+	mkdir "$subpkgdir"
```
ollieparanoid: `abuild` requires the `$subpkgdir` to exist, that's why `postmarketos-base` has this in the subpkg functions. But here the subfolder gets created already with `install -D`, so the line is redundant.

MartijnBraam:
I've removed the weston pixman type workaround now as it was indeed not needed anymore. I've also added the msm-fb-refresher which I was shure about I've added it already.

PR #925, opened 2017-11-21
[lg-d285] fix weston redscreen and xorg problems

zhenyolka:
Fix redsreen problem and xfce4 segfault problem.
Remove touchscreen calibration, because it not need now.(in past it need to sync screen image and touchscreen, now we can sync it by rotate screen in xorg.conf).


aports/device/device-lg-d285/APKBUILD
```diff
@@ -23,8 +24,11 @@ package() {
 		"$pkgdir"/etc/udev/rules.d/90-android-touch-dev.rules
 	install -D -m644 "$srcdir"/initfs-hook.sh \
 		"$pkgdir"/etc/postmarketos-mkinitfs/hooks/00-${pkgname}.sh
+    install -D -m644 "$srcdir"/xorg.conf \
```
drebrez: @zhenyolka please move the installation of the `xorg.conf` file in a `x11` subpackage with the `install_if` condition. See how it is done here => https://github.com/postmarketOS/pmbootstrap/pull/924/files#diff-0235b4cc2f5f00b650bcd942cee09d9c

ollieparanoid:
Does weston still work? I guess that you need to remove `deviceinfo_weston_pixman_type="2"` also, but please test it on the device. Thanks for making this PR!

drebrez:
@zhenyolka are you still planning on finishing this PR?

PR #927, opened 2017-11-21
Workaround for mdss-fb

### LINKED ISSUE #912: [sony-amami] backlight workaround for weston is broken
@ata2001 reported in the chat, that [fb-workaround.sh](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/device/device-sony-amami/fb-workaround.sh) doesn't work anymore since `weston` doesn't run as `root` anymore. Because it depended on autologin as `root`.

My proposed workaround was using an OpenRC service instead of autologin (which will go away when we have a display manager anyway).

ata2001:
Closes #912.

It could possibly work for other devices, that currently are using msm-fb-refresher in loop mode. One such device is sony-honami.


ollieparanoid:
Could you document this in the wiki?
https://wiki.postmarketos.org/wiki/Troubleshooting:display

(If you're busy currently, just tell me and I'll add it.)

Thanks a lot for making this, @ata2001! \o/

ata2001:
@ollieparanoid done

PR #931, opened 2017-11-22
Use mdss hack on Z1

limiter121:
Replace `msm-fb-refresher` with the new `mdss-fb-init-hack` from @ata2001


ata2001:
@limiter121 Did you get the same result after removing that line?

limiter121:
@ata2001 Nope, it doesn't work without that.

limiter121:
Since the hack doesn't work but `msm-fb-refresher` works just fine, I think it would be best if the efforts go into getting mainline working instead of this.

I will keep the branch on my fork if anybody wants to continue.

Issue #629, opened 2017-09-25
Package LuneOS user interface


zhuowei:
[LuneOS](https://en.wikipedia.org/wiki/LuneOS) is a continuation of webOS, HP's discontinued and open sourced mobile operating system. Currently they are focusing on [upgrading the user interface](http://webos-ports.org/wiki/Luna_Next) to use modern technologies such as Qt5 and Wayland, which would allow running the LuneOS interface on regular Linux. They provide [build instructions](http://webos-ports.org/wiki/Getting_Started_-_QML_Apps) for running these UI components outside of their OpenEmbedded-based distribution; those instructions may be useful to get LuneOS's UI running on postmarketOS.

ollieparanoid:
Just like we did it in #161 with plasma-mobile, I recommend making a list of all required dependencies first (as a wiki page if there are many), then packaging one after another.

We have the strict package building mode #256 now, which should make it a bit easier (`pmbootstrap build --strict package-name`). Also related: #62

PureTryOut:
Well, I don't like the look of the UI, but I love choice. Since I'm kinda stuck with Plasma Mobile right now anyway, I'll look into this. Seems easy enough from the little bit of reading I have done so far.

MartijnBraam:
The default icons for webos are kinda horrible but it has a very nice user experience. And if we port the HP touchpad then we kinda have the native shell running on it.

PureTryOut:
I'm a bit less optimistic now. LuneUI is quite tightly integrated with LuneOS itself. It has some services which has to run at boot, and it has some hard deps on systemd. I guess we can patch these out eventually, but it'll take a while as of course the devs aren't as interested in running systemd-less as we are.

ollieparanoid:
The services running at boot is something we can do, but the systemd dependencies really make it harder. Did you write down somewhere, which ones you encountered?

PureTryOut:
So far I've encountered 2, of which one only has a dep because it offers a .service file (so it's easily patched out):

* luna-next: easily patched, only provides a .service file
* luna-service2: this actually has code depending on systemd, and will need more work to patch out

ollieparanoid:
Thanks for listing that. Looks like there is [more interest](https://www.reddit.com/r/postmarketOS/comments/737u3f/great_project/) in getting the LuneOS UI running on pmOS.

@PureTryOut, how about putting what you have already in a branch in the `pmbootstrap` repo, so it 's a bit easier to collaborate on this?

[...]

PR #935, opened 2017-11-25
Refactor pmb/build/package.py (make depends work like in abuild)

### LINKED ISSUE #824: pmbootstrap doesn't figure out dependencies right when compiling KDE
@PureTryOut reported in IRC:
> if I tell it to compile a package which has deps on other packages that aren't compiled yet, it'll install all the deps for all packages combined before starting to compile the first in the list, even with `--strict`

ollieparanoid:
This PR will fix #824 (wrong depends handling), which is the **last blocker before merging the Plasma Mobile PR**. And it will allow extending package code again, because the code is split up in small functions and these have test cases, with new features like native cross compilation for autoconf packages (see @zhuowei's proof of concept in #665). Needless to say, testing and reviewing is greatly appreciated.

EDIT: ~~Test cases ran through locally, but fail with Travis.~~ Fixed! 👍 

### Change of behavior

`pmbootstrap build device-samsung-i9100` explains it all:
```
[01:18:04] Dependency handling in 'pmbootstrap build' has been changed.
[01:18:04] Previously we only built and installed the 'makedepends' from the APKBUILDs, now we use the 'depends', too.
[01:18:04]
[01:18:04] Your options:
[01:18:04] * Ignore depends (fast, old behavior, may cause problems with some packages):
[01:18:04]   pmbootstrap build device-samsung-i9100 -i
[01:18:04] * Build with depends (kernel!) and specify the right architecture:
[01:18:04]   pmbootstrap build device-samsung-i9100 --arch=armhf
[01:18:04]
[01:18:04] This change was necessary to be more compatible with Alpine's abuild.
[01:18:04] The default architecture is the native one (x86_64 in your case), so you need to overwrite
[01:18:04] it now to get the kernel dependency of your device package for the right architecture.
[01:18:04] Sorry for the inconvenience.
[01:18:04]
[01:18:04] ERROR: Missing -i or --arch parameter
[01:18:04] Run 'pmbootstrap log' for details.
[01:18:04] See also: <https://postmarketos.org/troubleshooting>
```

### How to test (easy!)
* Build packages like you would normally do
* Do full installations

### Detailed changes
* Rename pmb/build/package.py to pmb/build/_package.py, so we can
  access the functions it contains in testcases, and still use
  pmb.build.package()
* Refactor the entire file. Instead of one big function that does
  too many things, we have many small ones now, that are tested
  in the testsuite and easier to modify
* Whenever building a package, pmbootstrap does not only build and
  install the "makedepends" (like we did before), now it does the
  same for the "depends". That's required to be compatible with
  abuild. The old behavior can still be used with 'pmbootstrap
  build --ignore-depends'.
* Because of that change, noarch packages can no longer be built in
  the native chroot if we need them for a foreign chroot. A device-
  package depending on a kernel would pull in the same kernel for
  the native architecture otherwise.
* Running 'pmbootstrap build device-...' without '--ignore-depends'
  and without a matching '--arch' displays a note that explains
  this change to the user and tells how to use it instead.
* Noarch packages no longer get symlinked. That was only
  implemented for packages built in the native chroot, and now that
  is not always the case anymore. Symlinking these packages creates
  packages with broken dependencies anyway (e.g.
  device-samsung-i9100 can't be installed in x86_64, because
  linux-samsung-i9100 is armhf only).
* Rename "carch" to "arch" wherever used. Naming it "carch"
  sometimes is confusing with no benefit.
* Add a testcase for the aarch64 qemu workaround (because it failed
  first and I needed to know for sure if it is working again).
* Improved some verbose logging, which helped with development of
  this feature.
* Removed the old "build" test case (which was disabled in
  testcases_fast.sh) as the new "build_package" test case covers its
  functionallity.


zhuowei:
I ran into a dependency error when compiling [the luneos branch](https://github.com/postmarketOS/pmbootstrap/issues/629), as luna-service2 was in luna-sysmgr's depends but not makedepends. I'm currently testing if this PR fixes that issue.

PR #939, opened 2017-11-25
Add Wiko Lenny3 device

drebrez:
What works:

- Kernel compilation
- Flash Kernel (SP Flash Tool required)
- Flash System (SP Flash Tool required)
- Initramfs boot
- USB Network
- Framebuffer graphics (Splash screens)
- SSH Connectivity
- Weston on framebuffer
- Xwayland
- Touchscreen

unfortunately `osk-sdl` doesn't work yet, it reboots the device, I opened the issue postmarketOS/osk-sdl#72


drebrez:
@ollieparanoid I did like you suggested and created a mirror on github ([drebrez/kernel_wiko_lenny3](https://github.com/drebrez/kernel_wiko_lenny3)). 👍

ollieparanoid:
Regarding the GPL2 compliance:
> I think it's not a GPL violation
> the GPL states that you must include the "scripts to control compilation and installation" but it does not require those "scripts" to be licensed under GPL. 
> so I would consider the generator as such a "script" (even though it technicall isn't a script) and hence it is ok if it's redistributable but not GPL2

PR #943, opened 2017-11-27
Fix #941: Use the right arch for foreign-arch-only packages

### LINKED ISSUE #941: pmbootstrap doesn't automatically pick the right architecture for building when none is specified
Sorry, #935 caused the following regression.

Example:
`pmbootstrap build linux-samsung-i9100`

What should happen:
* `pmbootstrap` figures out, that this is only available for `armhf` and start building for `armhf`

What happens:
* `pmbootstrap` complains, that it can't be built for `x86_64` and that you should specify the architecture. It aborts there.

Workaround:
`pmbootstrap build --arch=armhf linux-samsung-i9100`

ollieparanoid:
This is a follow-up to #935.

* fix regression #941: pmbootstrap doesn't automatically pick the
  right architecture for building when none is specified
* remove obsolete --noarch-arch parameter


ollieparanoid:
Your understanding is right (except for: x86_63, wait what? 😉 ).
Thanks for testing @craftyguy!

Issue #936, opened 2017-11-25
"pmbootstrap qemu" broke with Linux 4.14 (we use Linux LTS as workaround now)


ollieparanoid:
@PureTryOut wrote:
> ugh, pmOS Qemu seems dead... with `dri-virtio`, it doesn't boot at all. with `dri-swrast`, the filesystem is readonly which makes it impossible to write to `/tmp` 

@PureTryOut does it work, when adding "rw" to the kernel command line (there's a switch in `pmbootstrap qemu` which lets you directly modify the kernel commandline)?

ollieparanoid:
I can reproduce this issue, and adding "rw" to the kernel command line does *not* fix it.
However, I have noted that OpenRC says:

```
...
 * Checking local filesystems ...
fsck.ext4: Unable to resolve 'LABEL="pmOS_root"'
 * Operational error
 * Remounting root filessystem read/write ...
mount: /: can't find LABEL="pmOS_root".
...
```

PureTryOut:
I wanted to wait for someone else to reproduce the error before making an issue out of it, I guess I was too late 😉

It's an interesting issue, I'm not exactly sure what commit caused this to happen.

ollieparanoid:
My checklist for the partition-by-label not found issue:
* ~~pmbootstrap doesn't write the label to the partition~~ checked, that does not happen
* caused by the last kernel update?
* openrc update? (at least there's no such open issue - and that would mean it happens with non-qemu, too)

ollieparanoid:
The Linux upgrade to 4.14 introduced the "read-only bug", it works fine with the last 4.13 kernel we had packaged.

**Workaround:** choose `none` as UI in `pmbootstrap init` or modify your `postmarketos-ui-*` profile script to drop to a shell. Then remount the root partition with `rw` as @magmastonealex wrote elsewhere.


Next steps:
* Does this also happen with 4.14.2? (yeah we should upgrade our Linux package!)


What I also found out: The errors in OpenRC come from `/etc/init.d/fsck` and `/etc/init.d/root`. Both use `/lib/rc/bin/fstabinfo`, which is a program shipped with OpenRC ([fstabinfo.c](https://github.com/OpenRC/openrc/blob/master/src/rc/fstabinfo.c)).

zhuowei:
Some other weirdness:

- Whenever we try to look for pmOS_root (e.g. with findfs), this shows in the kernel log:

```
[  976.261467] attempt to access beyond end of device
[...]

PR #954, opened 2017-11-29
device-qemu-amd64: Use LTS kernel

ollieparanoid:
Workaround for #936, which broke with the `4.14.0` release of the Linux kernel. This packages the LTS release and makes `device-qemu-amd64` use it. The kernel config is the same as `linux-postmarketos-stable`.

I want to find the commit in the kernel that caused the regression the next days, and e-mail the author so we can work this out properly in the long term.

Kernel compiles for me, it does *not* have the read-only issue and `dri-virtio` works out of the box again.


zhuowei:
@ollieparanoid Tested this by running a `pmbootstrap zap` on my qemu-amd64 rootfs, checking out this branch, building, and booting: qemu mounted the partitions and booted fine.

PR #957, opened 2017-11-30
[New Device] Google Glass

kaendfinger:
This implements support for Google Glass.

@ollieparanoid has also asked me to take photos of this, and I will get these to the project ASAP.


PureTryOut:
Thank you very much! Very cool to see Google Glass supported!

Please add the device to the [list of supported devices](https://wiki.postmarketos.org/wiki/Devices).

kaendfinger:
@PureTryOut I have updated the device list. I will be working on a detailed wiki page soon.

aports/device/linux-google-glass/APKBUILD
```diff
@@ -0,0 +1,79 @@
+# Kernel config based on: arch/arm/configs/(CHANGEME!)
```
ata2001: Please adjust this line.

aports/device/linux-google-glass/APKBUILD
```diff
@@ -0,0 +1,79 @@
...
+
+pkgname="linux-google-glass"
+pkgver=3.4.83
+pkgrel=0
+pkgdesc="Google Google Glass kernel fork"
```
ollieparanoid: How about replacing that with `pkgdesc="Google Glass kernel fork"`? (I know that this was auto generated)

aports/device/linux-google-glass/02_gpu-msm-fix-gcc5-compile.patch
```diff
@@ -0,0 +1,33 @@
+source:
```
zhuowei: This file isn't used: you might want to remove it.

kaendfinger:
Checked over everything again, let me know if there are any more things to fix 👍 

@ollieparanoid It's extremely tricky to get a good photo where you can see the screen, so my solution is to take the following photos / screenshots:

1) Screenshot of Weston with the terminal open, identifying the device as Google Glass (hardware ID if one exists, or a hostname of google-glass)
2) A photo of the device itself with a backdrop of the postmarketOS site
3) Once I make it, a screenshot of a Web Interface for viewing Glass sensor data

Also, it would probably be cool to mention that coinciding with getting pmOS running on Glass is making open source libraries for interacting with the Glass sensors, and in the future, maybe input methods consisting of a head-movement and blinking for Weston and other environments via libinput.

kaendfinger:
glass-pmos-1

ollieparanoid:
Nice!

Sorry for the stupid question, but why is it difficult to make a photo with that running on the google glass device? Since it runs already, I thought you would put down the glass on the table, then take out any (phone or not) camera, and make a photo of the whole thing. Even if you can barely see the display, together with the screnshot it would be good enough to present in my opinion.

kaendfinger:
@ollieparanoid Nah it's a great question :)
I can get a non-display image pretty easily, however the way the screen works on the Glass (beaming from the side onto film inside a prism), it is very dependent on your eye (in this case, the camera) having a very particular field of view, focus, and position



I've taken some more :smile:  I hope I can get a video done soon showing some feedback from the sensors.

img_7755
img_7752
img_7757

ollieparanoid:
If you don't mind me asking, how about a photo where you wear the Glass for the blog post? As I understand it should successfully display something then.

kaendfinger:
I will get one ASAP, sorry for the delay, been a busy week

ocdtrekkie:
@ollieparanoid In the case of default Google Glass software (which mind you, I presume doesn't apply here), Glass actually shuts off the screen when you set it on a table! But yeah, I can definitely confirm, near impossible to take a picture of the screen, it looks like a small square of light when on.

What Google does for "screenshots" is they take a screen capture and then put it as a layer with some opacity in the top right corner of a normal picture of whatever the person is looking at.

Like this: "screenshot"

PR #962, opened 2017-12-01
Fix #960: Disable root login (sudo still works)

### LINKED ISSUE #960: pmbootstrap install should set the root password
Currently we set the main user's password, but not the root user's.

To reproduce:

- run `su` on postmarketOS when logged in as a standard user, or log in as "root" in a qemu image with "none" ui

Expected:

- postmarketOS asks me for a root password

Actual:

- no password is needed.

ollieparanoid:
Currently, you can login on the tty0 with the root and no password. This PR disables root login.


zhuowei:
Tested by:
- checking out this branch
- doing a `pmbootstrap zap` on the old rootfs
- running an `install --no-fde` for qemu-amd64, setting the user password
- starting up QEMU
- tried to log in as root with an empty password: can't.
- tried to log in as user with my password: worked
- tried to `su` with empty password: can't
- tried to `sudo su` with my user's password: worked.

So it seems to work.

(Am I doing the code reviews correctly? Do I need to test each pull request by booting it before approving?)

ollieparanoid:
Thank you very much for testing. I think you've tested this PR perfectly. :tada:

> Am I doing the code reviews correctly? Do I need to test each pull request by booting it before approving?

That depends on what the PR changes. In general, the idea is to read through the code changes, and to confirm that it:
* really changes what it says it would
* does not introduce regressions

This PR was about disabling the root login on the `tty` (and with `su`, good that you checked that too!), so to verify that it works one needs to boot into the OS (in Qemu or on the device, if you can access the `tty` there) and try to login on the `tty`.

If that doesn't answer your question, let me know 👍

PR #972, opened 2017-12-03
Fix #968: cache which packages have been built

### LINKED ISSUE #968: dependency calculation takes much longer now (e.g. pmbootstrap build postmarketos-ui-plasma-mobile)
Since #935, the dependency resolving got much slower. This is probably, because we check now, if a package's dependencies need to rebuilt or not - and we don't cache it during one session.

This means for meta-packages like `plasma-mobile`, it will check the same dependencies over and over again, causing it to take minutes to complete, at least on my (rather low-end) laptop.

The fix would be caching the packages we already checked during one session (we already do lots of caching, e.g. for parsed APKBUILDs and that gave us good performance boosts).

ollieparanoid:
Fixes #968.
### How to test

Run this with and without this PR. It should me much faster.
```shell
pmbootstrap build postmarketos-ui-plasma-mobile
```


ollieparanoid:
Thanks, looking at the code is important!

I'd also like to see someone other than myself test this, if possible.

ollieparanoid:
Merging it anyway, because without this fix Plasma development isn't much fun. I've tested it myself before making the PR (it also has a test case) and also the past days while building various things, so it should be good enough.

PR #976, opened 2017-12-04
Upgrade Qt libs to 5.9.3

PureTryOut:
Alpine has updated their Qt to 5.9.3, and mixing versions causes weird behaviour.


ollieparanoid:
Thanks for making this PR, your changes look good!

I think we will hit this issue often, so I've added a test case, that simply checks if our qt5-packages have the same `pkgver` as `qt5-qtbase` from Alpine (checked by parsing their `APKINDEX`).

(In the long term, these `qt5-` packages should be upstreamed of course, but this could safe us from lots of silent breakage until then.)

While doing that, I've found that `qt5-qtsensors` from the `luna` folder also was outdated (CC @magmastonealex). It probably makes sense to move this to `main` some time soon, but we can do that in a new PR in my opinion (and I didn't want to interfer with the ongoing luna porting work).

@PureTryOut: do you think my changes on top are good for merging?

magmastonealex:
@ollieparanoid  seems like a good test to have!

I noticed yesterday that the luna qt modules were out of date, and spent a bunch of time updating my `qtwebengine` package & fixing a bad upstream `qtwebchannel`. Your change for `qtsensors` is identical to mine. I agree all of those should probably get moved out to main sooner rather than later, as both plasma & luna will depend on most of them.

It's also worth noting that Alpine has started packaging `qtquickcontrols2` as of December 2nd, and their package seems to work fine. I think that may mean we no longer need to package it as well.

ollieparanoid:
Awesome, thanks for the information! @PureTryOut, would it be OK for you if we add a new commit to remove `qtquickcontrols2`, and then merge this PR?

PureTryOut:
All sounds fine to me, go ahead!

PR #977, opened 2017-12-04
Re-add powerdevil and polkit-kde-agent as runtime deps to postmarketos-ui-plasma-mobile

PureTryOut:
Since `powerdevil` depends on `kwin`, it'll also be installed.

This PR combined with #976 makes Plasma boot again on Qemu x86_64.


ollieparanoid:
Thank you very much @PureTryOut!

I understand now why we are missing so many dependencies: as the `plasma-mobile` package was removed (not necessary anymore as I understood) in https://github.com/postmarketOS/pmbootstrap/pull/440/commits/87ad564af5afec2a97a6fc8ae8263ab45e34614b, it's dependencies did not get pulled in anymore:

```shell
depends="plasma-workspace solid solid-libs desktop-file-utils powerdevil polkit-kde-agent"
```

So... what about them, do we also need to add them again?

@MoreRobustThanYou and I only added `kwin` and `consolekit2`, because we didn't understand that they were getting pulled in by `plasma-workspace`, and that we should depend on that instead.

MoreRobustThanYou:
I added Consolekit2 because it wasn't being installed with postmarketos-ui-plasma-mobile

ollieparanoid:
@PureTryOut wrote in the channel:
> we're not missing anything

PR #982, opened 2017-12-06
Testsuite: Run UIs in Qemu and check running processes and more

ollieparanoid:
### Changes
* When `pmbootstrap qemu` gets killed, it now takes down the Qemu process with it
* `test/check_checksums.py` got a new optional `--build` parameter, which makes
  it build all changed packages instead of just checking the checksums
* We run this before running the testsuite now, so all changed packages get
  built before running tests (otherwise tests would hang without any output
  while a changed package is building)
* New testcase, that zaps all chroots, installs a specific UI (xfce4 and
  plasma-mobile currently, easy to extend), runs it via Qemu and checks the
  running processes via SSH.
* Version checking testcase: rewritten to include Alpine's testsuite file in
  our source tree, so we don't need to clone their git repo anymore. Now it
  is enabled for Travis.
* All this gives us a nice 8% code coverage boost
* Increased the `hello-world` pkgrel to verify that the Travis job is working.

### Note
Plasma doesn't work with Qemu's `--display=none` yet, so we can only test if `polkitd` is running.

### TODO
- [x] `pmbootstrap build device-samsung-i9100` and other device packages: automatically build for the right arch instead of telling the user that it needs to be specified. Only that way it works automatically, and that's more userfriendly anyway. We can add a NOTE in the debug log (pmbootstrap log), which tells that we've assumed that this is what the user means.
- [x] I bet this won't work on first try with Travis 😉 - for some reason, it seems to run the `hello-world` package building in background, so it clashes with the Python testcases.

### Reviewing
Read through the code, you don't necessarily need to run it locally (as it runs on Travis).


drebrez:
Nice to have this, did you also consider adding visual regression tests? for example take a screenshot and show it in the result or compare it with an expected one with some tolerance? something like they explain here => https://twolfson.com/2014-02-25-visual-regression-testing-in-travis-ci

ollieparanoid:
Yes, I would love to have that, and thanks for the link! I was aware that openSUSE does such visual regression testing, but I didn't know that someone already did it with Travis.

Just in case it's not obvious, such follow-up enhancements should be made in new PRs once we got this one working in my opinion.

ollieparanoid:
FYI: Here's the bugreport regarding "abuild -f doesn't verify checksums", which is why `--force` is disabled in `check_checksums.py` now (only relevant for Travis):
https://bugs.alpinelinux.org/issues/8257

ollieparanoid:
Travis fails installing `qemu-system-x86`, because of the following bug 😞 
https://bugs.launchpad.net/ubuntu/+source/dpkg/+bug/1730627

Ideas how to fix this are appreciated. I've already tried to use the upcoming Travis image, but it did not help.

ollieparanoid:
Here we go, it runs on Travis now! Please review.

pmb/build/autodetect.py
```diff
@@ -17,11 +17,37 @@
...
+    file.
+
+    :returns: None (no deviceinfo file)
+              arch from the deviceinfo (e.g. "armhf")
+    """
+    # Require a deviceinfo file in the aport
+    if not pkgname.startswith("device-"):
+        return
+    deviceinfo = aport + "/deviceinfo"
+    if not os.path.exists(deviceinfo):
```
drebrez: @ollieparanoid I also faced these kind of checks in the other feature i'm writing, what you think about adding something directly in the `pmbootstrap build` operation that doesn't let build a `device-` package if it doesn't contain a `deviceinfo` file? (of course in another PR)

ollieparanoid: Good idea, but this would make `pmbootstrap` incompatible with building Alpine's `device-mapper` package. This is not a real use case of course, but I would prefer to stay compatible with `abuild`/Alpine stuff as closely as possible.

drebrez: Oh the only package that's not a device :sweat:, what about if starts with `device-*` and it's not `device-mapper*` :laughing:

ollieparanoid: I know, I chose the `device-` prefix carefully (only one package, not that bad, right?) 😄 
It would be totally fine with me if we added the check together with a whitelist in `pmb/config/__init__.py`, something like:
```python
deviceinfo_not_required = ["device-mapper"]
```

What do you think about that? (and yeah, new PR of course 👍)

z3ntu: What about a regex like `device-\w+-\w+` ? Doesn't match `device-mapper` but matches `device-fairphone-fp2` and `device-asus-flo`

ollieparanoid: This would make it harder to adjust `pmbootstrap` if there was a new package in Alpine, where the regex matches, but which is not a postmarketOS `device-*` package. Anyway, this is all theoretical and far away from our real use cases, so I don't care that much.

@drebrez: If you (or someone else) thinks, that *letting `pmbootstrap build` not build a `device-` package unless it has a `deviceinfo` file* would make our codebase better, and that this weights more than this tiny incompatibility, I'm totally fine with doing it.

We can always change the behavior again if we notice that it causes problems with future use cases we didn't have or think of right now.

.travis.yml
```diff
@@ -5,16 +5,16 @@ addons:
...
-install: "pip install flake8 pytest-cov python-coveralls"
+install:
+  - sudo apt install shellcheck qemu-system-x86
+  - pip install flake8 pytest-cov python-coveralls
 script:
   - test/static_code_analysis.sh
   - yes "" | ./pmbootstrap.py init
   - ./pmbootstrap.py config timestamp_based_rebuild False
   - ./pmbootstrap.py kconfig_check
+  - test/check_checksums.py --build
```
drebrez: @ollieparanoid does this still works with the recent changes about the `--force` in the build? or it still works because it will build only the APKBUILDs with a newer version?

ollieparanoid:
Wow, thanks a lot for getting back to this PR and reviewing it!
I am not sure what you mean with recent `--force` change though, but I'll make sure that this still works when rebased on master before merging it.

drebrez:
@ollieparanoid Probably I've already answered myself, I was thinking about the change that a package doesn't get rebuilt if the version is the same (require `--force`), but in the checksum test we actually want to build the packages where the version is newer, so everything should be fine.

PR #983, opened 2017-12-06
Add support for i9505

rrooij:
Current status:

* Only works with exporting as recovery zip and flashing it with TWRP
* Currently boots Weston
* SSH access is possible on USB
* Touch screen does not work
* Screen has a red glare


aports/device/linux-samsung-i9505/APKBUILD
```diff
@@ -0,0 +1,89 @@
+# Kernel config based on: arch/arm/configs/(CHANGEME!)
```
ollieparanoid: Please replace the `(CHANGEME!)` part.

aports/device/linux-samsung-i9505/APKBUILD
```diff
@@ -0,0 +1,89 @@
...
+
+pkgname="linux-samsung-i9505"
+pkgver=3.4.112
+pkgrel=0
+pkgdesc="Samsung Samsung Galaxy S4 International kernel fork"
```
ollieparanoid: Samsung once is enough there (I know it's automatically generated).

rrooij:
Thanks for the comments. I will fix it ASAP.

rrooij:
Done 👍

rrooij:
Another thing: the demo `weston-simple-egl` runs fine. This means that OpenGL ES seems to be working fine.

PR #984, opened 2017-12-07
i9505: add touch screen support

rrooij:
Add support for the touch screen by adding a udev rule for the touch
screen device.


aports/device/device-samsung-i9505/90-android-sec_touchscreen.rules
```diff
@@ -0,0 +1,5 @@
+# use 'weston-calibrator' to calibrate the touchscreen!
+SUBSYSTEM=="input", ATTRS{name}=="sec_touchscreen", \
+ENV{ID_INPUT}="1", ENV{ID_INPUT_TOUCHSCREEN}="1" \
+ENV{WL_CALIBRATION}="0.982113 -0.006129 1.983124 -0.002560 1.000800 -6.012451"
```
ollieparanoid: Oftentimes we don't need the `ENV{WL_CALIBRATION}` line, could you check if it works without it?

```
SUBSYSTEM=="input", ATTRS{name}=="sec_touchscreen", \
ENV{ID_INPUT}="1", ENV{ID_INPUT_TOUCHSCREEN}="1"
```
If it does work without it, please also remove the comment at the top, as there's no need to calibrate your device.
If it does *not* work without it, just reply to this comment.

aports/device/device-samsung-i9505/APKBUILD
```diff
@@ -1,17 +1,21 @@
 pkgname="device-samsung-i9505"
 pkgdesc="Samsung Samsung Galaxy S4 International"
-pkgver=0.1
-pkgrel=0
+pkgver=0.2
+pkgrel=1
```
ollieparanoid: The `pkgrel` gets reset to `0` when the `pkgver` increases.
Increasing the `pkgrel` is okay here, because you've adjusted something else than `APKBUILD` (or `patch` files).

(In general, when you **only** change the `APKBUILD` or patches, then only increase the `pkgrel` and keep the `pkgver` unchanged.)

rrooij:
Done. Thanks for your comments!

ollieparanoid:
Almost! What we need is:
```shell
pkgver=0.2
pkgrel=0
```

Also, just to make sure, can you confirm that the touchscreen is still working without the calibration?

rrooij:
Thanks, I mixed the two up.

Touch still works without the calibration.

PR #986, opened 2017-12-07
Add 50 MB of free space to the root partition

### LINKED ISSUE #928: 'No space left on device' error during pmbootstrap.py install
Trying to build for i9100 fails when it gets to Fill Install Blockdevice

[21:26:43] ERROR: Command failed: (native) % cd /mnt/rootfs_samsung-i9100 && cp -a efs boot sys srv run root media sbin bin usr lib dev tmp etc mnt var proc /mnt/install/

cp: write error: No space left on device
cp: write error: No space left on device
cp: write error: No space left on device

ollieparanoid:
Fixes #928.

@drebrez, I know you've put a lot of thought into the calculation, feel free to suggest better adjustments.
@PureTryOut: this should also fix the no space left issue for you.


drebrez:
@ollieparanoid i'm fine as a temporary workaround, but at some time we may probably face the problem again. 
when I investigated it, if I remember correctly this has something to do with the filesystem where the pmbootstrap work folder is stored, something related to holes in ('sparse') files, internal fragmentation, indirect blocks,...
like explained here => https://unix.stackexchange.com/a/106277

PureTryOut:
Should be fine for now, still I'd like to see LVM as a long-term solution.

drebrez:
@PureTryOut the LVM solution will not solve this issue. Even with LVM we will have to create an image file and write the LVM layer into it.

ollieparanoid:
@drebrez: yeah, you're right, it's more of a workaround. I think the root cause to our problem is, that `du` doesn't really give us the actual folder size, see my comment from https://github.com/postmarketOS/pmbootstrap/pull/760#issuecomment-337290826:
> [...] I still get a different result on my PC compared to what you have.... I'm starting to doubt that it is possible to use `du` (which comes from "disk usage" apparently) to display the actual folder size, as if you would put it into an image file (although it says in the manpage that you can do just that with `--apparent-size` - it clearly does not work).
>
> However, I think the roughly estimated size we get from `du` is good enough for our use case. So I've adjusted the testcase to be more tolerant with the output.

@PureTryOut: in case it is not obvious, we're generating an image file with just about the size the chroot has - in contrary to making it as big as the partition where we want to flash it to. That way we don't need to know how big it will be in advance, and save disk space. On the first boot, the initramfs automatically resizes the partition to the maximum available space for that partition (the full partition space basically, @drebrez contributed the great code that does this).

So if you run out of space right now with your `system` partition, you can use the `userdata` partition instead, which might be a lot bigger (system: Android installation, userdata: photos, apps, everything from the user). LVM will make it possible to combine both partitions (and possibly more partitions, such as cache), and therefore use the maximum available space.

I'm merging this, as @drebrez mentioned it's fine for him as temporary workaround (ideas on how to resolve this for good are also welcome - but some experiments showed that walking through the folder in Python and calculating the size with that is a lot slower - but it might be worth it).

PR #990, opened 2017-12-08
[hildon-ui] Rotate display and touchscreen coordinates

### LINKED ISSUE #819: Rotate screen for Hildon by default
Hildon looks strange, when the screeen is not rotated by 90 degrees for most phones.

@montvid wrote in https://github.com/postmarketOS/pmbootstrap/pull/713#issuecomment-337570371:
> I propose adding xrandr and xinput apks to hildon and adding lines to sh:
> `xrandr -o right`
> `xinput set-prop '*Your touchscreen name' 'Coordinate Transformation Matrix' \`
> `0 1 0 -1 0 1 0 0 1`
> *if someone knows how to parse the name of the touch device please share.
> https://wiki.ubuntu.com/X/InputCoordinateTransformation

drebrez:
This adds an X11 configuration file (`x11-fbdev-rotate.conf`) to rotate the display by 90° degrees clockwise.

In order to rotate the touchscreen, in the `xinitrc_hildon.sh` script, it gets the `deviceinfo_dev_touchscreen` variable from the `/etc/deviceinfo` file and perform a query with `udevadm` in order to find out the corresponding name of the touchscreen device.
With the touchscreen device name it calls `xinit` to set the `Coordinate Transformation Matrix` in order to rotate the touch.
This part works but is a bit fragile and needs to be tested on various devices.
An alternative solution would be to find a way to set the coordinate transformation matrix using the `/dev/input/eventX` directly (with udev rule is possible).

Closes #819

### DIFF
--- a/aports/main/postmarketos-ui-hildon/xinitrc_hildon.sh
+++ b/aports/main/postmarketos-ui-hildon/xinitrc_hildon.sh
@@ -11,6 +11,14 @@ done
 # Start dbus and export its environment variables
 eval "$(dbus-launch --sh-syntax --exit-with-session)"
 
+. /etc/deviceinfo
+if [ ! -z "$deviceinfo_dev_touchscreen" ]; then
+	# walk through all the attributes of the parent devices and reverse the
+	# output to find and extract the name attribute of the top parent device
+	touch_name=$(udevadm info -a -n "${deviceinfo_dev_touchscreen}" | tac | grep -m1 "ATTRS{name}" | cut -d '"' -f2)
+	xinput set-prop "$touch_name" "Coordinate Transformation Matrix" 0 1 0 -1 0 1 0 0 1
+fi
+
 # Start X11 with Hildon
 export LC_MESSAGES=en_US.UTF-8
 exec hildon-desktop
--- a/aports/main/postmarketos-ui-hildon/APKBUILD
+++ b/aports/main/postmarketos-ui-hildon/APKBUILD
@@ -1,6 +1,6 @@
 pkgname=postmarketos-ui-hildon
 pkgver=1
-pkgrel=3
+pkgrel=4
 pkgdesc="(X11) Lightweight GTK+2 UI (optimized for single-touch touchscreens)"
 url="https://github.com/postmarketOS"
 arch="noarch"
@@ -11,11 +11,16 @@ depends="
 	hildon-theme-alpha
 	mesa
 	xorg-server
+	xinput
 "
 makedepends=""
 install="$pkgname.post-install"
 subpackages=""
-source="start_hildon.sh xinitrc_hildon.sh $install"
+source="
+	start_hildon.sh
+	xinitrc_hildon.sh
+	x11-fbdev-rotate.conf
+"
 options="!check"
 
 package() {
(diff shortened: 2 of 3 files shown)


ollieparanoid:
I didn't test it on any device yet.

Some thoughts:
* This rotates the screen for all devices. I think at least the N900 and Qemu doesn't need rotation.
* Can we do the screen rotation (not touch screen) with `xrandr` instead of writing a config file? Then we could do it for Hildon only, and also make it depend on a `deviceinfo` setting.

Thank you very much for this contribution @drebrez!

drebrez:
@ollieparanoid on my device it says that `xrandr` is not supported, maybe it's because of `mesa-dri-swrast`? no idea.
You are right, we need to find a way to distinguish the devices that doesn't need a rotation, maybe looking and the `deviceinfo_screen_width` and `deviceinfo_screen_height` variables in the `deviceinfo`? only when height is greater than width we need to rotate it

ollieparanoid:
I guess it's because of the framebuffer driver. Which command have you used for rotation?
And we might need to enable the [`ShadowFB` option](https://linux.die.net/man/4/fbdev), although that should be on by default. As I understand it, it doesn't directly write to the framebuffer, but to a copy, and therefore is able to rotate it that way (I'm guessing this from the "Rotate" description: *Implies use of the shadow framebuffer layer.*).

> only when height is greater than width we need to rotate it

Great idea!

drebrez:
Ok I found something in the logs:
```
localhost:~$ cat /var/log/Xorg.0.log | grep -i randr
[    15.349] (II) FBDEV(0): using driver rotation; disabling XRandR
[    15.349] (--) RandR disabled
```
this allows to use the `Option "Rotate" "CW"` in the config. If somehow I manage to disable the driver rotation, then I think I can use the xrandr. I've read somewhere that are mutually exclusive.

ollieparanoid:
Good research you did there! I have a new idea.
When searching the internet for Hildon and rotate, threads like [this one](https://answers.microsoft.com/en-us/mobiledevices/forum/mdnokian/full-portrait-desktop-on-the-n900/a4d88b47-4b5a-44c1-b70d-08a547e9187b?auth=1) turn up (*don't you love how after buying Nokia the N900 Linux related stuff is now on a Microsoft answers site?*).
This makes me wonder if there's a magic command that we can send to Hildon, that makes it rotate the screen by itself, without using any X API.

@NotKit, do you know if that is possible?

drebrez:
@ollieparanoid I've found this in the source code => https://github.com/fremantle-gtk2/hildon-desktop/blob/master/src/main.c#L798-L811

ollieparanoid:
That sounds promising! Yeah with some more digging in the source, we might also find the answer 😃

drebrez:
@ollieparanoid https://wiki.maemo.org/Community_SSU/Features ?

ollieparanoid:
That seems to be for enabling/disabling rotation in general. I am not sure if it correctly picks up your rotation sensor, but you could try it.
```shell
gconftool-2 -s /apps/osso/hildon-desktop/ui_can_rotate -t bool true
gconftool-2 -s /apps/osso/hildon-desktop/desktop_orientation_lock -t bool false
gconftool-2 -s /system/systemui/tklock/auto_rotation -t bool true
```

aports/main/postmarketos-ui-hildon/xinitrc_hildon.sh
```diff
@@ -11,6 +11,14 @@ done
 # Start dbus and export its environment variables
 eval "$(dbus-launch --sh-syntax --exit-with-session)"
 
+. /etc/deviceinfo
+if [ ! -z "$deviceinfo_dev_touchscreen" ]; then
+	# walk through all the attributes of the parent devices and reverse the
+	# output to find and extract the name attribute of the top parent device
+	touch_name=$(udevadm info -a -n "${deviceinfo_dev_touchscreen}" | tac | grep -m1 "ATTRS{name}" | cut -d '"' -f2)
```
drebrez: Another possible way to get the name of the touchscreen from the `/dev/input/eventX` in the deviceinfo is by changing it to `/devices/virtual/input/inputX` and look into the `/proc/bus/input/devices` file.

cat  /proc/bus/input/devices

drebrez: or even better:
```
localhost:~$ source /etc/deviceinfo
localhost:~$ cat ${deviceinfo_dev_touchscreen/\/dev/\/sys\/class}/device/name
sec_touchscreen
```

ollieparanoid: This is so much better than walking through the attributes etc, great solution!

aports/main/postmarketos-ui-hildon/xinitrc_hildon.sh
```diff
@@ -11,6 +11,14 @@ done
...
+	xinput set-prop "$touch_name" "Coordinate Transformation Matrix" 0 1 0 -1 0 1 0 0 1
```
michitux: Doesn't this override an already set calibration matrix? On some devices like the n900, the touchscreen is mirrored with the default matrix, on others like the Samsung Note 8.0 tablet, this rotation is needed with the default output rotation. So if this overrides the default matrix, this is breaking touch input on these devices.

PR #992, opened 2017-12-09
i9505: add wifi firmware

rrooij:
Add the wifi firmware for bcm4335. This is the blob that is being used
in LineageOS/CyanogenMod 14.1.


rrooij:
By the way, according to [this](https://wireless.wiki.kernel.org/en/users/drivers/brcm80211) the bcm4355 wireless Broadcom chipset is supported in the mainline kernel without any proprietary blobs.

aports/device/device-samsung-i9505/APKBUILD
```diff
@@ -1,12 +1,12 @@
 pkgname="device-samsung-i9505"
 pkgdesc="Samsung Galaxy S4 International"
-pkgver=0.3
+pkgver=0.4
 pkgrel=0
```
ollieparanoid: Here only the APKBUILD was modified, not the sources it gets built from.
```shell
pkgver=0.3 # stays the same
pkgrel=1 # gets increased
```

ollieparanoid:
> By the way, according to [this](https://wireless.wiki.kernel.org/en/users/drivers/brcm80211) the bcm4355 wireless Broadcom chipset is supported in the mainline kernel without any proprietary blobs.

As I understand it, they mean, that the kernel driver does not contain a long sequence of bytes in a `.c` or `.h` file, which is essentially a blob disguised as source code (there's the linux libre kernel that has a script which can automatically clear such blobs).

But you still need the proprietary firmware blob, that runs on the wifi chip. From the same page:
> Both brcmsmac and brcmfmac drivers require firmware files that need to be separately downloaded. Firmware is available from the Linux firmware repository at: [...]

aports/device/device-samsung-i9505/APKBUILD
rrooij: Ah, clear. I thought of it as a newer version of the package and therefore in need of a new pkgver. Thanks for the info.

rrooij:
Thanks for your remarks.

I think that the different conventions regarding pkgver and pkgrel can be confusing indeed. Documentation regarding some conventions we use for the APKBUILD's would be a nice addition.

ollieparanoid:
You're welcome!

Alright, it's documented here now: https://wiki.postmarketos.org/wiki/Create_a_package#When_should_pkgver.2Fpkgrel_get_increased.3F

PR #993, opened 2017-12-10
Update KDE Frameworks to 5.41.0

PureTryOut:
Some Plasma packages (e.g. `plasma-workspace` and others) will need rebuilding to use the updated KDE Frameworks.


ollieparanoid:
Thanks a lot! Could you please increase the `pkgrel` for the packages, that need to be rebuilt?

PureTryOut:
Is that what Alpine does? Sure I guess, I'll see if I can do that tomorrow.

PR #995, opened 2017-12-10
Add `devicepkg-dev` which generate the touchscreen udev rule based on the deviceinfo

drebrez:
Including `devicepkg-dev` package in `makedepends` of the device APKBUILD, allows to call:
- `devicepkg_default_build` during the `build` function which reads the `deviceinfo` and generate the udev rule for the touchscreen
- `devicepkg_default_package` during the `package` function will install the generated file

It will use the `deviceinfo_dev_touchscreen` and `deviceinfo_dev_touchscreen_calibration` variables of the `deviceinfo` file and creates a rule like this:
```
SUBSYSTEM=="input", ENV{DEVNAME}=="/dev/input/event2", \
ENV{WL_CALIBRATION}="1.043478 0.000000 -5.217407 0.014562 0.956938 2.798035", \
ENV{ID_INPUT}="1", ENV{ID_INPUT_TOUCHSCREEN}="1"
```

(PS: for the samsung i9070 I removed the calibration since it's not necessary)

commands used for testing:
```shell
sudo rm -f ~/.local/var/pmbootstrap/packages/*/devicepkg-dev-*
sudo rm -f ~/.local/var/pmbootstrap/packages/*/device-samsung-i9070-*
pmbootstrap index
pmbootstrap -y zap

pmbootstrap checksum devicepkg-dev && pmbootstrap build devicepkg-dev --arch armhf
pmbootstrap checksum device-samsung-i9070 && pmbootstrap build device-samsung-i9070 --arch=armhf

pmbootstrap -y zap
pmbootstrap chroot -r -- apk add device-samsung-i9070
pmbootstrap chroot -r -- cat /etc/udev/rules.d/90-touchscreen-dev.rules
```


aports/main/devicepkg-dev/devicepkg_default_build.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+if [ ! -f ./deviceinfo ]; then
+	echo "deviceinfo file missing!"
```
ollieparanoid: How about:
```shell
echo "NOTE: $0 is intended to be used inside of the build() function"
echo "of a device package's APKBUILD only."
echo "ERROR: deviceinfo file missing!"
```

aports/main/devicepkg-dev/devicepkg_default_package.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+if [ ! -f ./src/90-touchscreen-dev.rules ]; then
+	echo "90-touchscreen-dev.rules file missing!"
```
ollieparanoid: How about:
```shell
echo "NOTE: $0 is intended to be used inside of the package() function"
echo "of a device package's APKBUILD only."
echo "ERROR: 90-touchscreen-dev.rules file missing!"
echo "Please call devicepkg_default_build() inside your APKBUILD's"
echo "build() function first."
```

drebrez:
@ollieparanoid thanks for the review, yes I also wanted to improve the error messages. 
I also agree that we should change all the packages and make it the default also for the wizard.

ollieparanoid:
I did not notice it yet, good that you point it out. Passing `$pkgdir` as argument seems to be less fragile to me, than the `find`. And we can save one line 👍

drebrez:
@ollieparanoid to avoid random problems with the creation of the file depending on the current folder (as explained in chat, with --force it runs in a different location) I've added some arguments to both calls.
I also changed it as the default in the new device wizard.
There is only a problem left, `devicepkg_default_build ` creates the udev rule file only if the `deviceinfo_dev_touchscreen` variable is specified, but `devicepkg_default_package` always expect the existence of the generated file. How we proceed there?

extra: is it ok the file name `90-touchscreen-dev.rules`? I remember a discussion sometime ago about how to rename that file

ollieparanoid:
Does always passing `$startdir`, and constructing the other folders from that, work? [See also](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#Functions):
> Note: All functions should consider the current working directory as undefined, and should therefore use the abuild-defined directory variables to their advantage.

---

> `devicepkg_default_build` creates the udev rule file only if the `deviceinfo_dev_touchscreen` variable is specified, but `devicepkg_default_package` always expect the existence of the generated file. 

Not all devices have a touch screen (e.g. the Ouya). So I think it makes sense to only try to install the file if it was generated before.

What do you think about also installing the `deviceinfo` file? This one must always exist, so we can use it as check if the script was called inside the `APKBUILD`. And it makes the `package()` function shorter, because we don't need to specify it there.

drebrez:
@ollieparanoid as we discussed in chat, here what I've changed:

 - use `$startdir` and `$pkgname` arguments for both calls
 - install `deviceinfo` directly from package script
 - rules file renamed to be generic
 - rules file installed only if present (not needed for some devices)
 - updated default template

I've also added the new `deviceinfo_dev_touchscreen_calibration` variable in the template, we will also need to update the wiki accordingly.

ollieparanoid:
Thanks for updating it! I've made some further adjustments:
* Fix the testcase (now it does not fail and actually tests if `devicepkg-dev` builds and its functions run through)
* Removed the `_default_` in `devicepkg_build` and `devicepkg_package`. In my opinion, that made it look like it works similar to `default_prepare` - but in reality it works completely different since we're not calling a function available in the scope of the APKBUILD, but we're executing another script. So this version is shorter and may be a bit less confusing.
* Wrote a [wiki page](https://wiki.postmarketos.org/wiki/Device_specific_package), and linked to it in the deviceinfo (i9070 and aportgen). That way we can explain what the `devicepkg_build` function does without requiring the user to dig through the source code.

I hope you like the changes, and if you don't feel free to revert them. I've tested that everything is working as expected, before and after these adjustments, so if it's all cool to you we can merge it in my opinion.

drebrez:
@ollieparanoid looks good and the changes make sense 👍 . Also thanks for fixing the test case.

ollieparanoid:
You're welcome!
> I've also added the new `deviceinfo_dev_touchscreen_calibration` variable in the template, we will also need to update the wiki accordingly.

Do you have time and want to this?

drebrez:
@ollieparanoid wiki updated, now we need to update all the existing device packages :sweat_smile:

PR #996, opened 2017-12-10
Add qt5-qtbase with OpenGL ES2 enabled

### LINKED ISSUE #994: Alpine's QT5 is built with desktop OpenGL, but we need OpenGL ES2
Reported by @NotKit in `#postmarketOS`.
The quickest solution (to get a proof of concept plasma shell working on a real device) would be if we fork the package and adjust it (with pkgrel=9999), then we could look into upstreaming that change if Alpine agrees on it. We did the same approach for busybox and sdl2, and that worked out well.

PureTryOut:
Resolves #994.

Someone please test this on an actual device, as I could only test it in Qemu which worked fine without OpenGL ES2 as well. If this is confirmed working, we can upstream this pretty quickly to Alpine hopefully.

Note that qt5-qtdeclarative needed rebuilding, otherwise `kwin` wouldn't even launch.


PureTryOut:
Well, the tests fails on that new Qt version check. I can also just keep the `qt5-qtbase` `$pkgrel` just 1 higher than Alpine's, might work better.

ollieparanoid:
Thank you very much for the PR! I've updated the testcase, so it looks at `_pkgver` when `pkgver=9999`.

Testing on a real device this is good in any case (I'll start rebuilding the repo as soon as the KDE PR is merged), but it's not sure that Alpine would accept the change the way we made it. It makes sense for embedded devices, but not necessarily for `x86_64` desktops, we'll need to talk to them once we've tested that it works (maybe specify the flag depending on the arch).

PureTryOut:
I see you've removed the "desktop OpenGL" [here](https://github.com/postmarketOS/pmbootstrap/pull/996/commits/a62a789e45751156e3e34607d66011ed54751428#diff-cb632e66023b4c53d761b54652f56903L93), did it not work for you having them both compiled? I hoped having both the options compiled in would mean it would change behaviour depending on what was needed. If that worked, the change could be upstreamed quite easily. Of course just replacing desktop OpenGL with ES2 will not be accepted.

ollieparanoid:
Yeah, as I read the output of the configure it only allows one of them. And the second one overrides the first one, so I've kept only one to avoid confusion. (I've also mentioned in the commit message, but sorry if that was confusing.)

I'll try to find out how it works in the source to make sure.

ollieparanoid:
Here's what `./configure --help` says´:
```
  -opengl <api> ........ Enable OpenGL support. Supported APIs:
                         es2 (default on Windows), desktop (default on Unix),
                         dynamic (Windows only)
```

So I've found a `config.summary` file, that gets generated during the build. Here's how it looks like with what we have on `master`. In fact, it looks like it would allow both Desktop OpenGL and OpenGL ES.
```
...
  EGL .................................... yes
  OpenVG ................................. no
  OpenGL:
    Desktop OpenGL ....................... no
    OpenGL ES 2.0 ........................ yes
    OpenGL ES 3.0 ........................ yes
    OpenGL ES 3.1 ........................ yes
  Session Management ..................... yes
...
```

But when I built with the following `./configure` options...
```shell
        -opengl desktop \
        -opengl es2 \
```

...the `config.summary` stays the same, which means the `-opengl es2` simply overwrites the `-opengl desktop`.

I've also tried `-opengl es2,desktop`, which aborts the build with:
```
ERROR: Invalid value 'es2,desktop' supplied to command line option 'opengl'.
```

So my conclusion is, that it' not possible to have both :/

PR #999, opened 2017-12-11
Add qtwebengine & qtwebchannel from luna work

magmastonealex:
Qtwebengine takes a very long time to build, and has stabilized for now in terms of needed patches, etc. Getting it into the binary repository will save everyone quite a bit of time who needs to work with it.

There's a fair amount of luna-specific code here (taken nearly verbatim from webos-ports repository), but everything still builds without it (remove everything after `no-pvalloc.patch` to get a build with nothing luna-specific). It also doesn't seem like anything that will break other use cases.

qtwebchannel is in alpine aports, but qtdeclarative/qtquickcontrols2 were not added to the makedepends, which means qtwebengine can't build with it. It's a trivial change to upstream, but I don't know who to contact or how to make that happen. I followed @PureTryOut's pattern used for qt5-qtbase (pkgver=9999, _pkgver=5.9.3)

@ollieparanoid this one will take a while to build :) On my 4-core Intel box with 8gb of ram, it took ~6 hours the first time through. It's effectively all of Chrome.


ollieparanoid:
Two remarks:
* Please remove the original maintainer comment, we don't want him to get annoyed with questions about postmarketOS packaging he didn't even do. If you want to credit the maintainer, you can do so in the commit message (although I think it's obvious that this one is based on Alpine because of the pkgver=9999).
* These are LOTS of patches. What do you think about grouping them in subdirectories? To do that, you'd need to:
  * move the files to appropriate subfolders inside `qt5-qtwebengine` (e.g. patches-musl, patches-luna, ...)
  * change the path to the specific patches in the `source` variable
  * change the following `prepare()` function:
```shell
prepare() {
	for i in "$startdir"/patches*; do
		ln -sv "$i" "$srcdir/$(basename "$i")"
	done
	default_prepare
	mkdir -p "$srcdir"/build
}
```
  * run `pmbootstrap checksum`
  * run `pmbootstrap build` and abort after all patches apply successfully

Thank you very much for making this PR so quickly!
I hope I can compile it :)

magmastonealex:
@ollieparanoid  

I've removed the maintainer lines, I forgot those were there.

I agree with splitting the patches up. I tried at one point, but using a symlink hadn't occured to me. The patches all still apply in my latest commit.

ollieparanoid:
Here's my progress:
* `x86_64`: compiled, and is available in the repo now (I did not test using it, that would be helpful)
* `armhf` failed at the very end when linking `libQt5WebEngineCore` (might be distcc related?):
```
linking ..../lib/libQt5WebEngineCore.so.5.9.3
collect2: fatal error: ld terminated with signal 11 [Segmentation fault], core dumped
compilation terminated
distcc[12345] ERROR: compile (null) on localhost failed
```
* `aarch64`: hangs at forever (left it there for hours) on `[327/237] LINK gn` fairly early for some reason.

I have yet to find out why `armhf` and `aarch64` are failing. If someone has the time, it might be helpful if someone tried to compile it for `armhf` and `aarch64` too and maybe looked into more why it fails (if it's also failing). I gave it 50 GB of SWAP just in case, so that shouldn't be the problem.

Issue #655, opened 2017-09-29
Build matchbox-keyboard for Alpine


montvid:
It would be very good to have an on screen keyboard to use with xfce/mate or other X11 window managers but Alpine does not package one so far I know. My research shows that there is one widely used keyboard called matchbox-keyboard and there is an updated source for it. Would anyone help me compile it and test or create a package for Alpine? What other on screen keyboards can I install in Alpine?
https://www.modmypi.com/blog/matchbox-keyboard-raspberry-pi-touchscreen-keyboard
http://git.yoctoproject.org/cgit.cgi/matchbox-keyboard/tree/

ollieparanoid:
(Maybe we can extend osk-sdl to work in X11 and wayland - that will be tricky though - at some point.)

Could you make a work in progress `APKBUILD` for the `matchbox-keyboard`, which shows where you get stuck?

montvid:
I just don't have any idea what should I install in alpine to build this apk. In the readme it is said autotool but I can't find it in Alpine.

ollieparanoid:
It mentions, that you need to run `autogen.sh` when running from svn (or git in our case?), which generates the `./configure` script. Looking at `autogen.sh` it shows the `autoreconf` command, which is part of the [`autoconf`](https://pkgs.alpinelinux.org/contents?file=autoreconf&path=&name=&branch=&repo=&arch=) package in Alpine.

(Stable releases are preferred over using git commit hashes as versions, if there are any.)

craftyguy:
Looks like they have a release tagged here: https://git.yoctoproject.org/cgit.cgi/matchbox-keyboard/refs/

The readme lists some requirements for building: `xlibs`, `xft`, `libfakekey` and `expat`.

Welcome to the fun world of porting packages to Alpine, because you'll likely need to port some dependencies too! Alpine package search is terrible, so some of these missing ones could be there somewhere I just wasn't able to insert enough wildcards in the search to find it.

#### libfakekey
libfakekey doesn't seem to exist in Alpine, but [there's a PKGBUILD in Arch that you could use as  a starting point.](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/libfakekey). Luckily the one dependency this package has, `libxtst`, appears to be in Alpine!

#### xlibs
I can't find 'xlibs' in Alpine, but there is a `libx11` which *should*, I *think*, provide 'xlibs'.

[...]

PR #1001, opened 2017-12-12
aports: packaged matchbox-keyboard

drebrez:
I packaged `matchbox-keyboard` and tested it on an actual device (with hildon ui).
I've no idea why but it shows the keyboard 3 times on screen.
I've opened a terminal and tried to touch some keys and it wrote them on the terminal.
At least something works.

Closes: #655


rrooij:
Works pretty good on my i9505 with XFCE4. Thanks for this! It only starts once fortunately, not three times.

ollieparanoid:
More information about config guess / config sub [here](https://wiki.postmarketos.org/wiki/Troubleshooting#Packaging_for_aarch64:_..2Fconfig.guess:_unable_to_guess_system_type) (I've tested building for aarch64, simply because it's an older codebase and autotools based).

Thank you very much for this contribution @drebrez! 🎉 🚢 

If someone could add a photo, that would be cool.

rrooij:
Photo

drebrez:
@rrooij whoaa 🎉 looks good on xfce4
@ollieparanoid thanks for fixing up the packaging
on hildon maybe it's due the landscape mode, or because i'm starting it manually on the command line? no idea, @rrooij does the keyboard popup automatically when it detects an input field?

rrooij:
@drebrez it always stays on in normal mode. I tried running it in daemon mode but it doesn't seem to recognize GTK2 text inputs as far as I could test.

Issue #678, opened 2017-10-02
libhybris support?


NotKit:
libhybris is a compatibility layer for loading Android libraries (compiled against Bionic) by glibc-based Linux applications. However, with some changes it can be compiled with musl as well. Caveats yet to be found, but I was able to utilize Android OpenGL libraries on at least two devices (Moto Droid 4 and Moto Z Play).

Does it make sense to package libhybris for postmarketOS, and if yes, what's the best way to do it? Building libhybris requires extracted Android headers, ideally from device's Android tree. SailfishOS compiles libhybris for each device, while UBPorts/Hallium tried to keep single libhybris build for multiple ones, but now provide at least two libhybris packages - normal version and libhybris-caf for Qualcomm devices due to changes in gralloc headers, making it binary incompatible.

Another problem than would be packaging device-specific Android bits. SFOS uses files from CyanogenMod/LineageOS install, but overrides init scripts and libs. Halium instead builds minimal Android rootfs and runs it inside LXC.

ollieparanoid:
+CC: @bhush9 (bshah), @z3ntu from Halium

First of all, great research!

### Proprietary blobs in userspace
My personal opinion, as written in `#postmarketOS`:
> I don't like them and I prefer solutions like freedreno or tinygles. But I think as long as we keep them optional, it's better to allow them to get more people to contribute to the project instead of forbidding them. I'll write something like this down to the wiki, a "guidelines for proprietary blobs" page or something (as initial draft open for discussion as always).

### Chroot instead of libhybris?
I'm wondering if it is possible to run `ldd` on the blob in question, and then copy the blob with all libraries it links against (including bionic libc) to a folder on the device, and run it in a chroot. As I see it now, it would save us from the complicated compiling steps, the libhybris layer, and maybe we wouldn't even need to worry about using a specific kernel. But that would be too easy, so where is the catch?

### Hardening
I would like to restrict the blobs access as much as possible. This should be relatively easy with `firejail` for example (packaged in Alpine). I would chroot it, restrict network access, apply seccomp filters so it can do nothing but be the graphics driver.

### Packaging libhybris
If we really go down that path:
* having two versions of libhybris is better than one per device in my opinion, we could package both (in one aport, with a subpackage for `libhybris-caf`)
* I would prefer not running Android in LXC, but using native libraries/programs like SFOS (whatever that means)

z3ntu:
[...]

PR #1002, opened 2017-12-12
Add feature/hybris branch for experimenting with libhybris-based HW adaptations (issue #678)

NotKit:
libhybris is a compatibility layer that allows to use Bionic-based Android libraries and drivers in glibc systems.

The way it's implemented has no particular dependency on glibc though, and most of changes required for musl were compilation fixes. Yet I had to disable some hooks due to hooked functions missing in musl, which might cause issues if bionic versions of that functions don't work properly with musl structs/without proper bionic initialization, but so far OpenGL acceleration has been tested to work on Moto Droid 4, Moto Z Play and OnePlus 2.

Using Android libraries in most of cases requires base/core Android system to be up and running, which includes its init system, property service, logger, servicemanager, device specific services and so on. It's mostly lightweight though, since this is limited to native code, so Java VM and higher level Android components are never started.

Currently there are two approaches to packaging Android base. libhybris requires patched versions of some Android libraries and components. Mer (SailfishOS base) uses customized Android tree (https://github.com/mer-hybris) to build them, then packages modified libs and init scripts into droid-hal-$device packages. Unmodified files are used as-is by mounting /system partition from corresponding CyanogenMod/LineageOS install. Android base runs in same root with "Linux" system.

Halium instead builds minimal self-content Android rootfs with patched components and runs it inside LXC. This is a cleaner solution overall. While it has halium-5.1 branch, only halium-7.1 branch is actively maintained so far, so it's way easier to build for newer devices than for older ones, especially if they don't have LineageOS 14.1 port.

Due to this I went packaging "droid-hal-maserati" from SailfishOS community port for Droid 4 and added simple OpenRC script to start droid-hal-init (renamed Android init). The device has old 3.0.8 kernel, so I wasn't sure if it can start LXC at all, but LXC website tells that it requires at least 2.6.32.

The issue is that libhybris must be built with Android headers extracted from device, or at least Android headers matching Android version on device and SoC vendor. I'm not sure how to go around creating multiple packages like libhybris-4.4, libhybris-7.1, libhybris-7.1-caf and so on. What's worse, software utilizing libhardware through libhybris (for example, interfacing with gralloc and hwcomposer) currently has the same kind of dependency on Android headers it's built with, so with current approach there will be say kwin-4.4, kwin-7.1, xf86-video-hwcomposer-4.4 and so on.

About the current status of Droid 4 port, which I included as example. It boots if encryption is disabled, osk-sdl won't display yet during initramfs stage, but will display after starting and finishing Xorg with xf86-video-hwcomposer backed. It seems PowerVR driver does some calls to activate/reset framebuffer. It can start Xorg server with mentioned driver and utilize 3D acceleration via libhybris (tested with hildon-desktop). Once Qt QML JIT bug fixed on armhf, it should be able to start LuneOS with 3D acceleration as well, if needed middleware is packaged. Wi-Fi works when wpa_supplicant is started with wext backend.


ollieparanoid:
I've merged this into a new `feature/hybris` branch, as you suggested in the title. And I've written down my thoughts about proceeding further in #678.

> About the current status of Droid 4 port [...]

Would you mind creating wiki pages for the devices you have been experimenting with, which contain links to the packaging you did for them? If you don't have a version that works without libhybris for them, you could add them to the wiki as "Not booting", and write something like "libhybris experiments (boots with 3D acceleration, not in master branch)" in the description.

**@NotKit, thank you very much for your work and for the detailed and well structured explanation. It's obvious that you put a lot of effort into this!**

PR #1004, opened 2017-12-12
build necessary logic: look at the package with the highest version

### LINKED ISSUE #955: Some packages will still build despite the same version being on the binary repo
Examples: ```armhf/breeze-icons-5.40.0-r0.apk```,  ```x86_64/binutils-armhf-2.28-r3.apk```, ```armhf/kcoreaddons-5.40.0-r0.apk```

I did pmbootstrap zap, update, then index, but it STILL tries to locally build these rather than fetching them from binary repo. Ubuntu 17.10, pmbootstrap up to date with master.

ollieparanoid:
Fixes #955. Previously we did not look through all APKINDEXes while
looking for the package with a given name and the highest version.
This caused pmbootstrap to build packages even if they are in the
binary repo and up to date.


ollieparanoid:
@MoreRobustThanYou reported that it still doesn't work for him (thanks for testing!) - now I only need verbose logs to debug this further.

MoreRobustThanYou:
here's the verbose output for ```pmbootstrap -v build binutils-armhf```: https://pastebin.com/KaSUiQuz

ollieparanoid:
Thanks for posting the log! I've answered in new issue #1023, because this is a related, but different bug that can not be fixed with this PR.

However, this PR _does_ fix the bug that @zhuowei analyzed (and probably @z3ntu was having?) https://github.com/postmarketOS/pmbootstrap/issues/955#issuecomment-350794101:
> It seems that the check looks at the local repo first, and it's picking up my locally compiled 5.9.1-r0 and concluding that the package needs to be recompiled, without checking the online binary repo.

This means reviewing/testing would still be good 👍

magmastonealex:
I've been using this patch for a few days, it's saved me a lot of time and frustration.

I haven't yet encountered any issues where packages are being built when they shouldn't be, or packages not being built when they should be, so this patch appears to be working properly for me.

ollieparanoid:
Thanks for the feedback! 👍

PR #1012, opened 2017-12-15
Add MATE Desktop Environment

drebrez:
TODO:

- [ ] Solve loading delay (~30s) issue with `marco` (MATE window manager) 
```
mate-session[1377]: WARNING: Application 'marco.desktop' failed to register before timeout
```
- [ ] Add pmos background by default (which one?)
 - https://github.com/postmarketOS/artwork/blob/master/wallpapers/pmos.png
 - https://github.com/postmarketOS/artwork/blob/master/wallpapers/postmarketos.png
 - other?
- [ ] ?? Start in landscape mode ??


aports/main/postmarketos-ui-mate/APKBUILD
```diff
@@ -0,0 +1,21 @@
+pkgname=postmarketos-ui-mate
+pkgver=1
+pkgrel=0
+pkgdesc="(X11) MATE Desktop Environment, fork of GNOME2 (stylus recommended)"
+url="http://mate-desktop.org/"
+arch="noarch"
+license="GPL"
```
ollieparanoid: GPLv1? This is the license of the `postmarketos-ui-mate` package, so it doesn't need to match MATE's licenses (which [seem to be](https://github.com/mate-desktop) GPL2, GPL3, LGPL-2.1 depending on the component).

montvid:
I don't think I had any problems with loading delay of Mate as I was using lxdm to boot into Mate. Mate gui on Galaxy S2 is cool, thanks drebrez.

drebrez:
@ollieparanoid which license should we use by default for postmarketos packages? (It should be documented somewhere also)
@montvid I'll try to run MATE through lxdm to see if it solves that issue.

ollieparanoid:
@drebrez: I mostly went with MIT as default license for postmarketOS-specific packages, because that seems to be what Alpine usese for most of their Alpine-specific packages (e.g. `alpine-base`).

(pmbootstrap is GPLv3+.)

aports/main/postmarketos-ui-mate/APKBUILD
drebrez: @ollieparanoid 
postmarketos-ui-hildon `GPL3+`
postmarketos-ui-luna `GPL3+`
postmarketos-ui-plasma-mobile `GPL3+`
postmarketos-ui-weston `GPL3+`
postmarketos-ui-xfce4 `GPL3`

should I also set it to `GPL3+`?

ollieparanoid: yeah, that makes sense 👍

drebrez:
I tried with lxdm like @montvid suggested but it still has that warning and the delay. I even tried installing alpine from scratch in a x86_64 virtual machine following the https://wiki.alpinelinux.org/wiki/MATE guide and I have the same result.
Setting the default background I've no idea how to do it and starting in landscape is not easy too (see hildon for example).
At this point I suggest to merge this PR and the other things can be done in subsequent PRs.

Imerion:
Impressive and awesome!

drebrez:
@ollieparanoid wiki updated 👍

PR #1014, opened 2017-12-16
Upstream compatibility test: also test generated aports, e.g. gcc-armhf

ollieparanoid:
We have a few packages for cross-compiling, that we automatically generate from Alpine's packages.
For example `gcc-armhf`, which is the same as Alpine's `gcc`, but built as cross-compiler.
This PR extends the `test_upstream_compatibility.py` testcase (which was introduced to verify that our QT package versions match the one from Alpine), to test these versions, too. That way we can recognize fast when we need to regenerate these abuilds.

### How to test
* Run the testcase (`pytest test/test_aportgen_device_wizard.py`)
* Modify the `pkgver` or `pkgrel` in any of the `aports` it tests (e.g. `gcc-armhf`) and run it again, it should fail.


magmastonealex:
I tested this on two machines (using `test/test_upstream_compatibility.py` not `test/test_aportgen_device_wizard.py` :) ). Worked properly on both.

The code looks good to me too.

ollieparanoid:
Thanks for reviewing and testing! 👍 🚢

Issue #878, opened 2017-11-05
JIT is disabled for QML


NotKit:
Starting kwin_wayland or any QML example (like qmlscene-qt5 with minimer, see https://public.etherpad-mozilla.org/p/faq-hadk#lineNumber=40) on armhf device currently results in crash with:
```
JIT is disabled for QML. Property bindings and animations will be very slow. Visit https://wiki.qt.io/V4 to learn about possible solutions for your platform.
Illegal instruction
```
I investigated this and found out that qt5-qtdeclarative is built without JIT support due to usage of armv6 toolchain for armhf by Alpine. Would it be possible to build and use armv7 toolchain for it instead?

PureTryOut:
Are you sure JIT being disabled has anything to do with the crash? Like it says in the next sentence, it should only be slow, not completely break. `Illegal instruction` seems to be caused by something different.

Furthermore, I thought JIT being disabled for QML was because of licensing issues, but I may recall that wrong.

ollieparanoid:
I thought the same thing, regarding that this message reads like it should not cause the crash.

`armv7` might come at some point for Alpine, see #238. But I don't think this will happen in the next few months, because they are focusing on the `3.7` release now, and that surely needs a big amount of work.

NotKit:
> Are you sure JIT being disabled has anything to do with the crash? Like it says in the next sentence, it should only be slow, not completely break. Illegal instruction seems to be caused by something different.

Yes, I used gdb to inspect the crash, and it crashes in intepreter (which wouldn't be used if JIT was enabled). Understanding why intepreter crashes is an option, but I'd prefer to get JIT working, since we don't want to use slow path for already pretty heavy stuff (Qt Quick).

ollieparanoid:
As I understand the terminology from the [V4 wiki page](https://wiki.qt.io/V4) linked in the error message, using the _interpreter_ is the opposite of using _JIT_.

There seems to be a flag to disable the JIT (also from that page): `export QV4_FORCE_INTERPRETER=1`

I guess when you run it with that, it still crashes the same way?

It's not that I'm against getting the JIT running. But I doubt that it will solve the problem and AFAIK that would only be possible in Alpine if we compile the whole `armv7` arch ourselves (because we can't cherry-pick some packages to be compiled with a `armv7` toolchain - and even if we could, that sounds like a horrible hack and would break everything for actual `armv6` devices, such as the first ipod).

[...]

PR #1022, opened 2017-12-18
qt5-qtdeclarative (QML) compile for armv7 to enable JIT and make it work at all

ollieparanoid:
Works for me on `i9100`. I was able to start the hello world demo from QML with weston (it takes seconds to show up though, even with the JIT).

Thanks a lot for @magmastonealex for providing instructions on how to do this. Could you please review this implementation (other reviews also welcome)?

PS: @magmastonealex, do you mind if I add you to the GitHub project, so you can be selected as reviewer for PRs? There are no special rights attached in votes, it's only responsibility 😉 

Workaround for #878.


magmastonealex:
@ollieparanoid  Looks good to me. The seconds to go from black window to first draw was the same for me - My theory is that it's related to mesa/swrast, but I have no good evidence for that.

And I certainly don't mind being added to the GitHub project!

PureTryOut:
I guess this is fine because `qt5-qtdeclarative` didn't work on armhf before anyway right?

ollieparanoid:
Exactly. And since this is only a workaround, the issue stays open until it is fixed properly.

PureTryOut:
Cool! Hopefully we can fix it eventually for armv6 devices as well. It might run slower than preferred, but it shouldn't crash. I'll move back to the issue then.

magmastonealex:
@PureTryOut There's a half-fix for armv6 devices, which is to build in debug mode. It's not a great option, as it bloats the package significantly, but it should let the interpreter work on armv6. The fact that it doesn't have the same issue in debug mode makes it rather hard to track down the actual root cause.

PR #1026, opened 2017-12-20
ccache: Fix for distcc cross-compiling / various improvements

### LINKED ISSUE #716: Is ccache broken in distcc compilation mode?

ollieparanoid:
* Make ccache work when cross-compiling with distcc (fix #716)
* Allow to configure the ccache size in "pmbootstrap init"
* Moved ccache stats code from pmb/build/other.py to
  pmb/helpers/frontend.py
* Grouped job count, ccache size and timestamp based rebuilds
  together to "build options" and allow to skip them
* Sorted config options that had to be modified anyway
  alphabetically

### How to test
* Run `pmbootstrap init` and check out the new grouped options there
  * Change the ccache size there
* Verify that ccache works, and that the cache size gets set properly:
  * `pmbootstrap stats --arch=armhf`
  * `pmbootstrap build --arch=armhf hello-world`
  * `pmbootstrap stats --arch=armhf`
  * `pmbootstrap build --arch=armhf hello-world`
  * `pmbootstrap stats --arch=armhf`


ollieparanoid:
I've rebased the PR, fixed what you've commented and also fixed ccache usage for non-native chroots in #1026. Could you test/review it again?

magmastonealex:
Thanks for making this! This will save me hours of waiting for code to compile. 

I tested it with a bunch of packages, as an example, `qt5-qtdeclarative` went from a 2 hour build to an 8 minute build.

The code looks good too, though I don't really know too much about the `pmbootstrap` internals.

ollieparanoid:
I'm glad to hear that it saves so much time 😃
Thanks a lot for testing and reviewing!

PR #1027, opened 2017-12-20
Added new device: samsung-espresso10 (Samsung Galaxy Tab 2 10.1)

WilliamO7:
Information:

- **GSMArena:** https://www.gsmarena.com/samsung_galaxy_tab_2_10_1_p5100-4567.php
- **LineageOS:** https://wiki.lineageos.org/devices/espresso3g

Notes:

- This _should_ work on both versions of the Galaxy Tab 2 10.1" (3G and Wifi), but this has only been tested on the 3G version.
- Since the kernels have been unified between all devices in the Tab 2 range (including the 7.0" devices), I have made the name of the kernel more general. This also trivialises ports to the 7.0" edition (espresso7) as well, but I do not own that particular device.
- Nomenclature has been tweaked from LineageOS's naming due to our unique situation here (as in, we cannot unify the different sized devices as easily)

Observations:

- In the repository's current state, PMOS builds, but it has to be installed in two stages: First the system partitions via android recovery (TWRP), then the kernel via the usual heimdall method. Doing it all with one tool yields flash errors on both sides with boot.img and system.img/FACTORYFS (where it gets to 9% progress then stops) respectively.
- The device boots to weston at the correct resolution, with everything visible as intended, so no problems there. The backlight doesn't turn off on sleep though, (although I cannot fix this now, see the next point.)
- USB does not work **at all.** As in it doesn't even show up in dmesg or lsusb when connected. I tried two linux distros (Artix and Ubuntu 16.04) to test this, and both yielded the same result. (The tablet does have an Apple 30-pin-esque connector, which should be compatible with USB, but also needs extra circuitry for OTG mode, so this could be why?)
- As a result, I cannot enable ssh networking or telnet, and so my efforts end here. I can't even debug this, since I cannot access the logs.

If anyone could help getting USB working on this device, that would be amazing.
Thanks in advance.

UPDATE 2017/12/21 - ~~I tried building the OS with my current tree, and now it bootloops? Like WTF?~~ Realised that my '_flavor' variable was set wrongly, so I'm not going to unify the kernel this time around. Will push an update ASAP.


PureTryOut:
Awesome work! Make sure you join our [Matrix room](https://matrix.to/#/#postmarketos:disroot.org) if you haven't already, this will speed up debugging.

Please also create a device page on [the wiki](https://wiki.postmarketos.org/wiki/Devices) for this device!

zhuowei:
@WilliamO7 Would getting the kernel dmesg help? You can access the dmesg by rebooting to Android and grabbing the /proc/last_kmsg file.

WilliamO7:
There is no last_kmsg file when rebooted into revovery :(

ollieparanoid:
> Realised that my '_flavor' variable was set wrongly, so I'm not going to unify the kernel this time around. Will push an update ASAP.

Okay, adding a `dont_merge_pr` label for now.

Regarding networking, I have diffed your kernel config against the `i9100` kernel config, because it is based on the same kernel version (there are quite a few based on `3.0.101`, that we have in our aports - just grep for them). I have noted that the `i9100` has the following options *unset*:

* `CONFIG_USB`
* `CONFIG_USB_USBNET` (which probably depends on the above?)

While it sounds counter-intuitive to disable those to make network functional, it might be that these were interfering with Androids own network drivers or something. So I would just try to disable these in `menuconfig` and see if it works then in case you didn't try this yet.

Thanks for trying out postmarketOS, doing the port, making the PR and for documenting in the wiki \o/

aports/device/linux-samsung-espresso/02-Fixed-compilation-error-for-rndis_bind_config.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/drivers/usb/gadget/ether.c b/drivers/usb/gadget/ether.c
+index 1690c9d..53ee2c5 100644
+--- a/drivers/usb/gadget/ether.c
++++ b/drivers/usb/gadget/ether.c
+@@ -246,7 +246,7 @@ static int __init rndis_do_config(struct usb_configuration *c)
+ 		c->bmAttributes |= USB_CONFIG_ATT_WAKEUP;
+ 	}
+ 
+-	return rndis_bind_config(c, hostaddr);
++	return rndis_bind_config(c, hostaddr, 12345, "android-arm");
```
ollieparanoid: This change might also be relevant to usb networking, in case the other advice doesn't help. I would look at the definition of `rndis_bind_config` and look up what the parameters do, and how this function is called in other device kernels.

adrianparvino:
I have a Samsung Galaxy Tab 2 7.0, how should I port this? Is there any non-trivial differences aside from the screen?

ollieparanoid:
Welcome to postmarketOS, @adrianparvino!

I've created a new issue and answered your question there: #1606

PR #1028, opened 2017-12-20
Feature debug-shell

### LINKED ISSUE #885: Provide a method to wait in the initramfs
From reading the chatlog, that would have been useful for ayr3s, because the booting process got stuck after killing the usb-shell `telnet` daemon and before starting `sshd`.

Idea: In the `usb-shell` hook, check for a new kernel parameter, and if it is present, simply wait forever (or drop to a shell?).

What do you think?

drebrez:
The old `usb-shell` hook has been replaced with this new `debug-shell` which:

- opens a telnet deamon on port 23 (not 24 anymore)
- shows a dedicated splashscreen with: `WARNING debug-shell is active`
- if there is a tty connected it opens a shell (exit shell to continue boot)
- otherwise it will loop until `pmos_continue_boot` is called from a telnet session

I've updated the url to `https://postmarketos.org/debug-shell`, we should redirect it to a dedicated wiki page which explains it.

Close #885


aports/main/postmarketos-mkinitfs-hook-debug-shell/20-debug-shell.sh
```diff
@@ -0,0 +1,47 @@
...
+echo "This is a security hole! Only use it for debugging, and"
+echo "uninstall the debug-shell hook afterwards!"
+echo "---"
+
+if tty -s; then
+	echo "Exit the shell to continue booting:"
+	sh
+else
+	echo "No shell attached, looping forever."
+	/usr/bin/pmos_loop_forever
```
ollieparanoid: Why is the absolute path used here?

ollieparanoid:
I'll try this out later today. Code looks good to me, only added one comment.
Awesome work, this finally makes the USB debug shell user friendly again 😃

aports/main/postmarketos-mkinitfs-hook-debug-shell/20-debug-shell.sh
drebrez: you are right, there's probably no need to use the absolute path, I'll change it

aports/main/postmarketos-mkinitfs-hook-debug-shell/20-debug-shell.sh
```diff
@@ -0,0 +1,47 @@
...
+
+setup_usb_network
+start_udhcpd
+
+show_splash /splash-debug-shell.ppm.gz
+
+echo "Create 'pmos_continue_boot' script"
+{
+	echo "#!/bin/sh"
+	echo "pkill -f pmos_loop_forever"
```
drebrez: `pkill -f pmos_loop_forever`, probably the `-f` is not necessary here

aports/main/postmarketos-mkinitfs-hook-debug-shell/20-debug-shell.sh
```diff
@@ -0,0 +1,47 @@
...
+	echo '. /init_functions.sh'
+	echo "loop_forever"
+} >/usr/bin/pmos_loop_forever
+chmod +x /usr/bin/pmos_loop_forever
+
+echo "Start the telnet daemon"
+{
+	echo "#!/bin/sh"
+	echo "echo \"Type 'pmos_continue_boot' to continue booting:\""
+	echo "/bin/sh"
```
drebrez: probably also here is not necessary the absolute path

ollieparanoid:
Please remove the label when it's ready to be tested (I probably won't be able to do it today thought).

aports/main/postmarketos-mkinitfs-hook-debug-shell/20-debug-shell.sh
drebrez: I was wrong, this is actually necessary otherwise it doesn't kill it

drebrez:
Ok, I've removed the absolute paths and tested it again on device and with qemu (tty attached and not).
Also updated the comment which now says `No tty attached` instead `No shell attached`.
Added the script to the static code analysis also.
For me it's ready to go 👍

ollieparanoid:
I've just tried it in `qemu-amd64` with:
```shell
$ yes | pmbootstrap install --no-fde --add postmarketos-mkinitfs-hook-debug-shell
```

...and connecting to telnet works, but I do not get a shell. It hangs here:
```
$ pmbootstrap qemu
$ telnet localhost 2223 # 2nd terminal
Trying ::1...
Connection failed: Connection refused
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
```

Any idea?

drebrez:
@ollieparanoid yes, the network is not up in the initramfs for the qemu, we need to change something in `init_functions.sh` to also bring the `eth0` up and set the ip address

drebrez:
@ollieparanoid now connecting with telnet also works for qemu, but calling `pmos_continue_boot` it kills the telnet deamon but it doesn't continue booting because there is the shell open on tty

drebrez:
@ollieparanoid resolved adding a `pmos_shell` script which simply start the `sh`, in this way the `pmos_continue_boot` script can kill it and the boot process can continue.
I tested all the use cases (on device, with qemu exit shell and with telnet) and seems to work fine now.

ollieparanoid:
I must be testing your PR wrong, I still get no shell, like before it connects but I can't type anything. I have rebuilt the hook with --force.

drebrez:
that's strange, did you rebuilt also the `postmarketos-mkinitfs`?
Just re-tested and it works, here the commands I used:

```
pmbootstrap -y zap

pmbootstrap checksum postmarketos-mkinitfs
pmbootstrap build postmarketos-mkinitfs --force 
pmbootstrap checksum postmarketos-mkinitfs-hook-debug-shell
pmbootstrap build postmarketos-mkinitfs-hook-debug-shell --force

pmbootstrap initfs hook_add debug-shell

yes | pmbootstrap -y install --no-fde

pmbootstrap qemu
```
on 2nd terminal:
```
drebrez@drebrez-laptop:~/pmbootstrap$ telnet localhost 2223
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.

Type 'pmos_continue_boot' to continue booting:
/ # pmos_continue_boot
Connection closed by foreign host.
drebrez@drebrez-laptop:~/pmbootstrap$
```

ollieparanoid:
> did you rebuilt also the `postmarketos-mkinitfs`?

Ah, that was my mistake. I did not. Let's try again...

PR #742, opened 2017-10-10
Disable ANDROID_PARANOID_NETWORK for all devices

z3ntu:
Fixes #737

PR #1030, opened 2017-12-21
Rename deviceinfo variable flash_methods to flash_method

### LINKED ISSUE #460: Multiple flash methods
Some devices support multiple flash methods by default. Since #404 is merged, you can specify the flash method manually from the command line with the `--method` option. However it is not safe to allow using every flash method on any device.

The plan is to only allow flash methods, which are enabled in the device's `deviceinfo` file (`deviceinfo_flash_methods`).

@ollieparanoid said we should add adb (recovery flashing) to the `deviceinfo` too, but I disagree, because TWRP is not something installed by default, hence I think it doesn't belong in the `deviceinfo` file.

We would like to hear the community's opinion on this, so we could fix this.

ata2001:
This would close #460.

Thanks to @z3ntu for sharing his [sed commands and for loops](https://github.com/postmarketOS/pmbootstrap/pull/742#issuecomment-335608932).


ollieparanoid:
Do you have time to go through the wiki and adjust it? (e.g. by using `grep -r flash_methods` in the wiki repository so we don't miss any, and then editing the relevant articles in the web UI.)

PR #1034, opened 2017-12-23
samsung-espresso10: Added wifi, touch and usb support

WilliamO7:
Haven't fully tested the wifi yet, I've been trying different interfaces mostly and I'll document my findings on the wiki page soon.


aports/device/device-samsung-espresso10/90-android-touch-dev.rules
```diff
@@ -0,0 +1,4 @@
+# use 'weston-calibrator' to calibrate the touchscreen!
```
ollieparanoid: We have a relatively new feature from #995, that makes the `90-android-touch-dev.rules` file in the device packages obsolete. Because the `devicepkg-dev` package (which your `APKBUILD` already uses since) can generate the udev rule on the fly from the `deviceinfo` file.

To use it, you should only need to fill out the `deviceinfo_dev_touchscreen` variable (which you did already) and also the `deviceinfo_dev_touchscreen_calibration`:
```shell
deviceinfo_dev_touchscreen_calibration="1.007935 -0.019309 18.065613 -0.003218 1.021773 4.029175"
```

Could you try if this works for you? I know that this is not obvious at all when looking at the other devices, because it is not used there yet.

aports/device/device-samsung-espresso10/APKBUILD
```diff
@@ -2,21 +2,25 @@
...
 makedepends="devicepkg-dev"
-source="deviceinfo"
+source="deviceinfo 90-android-touch-dev.rules"
 
 build() {
 	devicepkg_build $startdir $pkgname
 }
 
 package() {
-	devicepkg_package $startdir $pkgname
```
ollieparanoid: Please keep the `devicepkg_package $startdir $pkgname` line and remove the install lines below, the function should do just that (see comment above).

aports/device/linux-samsung-espresso10/APKBUILD
```diff
@@ -2,7 +2,7 @@
 
 pkgname="linux-samsung-espresso10"
 pkgver=3.0.101
-pkgrel=0
+pkgrel=4
```
ollieparanoid: `pkgrel=1`

aports/firmware/firmware-samsung-espresso10/APKBUILD
```diff
@@ -0,0 +1,33 @@
+pkgname=firmware-samsung-espresso10
+pkgver=1
+pkgrel=1
```
ollieparanoid: `pkgrel=0`

WilliamO7:
touchscreen still works btw

ollieparanoid:
Excellent, thanks for changing it and confirming that it still works!

PR #1039, opened 2017-12-25
Motorola Droid 4 (maserati)

NotKit:
This adds basic support for Motorola Droid 4 (maserati). postmarketOS boots and can be accessed through SSH via USB.

osk-sdk would require this hack for framebuffer to display: https://github.com/TeamWin/Team-Win-Recovery-Project/blob/twrp2.7/minuitwrp/graphics.c#L165. droid-hal-maserati and xf86-video-hwcomposer from libhybris branch can be used for accelerated Xorg server and starting hildon-desktop.

LuneOS UI can be launched with qt5-qpa-hwcomposer and changes to init scripts mentioned in #629


aports/device/device-motorola-maserati/10-safestrap-maserati.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
```
ollieparanoid: What does this script do, is this for booting postmarketOS off an image file?

If that was the case, I would recommend to integrate this in a way that it can easily be used by all devices (changing the initramfs instead of providing a hook), and I would do it in its own PR.

aports/device/device-motorola-maserati/APKBUILD
```diff
@@ -0,0 +1,28 @@
...
+	install -Dm644 "$srcdir"/deviceinfo \
+		"$pkgdir"/etc/deviceinfo
+
+	mkdir -p "$pkgdir"/etc/postmarketos-mkinitfs/hooks/
+	install -Dm644 "$srcdir"/10-safestrap-maserati.sh \
+		"$pkgdir"/etc/postmarketos-mkinitfs/hooks/
+}
+
+hybris() {
+    depends="droid-hal-maserati libhybris-4.4"
```
ollieparanoid: We don't have the depends in this branch, so I would remove this subpackage to keep `master` consistent.

aports/device/device-motorola-maserati/deviceinfo
```diff
@@ -0,0 +1,29 @@
...
+deviceinfo_keyboard="true"
+deviceinfo_external_disk="true"
+deviceinfo_screen_width="540"
+deviceinfo_screen_height="960"
+deviceinfo_dev_touchscreen=""
+deviceinfo_dev_keyboard=""
+deviceinfo_msm_refresher="true"
+
+# Bootloader related
+deviceinfo_flash_method="none"
```
ollieparanoid: Could you explain that?

aports/device/device-motorola-maserati/10-safestrap-maserati.sh
NotKit: The device doesn't have unlocked bootloader, so SafeStrap recovery and kexec is used to boot custom kernels/OSes. SafeStrap supports 4 ROM slots on sd-card, +stock (can be used for LineageOS, but installing pmOS there would brick the device) and safe (installs to unused partitions). That script tries to read active SafeStrap slot and boot off correct image file. So it's not done in exactly generic way to be integrated for all devices.

aports/device/device-motorola-maserati/deviceinfo
NotKit: deviceinfo_msm_refresher - need to remove that, it doesn't actually have any effect on device (I thought it could help at some point).

deviceinfo_flash_method - due to need to use SafeStrap, no existing flash method would work. I installed it manually when testing, but Ideally a zip needs to be generated, containing at least the kernel and kexec scripts to be installed onto slot's virtual /system partition.

ollieparanoid:
Thanks for the explanation! I think these points should be addressed before merging:
- [ ] remove the droid/libhybris depends, since this is the `master` branch
- [ ] write a wiki page for the device, that explains how to perform the installation

jja2000:
Have you tried using the Mainline kernel instead of the one that ships with the device?
SRE still seems to be working on it and the current hw compatibility seems to be nearing completion.

http://elektranox.org/droid4/

NotKit:
@jja2000 yes, I did. Ideally I would package both, since mainline support and running newer kernel is cool, but it lacks 3D acceleration, cameras and voice calls, which is possible with Android one with libhybris and proper middleware.

tmlind:
Cool to see this :) I sure hope you get things booting with droid4-kexecboot as SafeStrap has quite a few non-standard dependencies we can completely avoid that way. So the comments above about not having and unlocked bootloader above are no longer really true :p

NotKit:
@tmlind after droid4-kexecboot is installed, can system/data/userdata (? - mmcblk1p25) partitions be safely overwritten?

tmlind:
@NotKit you should first check if mmcblk1p25 is vfat and has SafeStrap config files and ask for confirmation before reformatting and installing to it naturally.

But yeah I've been using mmcblk1p25 as my ext4 rootfs with mainline kernel and alpine for past three or four months, so that's 8GB right there on the faster eMMC so worth the extra effort for sure :) The logic for mounting partitions was discussed a bit few days ago related to the SailfishOS support:

[https://talk.maemo.org/showpost.php?s=2769ccef9c9bf887a6d85749576f6d6b&p=1539649&postcount=113](https://talk.maemo.org/showpost.php?s=2769ccef9c9bf887a6d85749576f6d6b&p=1539649&postcount=113)

tmlind:
Looks like the legacy /usr/sbin/kexec-droid4 has some bug and does not properly load the dtb the same way as kexec in SafeStrap. So to kexecboot v3.0.8 based kernels with droid4-kexecboot, please use the SafeStrap /system/etc/kexec/kexec binary. Or just update /usr/sbin/kexec wrapper script to use kexecboot mounted location for the SafeStrap kexec binary as the 4MB droid4-kexecboot image is out of disk space. Or just do it manually for now until I have fixed droid4-kexecboot legacy kexec.

Then for the v3.0.8 booted kernel, you should disable CONFIG_CMDLINE_FORCE=y I think so the passed kernel command line is used instead of the one in atags. If that does not help, then I think we need to generate the legacy atags dynamically in the legacy kexec binary.

aports/device/device-motorola-maserati/deviceinfo
```diff
@@ -0,0 +1,28 @@
...
+deviceinfo_dev_touchscreen=""
+deviceinfo_dev_keyboard=""
+
+# Bootloader related
+deviceinfo_flash_method="none"
+deviceinfo_safestrap_partition="/dev/mmcblk1p25"
+deviceinfo_data_partition="/dev/mmcblk1p24"
+deviceinfo_system_partition="/dev/mmcblk1p20"
+deviceinfo_preinstall_partition="/dev/mmcblk1p22"
+deviceinfo_webtop_partition="/dev/mmcblk1p23"
```
ollieparanoid: `webtop_partition` and `preinstall_partition` are not used yet, are they necessary?
If it's just for information purpose, I would recommend to write that into the [device's article in the wiki](https://wiki.postmarketos.org/wiki/Motorola_Droid_4_(motorola-maserati)).

aports/device/device-motorola-maserati/APKBUILD
```diff
@@ -0,0 +1,22 @@
...
+pkgdesc="Motorola Motorola Droid 4"
+pkgver=0.1
+pkgrel=1
+url="https://postmarketos.org"
+license="MIT"
+arch="noarch"
+options="!check"
+depends="linux-motorola-maserati"
+source="deviceinfo 10-safestrap-maserati.sh"
+subpackages="$pkgname-hybris"
```
ollieparanoid: This causes the build to fail currently:
https://travis-ci.org/postmarketOS/pmbootstrap/builds/344322059#L5301-L5305

MartijnBraam:
@NotKit I can't seem to figure out how to make safestrap boot postmarketos since I can only find how to make it switch /system and /data and nothing about the kernel/initfs. How do you boot it?

ollieparanoid:
@MartijnBraam: From reading the initramfs hook (which you guys probably need to boot pmOS off it with safestrap), I would guess that you put the `boot.img` file next to the `system.img` file on the partition.

MartijnBraam:
What I found so far by looking at the sailfish port for maserati is that I need to put the kernel and the initfs in `/system/etc/kexec/` inside the system.img file and then put a system.img and data.img in a zip to get it on the phone. Which is kinda messy and makes it impossible to update the kernel from inside pmos.

NotKit:
If SafeStrap is used, there is no boot.img, but kernel/initramfs in system partition. Sorry, I'm currently in another country, so can't look into this properly till I return on Friday.

NotKit:
To go with the direction proposed in #1275, I addressed @ollieparanoid open points in review and rebased it to master to start with.

NotKit:
Can confirm it's still booting after rebase. Any idea if I should somehow package https://gist.github.com/NotKit/7d233dd21ff1c4f34947d6a56ace3ce0? This allows it to activate framebuffer output on Android kernel.

drebrez:
@NotKit nice small tool, maybe we should add that activate flag also in the `fbdebug` tool is case some other device has a similar situation. 
Or if it will be packaged as a tool itself (maybe something like `fbactivator`), then we should include it to the `debug-shell` hook.

NotKit:
@drebrez Checked fbdebug, yes, that would be nice to have. Can we add fbdebug to initramfs to call it from device-specific hook?

Initial attempt at documenting installation: https://wiki.postmarketos.org/wiki/Motorola_Droid_4_(motorola-maserati)#Installation. Feels way more complicated than it needs to be.

drebrez:
@NotKit the `fbdebug` is added to the `initramfs` if it's present during the creation of the `boot.img`, this means that adding `fbdebug` to the `depends` of the `device-aaa-bbb` package it will be automatically included, and should work calling it from a device hook.
Maybe we can rename the `fbdebug` to `fbtools` or something, and change it's purpose, also adding the pan-loop operation and replace the `msm-fb-refresher`.
We will have our own tool to fix all kind of stuff related to the framebuffer. @ollieparanoid what you think?

ollieparanoid:
> We will have our own tool to fix all kind of stuff related to the framebuffer. @ollieparanoid what you think?

Sounds good to me (see #1336), but I would try to get this initial booting version merged first, even if the screen does not show anything yet (because it has been open for such a long time, and then we have a good base to do further PRs from).

ollieparanoid:
me:
> NotKit: so @V13Axel confirmed that your PR works. Do you want to get it merged now, or should we wait some more for an additional review?

@NotKit:
> ollieparanoid[m], let's get it merged then

PR #1040, opened 2017-12-25
Add NetworkManager and PulseAudio applets to Plasma

### LINKED ISSUE #979: [Plasma] Top drawer and background missing on first boot
When booting plasma mobile directly after the installation, the top drawer and the background are missing. The top drawer appears when booting the second time. I've tested all this in Qemu.

@PureTryOut wrote:
> if you also want the background to appear, remove ~/.config/plasma-org.kde.plasma.phone-appletsrc once and reboot. that will however make the top drawer disappear...
> yeah it's all very strange, I know

@bhush9 and @PureTryOut (or whoever else knows):
* Is this an upstream bug?
* How can we assist with fixing this?
* Is there a method to get both the background and the top drawer working at the same time?

@bhush9, if you have time and disk space, you could reproduce this in a few minutes (as we have binary packages now):
```shell
cd pmbootstrap
./pmbootstrap.py init # select qemu-amd64 and plasma-mobile, leave defaults for the rest
./pmbootstrap.py install --no-fde
./pmbootstrap.py qemu --image-size=2G
```

PureTryOut:
Resolves (partially) #979, the black background on first boot still happens however.

Right now Plasma Shell seems to crash after a second or two on my Amami's. This does not happen on x86_64 Qemu, so I'd like others to verify this. Also, the NetworkManager applet seems to be always extended, making it completely unusable. @bhush9 could you look into this?


PureTryOut:
Well I just realized why this causes the Plasma Shell to crash on an actual device. This probably only happens on devices without hardware acceleration (which my Amami's are part of), as it's known that QML apps crash on those. And let it just be that the applets both are using QML.

So, before merging this, we need to figure out how to reliable run QML apps on devices without hardware acceleration.

bhush9:
> Well I just realized why this causes the Plasma Shell to crash on an actual device. This probably only happens on devices without hardware acceleration (which my Amami's are part of), as it's known that QML apps crash on those. And let it just be that the applets both are using QML.

No, applets are loaded as part of plasmashell process, and QML does work completely fine in plasmashell since we enable software rendering.. however I've theory on what's causing crash...

PureTryOut:
So, according to @bhush9 on Matrix, the shell shouldn't crash on devices with DRM available. @opendata26 could you give this PR a shot on your Castor with hardware acceleration?

PureTryOut:
So, armhf seems to be the only arch having problems with this. Honestly, I don't think we should hold this PR just waiting for a fix on the Qt side which might or might not come. Is it an idea to add a "(doesn't work on armhf)" warning to the UI description, and just go with aarch64 for now? I could even add "!armhf" to `$arch` on some Plasma packages.

ollieparanoid:
> Honestly, I don't think we should hold this PR just waiting for a fix on the Qt side which might or might not come.

I think you are right, it isn't usable on downstream armhf kernels anyway (performance and QML issues), so we don't need to keep that case supported at the cost of having features missing and bugs introduced to armhf devices using mainline and aarch64 devices in general.

> Is it an idea to add a "(doesn't work on armhf)" warning to the UI description, and just go with aarch64 for now?

If I understood @bhush9 right, it only does *not* work on `armhf` devices with framebuffer rendering (so all downstream kernels). It should work on the armhf hammerhead with mainline kernel (drm + freedreno instead of fb), right?

So maybe a good description for `postmarketos-ui-plasma-mobile` would be:
`pkgdesc="(Wayland) Mobile variant of KDE Plasma (optimized for touchscreen, aarch64 or mainline required!)"`

With that being said, I think doing it with such a description change is a good idea.

PureTryOut:
So, I fixed the compilation issue I was having with plasma-pa thanks to @<insert name>. It would constantly complain that `gconf/gconf-value.h` was missing while it's not. This ended up being a problem in the upstream `gconf` package (from the Alpine repo's) where it was built with the `orbit2` dependency, but it wasn't added to `$depends` as well. Because of this I now include our own gconf, but I'll upstream this to Alpine as soon as possible.

Note that plasma-nm looks quite buggy right now.

screenshot_2018-02-27_19-19-51
screenshot_2018-02-27_19-20-01

I'm not sure what we can do about this, any ideas @bhush9 ?

PureTryOut:
This PR is slowly becoming a mess :disappointed: 

Instead of adding `orbit2` as a runtime dep to gconf and having to patch upstream (while we're probably the only one ever encountering this problem), I decided to add it as a buildtime dep to `plasma-pa` instead. This works as well, so we'll go with this.

PureTryOut:
Ok, fully cleaned up the PR. Upstream `gconf-dev` is fixed, so no need for `orbit2` anymore in our own packages. This is in my opinion ready to merge. Of course still doesn't work on armhf without hardware acceleration.

aports/kde/qca-qt5/APKBUILD
```diff
@@ -0,0 +1,37 @@
+pkgbase=qca
+pkgname=qca-qt5
+pkgver=2.1.3
+pkgrel=0
+pkgdesc='Qt Cryptographic Architecture'
+arch="all"
+url='ttp://delta.affinix.com/qca/'
```
ollieparanoid: missing a "h" here

aports/main/postmarketos-ui-plasma-mobile/APKBUILD
```diff
@@ -1,7 +1,7 @@
 pkgname=postmarketos-ui-plasma-mobile
 pkgver=1
 pkgrel=9
```
ollieparanoid: pkgrel needs to be increased

PureTryOut:
> but I'm wondering what happens right now, as we don't have the networkmanager daemon started at all

Yup I know. The applet shows, it just doesn't show anything *useful*. It seems like it's current contents are more of a placeholder (notice the "VPN work" and "University" wifi entry). However, it was my idea to just go with it for now, and get NetworkManager itself in later on.

Issue #756, opened 2017-10-12
Make proprietary firmware and userspace optional


ollieparanoid:
The built-in wifi and baseband will not work without proprietary firmware on practically all phones. However, we could allow the user to avoid installing these blobs, just like libre distributions do it.

### Implementation idea

1. Change the device package recipes to generate the following subpackages:
* `device-vendor-name-nonfree-userland` (we don't have proprietary userland right now, so this is not needed)
* `device-vendor-name-nonfree-firmware` (depends on the firmware package(s))

(This is really easy in APKBUILDs and can be done with very few lines.)

2. Ask in `pmbootstrap init`:
```
(Explanation text)
(pkgdesc from the nonfree packages, so they can tell which hardware components they enable)
Use non-free components (none, firmware, userland, all)? [firmware]: 
```

We could directly skip the question, in case there is no nonfree subpackage for the device, and only show the firmware/userland choice when such a package exists.

3. Depending on what the user chose, the `nonfree` packages and its dependencies get built and installed in `pmbootstrap install`.

### Advantages
* Make people aware, that there are proprietary components and what they are for
* People who prefer a full libre stack can disable the firmware files which are enabled by default right now (they could use an external modem via usb etc.)
* It would allow packaging proprietary userspace drivers with libhybris (#678) in theory, while keeping it cleanly separated for users who do not wish to install them.
* Once we can run proprietary components in pmOS, it might be a lot easier to reverse engineer them.

Thoughts?

drebrez:
The idea to let the user choose during the `pmbootstrap init` looks good.

What is not really clear, at least for me, are the options of the question.
`none` and `all` are fine, but what exactly means `firmware` and `userland`?
With `firmware` I suppose proprietary blobs in the kernel, right? 
Choosing `firmware` means also having proprietary blobs in `userland`?
Use something like `kernel-land only` and `user-land only` makes it more clear?

ollieparanoid:
### Firmware

With firmware I mean blobs, that other chips on the phone require to function. Such as the Wifi chip. They are not really part of the kernel, but the kernel loads them from the disk and passes them to these chips.

Speaking in package names, this would be any `firmware-` package from postmarketOS, and the `linux-firmware` package provided by the Kernel developers (they only collect the firmwares, the binaries come from the vendors).


### Userland

[...]

PR #1046, opened 2017-12-27
Add basic support for Nokia N9

pavelmachek:
Filip Matijević did all the work here, I'm just attempting to merge it (with his permission).

N9 has pretty good mainline support, and it is being improved, so it is good target. OTOH it does not have a u-SD card slot, and battery is not removable, which makes it tricky. Currently, it is usable with root mounted over NFS over USB, including working xfce4.


aports/device/device-nokia-n9/APKBUILD
```diff
@@ -0,0 +1,45 @@
+pkgname=device-nokia-n9
+pkgver=1
+pkgrel=14
```
zhuowei: You might want to reset the pkgrel to 0, since it's the first version in master.

aports/device/device-nokia-n9/deviceinfo
```diff
@@ -0,0 +1,19 @@
...
+deviceinfo_date=""
+deviceinfo_keyboard="false"
+deviceinfo_keymaps=""
+deviceinfo_nonfree="????"
+deviceinfo_dtb="omap3-n9"
+deviceinfo_modules_initfs="omap_wdt twl4030_wdt"
+deviceinfo_external_disk="false"
+deviceinfo_external_disk_install="false"
+deviceinfo_arch="armhf"
+
```
ollieparanoid: How about adding `deviceinfo_flash_method="0xffff"` here?

aports/device/device-nokia-n9/initfs-hook.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
ollieparanoid: This script exists twice basically, once for the initfs-hook and once for the `watchdog-kick` package. I prefer the packaged version, because:
* `#!/bin/sh`
* doesn't check if `sleep` exists, and that check isn't necessary as `busybox` always provides it
* tabs instead of spaces is in line with the rest of Alpine and postmarketOS shell scripts

I think kicking the watchdog is a feature that other devices might need as well, and I would also like to reduce the duplication. How about we add and run the `watchdog-kick` script to/in the initramfs, when a new `deviceinfo_kick_watchdog` option in the `deviceinfo` is set?

You could integrate it just like `msm-fb-refresher` is integrated to the initramfs (grep the `aports/postmarketos-mkinitfs/` for msm).

(CC @Halamix2, your device had a watchdog problem, maybe such a script helps you as well.)

aports/device/linux-nokia-n9/APKBUILD
```diff
@@ -0,0 +1,128 @@
+# APKBUILD based on linux-vanilla aport. Changes:
```
ollieparanoid: Ideally we would use `linux-postmarketos-stable` (which the N900 uses, or `-mainline`/`-lts`) instead, so we can use the same kernel package across multiple devices. Would it be feasible for you to change it in this PR or in a follow up PR?

When merging with a `linux-postmarketos` package, we could put the N9 specific patches in a subfolder in the aport (like it's done in our `qt5-qtwebengine` aport).

aports/device/linux-nokia-n9/APKBUILD
```diff
@@ -0,0 +1,128 @@
...
+	0003-SEB.patch
+	0004-drm-add-rename-MIPI_DCS_SET_PARTIAL_XXX.patch
+	0005-drm-omap-panel-dsi-cm-use-defines-from-mipi_display..patch
+	0006-ARM-dts-n9-add-display-support.patch
+	0007-Revert-drm-omap-work-around-for-omap3-display-enable.patch
+	0008-HACK-drm-omap-panel-dsi-cm-force-0xff-for-brigthness.patch
+	0009-ARM-dts-N9-N950-Add-touchscreen-support.patch
+	0010-ARM-N9-N950-Add-PVR-drivers.patch
+	$_config
+	compiler-gcc6.h
```
ollieparanoid: This should not be necessary with a 4.14 kernel

aports/device/linux-nokia-n9/APKBUILD
```diff
@@ -0,0 +1,128 @@
...
+	unset LDFLAGS
+	make ARCH="$_carch" CC="${CC:-gcc}" \
+		KBUILD_BUILD_VERSION="$((pkgrel + 1 ))-Alpine" \
+		|| return 1
+}
+
+package() {
+	cd "$srcdir/build/arch/arm/boot"
+
+	cat zImage dts/omap3-n9.dtb > zImage-dtb
```
ollieparanoid: `postmarketos-mkinitfs` can also pick the `dtb` file as specified in the `deviceinfo` and attach it to the kernel. That would be necessary when using a `linux-postmarketos-*` package instead.

aports/main/watchdog-kick/APKBUILD
```diff
@@ -0,0 +1,25 @@
...
+url="https://github.com/postmarketOS"
+arch="all"
+license="GPL2"
+install="$pkgname.post-install"
+source="
+	watchdog-kick.init
+	watchdog-kick
+"
+
+build() {
```
ollieparanoid: empty `build()` functions can be removed

aports/main/watchdog-kick/watchdog-kick
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+while :
+do
+	for wd in /dev/watchdog?; do
+		if [ -c $wd ]; then
+			echo X > $wd
+		fi
+	done
+	sleep 10s
```
ollieparanoid: any reason why this version has 10s and the one from the initramfs has 2s?

aports/main/watchdog-kick/watchdog-kick.init
```diff
@@ -0,0 +1,26 @@
+#!/sbin/openrc-run
+
+cmd=/usr/sbin/watchdog-kick
+pid=/run/watchdog-kick.pid
+
+depend() {
+	after udev
```
ollieparanoid: Shouldn't the watchdog be started as early as possible, without waiting for udev?

filippz:
@ollieparanoid Thank you for review. GPU works for a handful of basic demos due to issues with glibc vs musb. There is X11 support in the GPU drivers but I haven't tried it yet - I'll try using https://github.com/nemomobile/ti-omap3-sgx-wayland-wsegl for wayland first and fall back to X11 if that fails. Currently I'm working on packaging both kernel modules and userspace libraries but it's a slow process. In any case we have working  bootsplash and xfce4 using just fb so we can use either one for pictures.
As for the rest of the things you commented on I'll try to reply sometime tomorrow if time permits.

ollieparanoid:
Awesome, take your time! I think the goal of `gcompat` is to make all glibc applications work with musl (right @awilfox?), so they would probably be happy to receive [bug reports](https://bts.adelielinux.org/) if you have time for it at one point. Ideally the issues get ironed out quickly then.

pavelmachek:
Yes, I also have N9 (besides N900, N950, and few older ones).

Photos for the blog post: I don't think I'll find time for that.

Wiki update: I'll add it to the todo list. There's not much to be said
at the moment... besides listing what does not work. ... same goes for
flasher boot update. I guess that would be interesting for N900,
too...

(0xffff does not work too well on either N900 or N9; but I guess
that's separate topic).

maxtouch.cfg: Its not really a firmware. More like set of calibration
parameters for the touchscreen. Not copyrightable, AFAICT. @filippz
even understands what the parameters mean.

Switching to linux-vanilla -- I don't think that's good idea at the
moment. We'll need quite a few patches for now, and it would be bad to
break some other phone with patches relevant for N9. Long-term it is a
goal, and we are not that far, but I don't think we should do it now.

Watchdog timeout is 30 seconds, iirc, so it is not _too_ critical.

Let me go through the rest of comments...

aports/device/device-nokia-n9/APKBUILD
```diff
@@ -0,0 +1,45 @@
...
+}
+
+package() {
+	install -D -m644 "$srcdir"/deviceinfo \
+		"$pkgdir"/etc/deviceinfo
+	install -D -m644 "$srcdir"/initfs-hook.sh \
+		"$pkgdir"/etc/postmarketos-mkinitfs/hooks/00-${pkgname}.sh
+
+	#Atmel TS
+	install -D -m644 "$srcdir"/RM-696_Pyrenees_SMD_V1_6.raw \
```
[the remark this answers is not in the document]
filippz: Well, this is a strange one. The "firmware" is in fact configuration file that can be dumped from running hw by using https://github.com/atmel-maxtouch/mxt-app, or simply c/p over from previous versions etc. The thing is that is simply has to be called maxtouch.cfg (kernel drivers can't be told the file name by using dts property or something like that) and since we expect using N950 that has the same TS controller but slightly different screen orientation that has to be matched by TS controller we also need a bit different (just 2 bytes) maxtouch.cfg for it but named the same. I'm targeting for using the same kernel package for N950 by means of subpackages so the names in the package should be more meaningful that maxtouch.cfg (ie RM-696_Pyrenees_SMD_V1_6.raw)

aports/device/device-nokia-n9/deviceinfo
filippz: I haven't used 0xffff (I'm using ubiboot via serial cable), so @pavelmachek might be the one to have more info. Currently, there is no easy way to "flash" N9/N950 with pmOS.

aports/device/device-nokia-n9/initfs-hook.sh
filippz: Ah... My N9 has damaged MTD so I get lots of ECC errors that slow the boot down. Once kernel starts using WD they need to be kicked every 30sec, but OpenRC from root partition doesn't get started early enough to be kicked from there (especially when first boot resizes FS and generates crypto keys), hence the solution to have a similar script in initramfs. IIRC checking for sleep was also needed - but I can't remember why... I've tried to make both scripts reusable on other devices so they might the base for something better in  the future.

aports/device/linux-nokia-n9/APKBUILD
filippz: As @pavelmachek already commented - I'd leave it as it is for time being.

aports/device/linux-nokia-n9/APKBUILD
filippz: We can try without it - right @pavelmachek?

aports/device/linux-nokia-n9/APKBUILD
filippz: OK, we' can try without it (CC @pavelmachek)

aports/main/watchdog-kick/watchdog-kick
filippz: We need to be sure that the time between when initramfs script kicks them for the *last* time and the one that is started from root partition kicks them for the *first* time is no more that 30sec. If we have too high number in here the time from last kick by initramfs script to the time root partition script is started could be longer, so the 2s seemed reasonable.

aports/main/watchdog-kick/watchdog-kick.init
filippz: Yes, but we need `/dev` to be populated so if we start it too early and there is no `/dev/watchdog?` devices to kick and we retry after 10s - that might me just too long and the device dies.

aports/main/watchdog-kick/APKBUILD
filippz: IIRC there was a warning w/o it - I'll retry and see.

pavelmachek:
I updated the device web page. I can mark myself as a maintainer, but
perhaps filippz would be better maintainer.

-- 

aports/device/linux-nokia-n9/APKBUILD
ollieparanoid: Seems like you guys are busy, so I would say since we're not merging the kernel together with `linux-postmarketos-*` yet, we can leave this one as it is for now.

aports/device/device-nokia-n9/APKBUILD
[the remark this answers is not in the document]
ollieparanoid: Would it make sense to rewrite the driver to get the config file from `/etc/` then?

I think you don't even need a subpackage - you could make one `device-...` package for the N9 and one for the N950, and let each one ship their own config file, but depend on the same kernel package.

ollieparanoid:
Thanks to both of you for answering the questions in such detail, and thanks for expanding the wiki page! In my opinion, only the following is left to do before we can ship it 🚢:
- [ ] remove `compiler-gcc6.h`
- [ ] shorten the initramfs watchdog-kick hook (we could also integrate watchdog-kick hook into the initramfs like msm-fb-refresher, but that doesn't seem to be beneficial enough at this point). This should do it (untested):
```shell
#!/bin/sh
watchdog_kick() {
	while true; do
		for wd in /dev/watchdog*; do
			[ -c $wd ] && echo X > $wd
		done
	done
	sleep 2s
}
watchdog_kick &
```
- [ ] adjust the package and kernel name to use the codename. The wiki page states, that the codename is Lankku/rm-969, so we should use one of those for consistency (which one do you prefer? if a device has multiple code names, we just chose one). The N900 has `device-nokia-rx51` as device package name.

aports/device/device-nokia-n9/APKBUILD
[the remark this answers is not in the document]
filippz: There was a discussion at [LKML](https://lkml.org/lkml/2013/6/5/581) where Atmel guys were trying to push some changes but the whole thing stalled when maintainers requested to rewrite driver to use regmap - the discussion wasn't productive after that (keywords: *proprietary*, *custom*, *NDA*, ...) One idea in there was to use DT entry instead of special file - which would be great for our usercase, but I'm reluctant to bring this up on LKML once more :)

Actually, putting `maxtouch.cfg` in device- package seems to be a good idea. Maybe we can do that once we decide to introduce support for N950 (CC  @pavelmachek)

filippz:
My vote goes to rm-696 (N950 is rm-680) - these are mentioned in early kernels and device labels etc. I've managed to get SGX stuff packages and I'm off to see if it boots, after that I'll retry with gcc6 removed and changed watchdog-kick hook. Thanks @ollieparanoid & @pavelmachek!

awilfox:
I'm not sure what gcompat has to do with anything after a cursory read of the comments here, but that is correct.  It's a binary library and won't be any help compiling glibc software, but will help run binaries built against glibc.  Bug reports or missing APIs go on [the BTS](https://bts.adelielinux.org/enter_bug.cgi?product=gcompat) and merge requests go on [the official GitLab](https://code.foxkit.us/adelie/gcompat/merge_requests).

pavelmachek:
Concerning the consistency of package names...

Yes, I know n900 is referred as rx51 in pmos, but maybe we should
rename that to n900 and keep the n9. In nokia case, code names are
both hardto remember and useless :-(.

-- 

ollieparanoid:
@pavelmachek: we use the code names for *all* devices, so we would have to change all of them to keep it consistent. I think it makes more sense to adjust the N9 packaging to `device-nokia-rm-696` and `linux-nokia-rm-696`.

filippz:
I've completed SGX packaging (includes downloading of extracted content of `Graphics_SDK_setuplinux_4_10_00_01_hardfp_minimal_demos.bin` since that file requires i386 libs to be executed and I wasn't able to find a workaround). Only strange thing I'm struggling with is that I need to explicitly preload libgcompat (`sudo env "LD_PRELOAD=/lib/libgcompat.so.0" eglinfo`) and even then not all binaries work due to missing symbols, but at least some demos do run OK.

I can confirm that the `compiler-gcc6.h` is not needed and can be dropped.

I tried your (@ollieparanoid ) watchdog-kick hook but it didn't work, as `sleep` should be moved one line up, but the thing that breaks the boot is `/etc/postmarketos-mkinitfs/hooks/00-device-nokia-n9.sh: line 12: sleep: not found`. I'm not an expert by any means, but only viable workaround I've found is to check for existence of `/bin/sleep` before calling it. In any case - both *watchdog* scripts are modified (the OpenRC one now can accept optional `-1` arg that is used when shutting down to call explicitly *kick* the watchdogs *goodbye* one last time)

The last thing remaining is the renaming of the packages to rm-696 and maybe we can also consider including SGX immediately, since I haven't made a separate commit for it in `linux-nokia-n9` package (it's all in my [repo](https://github.com/filippz/pmbootstrap)) - but I'm open to suggestions.

awilfox:
Can you tell me what `eglinfo` (etc) have for `DT_INTERP`?  If you have `scanelf` installed, this should be as simple as running `scanelf -i /path/to/eglinfo`.

Please do report any missing APIs to the [gcompat tracker](https://bts.adelielinux.org/enter_bug.cgi?product=gcompat).  We can try to implement them.

filippz:
`eglinfo` uses `ld-linux.so.3`, but I've linked that missing one to `ld-musl-armhf.so.1` instead of `ld-linux-armhf.so.3`, so that is not a problem of `gcompat`.

Other binaries that don't work are complaining about missing `__vsprintf_chk`  and I've made a report [#43](https://bts.adelielinux.org/show_bug.cgi?id=43) for that.

Thank you for your time!

ollieparanoid:
@filippz: Thanks for the status update!

I can't explain right now why the sleep binary would be missing at this point, as I've just confirmed in QEMU that the sleep binary exists with the debug-shell hook (which gets executed in the same line as all other hooks - including the N9 hook - in `init.sh`). Agreed, leaving the check for `/bin/sleep` in place seems to be the only way to make it work right now. The modifications you've made to the watchdog sound good!

Regarding the SGX blob: Right now the only proprietary packages we have are in the `firmware` folder, and since we're trying to make proprietary firmware and userspace optional (#756) I think it makes sense to discuss in that issue where exactly to package userland blobs before integrating it. I'll write my ideas down in the issue and recommend for now that we leave the SGX blob out of this PR and package it in a new one, when the details are figured out.

The [TODO](https://github.com/postmarketOS/pmbootstrap/pull/1046#issuecomment-354457397) list from above has been updated. Thanks for the great work everyone!

@pavelmachek: Would you mind updating this pull request when you find time for it?

*PS: The [blog post](https://postmarketos.org/blog/2017/12/31/219-days-of-postmarketOS/) is out and mentions the N9 progress.*

filippz:
@ollieparanoid Is it OK to compile kernel modules together with `linux-` package or is that something you'd like to see in separate `APKBUILD`? It way more easier to build it in same `APKBUILD` but we need to pull the TI SDK in to build them so I need to ask. I can make a supbackage for them and make userspace libs package to depend on that subpackage if that makes any difference.

ollieparanoid:
How does the SDK look like? If it's just a bunch of source code files (kernel modules?), then I would recommend that we put it in the `linux` `APKBUILD` for now (we don't use a shared kernel anyway). In that case we could possibly split it in its own `APKBUILD` when we use the shared kernel in the future.

But if it contains binaries, just like a firmware package, then I think it we should split it right now. The reasoning is, that we don't have any firmware outside of the `aports/firmware` folder.

Could you post a link to the SDK?

filippz:
Well, it not just kernel modules source, but also a bunch of .so and binary files. For this package I only need source files, and the rest is not used.  Since those kernel modules are built using existing kernel sources that have been already built `linux-` package is a natural place for them. Once  (or better *if*) I get userspace to use GPU, additional package will contain those libs and binaries - but to make things more complicated those libs/binaries *should* be placed in /usr/local/lib & /usr/local/bin. I guess modifying `LD_LIBRARY_PATH ` is also an option if they *must* be placed elsewhere. I've already tried making that package and it's providing to be problematic due to a fact those libs should be used instead of `mesa`  libs that are pulled in due to dependencies of various pmOS packages - but we'll see what to do *if* and when I get something usable.

The full download is [here ](http://software-dl.ti.com/dsps/dsps_public_sw/sdo_sb/targetcontent/gfxsdk/4_10_00_01//exports/Graphics_SDK_setuplinux_4_10_00_01_hardfp_minimal_demos.bin), but since it's binary file that needs x86 libs to be extracted, I've simply repacked it to `tar.gz` and uploaded to Dropbox [here](https://www.dropbox.com/s/m3v4w36lb4edbi9/SGX.tar.xz?dl=1).

ollieparanoid:
Hosting on dropbox doesn't look very stable to me, how about we split up the package and put the sources on github? We have [this](https://github.com/pmsourcedump) organization created for the purpose where the upstream source code isn't packaged in an useful way for us. So for this case, how about we put *just the header files* in an extra repo there and tag it with the version number of the SDK. Then you could use that as source and let GitHub generate a tarball for you - while at the same time it's convenient to browse the source online.

Please note that we can only do that if the license allows us to do this - are the kernel module sources under the GPLv2, just like the kernel? 

And to answer your question: yes, I think it makes sense to put it in the kernel package then, especially if it makes it easier 👍 

(I am not sure where exactly they are in the archive and I kind of need to go - but I could look into it more later.)

filippz:
I've decided to drop support for SGX for time being, as it seems that the guys from [Maemo Leste](https://github.com/maemo-leste) managed to reuse old Nokia drivers on N900 with mainline kernel. In theory they are better choice since we know those were able to run X11 as is, and Nemo was running Wayland with those and an additional [WSEGL plugin](https://github.com/nemomobile/ti-omap3-sgx-wayland-wsegl). I haven't had a chance to try it out, but if it works there will be a separate PR for it or TI ones if those turn out to be a better choice.

filippz:
@pavelmachek - could you please update my commits in this PR with fresh versions from my repo? I don't think I can do it myself.

pavelmachek:
You should be able to do the pull request yourself. It will get new PR
number, but I guess that's ok.

If not, yes, I can do that.

-- 

filippz:
@ollieparanoid I guess it's up to you to decide. If you want you can then close this PR and I'll make another one, or @pavelmachek  will update this one.

ollieparanoid:
Yeah I think it's easier when you reopen a new PR coming from your account. Looking forward to it!

PR #1048, opened 2017-12-28
Replace dtbTool with the Skales dtbTool

MoreRobustThanYou:
Should work the same, except it should also work in mainline, because it uses libfdt (CM dtbtool often errors on mainline)


aports/main/dtbtool/APKBUILD
```diff
@@ -1,21 +1,19 @@
+_commit="c2bfa15aa5bfc4e3f1b32de28e327b80b37db1cf"
+
 pkgname=dtbtool
-pkgver=3
-pkgrel=3
-pkgdesc="dtbtool"
-url="https://github.com/LineageOS/android_device_qcom_common/tree/cm-14.1/dtbtool"
+pkgver=9.6.17
```
MoreRobustThanYou: Since I couldn't spot a version anywhere, I used the date of the latest change to it

drebrez:
I'll try to build the kernel for my huawei y530 which needs the dtbtool and see if it works with this one

aports/main/dtbtool/APKBUILD
```diff
@@ -1,21 +1,19 @@
...
+pkgrel=0
+pkgdesc="skales dtbtool"
+url="https://github.com/MoreRobustThanYou/skales-mirror/blob/master/dtbTool"
 arch="all"
-license="MIT"
-source="dtbtool.c::https://raw.githubusercontent.com/LineageOS/android_device_qcom_common/cm-14.1/dtbtool/dtbtool.c Makefile"
+license="BSD-2"
+source="https://raw.githubusercontent.com/MoreRobustThanYou/skales-mirror/${_commit}/dtbTool"
```
drebrez: why not getting it directly from `https://source.codeaurora.org/quic/kernel/skales/plain/dtbTool?id=c2bfa15aa5bfc4e3f1b32de28e327b80b37db1cf` ?

MoreRobustThanYou: didn't know about that, thanks!

aports/main/dtbtool/APKBUILD
```diff
@@ -1,21 +1,19 @@
+_commit="c2bfa15aa5bfc4e3f1b32de28e327b80b37db1cf"
```
drebrez: @MoreRobustThanYou instead of the commit you can use the tag, which can also be used as version.
They have tag `1.6.0` and you can have `pkgver=1.6.0` and the url for the file will be `https://source.codeaurora.org/quic/kernel/skales/plain/dtbTool?h=${pgkver}`

FYI: tag 1.6.0 points to commit c2bfa15aa5bfc4e3f1b32de28e327b80b37db1cf

drebrez:
@MoreRobustThanYou I've rebuilt the `linux-huawei-y530` package with this new dtbtool and the resulting `dt.img` is exactly the same as the one generated with the lineageos dtbtool (compared using `diff`).
The only problem I see now is the downgrade of the version. since skales properly have tags and it's version `1.6.0` while the previous dtbtool was version `3`. 
@ollieparanoid how we handle the version downgrade?

ollieparanoid:
@MoreRobustThanYou: thank you very much for making this package!

@drebrez: great reviewing and testing!

Good catch with the version downgrade. I think the only mechanism `apk` provides for this case is [`replaces`](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#replaces). So you *would* need to rename the new package to `dtbtool-skales`, and make it replace `dtbtool`.

But than again we don't handle this consistently yet (i.e. we have qt packages with `pkgver=9999` to overrule the Alpine packages), so I would say it's good enough for now if we delete the old `dtbtool` package from the repo (which I just did, run `pmbootstrap update` to update your APKINDEXes).

PR #1050, opened 2017-12-28
Add Teclast X80 Pro (EFI/x86_64 tablet)

magmastonealex:
This appears to be the first x86_64 device added. It boots with standard EFI instead of a custom bootloader, so installation is a little different (but documented [on the wiki](https://wiki.postmarketos.org/wiki/Teclast_X80_Pro_(teclast-x80pro))).

Because it's an Intel device with Intel graphics, this device is already effectively mainlined, with full 3D acceleration support as well. Unfortunately, our existing `linux-postmarketos-*` packages don't include enough things in the kernel config to boot on any real x86_64 machine I've tried. I spent quite a bit of time experimenting, trying to find the right options to make things work with no luck. So, I created `linux-teclast-x80pro` as a temporary workaround. It's an exact clone of `linux-postmarketos-stable`, but using the Arch Linux kernel config. This results in a very large kernel package, with lots of modules we don't need, but it does boot. At some point in the future the two packages should be merged, removing the clearly unnecessary modules.

Interestingly, very little of this work is specific to the Teclast device. I also tested this procedure on a similar cheap Windows tablet, and a Surface Pro 3 with success. I think that this work, with small modifications (probably just deviceinfo) is likely to work on most other EFI-booting x86_64 devices.


aports/main/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -48,7 +48,13 @@ mount_subpartitions() {
 			2)
 				echo "Mount subpartitions of $i"
 				kpartx -afs "$i"
-				break
+				# Ensure that this was the *correct* subpartition
+				# Some devices have mmc partitions that appear to have
+				# subpartitions, but aren't our subpartition.
+				if blkid | grep -q "pmOS_boot"; then
+					break
```
drebrez: @magmastonealex if it's not the *correct* subpartition maybe you can add `kpartx -d "$i"` to unmount it. What you think?

aports/firmware/firmware-teclast-x80pro/APKBUILD
```diff
@@ -0,0 +1,22 @@
...
+
+# Source
+_commit="cc77e7b6092c54500058cd027b679421b9399905"
+source=$pkgname-$_commit.tar.gz::$url"/archive/${_commit}.tar.gz"
+
+package() {
+	cd "$srcdir/rtl8723bs-${_commit}"
+
+	# Wifi
+	install -Dm644 "rtl8723bs_nic.bin" \
```
ollieparanoid: That file is already provided by `linux-firmware`:
https://pkgs.alpinelinux.org/contents?file=rtl8723bs_nic.bin&path=&name=&branch=&repo=&arch=

So we don't even need to add a custom firmware package 👍

magmastonealex: I noticed that. I just hesitated to add 300mb of binaries which won't get used for a single file. Though looking at the math, that's under 5% of the total space used on the install partition. I've updated the PR to switch over the package.

aports/main/postmarketos-mkinitfs/init_functions.sh
magmastonealex: Good point. Updated the PR to do so.

magmastonealex:
@ollieparanoid I did try `lsmod` to figure out the exact modules needed, that's also how I came up with the list in the `deviceinfo`. 

Unfortunately, I'm not sure it's a module issue at the root of this - I can't even get into the initramfs, it hangs somewhere before that, with nothing on screen. I have applied the platform specific options with no luck. I agree a  diff is a good option moving forward, but it's more than 3000 lines to sort through :(

ollieparanoid:
Ouch, that is no fun. I recommend diffing side by side, because kernel configs are divided in sections. So you can see when sections begin, and skip irrelevant sections.

Alright, I recommend that we merge this one as it is, but after we tested that the initramfs is still working in Qemu and on another device, just to make sure. Then you could make another PR at some point in the future to merge the kernel back, when you have time. What do you think?

magmastonealex:
Sounds good to me. I've tested in QEMU and on my Samsung i747, and it's worked on both. It might be a good idea to have someone that's not me give it a go though :)

Thanks for the tips for diffing kernel configs, that's good to know!

ollieparanoid:
Tested on `samsung-i9100` with FDE and boots up fine 👍

Issue #1055, opened 2017-12-29
linux-xiaomi-ido doesn't compile


ollieparanoid:
It fails for `linux-xiaomi-ido` with:
```
...
  OBJCOPY arch/arm64/boot/Image
  GZIP    arch/arm64/boot/Image.gz
No valid dtbs found
```
@lawl: do you know how to fix this directly / could you make a PR that fixes it?

lawl:
fairly sure i found the error, old version of kernel config, PR comming after this compile cycle is trough and verifies it.

lawl:
nope, failure seems to come from dtbTool, investigating

lawl:
Jumping one commit behind 19e06cf7f280c220bc6851a6af648a4ca0f134f2 and it compiles again, I haven't fully `bisect`ed it but I strongly suspect this is the commit to blame.

Have we compiled all other kernels since that commit? Is this the only one that breaks?

z3ntu:
I just tried compiling linux-fairphone-fp2 and the image still boots (= I can access the debug shell).

ollieparanoid:
So much for drop-in replacement 😕 

@drebrez compiled `linux-huawei-y530` with it and compared the resulting `dt.img` file, and it [was the same one](https://github.com/postmarketOS/pmbootstrap/pull/1048#issuecomment-354310805) as with the old `dtbTool`.

The following devices depend on `dtbtool`, I will check if they still compile:
- [x] linux-amazon-thor (OK, tested by @MoreRobustThanYou)
- [x] linux-fairphone-fp2 (OK, tested by @z3ntu)
- [x] linux-huawei-y530 (OK, tested by @drebrez)
- [ ] linux-oneplus-bacon
- [ ] linux-oneplus-onyx
- [ ] linux-t2m-flame
- [x] linux-xiaomi-ido (NOK)

@lawl: thanks for investigating this and for the PR!
@z3ntu: thanks for confirming that the `linux-fairphone-fp2` still works!

z3ntu:
```
 /m/h/p/pmbootstrap   master $…  ./pmbootstrap.py build linux-oneplus-bacon --arch=armhf --force   1148ms  Sam 30 Dez 2017 13:55:02 CET
[13:55:05] (native) install dtbtool python2 xz
[sudo] password for luca: 
[13:55:07] (native) install
[13:55:07] (native) build armhf/linux-oneplus-bacon-3.4.113-r7.apk
[14:02:55] Done
```
[...]

PR #1056, opened 2017-12-29
Clean up xiaomi-ido kernel config

lawl:
Note, this does **not** fix https://github.com/postmarketOS/pmbootstrap/issues/1055, as I said, the issue is with 99% certainty the dtbTool update, since one commit before that, I have just verified as known good. But I came across this when checking out https://github.com/postmarketOS/pmbootstrap/issues/1055.


ollieparanoid:
@lawl: does it still boot with your PR and the new `dtbTool`?

Iam not sure I understand the fix. Looks like you reset this option to the default:
```
config BUILD_ARM64_APPENDED_DTB_IMAGE
	bool "Build a concatenated Image.gz/dtb by default"
	depends on OF
	help
	  Enabling this option will cause a concatenated Image.gz and list of
	  DTBs to be built by default (instead of a standalone Image.gz.)
	  The image will built in arch/arm64/boot/Image.gz-dtb
```
Which was disabled before in the case of your device meaning it did not create the `arch/arm64/boot/Image.gz-dtb` file. So this means the new `dtbTool` requires this file to exist?

Thanks for making this PR!

lawl:
@ollieparanoid no it doesn't and it doesn't fix anything. But on the current version I had edited the config manually and set

    CONFIG_BUILD_ARM64_APPENDED_DTB_IMAGE=n

all the other build options depend on this, so calling menuconfig does exactly the modification in this PR. This is only cosmetic and changes nothing.

ollieparanoid:
Thanks for your quick reaction and submitting a workaround. But it's obvious now that we can't just use the new version for all kernels - so I'll make a new PR that packages both versions and uses the old one for ido again.

PR #1057, opened 2017-12-30
[dtbTool] Fix version detection and find dtb in subfolders

drebrez:
With a simple fix seems to work fine.

This fixes the compilation of `linux-t2m-flame` and the generated `dt.img` is exactly the same as the one generated with the old dtbtool.

I'm compiling the `linux-xiaomi-ido` right now but I don't have the `dt.img` generated with the previous dtbTool to compare with

Close: #1055


drebrez:
@ollieparanoid I've added a patch to find dtb files generated in subfolders. Now also the `linux-xiaomi-ido` builds correctly

ollieparanoid:
Awesome work @drebrez! \o/
I'm rebuilding all kernels depending on dtbtool with the one from this branch 👍

PR #1058, opened 2017-12-30
Package the dtbtool from lineage next to the skales one

ollieparanoid:
Workaround for #1055.

We ship a new `dtbTool` from Skales in the current master branch, because it works better with mainline kernels. That tool should act like a drop-in replacement for the `dtbTool` from LineageOS we used before.

However we found out, that two kernel forks we ship don't compile with it anymore.
As a workaround, this PR packages both versions, as `dtbtool-skales` and `dtbtool-lineage`. All packages use the Skales version (it seems to be maintained better), except for the two kernels which fail with it. Both can be installed side by side, as the executable has been renamed as well.

I have compiled all kernels with this PR and it runs through.

@drebrez, @lawl, @MoreRobustThanYou, @z3ntu: Could one of you review the PR?


z3ntu:
Btw Debian calls it skales-dtbtool, not dtbtool-skales (https://packages.debian.org/sid/amd64/skales/filelist)

ollieparanoid:
Thanks for the comments! I think fixing up the skales' version is better than shipping both, which makes this PR obsolete. See https://github.com/postmarketOS/pmbootstrap/issues/1055#issuecomment-354558853 for ideas how to move on with this issue.

PR #1062, opened 2017-12-30
Modified Plasma launch scrict to (crudely and unreliably) detect if it has to run on the framebuffer or not

PureTryOut:
It would be nice if someone could test this on a non-hardware accelerated device. So far I only had the possibility to test on Qemu, which the check doesn't reliably work for (it always enters the first option).

If everything works, Plasma should appear first boot without modifications.


MoreRobustThanYou:
tbh we should just add "deviceinfo_has_hardware_acceleration"

PureTryOut:
I rather have a solid way to detect it on the device itself. `glxinfo` from the `mesa-demos` package would've been nice, but it depends on a running X server. I'd love an utility that can detect it without running either X or Wayland. Possibly we can write one ourselves?

PureTryOut:
Well, at least I can confirm this works on my Amami's. It launches kwin succesfully first boot. I think it's good enough for now, maybe we can think of a better way to detect this in the future.

MoreRobustThanYou:
Still doesn't work on amazon-thor. Nothing new, it didn't work before, after manually changing start_plasma.sh, either.

PureTryOut:
No wonder, this PR only changes what you have already done manually before. It wouldn't make sense if it suddenly started working for you.

PR #1063, opened 2017-12-30
Packages swiftshader

MoreRobustThanYou:
https://github.com/google/swiftshader

Currently doesn't compile.... due to musl related issues. Gcompat doesn't really help.

Log: https://pastebin.com/5emAqVKv


PureTryOut:
As we discussed before, could you please wait with making a PR till there is *something* working? PR's are meant to be reviewed and tested by others when there is something to test, none of which is the case here. If you want help from others, just push it to a public branch (you can push it to a separate branch in this repository if you want too) and mention it in Matrix/IRC.

Now to figure out the issue, could you compile with `-j1`? So `pmbootstrap -j1 build --force --strict swiftshader`. This will make pmbootstrap compile it with just 1 core. It'll be slower, but the error will be *way* easier to spot (basically at the bottom of the log).

PR #967, opened 2017-12-02
[linux-postmarketos-mainline] update to 4.15-rc1

limiter121:
**Note:** This includes a fix (hack?) for #950.

PR #1067, opened 2018-01-01
Update linux-postmarketos-mainline to 4.15-rc6

z3ntu:
cgit doesn't generate proper patches for binary files (one was added in
v4.15-rc1 and removed in v4.15-rc5 because it also introduced a new
build-time dependency which is now solved. For 4.16 the "old" source
lines can be re-added but until then a full diff skipping v4.15-rc1
is needed.

Replaces #967 

Anami patch was regenerated by git btw


z3ntu:
I would rather use the patch in linux-next instead of the APKBUILD workaround.

ollieparanoid:
(build runs through)
That' fine for me as well, could you adjust the APKBUILD for it?

PR #1070, opened 2018-01-02
[initramfs] Set framebuffer mode earlier in the script (before hooks)

drebrez:
This small change mainly because the `debug-shell` hook shows the splash-screen and without the framebuffer mode set nothing gets displayed


lawl:
For `xiamo-ido` this doesn't matter since the original framebuffer setting doesn't work anyways and my custom hook doesn't care about this. So fine with me.

PR #1074, opened 2018-01-03
Added new device: Sony Xperia J

WilliamO7:
So this is the second device I've ported, and boy was this one difficult!
- Two new packages were added, unpackelf and mkelf to deal with the non-standard boot images the phone uses (don't ask me why, blame Sony.)
- Therefore, pmb and the mkinitfs programs were modified to accomodate this.
- Info will go up into the wiki soon(tm) - need to do other stuff first.

 Device status:  Convinced the phone does not boot at all. I cannot get a usb shell and the screen stays turned off.


pmb/parse/bootimg.py
```diff
@@ -42,30 +42,69 @@ def bootimg(args, path):
             raise RuntimeError("File is a Kernel image, you might need the 'heimdall-isorec'"
                                " flash method. See also: "
                                "<https://wiki.postmarketos.org/wiki/Deviceinfo_flash_methods>")
+        elif "elf" in file_output.lower():
```
drebrez: @WilliamO7 are you sure it's enough to check for `elf` in the output of the `file` command? because every executable file will go through this

aports/main/mkelf/APKBUILD
```diff
@@ -0,0 +1,21 @@
+pkgname=mkelf
+pkgver=10.1
+pkgrel=1
+pkgdesc="Sony ELF (boot.img equivalent) creation tool"
+url="https://github.com/LineageOS/android_device_sony_tamsui-common/blob/jellybean-3.4/tools/mkelf.py"
+arch="noarch"
+license="APACHE2"
+depends="python2"
+subpackages=""
+source="$pkgname::https://raw.githubusercontent.com/LineageOS/android_device_sony_tamsui-common/jellybean-3.4/tools/mkelf.py"
```
drebrez: @WilliamO7 you are pulling the file directly from a branch, if they push new changes to that branch the package will be broken. I suggest using tags or commit directly to avoid it

aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -188,7 +188,7 @@ create_uboot_files()
 # Android devices
 create_bootimg()
 {
-	[ "${deviceinfo_generate_bootimg}" == "true" ] || return
+	[ "${deviceinfo_generate_bootimg}" == "true" ] && [ "${deviceinfo_bootimg_sonyelf}" == "false" ] || return
```
drebrez: @WilliamO7 if the `deviceinfo_bootimg_sonyelf` variable is not defined the functions returns. This will break all the other devices.

pmb/aportgen/device.py
```diff
@@ -100,6 +100,7 @@ def generate_deviceinfo_fastboot_content(args, bootimg=None):
     if bootimg is None:
         bootimg = {"cmdline": "",
                    "qcdt": "false",
+                   "sonyelf": "false",
```
ollieparanoid: I propose to refactor the entire function as follows, so it only prints lines given in the `bootimg` parameter.
(I've changed `qcdt` and `sonyelf` defaults from `"false"` to `""`, because when the user skipped the boot.img analysis, we don't know if it is true or false - so setting it to `"false"` could be misleading)

```python
# Left: deviceinfo key, right: bootimg key
mapping = {"kernel_cmdline": "cmdline",
           "generate_bootimg": "generate_bootimg",
           "bootimg_qcdt": "qcdt",
           "bootimg_sonyelf": "sonyelf",
           "flash_offset_base": "base",
           "flash_offset_kernel": "kernel_offset",
           "flash_offset_ramdisk": "ramdisk_offset",
           "flash_offset_second": "second_offset",
           "flash_offset_tags": "tags_offset",
           "flash_pagesize": "pagesize"}

# Defaults in case the user skipped boot.img analysis
ret = "\n"
if not bootimg:
    ret += "# Run 'pmbootstrap bootimg_analyze' to get the values, then remove this comment\n"
    bootimg = {"cmdline": "",
               "qcdt": "",
               "sonyelf": "",
               "base": "",
               "kernel_offset": "",
               "ramdisk_offset": "",
               "second_offset": "",
               "tags_offset": "",
               "pagesize": "2048"}

# Return the deviceinfo block
for deviceinfo_key, bootimg_key in mapping.items():
    if field in bootimg:
        ret += "deviceinfo_" + deviceinfo_key + "=\"" + bootimg[bootimg_key] + "\"\n"
return ret
```

aports/main/mkelf/APKBUILD
```diff
@@ -0,0 +1,22 @@
+pkgname=mkelf
+pkgver=10.1
```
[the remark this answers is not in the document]
drebrez: @WilliamO7 where did you get the version `10.1`? That files seems it doesn't have a versioning

aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -361,7 +383,11 @@ replace_init_variables
 create_cpio_image "$tmpdir" "$outfile"
 append_device_tree
 create_uboot_files
-create_bootimg
+if "${deviceinfo_bootimg_sonyelf}" == "true"; then
```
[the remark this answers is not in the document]
drebrez: @ollieparanoid I agree with your solution, I didn't advised against this, it was just breaking the other devices

pmb/aportgen/device.py
drebrez: @ollieparanoid I like your solution with the mapping 👍 
I didn't tried it, but the `deviceinfo_generate_bootimg="true"` gets added correctly? since it's not in the `bootimg` it might get skipped.
Changing from `"false"` to `""` is good, I agree that could be misleading.

pmb/parse/bootimg.py
```diff
@@ -21,7 +21,18 @@
...
 
 
+def check_sonyelf(path, file_output):
+    # Check the file output for the expected types
+    case1 = "ELF 32-bit LSB executable, ARM"
+    case2 = "ELF 64-bit LSB executable, ARM"
+    if file_output != case1 or file_output != case2:
+        return False
+    else:
+        return True
```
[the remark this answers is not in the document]
drebrez: @WilliamO7 can you upload somewhere the boot.img you have? I'm curious about it

pmb/aportgen/device.py
ollieparanoid: I've proposed to change the `bootimg` variable in the [comment below](https://github.com/postmarketOS/pmbootstrap/pull/1074#discussion_r159571214) so it gets included.

aports/main/unpackelf/APKBUILD
```diff
@@ -0,0 +1,26 @@
...
+pkgver="21"
+pkgrel=1
+pkgdesc="A single tool to unpack Sony ELF kernel format image variants"
+url="https://github.com/osm0sis/unpackelf/"
+arch="all"
+license="APACHE2"
+depends=""
+_commit="2735dd0b35d2e40a03008eb609b465acb83190db"
+source="
+$pkgname-$_commit.tar.gz::https://github.com/osm0sis/unpackelf/archive/${_commit}.tar.gz
```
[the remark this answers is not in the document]
osm0sis: @ollieparanoid tagged at 2017.12.13, should be ready to go for you now. Cheers!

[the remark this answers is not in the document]
ollieparanoid: Awesome, thanks!

mirh:
Man, I don't want to be disrespectful, and all, but I think like you canned the tree to base. 
This is the [latest one](https://github.com/TamsuiCM11/android_kernel_sony_msm7x27a/tree/jellybean-3.4) from "the team", and [here](https://github.com/TamsuiCM11/local_manifest/wiki/Useful-links-and-patches-for-fixing-stuff)'s more plot. 
You (all) should frequent a bit more XDA forums (totally absurd how this doesn't even seem to be a thing)

Anyway I guess this is even of a bigger feat, if you got it to boot? Gg.

ollieparanoid:
Thanks for the input @mirh. You're saying you would recommend the *TamsuiCM11* kernel and linked to the most important commits of it, right? Why do you recommend it?

PS: If this becomes a longer discussion, I'd recommend that we put it into a new issue in the tracker, so we don't clutter up this PR.

mirh:
Because it would be quite newer than the other. I mean, it isn't freexperia official one*, and I'm not sure how much of that work was half-assed rather than quality one, but (for as much as I don't own the device) certainly that would be my first rom choice. 
And I just linked to the branch they say to use with J in the readme. 

*whose latest branch is [this](https://github.com/LineageOS/android_kernel_sony_msm7x27a/tree/jellybean) btw

drebrez:
@WilliamO7 @ollieparanoid I think that almost everything have been addressed, only few things are left:
- proper versioning of `unpackelf` with tagged release
- version of `mkelf`?
- resolve merge conflict
- perform some manual test again
- add test case with boot.img [here](http://www.mediafire.com/file/0pqh9g6tj3v6ndj/boot.img)

ollieparanoid:
@William07: if you have enough time it would be awesome if you could resolve the outstanding issues. However, if you have no time, let us know and we'll take a shot at it.

PureTryOut:
I believe @WilliamO7 has been offline from Github for a while. Should we fix the remaining issues ourselves?

Also, @ollieparanoid, you notified the wrong William07 :wink:

WilliamO7:
I'm still here, but uni work, and the fact that my laptop works on a btrfs filesystem (and I didn't bring the phone up here either) means i've been out of action for a while.
You're free to tie up the odds and ends though - I probably won't be back to this for another two weeks at least.

ollieparanoid:
I think this makes more sense, when there's someone with the device to actually test it. So maybe we can try this in a few weeks if you're still interested and have access to your device again, I'd be happy to finish this up together :+1: 
We won't run out of work until then, no worries :wink:

PR #1080, opened 2018-01-04
Update Plasma to 5.11.5 and make QML use software rendering in case of framebuffer

PureTryOut:
This should prevent QML apps from crashing on framebuffer devices.

I've also added a patch to stop Plasma from setting the brightness to 1 at boot. This comes from upstream, and can be removed once Plasma 5.12 is out.


aports/main/postmarketos-ui-plasma-mobile/start_plasma.sh
```diff
@@ -18,6 +18,7 @@ if test -z "${XDG_RUNTIME_DIR}"; then
 		else
 			export KWIN_COMPOSE=Q
 			ck-launch-session kwin_wayland --framebuffer --xwayland plasma-phone 2>&1 | logger -t "$(whoami):plasma-mobile"
+			export QT_QUICK_BACKEND=software
```
ollieparanoid: I think this doesn't have any effect if you put it after the `ck-launch-session kwin_wayland...` line.

PureTryOut: Well it's not needed for `kwin`, only for any QML apps launched once the phone has finished booting.

ollieparanoid:
*Packages built in the repo for all architectures.*

Issue #923, opened 2017-11-20
postmarketos-demos is broken


ollieparanoid:
* The shutdown button is broken (we don't run weston as root anymore, hence it doesn't work)
* Firefox is not installed by default (maybe note that in the button)?

PabloCastellano:
> The shutdown button is broken (we don't run weston as root anymore, hence it doesn't work)

What do you think about adding the user to a group that can run `poweroff` without being root?
In UNIX systems it may make sense that any user cannot shutdown the machine where other users may be working on. But for a smartphone I don't think this is a major drawback

Issue #464, opened 2017-08-25
Adding postmarketos-artwork


MartijnBraam:
I guess we need a postmarketos-artwork package to provide default wallpapers for this distro

The *-artwork package name is based on how debian handles artwork for different distro variations, these packages provide a bunch of images in /usr/share (https://packages.debian.org/sid/all/debian-edu-artwork-lines/filelist)

I think we should provide wallpapers and a icon to be used for "start" buttons

Photo's that I made that can be used: http://imgur.com/a/73XW6 (I haven't added a licence to them yet, anyone knows what would be the best one?)

PabloCastellano:
About the licensing I would choose some creative commons variant.
See https://creativecommons.org/choose/

MartijnBraam:
Here is a mint-style wallpaper created with blender

![pmos](https://user-images.githubusercontent.com/6928199/29723289-4ce97d9c-89c4-11e7-9d53-a5f84f3986d5.png)

The blender source:
[pmos.zip](https://github.com/postmarketOS/pmbootstrap/files/1252432/pmos.zip)

MartijnBraam:
I found the solution for the stem of the "p" when using the full name. Smallcaps:

![postmarketos](https://user-images.githubusercontent.com/6928199/29728159-235fbd0c-89d7-11e7-814c-f1482bbcb8bc.png)

Jgjot-Singh:
I have some photos(my own work) that I'd like to make available. Should I just create an imgur gallery and link to it here?

MartijnBraam:
We've moved the artwork to https://github.com/postmarketOS/artwork, you can create a PR there with your wallpapers

moondeck:
I like the idea of a common package, but doesn't that require to either:

1. Have individual packages for each DE (as they store wallpapers in different places)

2. Have some sort of DE detection

PabloCastellano:
I ignore if different DE have different locations for wallpapers. However for this case, we could achieve it this way:

1. Create a postmarketos-artwork package that install the wallpapers to some location like `/usr/share/wallpapers`
2. By using the [install_if](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#install_if) variable, copy or symlink the files to the proper location for the installed DE.

ollieparanoid:
[...]

PR #1082, opened 2018-01-04
Modified postmarketos-demo as discussed on IRC:

moondeck:
Removed shutdown (as weston isn't running as root anymore)
Changed the firefox button to state that firefox needs to be installed

Fixes https://github.com/postmarketOS/pmbootstrap/issues/923


moondeck:
That would be great... if i knew how to implement that.

ollieparanoid:
I *think* the proper way to do this would be with `consolekit2` and sending `d-bus` messages:
https://wiki.archlinux.org/index.php/ConsoleKit#Use_D-Bus_for_power_operations

The upside is, if you could implement it that way, then it should work for XFCE4 and other desktop environments that expect it to work the same way, too. So if you want to do it that way, I recommend to test if it allows shutting down in XFCE4 etc., too (e.g. in Qemu).

We will probably need something like this in the long run anyway, so we can really shutdown as user from everywhere.

moondeck:
Looks good. Should i just make the program execute a command like that, or is there a more elegant way to do it? I also assume we need to give the user shutdown permission by default in the build system?

ollieparanoid:
I have edited my comment afterwards :) If you did implement it, then yes you could call the d-bus command like that. But right now we don't even start weston with consolekit2 (we start kwin with it though), so you would need to modify the weston start script etc - which would be solved more elegantly by using a real display manager such as sddm (which will probably work soon with Alpine Linux anyway).

moondeck:
Ah, yeah, sorry, didn't see that. How about using something like nodm?

ollieparanoid:
That one only works with X11, not with Wayland.

moondeck:
Well, then sddm would be perfect... but what about older devices? I imagine SDDM and the whole KDE stack takes quite a bit of RAM and processor power. We could always go ahead and reinvent the wheel... but that's a task even bigger.

ollieparanoid:
It is relatively lightweight on dependencies, and I think it doesn't load any of it at runtime if you use the autologin mode (which makes sense if you only have one user and one desktop environment). In contrary to KDM it does not depend on the KDE stack. Adelie Linux (which is also based on Alpine Linux) has packaged it [here](https://code.foxkit.us/adelie/packages/blob/master/user/sddm/APKBUILD) (it's highly likely that this gets upstreamed to Alpine later), so you can take a look at the depends and makedepends (libraries it links against are automatically added to the depends after compilation).

moondeck:
Thanks! Should have mentioned it was tested on my Z1. I'm gonna try to do https://github.com/postmarketOS/pmbootstrap/issues/464

PR #1083, opened 2018-01-04
Revert "init: Add manufacturer to pkgdesc when creating new device (#913)"

Halamix2:
This reverts commit 99d7b58ee5f00ad9f628691ed5720bd501df2779.

People usually add manufacturer name in the phone full name which
results in having manufacturer written double in pkgdescr (e.g. `Samsung
Samsung Galaxy Mini 2`)


PabloCastellano:
For me it makes more sense if we educate people to write the manufacturer name only in the manufacturer variable :smile:

ollieparanoid:
Well, the manufacturer name is *not always* included in the product name.
Example: "Google Nexus 4" (LG is the manufacturer)

Right now the `deviceinfo_name` is the official device model name ("Google Nexus 4" (LG) or "Samsung Galaxy SII" (Samsung)), as it is in the [device table](https://wiki.postmarketos.org/wiki/Devices). That name may or may not include the manufacturer as seen in both examples.

So... if we *don't* include any company name in the device model name, that information is "lost" from the `deviceinfo` file, that's why I would recommend to include it and accept this PR (we discussed the different options for the "Samsung Samsung"-in-pkgdesc-problem in the channel).

What do you think about that @PabloCastellano?

PabloCastellano:
@ollieparanoid I still don't see the point and I am not able to find that discussion in the channel, but if you approve it, then I do as well

ollieparanoid:
@PabloCastellano: sorry that I couldn't explain it better. The logs are scattered across the channel unfortunatelly, and the search in matrix doesn't really work. I have my "wall of posts" which answer the backlog saved in an extra file though, so basically it was:

(halamix2 asking if we could fix the "samsung samsung" pkgdesc somehow from not getting generated that way)

me:
> halamix2: re: full name without manufacturer: sometimes the manufacturer isn't in the full name, e.g. lg-hammerhead is Google Nexus 5. The idea was to put the full name in there, which may or may not contain the vendor. Not sure if that was your point though, please elaborate on what you would like to change - and whenever you have an idea for a improvement, no need to ask me, the community must like it ;)

(halamix2 went on explaining)

me:
> halamix2: do you recommend to revert #913, or that we always leave out the company name in deviceinfo_name? (again, there's the Google Nexus example where the device is actually produced by LG, so the "Google" would be missing from the name).

(halamix2 said, that he'd like to revert #913, I thought about it and found that this makes more sense for the points listed above - tl;dr: the "Google" in the "LG Nexus 5" would get lost otherwise - and he made this PR.)

PR #1084, opened 2018-01-05
Add support for Motorola Google Nexus 6 (motorola-shamu)

EricJ2190:
Flashing and USB networking work. This display only works if the framebuffer is refreshed manually or if msm-fb-refresher is started after the device has finished booting. If msm-fb-refresher starts at boot, the postmarket splash is replaced by a black screen and further refreshing does nothing. The touch screen does not work, and none of the /dev/input/eventX devices respond to touch input.


aports/device/linux-motorola-shamu/APKBUILD
```diff
@@ -0,0 +1,81 @@
+# Kernel config based on: arch/arm/configs/(CHANGEME!)
```
ollieparanoid: Please adjust this.

EricJ2190:
I have added the defconfig path and added the device to the wiki.

ollieparanoid:
Thank you very much! I'll merge it as soon as I can confirm that it builds.

PR #1086, opened 2018-01-05
Let new and most old devices depend on mesa dri swrast

### LINKED ISSUE #1013: Most devices need mesa-dri-swrast, but we don't install it yet. How to proceed?
Most devices do not have 3d acceleration yet, so almost all of them would need to install `mesa-dri-swrast` when `mesa` is installed:

```shell
subpackages="$pkgname-mesa"
mesa() {
	install_if="$pkgname mesa"
	depends="mesa-dri-swrast"
	mkdir "$subpkgdir"
}
```
Currently for `armhf`, the package is `3.15` MB in size, and `7.47` MB when installed ([source](https://pkgs.alpinelinux.org/package/edge/main/armhf/mesa-dri-swrast)).

But most devices don't have it, so Hildon for example will not work out of the box for them.
I think we have two options:
a) add this block to all devices, that do not have 3D acceleration, and add it to the `aportgen` code, so it is there by default for new devices
b) add this block to `postmarketos-base` and remove it from all device packages that currently have it

What does everybody think about it?

ollieparanoid:
Closes #1013.


drebrez:
@ollieparanoid Do you want me to try all the various UI with a specific device? just to see if they all works? let me know

ollieparanoid:
@drebrez, @magmastonealex: Thanks for reviewing!
@drebrez: yes, that would be nice!

drebrez:
@ollieparanoid tested the following UI and all works fine out of the box:
- `weston`
- `xfce4`
- `mate`
- `hildon`
- `plasma`

I never tried yet the Luna UI, I will do it in the next days

ollieparanoid:
Awesome, thank you very much \o/

PR #1087, opened 2018-01-06
[WIP] Basic support for Sony Xperia Arc ("semc-anzu")

asiekierka:
So far, it boots successfully to Weston and supports the touch screen. I can also SSH into the device over USB. More support in progress.

There is no TWRP version capable of installing the sideload package (I might be able to get a port of TWRP 3.1.0 working, however), so you'll have to install it using a computer directly to the SD card. As my Xperia Arc is officially unlocked, I used fastboot to install the kernel.

Most 2011-era Xperia devices seem to share a lot of similarities internally, so hopefully getting one to work will make it easy to get all of the other ones to work.


aports/device/device-semc-anzu/APKBUILD
```diff
@@ -0,0 +1,22 @@
+# Reference: <https://postmarketos.org/devicepkg>
+pkgname=device-semc-anzu
+pkgdesc="Xperia Arc"
+pkgver=1
+pkgrel=10
```
drebrez: @asiekierka please reset `pkgrel` back to **0**, since in master branch will be the initial version

aports/device/device-semc-anzu/deviceinfo
```diff
@@ -0,0 +1,35 @@
...
+deviceinfo_external_disk_install="true"
+deviceinfo_arch="armhf"
+
+# Device related
+deviceinfo_keyboard="false"
+deviceinfo_external_disk="true"
+deviceinfo_screen_width="480"
+deviceinfo_screen_height="854"
+deviceinfo_dev_touchscreen="/dev/input/event2"
+deviceinfo_dev_touchscreen_calibration="1.051104 0.019693 -28.753983 0.039463 1.001910 -21.101807"
```
drebrez: @asiekierka did you try if it works without the calibration? most devices doesn't need it and your looks like the default value

asiekierka:
Okay, I checked and applied the requested changes. I also added everything which is, in theory, necessary to support Wi-Fi - unfortunately, my Xperia Arc is not in good shape and, while it authenticates correctly with wpa_supplicant, it fails to send any subsequent packets; the chip is also rather unstable. I'm pretty sure this is a problem with my hardware, though.

asiekierka:
Notable issue: Plasma Mobile does not work:

libevdev error in fix_invalid_absinfo: BUG: Device "cy8ctma300_touch" has invalid ABS_MT_TRACKING_ID range

mirh:
Uh, well, aside of recommending him to check what we were talking about in #346 I wouldn't know. 
I found there's another version of touchscreen driver [here](https://android.googlesource.com/kernel/msm.git/+/android-msm-sony-cm-jb-3.0/drivers/input/touchscreen/cy8c_ts.c) (it doesn't seem that rare of a model). 

Anyway I see they are progressing in #1088

asiekierka:
One question I would like to ask: Is it worth pursuing separate, smaller firmware packages for things which are contained in linux-firmware? I don't think it's worth pulling in >200MB of firmware just for a few files.

ollieparanoid:
> Is it worth pursuing separate, smaller firmware packages for things which are contained in linux-firmware?

I like this idea, ideally this would happen upstream in Alpine. I've asked on the [alpine-devel](https://lists.alpinelinux.org/alpine-devel/6005.html) mailing list, let's see what they think about it.

mirh:
Just for the records, at least for wi-fi firmware, linux-firmware [sucks](https://bugzilla.kernel.org/show_bug.cgi?id=196541), compared to android.

ollieparanoid:
Thanks for opening that bug report, I hope it gets fixed upstream (so all Linux distros benefit from it). Maybe you can push this further by writing to LKML.

On the other hand, that 2013 firmware version will be affected by broadpwn as well, right? Alpine [ships](https://github.com/alpinelinux/aports/blob/8c2d539d41c31a8eef29d5308373f686fd381e37/main/linux-firmware/APKBUILD#L18) an unofficial version of the `43430` firmware with a fix for broadpwn at least. In my opinion, what we really need (for postmarketOS) is getting nexmon going #592 so we can binary patch out such vulnerabilities by ourselves.

mirh:
43430 is not included in google's repos (and for as much as I can see seems almost a [RPi](https://github.com/RPi-Distro/firmware-nonfree/tree/master/brcm80211/brcm)-only thing)
Nonetheless, I guess it's pretty recent enough that broadpwn is there. 

\*On the other hand\*, back to us, for example brcm4330 (for as much as latest firmware version is dated 2013) [doesn't seem](http://boosterok.com/blog/broadpwn/) to be vulnerable. 

Based on that blog, also, my educate guess is that you can consider safe anything that received an [update](https://android.googlesource.com/platform/hardware/broadcom/wlan/+log/master/bcmdhd/firmware) from June 2017 onwards.

Issue #1088, opened 2018-01-06
Plasma Mobile (and libevdev) fails to load with Xperia Arc touchscreen driver


asiekierka:
Jan  1 01:59:59 localhost user.notice user:plasma-mobile: kglobalaccel-runtime: Failed to register service org.kde.kglobalaccel
    Jan  1 01:59:59 localhost user.notice user:plasma-mobile: libevdev error in fix_invalid_absinfo: BUG: Device "cy8ctma300_touch" has invalid ABS_MT_TRACKING_ID rangeCouldn't start kglobalaccel from org.kde.kglobalaccel.service: QDBusError("org.freedesktop.DBus.Error.Disconn
    Jan  1 01:59:59 localhost user.notice user:plasma-mobile: FATAL ERROR: backend failed to initialize, exiting now
    Jan  1 01:59:59 localhost user.notice user:plasma-mobile: Couldn't start kglobalaccel from org.kde.kglobalaccel.service: QDBusError("org.freedesktop.DBus.Error.Disconnected", "Not connected to D-Bus server")

bhush9:
Is this all information you see in log? Maybe can you attach full log with logread?

asiekierka:
@bhush9 Updated with all lines pertaining to plasma-mobile's execution. I am pretty confident this is an issue in the touchscreen driver, though.

bhush9:
Do you have ~/.config/QtProject/qtlogging.ini file?

asiekierka:
@bhush9 No, I don't. Sorry. I'll correct that now.

asiekierka:
My confidence has failed me. Anyhow, here's the complete log.

bhush9:
Okay it seems kwin is missing the support for framebuffer format of your device, I'll fix it

asiekierka:
@bhush9 You might want to wait with that - I just discovered that the red and blue channels are swapped. Haven't noticed before - perhaps a "testcard image viewer" program would be useful in postmarketOS demos?

asiekierka:
#1089 contains a fix, though I still can't get it to work very reliably - this might be related to the lack of hardware acceleration, though. Either way, it does show something!

mirh:
@robclark so.. would you still be up for that A2xx kernel work you talked about in the past?
If yes, what kernel would be cool enough? 4.4? Or already 3.10?

(yes, I guess like I'm a pro at hijacking threads)

ollieparanoid:
[...]

PR #1089, opened 2018-01-06
semc-anzu: correct pixman type to match framebuffer

asiekierka:
Apologies for the omission.

If #1088 requires any additional changes in device-semc-anzu, it's probably a good idea to delay this pull request until said changes are done.

### DIFF
--- a/aports/device/device-semc-anzu/APKBUILD
+++ b/aports/device/device-semc-anzu/APKBUILD
@@ -2,7 +2,7 @@
 pkgname=device-semc-anzu
 pkgdesc="Xperia Arc"
 pkgver=1
-pkgrel=0
+pkgrel=1
 url="https://postmarketos.org"
 license="MIT"
 arch="noarch"
--- a/aports/device/device-semc-anzu/deviceinfo
+++ b/aports/device/device-semc-anzu/deviceinfo
@@ -19,7 +19,7 @@ deviceinfo_dev_touchscreen="/dev/input/event2"
 deviceinfo_dev_touchscreen_calibration=""
 deviceinfo_dev_keyboard=""
 deviceinfo_msm_refresher="true"
-deviceinfo_weston_pixman_type="2"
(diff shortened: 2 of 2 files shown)


asiekierka:
Checksum updated.

asiekierka:
I am still looking for a better solution, as Xorg also seems to have issues (xfce4 segfault).

ollieparanoid:
There's also a kernel a patch you could try out. But that might cause the [wrong colors](https://wiki.postmarketos.org/wiki/Troubleshooting:display#Kernel_patch) to appear in kwin. You might also need a X11 config file.

So I'm merging this not yet if you still want to experiment with it - just let me know when you think it is ready 😉

asiekierka:
Yes, I think I can come up with something better. Managed to add a new display mode to the framebuffer in an experiment and this will hopefully let me support everything more properly.

asiekierka:
Initial progress - I have managed to patch the Qualcomm framebuffer driver to add multiple missing framebuffer modes (BGR_888 and ABGR_8888, as well as many modes which weren't fully implemented in the code), however autodetection seems to be broken in both Weston and Xorg. As such, for the time being, I have to call *both* fbset (different modes for Weston and Xorg!) and pass a pixman_type to Weston.

TODO:

- [ ] Consult documentation and check if endianness differences aren't the issue for Weston's detection problems
- [ ] Test plasma-mobile

asiekierka:
Okay, this works with Weston, Xorg-based environments *and* KWin (plasma-mobile). All that needs to be tested now is osk-sdl, but I won't be able to do so for the next week or so.

asiekierka:
Please do not merge this PR until I verify that the patch in https://github.com/postmarketOS/pmbootstrap/pull/1005 isn't an infinitely simpler way to accomplish the same thing.

PureTryOut:
So, @asiekierka, did you manage to verify that patch yet?

PR #1092, opened 2018-01-07
Show UI selection ordered by name, `none` remains at the beginning

drebrez:
I've noticed that the UI selection order was different every execution of `pmbootstrap init`.
This was due the usage of a dictionary which doesn't maintain the order.
Since `pmb.helpers.ui.list(args)` function returned this dictionary it was possible to order it later by the keys but this was showing the `none` ui in the middle of the list.
I changed the `pmb.helpers.ui.list(args)` function to return a list of tuples instead, so that the order of the data inserted is maintained.
In the `ask_for_ui(args)` function I had to change the way to validate if the user input is valid, and the simplest way is to convert the list of tuples to a dictionary and get the list of keys from it, and just check if the user input is in this list.

Result:
```
[23:10:42] Available user interfaces (6):
[23:10:42] * none: No graphical environment
[23:10:42] * hildon: (X11) Lightweight GTK+2 UI (optimized for single-touch touchscreens)
[23:10:42] * luna: (Wayland) webOS UI, ported from the LuneOS project (Not working yet)
[23:10:42] * mate: (X11) MATE Desktop Environment, fork of GNOME2 (stylus recommended)
[23:10:42] * plasma-mobile: (Wayland) Mobile variant of KDE Plasma, optimized for touchscreen
[23:10:42] * weston: (Wayland) Reference compositor (demo, not a phone interface)
[23:10:42] * xfce4: (X11) Lightweight GTK+2 desktop (stylus recommended)
```


drebrez:
`Python 3.4+`? 

```
drebrez@drebrez-laptop:~/pmbootstrap$ python --version
Python 2.7.12
```

PR #1094, opened 2018-01-07
qemu-amd64: make xfce4, mate work out of the box with SDL display

### LINKED ISSUE #672: [qemu-amd64] Xserver does not work
I added xfce4 and xf86-video-modesetting,pciutils,xf86-input-libinput,xf86-server-vesa packages. Unfortunately, I'm getting " vesa: ignoring device with bound kernel driver" message. According to google, it seems to be bug in X server, and newer X.org would solve that..?

ollieparanoid:
Fixes #672.
* `mate`, `xfce4`: depend on `mesa-egl`
* `xfce4`: update config, so it does not have the hardcoded `/home/user` reference anymore (https://github.com/postmarketOS/xfce4-phone/pull/2)

### How to test

This should start QEMU without problems:
```shell
pmbootstrap config ui xfce4 && \
  pmbootstrap -y zap && \
  yes | pmbootstrap -y install --no-fde && \
  pmbootstrap qemu
```


drebrez:
@ollieparanoid last week when I tried MATE before closing the PR, I noticed that `mesa-egl` was necessary only if there was the `mesa-dri-swrast` package installed. 
Don't know exactly why, maybe we can have the `install_if="mesa-dri-swrast"` here?
I tried to test it again but I zapped the chroot with all the possible options and also executed `apk cache clean --purge` in all the chroots and after installing again in the samsung i9070 the X11 wasn't working anymore, I still have to investigate why

ollieparanoid:
> last week when I tried MATE before closing the PR, I noticed that `mesa-egl` was necessary only if there was the `mesa-dri-swrast` package installed.

I did a lot of Qemu debugging lately, and with `qemu-amd64` + `mesa-dri-virtio`, X11 didn't start for XFCE4 and MATE because `/usr/lib/libEGL.so.1` was missing (it said so in `logread`). That file is provided by `mesa-egl`. What do you think?

> I tried to test it again but I zapped the chroot with all the possible options and also executed `apk cache clean --purge` in all the chroots and after installing again in the samsung i9070 the X11 wasn't working anymore, I still have to investigate why

Sorry to hear that 🙁 
I've just flashed my `i9100` from `master` with `XFCE4` and FDE enabled, and it worked. Then I've added `mesa-egl` and rebooted, and it still booted into `XFCE4`. Let me know if I can help you with debugging this issue.

drebrez:
@ollieparanoid with my last comment I wanted to say that both UI they worked fine even without the `mesa-egl` IIRC, I'm not sure about qemu. But with `mesa-dri-swrast` installed, those ui also requires the `mesa-egl` right?
My question was if we should install `mesa-egl` only if there is `mesa-dri-swrast` installed (or `mesa-dri-virtio` also?).

About X11 not working on the i9070 it's something not related to this, I even installed pmos with `none` ui and installed xorg manually and it wasn't working. no idea why but we can discuss it in chat

ollieparanoid:
> with my last comment I wanted to say that both UI they worked fine even without the `mesa-egl` IIRC, I'm not sure about qemu. But with `mesa-dri-swrast` installed, those ui also requires the `mesa-egl` right?
My question was if we should install `mesa-egl` only if there is `mesa-dri-swrast` installed (or `mesa-dri-virtio` also?).

Thanks for explaining again. So right now all devices depend on either `mesa-dri-swrast` or another `mesa-dri` package. If we added an `install_if` here for the very specific case that no `mesa-dri-*` package is installed, then we would not use that code path and we wouldn't notice if it breaks. For that reason I recommend we always install `mesa-egl`.

I'll try to join the chat later 👍

drebrez:
@ollieparanoid Only QEMUs have always a `mesa-dri` package, a device with working gpu probably doesn't have it so probably they don't really need that `mesa-egl` right?
Anyway, for me it's completely fine to always install them since most devices requires it, but it would be nice to understand if it's a problem for devices with working gpu or not.

ollieparanoid:
> a device with working gpu probably doesn't have it so probably they don't really need that mesa-egl right?

As I understand it, devices with a working GPU just have a different mesa-dri package, e.g. `mesa-dri-freedreno`:

https://github.com/postmarketOS/pmbootstrap/blob/12340fe5f60ced7d8917ad8fc8d24100f511a37f/aports/device/device-sony-castor-windy/APKBUILD#L8

...and with #1086 all devices without GPU acceleration depend on `mesa-dri-swrast`.

> if it's a problem for devices with working gpu or not.

I can't imagine that this package is a problem, because it [gets pulled in by weston anyway](https://pkgs.alpinelinux.org/package/edge/testing/armhf/weston) (mesa-egl is listed as dependency on the right).

ollieparanoid:
Thank you very much. The countless reviews you have been doing are very important and I highly appreciate them!

PR #1095, opened 2018-01-07
Improved pmbootstrap init

### LINKED ISSUE #472: Check if it is possible to create device nodes in the work folder
We've had a few cases where the device node creation did not work. It would be nice if we check that during `pmbootstrap init` and tell the user what could cause this and what to do about it (use another location, link to a wiki page with details).

Possible causes:
* Shared VM folder
* NTFS folder?
* Old versions of ecryptfs
* maybe encfs

ollieparanoid:
* Check if it's possible to create and read from a device node directly when initializing a chroot (closes #472)
* Copy the Qemu binary into the forign-arch chroots before initializing them, so the post-install scripts directly work during the chroot setup and we don't need to call `apk fix` afterwards
* Use `pmb.helpers.repo.update()`, which only updates the `APKINDEX` files if they are older than 4 hours, instead of using `apk`'s repo update function which always downloads the `APKINDEX` files
  *  Chroot initialization
  *  Getting the initial `APKINDEX` to download `apk-tools-static`
  * Updating the `APKINDEX` at the start of `pmbootstrap install`
* Fixed a bug in `from_chroot_suffix`: the `buildroot_x86_64` has architecture `x86_64`, not `x86`. This was randomly discovered because Travis didn't have the APKINDEX files in `buildroot_x86_64`, because they didn't get mounted right.

All in all this PR will make the workflow a bit faster and `pmbootstrap` will produce less traffic.

### DIFF
--- a/pmb/chroot/init.py
+++ b/pmb/chroot/init.py
@@ -70,20 +70,21 @@ def create_device_nodes(args, suffix):
 
         # Verify major and minor numbers of created nodes
         for dev in pmb.config.chroot_device_nodes:
-            stat_result = os.stat(chroot + "/dev/" + str(dev[4]))
+            path = chroot + "/dev/" + str(dev[4])
+            stat_result = os.stat(path)
             rdev = stat_result.st_rdev
-            assert os.major(rdev) == dev[2]
-            assert os.minor(rdev) == dev[3]
+            assert os.major(rdev) == dev[2], "Wrong major in " + path
+            assert os.minor(rdev) == dev[3], "Wrong minor in " + path
 
         # Verify /dev/zero reading and writing
         path = chroot + "/dev/zero"
-        logging.debug("Test device node: " + path)
         with open(path, "r+b", 0) as handle:
-            assert handle.write(bytes([0xff]))
-            assert handle.read(1) == bytes([0x00])
+            assert handle.write(bytes([0xff])), "Write failed for " + path
+            assert handle.read(1) == bytes([0x00]), "Read failed for " + path
 
     # On failure: Show filesystem-related error
     except Exception as e:
+        logging.info(str(e) + "!")
         raise RuntimeError(error)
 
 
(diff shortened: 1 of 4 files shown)


ollieparanoid:
Added `dont_merge_pr` because Travis is failing.

pmb/parse/arch.py
```diff
@@ -42,7 +42,7 @@ def from_chroot_suffix(args, suffix):
     if suffix == "rootfs_" + args.device:
         return args.deviceinfo["arch"]
     if suffix.startswith("buildroot_"):
-        return suffix.split("_", 2)[1]
+        return "_".join(suffix.split("_", 2)[1:])
```
drebrez: @ollieparanoid as mentioned in chat, split function already handles it without the need to join the multiple results. Specifying `1` it will always split at the first underscore only.
```
suffix.split("_", 1)[1]
```

pmb/chroot/init.py
```diff
@@ -45,6 +45,41 @@ def copy_resolv_conf(args, suffix="native"):
...
+                                            ])
+            except Exception as e:
+                raise RuntimeError(error)
+
+    # Test /dev/zero, so we know the device nodes are working
+    path = chroot + "/dev/zero"
+    logging.debug("Test device node: " + path)
+    try:
+        with open(path, "rb", 0) as handle:
+            handle.read(1)
```
drebrez: @ollieparanoid in the previous block it creates the devices nodes only if they don't exists, so it may already exist a file called `.../dev/null` and it will not create it.
Here we only check if we are able to read that file, but it might be a real file.
Maybe we can check if that file is actually a device node with the correct major/minor.
Or what about writing a byte different than `0` and read it again checking that it actually returns `0`?

ollieparanoid: Good ideas, implemented all improvements you've suggested :+1:

ollieparanoid:
That sounds nice, implemented the detailed errors as suggested. Thanks again!

PR #1096, opened 2018-01-07
Added artwork packages

moondeck:
It works, but not all environments use /usr/share/wallpapers (even though they probably should use something standard), but i think a way to solve it is to ship some kind of pmOS specific config files with all environments, just like other distros. Let me know what you think and if that is what we're doing i shall do that next.


aports/main/postmarketos-artwork-wallpapers-extra/APKBUILD
```diff
@@ -0,0 +1,22 @@
...
+_hash="2cad2af027fbec6aef0d1f16226a41ccf427f44e"
+
+pkgname="postmarketos-artwork-wallpapers-extra"
+pkgver=1
+pkgrel=0
+pkgdesc="Wallpapers for postmarketOS"
+url="http://postmarketos.org/"
+arch="noarch"
+license="CC-BY-SA"
+source="$pkgname-$_hash.tar.gz::https://github.com/postmarketOS/artwork/archive/${_hash}.tar.gz"
```
MartijnBraam: The $pkgname-$_hash on this line wil make it download the wallpapers twice from github to build both packages, I it can be changed to postmarketos-artwork-wallpapers-$_hash to prevent that and maybe the two apkbuilds could be merged into a single apkbuild with a subpackage for -extra

aports/main/postmarketos-artwork/APKBUILD
```diff
@@ -0,0 +1,14 @@
+# Contributor: Olgierd "moondeck" Nowakowski <olgierd AT novakovsky.eu>
+# Maintainer: Olgierd "moondeck" Nowakowski <olgierd AT novakovsky.eu>
+pkgname="postmarketos-artwork"
+pkgver=1
+pkgrel=0
+pkgdesc="meta-package for the postmarketOS artwork"
+url="http://postmarketos.org/"
+arch="all"
+license="CC-BY-SA"
```
MartijnBraam: Do empty packages have a licence? @ollieparanoid

aports/main/postmarketos-artwork/APKBUILD
```diff
@@ -0,0 +1,14 @@
+# Contributor: Olgierd "moondeck" Nowakowski <olgierd AT novakovsky.eu>
+# Maintainer: Olgierd "moondeck" Nowakowski <olgierd AT novakovsky.eu>
+pkgname="postmarketos-artwork"
+pkgver=1
+pkgrel=0
+pkgdesc="meta-package for the postmarketOS artwork"
+url="http://postmarketos.org/"
+arch="all"
```
MartijnBraam: this should probably be noarch instead of all

aports/main/postmarketos-artwork/APKBUILD
moondeck: I would guess yeah, because its packaging "work" (not that i care)

MartijnBraam:
Closed in favor of the other PR #1097

PR #1099, opened 2018-01-07
i9195: add initial support

rrooij:
Current status:

* Red glare on screen (probably need kernel patch)
* No touch support
* Networking on USB with SSH does not work, but the device boots normally and does not freeze, so this is probably just a kernel config that needs to be set


rrooij:
dmesg log when connecting it with USB:

```
[12575.946616] usb 2-1.2: new high-speed USB device number 12 using ehci-pci
[12576.055903] usb 2-1.2: New USB device found, idVendor=04e8, idProduct=685d
[12576.055910] usb 2-1.2: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[12576.055915] usb 2-1.2: Product: MSM8960
[12576.055918] usb 2-1.2: Manufacturer: Sasmsung
[12615.173769] usb 2-1.2: USB disconnect, device number 12
```

zhuowei:
@rrooij Try disabling CONFIG_USB_ANDROID_SAMSUNG_COMPOSITE: it looks like that disables non-mtp modes in the USB composite driver.

rrooij:
@zhuowei Thanks for your suggestion, I will take a look at it.

rrooij:
@zhuowei The kernel fails to compile when `CONFIG_USB_ANDROID_SAMSUNG_COMPOSITE` is turned off. The kernel does compile when I disable both `CONFIG_USB_ANDROID_G` and the Samsung Composite option, but it has the same problem with no SSH.

Using the ethernet USB Gadget driver does not work either.

drebrez:
as discussed in chat, the usb issue might be something like this https://github.com/postmarketOS/pmbootstrap/blob/master/aports/device/device-htc-bravo/initfs-hook.sh#L8

rrooij:
@ollieparanoid It's been added to the device overview and has now a entry on the wiki:
https://wiki.postmarketos.org/wiki/Samsung_Galaxy_S4_Mini_LTE

I already made a patch for this but haven't included it yet, as I was afraid that it might lead to other problems that I can't foresee.

aports/device/device-samsung-i9195/deviceinfo
```diff
@@ -0,0 +1,35 @@
...
+deviceinfo_bootimg_qcdt="false"
+deviceinfo_flash_offset_base="0x80200000"
+deviceinfo_flash_offset_kernel="0x00008000"
+deviceinfo_flash_offset_ramdisk="0x02000000"
+deviceinfo_flash_offset_second="0x00f00000"
+deviceinfo_flash_offset_tags="0x00000100"
+deviceinfo_flash_pagesize="2048"
+deviceinfo_flash_heimdall_partition_kernel="BOOT"
+deviceinfo_flash_heimdall_partition_system=""
+
```
Halamix2: I would suggest adding `deviceinfo_weston_pixman_type="2"`line
https://wiki.postmarketos.org/wiki/Deviceinfo_reference#weston
(kernel patch will probably break KDE so for now adding this line is the preferred way of fixing Weston)

zhuowei:
@rrooij The kernel linked on the Wiki page seems to be wrong: you might want to change it to link to `android_kernel_samsung_msm8930-common` instead.

rrooij:
@zhuowei thanks, the wiki entry is fixed

PR #1101, opened 2018-01-08
Add basic support for Nokia N9

filippz:
Basic support includes working display (no GPU support at this time) and touchscreen with xfce4 and USB networking.
Contains watchdog support for devices with more than one watchdogs.


filippz:
Thanks for the fixes - they are fine by me.
Also a big thanks to @pavelmachek for his help!

ollieparanoid:
@filippz, @pavelmachek: I've moved the Nokia N9 and N950 up in the [devices](https://wiki.postmarketos.org/wiki/Devices) page from non-booting to booting devices. Is the table filled out correctly for it? If not, please change it (or if you have no time, write here what's wrong). Thanks!

PR #1103, opened 2018-01-08
tf101: add initial support

rrooij:
Initial support for the TF101.

Boots into Weston, but doesn't support USB Networking.

Status:

* Boots into Weston
* Screen colors are fine
* Flashing with normal tools does not work, flashing needs to be done with dd inside recovery (see the [wiki entry](https://wiki.postmarketos.org/wiki/ASUS_Eee_Pad_Transformer_(asus-tf101)) for this device)
* USB networking does not work, neither seems touch screen to be working. It doesn't freeze after loading Weston though, since the device is able to go in standby and be waked up (although the Weston lock screen can't be unlocked since touch does not seem to be working)

bootscreen
weston


aports/device/device-asus-tf101/APKBUILD
```diff
@@ -0,0 +1,22 @@
+# Reference: <https://postmarketos.org/devicepkg>
+pkgname="device-asus-tf101"
+pkgdesc="Asus Asus Eee Pad Transformer"
```
ollieparanoid: One "Asus" is enough. (I know, autogenerated. `pmbootstrap` is patched now so this doesn't happen anymore 😉 )

aports/device/linux-asus-tf101/APKBUILD
```diff
@@ -0,0 +1,98 @@
+# Kernel config based on: arch/arm/configs/tf101-linux_defconfig
+
+pkgname="linux-asus-tf101"
+pkgver=3.1.10
+pkgrel=0
+pkgdesc="Asus Asus Eee Pad Transformer kernel fork"
```
ollieparanoid: One "Asus" is enough. (I know, autogenerated. `pmbootstrap` is patched now so this doesn't happen anymore 😉 )

ollieparanoid:
I can confirm that it builds successfully!

aports/device/device-asus-tf101/APKBUILD
rrooij: Done! Thanks for your remarks.

LongnoseRob:
Thanks for the work on the TF101,
unfortunatelz the wiki-page  on the TF101 is not clear to me:
Step 4: ./blobpack blob /tmp/postmarketOS-export/boot.img-asus-tf101 LNX
What should LNX be? the kernel-file?

rrooij:
@LongnoseRob 
LNX is not the kernel file. `boot.img-asus-tf101` is the kernel file and `LNX` is the name of the partition. It is needed for some reason to let the tablet know that the blob should be on the LNX partition.

ollieparanoid:
If you guys are interested, it should be possible to streamline this a bit more with `pmbootstrap`. Here are ideas that build on each other and could be done in independent PRs:
* Package `blobpack`
* Create a new deviceinfo variable `deviceinfo_generate_blobpack` and let the initramfs script generate the blob file automatically, when it is set (just like `deviceinfo_generate_bootimg`).
* Export the blob file along with the rest with `pmbootstrap export`
* Implement a new flash method, which executes the commands described in the wiki when running `pmbootstrap flasher flash_system` and `pmbootstrap flasher flash_kernel`.

drebrez:
@rrooij @ollieparanoid I've found on internet that it has an APX mode which is something like fastboot and allows to flash the device. More info here => https://forum.xda-developers.com/wiki/APX_mode

rrooij:
@drebrez 
cc @ollieparanoid 
I know about APX mode, but you have to use the proprietary `nvflash` in order to flash it. It is comparable to Odin, but a free software equivalent like Heimdall hasn't been created yet.

And my exact revision of this device doesn't support nvflash.

rrooij:
The problems with touch and USB should be fairly easy to solve by the way. We just have to look at what the Ubuntu installations do differently than we:

https://forum.xda-developers.com/showthread.php?t=2648862

I took the same kernel as used in this link.

rrooij:
@LongnoseRob I made an error in the wiki. LNX should come before the kernel image. I switched the command line arguments by accident. It is fixed now.

PR #1105, opened 2018-01-09
i9195: fix weston pixman type

rrooij:
Fix the red screen by defining the pixman type in the deviceinfo
file.
  
pixman_fix


aports/device/device-samsung-i9195/deviceinfo
```diff
@@ -33,3 +33,6 @@ deviceinfo_flash_pagesize="2048"
 deviceinfo_flash_heimdall_partition_kernel="BOOT"
 deviceinfo_flash_heimdall_partition_system=""
 
+# Weston pixman setting
+deviceinfo_weston_pixman_type="2"
+
```
PureTryOut: Trailing empty lines should be removed.

rrooij:
X does not boot with XFCE4, and I'm almost certain that it has to do with the pixman stuff, as I've experienced before. I think a kernel patch is needed as long term solution.

PR #1106, opened 2018-01-09
[samsung-i9070] Add proper fix for framebuffer

drebrez:
- Removed the framebuffer configuration in the initramfs hook
- Added patch for fixing the framebuffer mode in the kernel source
- Cleanup linux-samsung-i9070 APKBUILD


drebrez:
@ollieparanoid yes it fixes the X11 problem, when the `cat /sys/class/graphics/fb0/modes > /sys/class/graphics/fb0/mode` is performed before the hook, setting the `bits_per_pixel` and the `virtual_size` after, was useless. So I debugged the fb driver and discovered how to set the proper RGB value and how to compute the correct virtual width/height.
Everything works perfectly except (just discovered now) when X11 is rotated (with [this](https://github.com/postmarketOS/pmbootstrap/blob/88bea7534f242149296a3703ec80de94b3061326/aports/main/postmarketos-ui-hildon/x11-fbdev-rotate.conf) config), something trigger the driver to set some incorrect values somewhere and the colors/alignement is a little bit messed up.
I will continue with the debugging to find out what happens there. Or I'll look into mainlining it :smile:

PR #1110, opened 2018-01-10
Spaces -> Tabs in artwork package 

moondeck:
I apologize for this, should haven't missed that. Committed to master as i think creating another branch is unnecessary.

### DIFF
--- a/aports/main/postmarketos-artwork-wallpapers/APKBUILD
+++ b/aports/main/postmarketos-artwork-wallpapers/APKBUILD
@@ -20,9 +20,9 @@ package() {
 }
 
 extra() {
-  mkdir -p "$subpkgdir"/usr/share/wallpapers/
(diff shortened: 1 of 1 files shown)


PureTryOut:
> Committed to master as i think creating another branch is unnecessary.

Well, that's up to you, we don't really care 😛 

It's just that making a separate branch for everything you do makes sure you don't accidentally mix commits, as long as you're in the correct branch. And keeping `master` up-to-date with just upstream, allows you to use it as base branch for those other branches.

So say you now think of something new to do while this pull request isn't merged yet. If you continue working in `master`, the commit from this PR will also end up in the new PR you'll make, while it shouldn't be there. Making a separate branch at this point wouldn't matter, as the commit you made to `master` will just follow you along to the new branch. By keeping the `master` branch clean from your own commits, you can easily start working on something new: you change the active branch to master, checkout a fresh new branch, and start working.

PR #1111, opened 2018-01-11
Luna: Fix startup, add default background image

magmastonealex:
I'm in the midst of adding on screen keyboard support, and making startup easier on real devices. I noticed that something upstream has changed, and the Node.js modules wouldn't link at runtime anymore. This causes lots of issues. Bumped pkgrel on all three to fix.

I've also included a small fix to get a default background image from postmarketos-artwork.

### DIFF
--- a/aports/luna/luna-init/APKBUILD
+++ b/aports/luna/luna-init/APKBUILD
@@ -1,24 +1,16 @@
 pkgname=luna-init
 pkgver=5.0.0_git20171117
-pkgrel=2
+pkgrel=1
 _commit=5fff640fb0b9f3e973c13d1620889b2da4c11766
-_artwork_commit="2cad2af027fbec6aef0d1f16226a41ccf427f44e"
 pkgdesc="Initialization and font setup files used by luna-sysmgr."
 arch="all"
 url="http://webos-ports.org"
 license="GPL-3.0+"
 depends=""
-makedepends="cmake-modules-webos python2 py2-tz imagemagick"
-source="$pkgname-$_commit.tar.gz::https://github.com/webOS-ports/luna-init/archive/$_commit.tar.gz
-postmarketos-artwork-$_artwork_commit.tar.gz::https://github.com/postmarketOS/artwork/archive/${_artwork_commit}.tar.gz"
+makedepends="cmake-modules-webos python2 py2-tz"
+source="$pkgname-$_commit.tar.gz::https://github.com/webOS-ports/luna-init/archive/$_commit.tar.gz"
 options="!check"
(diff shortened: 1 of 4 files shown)


drebrez:
@magmastonealex for the pmos wallpaper you can add the dependency to the `postmarketos-artwork-wallpapers` package, which installs them in `/usr/share/wallpapers/`, and in the luna package you can just create a symlink at the correct location. All this has been discussed in #464.
I also noticed that you need to convert it to jpg, maybe is better to have the jpg version directly in the artwork repository? or when packaging `postmarketos-artwork-wallpapers` create these variants on the fly? @MartijnBraam what you think about it?

aports/luna/luna-init/APKBUILD
```diff
@@ -34,6 +46,12 @@ package() {
 	install -v -m 644 "$builddir"/files/conf/default-dock-positions.json "$pkgdir"/etc/palm
 	install -v -m 644 "$builddir"/files/conf/default-launcher-page-layout.json "$pkgdir"/etc/palm
 	install -v -m 644 "$builddir"/files/conf/locale.txt "$pkgdir"/etc/palm
+	
+	# Install a default background image.
+
+	install -d "$pkgdir"/media/internal/wallpapers
+	install -v -m 644 "$srcdir"/build/postmarketos.jpg "$pkgdir"/media/internal/wallpapers/LuneOS.jpg
```
drebrez: @magmastonealex FYI: there's no need to create the folder first, you can use `install -D` which creates them automatically.
e.g: `install -Dm 644 <source> <target>`

magmastonealex:
@drebrez Yes, a symlink would be preferable, and that's what I had originally, but I needed to convert to jpg to make it work.

 If postmarketos-artwork-wallpapers packages JPGs, that would solve the problem in a much nicer way.

magmastonealex:
@ollieparanoid  Sounds good to me.

To your second PR suggestion though, does it make sense to purposefully exclude the PNG images? I feel like we may wind up in an identical situation soon enough, when another desktop comes along and expects a PNG by default.

ollieparanoid:
Is there any? As far as I know, JPG is suited better than PNG for wallpapers anyway, because the compression ratio works slightly better. [`archlinux-wallpaper`](https://www.archlinux.org/packages/community/any/archlinux-wallpaper/files/) also only consists of JPGs.

If you have an example where a desktop expects PNG as wallpaper format, then we should include both in my opinion - but if we don't know any, then I would go with JPG only until we really need PNG.

z3ntu:
jpg is better for "real-life" pictures but for example for solid colors (like a black or blue background) png is normally better

PR #1112, opened 2018-01-11
Fix black screen after splash on motorola-shamu

EricJ2190:
This fix allows msm-fb-refresher to be used with getting stuck on a black screen. It also re-enables the 3D driver. I originally disabled it in attempt to fix the framebuffer issues but that was unnecessary.


drebrez:
@EricJ2190 nice that you fixed it. Please increase the `pkgrel` number of both packages.

EricJ2190:
I bumped `pkgrel` on the kernel and `pkgver` on the device, following [the packaging advice on the wiki](https://wiki.postmarketos.org/wiki/Create_a_package#When_should_pkgver.2Fpkgrel_get_increased.3F). Please let me know if this is incorrect.

PR #1115, opened 2018-01-12
Add blobtools as package for tf101

rrooij:
This is for making a kernel image for the TF101 easier. See the
pull request comment by Ollieparanoid for more information:

https://github.com/postmarketOS/pmbootstrap/pull/1103#issuecomment-357035481

The TF101 does not support normal Android boot images and instead needs
a blob to be created.

This is my first package outside the device directory, so I hope I did it somewhat good. I can confirm that it builds successfully.

### DIFF
--- a/aports/main/blobtools/APKBUILD
+++ b/aports/main/blobtools/APKBUILD
@@ -1,5 +1,5 @@
 pkgname=blobtools
-pkgver=6186e33baa10ce6f1c738d7b91eb5153743105af
+pkgver=1.0
 pkgrel=0
 pkgdesc="Tools for unpacking & repacking blobs used for updating "hidden" partitions on ASUS Transformer"
 url="https://github.com/AndroidRoot/BlobTools"
@@ -8,7 +8,7 @@ license="Apache-2.0"
 depends=""
 makedepends="cmake"
 subpackages=""
-source="$pkgname-$pkgver.tar.gz::https://github.com/AndroidRoot/BlobTools/archive/${pkgver}.tar.gz"
+source="$pkgname-$pkgver.tar.gz::https://github.com/rrooij/BlobTools/archive/${pkgver}.tar.gz"
 options="!check"
 
 builddir="$srcdir"/BlobTools-$pkgver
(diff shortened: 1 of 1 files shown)


ata2001:
@rrooij I think you should ask the developer to tag a release on github, so we'll have a sane pkgver.

aports/main/blobtools/APKBUILD
```diff
@@ -0,0 +1,26 @@
+pkgname=blobtools
+pkgver=1.0
+pkgrel=0
+pkgdesc="Tools for unpacking & repacking blobs used for updating "hidden" partitions on ASUS Transformer"
```
PureTryOut: Wrapping the word `hidden` in `"` breaks this description. Wrap it in single `'` tags instead.

aports/main/blobtools/APKBUILD
```diff
@@ -0,0 +1,26 @@
...
+pkgver=1.0
+pkgrel=0
+pkgdesc="Tools for unpacking & repacking blobs used for updating 'hidden' partitions on ASUS Transformer"
+url="https://github.com/AndroidRoot/BlobTools"
+arch="all"
+license="Apache-2.0"
+depends=""
+makedepends=""
+subpackages=""
+source="$pkgname-$pkgver.tar.gz::https://github.com/rrooij/BlobTools/archive/${pkgver}.tar.gz"
```
ollieparanoid: Having a packaged version is nice, but personally I prefer using the upstream version over forking the package just to tag the version. The point of postmarketOS is avoiding unnecessary forks after all.

How about packaging it with the [git scheme](https://wiki.postmarketos.org/wiki/Create_a_package#Packaging_git_commits) for now, and asking upstream to tag the version in an issue. When/if upstream does the tagging, you could update the package accordingly.

And if it turns out, that upstream really does not answer, and we need to make big changes to the package (i.e. more than a small patch file), then I'd suggest to fork it.

Keep in mind that all this is only my own opinion, feel free to discuss everything 👍

rrooij:
I agree @ollieparanoid. I've changed it to use upstream with git versioning. I also made an issue on upstream.

rrooij:
@ollieparanoid thanks for merging and for fixing my mistake! I wasn't sure what to do with the version number as there were no release numbers.

Issue #978, opened 2017-12-04
Raise a flag when we must rebuild KDE packages because of a QT update


ollieparanoid:
Whenever Alpine updates its `qt5` packages, we must upgrade the `qt5-*` packages we have in our `aports` folder as well, otherwise they become incompatible. In #976 I have added a `pmbootstrap` testcase for that, so we should notice it pretty soon at least (the testsuite runs with every new commit and PR, so at the current rate at least once a day).

According to @bhush9, we need to update the following KDE packages, whenever a `qt` package gets upgraded, too:
* kwin
* plasma-integration
* kwayland

@bhush9 wrote:
> there is rebuild.list in qt5-base packaging (at least for arch)

@awilfox: Do you know if Alpine has a similar mechanism to such a `rebuild.list` file? And how does Adelie handle this?

If we can't find a better solution, the `APKINDEX` files store the timestamp when a package was built. We could add a `pmbootstrap` test, that checks if `kwin` etc. have a more recent timestamp than `qt5-qtbase` from Alpine.

awilfox:
Adélie handles this manually right now.

Something like `rebuild.list` is fragile.  It would be much better if APK was able to read a tree, pick out deps, and rebuild dependent packages.  Basically, `revdep-rebuild` from Gentoo.  That is low on my todo list, after "shipping 1.0" definitely; so if you want to work on something like this or talk to Alpine people about it, go right head.

bhush9:
(FWIW, rebuild.list on archlinux is not automated thing, it's just a documentation basically)

ollieparanoid:
Ah, thanks for the input! I didn't realize that we can simply look at missing sonames to figure out which packages need to be rebuilt.

I think `apk` is already capable of [detecting missing dependencies](https://github.com/alpinelinux/apk-tools/blob/0700e8318fc5c88c5965320f6085f973e25dbeda/src/index.c#L91-L108) 
in an index file, and sonames automatically get written into the indexes as dependencies.

So I'll look into how it makes sense to handle this in `pmbootstrap` (we do dependency resolving ourselves before passing anything to `abuild` or `apk`, because then we can do it across multiple chroots of different architectures and before `apk`/`abuild` are even installed). Unfortunatelly that means, that the way it will be solved for postmarketOS, it will probably not work for Alpine - but if it does, I'll write to the mailing list.

Duplicate of #893.

awilfox:
[...]

PR #1116, opened 2018-01-12
Gracefully handle packages breaking because of soname bumps

### LINKED ISSUE #893: Properly handle soname bumps
Alpine just updated their `libressl` to a new [soname](https://en.wikipedia.org/wiki/Soname) version. Currently, `pmbootstrap` does not handle that, and users need to manually delete packages, that depend on `libressl` to build them again (e.g. `weston`).

After generating a package, Alpine automatically adds the `.so` versions a package links against to the `depends`, which should make it possible to handle this problem.

NOTE: Alpine handles this by bumping the pkgrel of all packages, which link against the library in question. That way, it causes a rebuild on their buildbots.

ollieparanoid:
Fixes #893. I ended up writing a lot here in the description in the hope that this makes it easier to understand the problem and how it's solved. Feedback whether that is good or bad is welcome (but actual reviewing is more important :wink:).

## Introduction: soname bumps break our packages
Linux programs as packaged in Alpine (and most other distributions) link against dynamic libraries (`.so` files). These libraries may release incompatible upgrades, which "bump" (increase) the version of the library file (the "soname").

Whenever you build a package with Alpine's `abuild` (directly or e.g. via `pmbootstrap build`), and it uses such dynamic libraries, `abuild` will automatically scan for those and add the exact soname as dependency.

Here's one example. All `depends` starting with `so:` are such sonames:
```shell
$ pmbootstrap parse_apkindex ~/.local/var/pmbootstrap/packages/x86_64/APKINDEX.tar.gz testapp
{
    "pkgname": "testapp",
    "version": "0-r1",
    "arch": "x86_64",
    "timestamp": "1500000000",
    "depends": [
        "testlib",
        "so:libc.musl-x86_64.so.1",
        "so:libtestlib.so.0"
    ],
    "provides": [
        "cmd:testapp"
    ]
}
```

postmarketOS uses its own binary package repository together with the ones from Alpine. That means: When a postmarketOS package links against a dynamic library from Alpine, it's currently possible that said library gets a soname bump in Alpine, and then you can't install the binary postmarketOS package anymore because it was built against the old version.

Alpine's `apk` package manager will write an error - but it is **not obvious how the user can work around that** if we didn't notice it and fix it in the binary repo: You'd need to mark the postmarketOS aport, which has the outdated dependency, for rebuild. To do it properly, one would increase the `pkgrel` (that's also how Alpine handles rebuilds for soname upgrades, I've asked in their IRC channel).

For reference, this is what `apk` says when a dependency had a soname bump:
```
  so:libtestlib.so.1 (missing):
    required by:
                 testapp-1.0-r0[so:libtestlib.so.1]
                 testapp-1.0-r0[so:libtestlib.so.1]
```

### New command: `pmbootstrap pkgrel_bump --auto`
Automatically find all packages from the postmarketOS binary repository (disable with `pmbootstrap --mirror-pmOS=""`) as well as the locally built packages, that are broken because of soname bumps. Increase the `pkgrel` in each package's `APKBUILD` to mark it for rebuilt.

#### Variations
* `pmbootstrap pkgrel_bump --auto --dry`: dry run mode, doesn't change anything and yields a non-zero exit code in case there's a package that would be bumped (this was `--auto-dry` before and got changed, see below)
* `pmbootstrap pkgrel_bump hello-world`: bump the `pkgrel` of one specific package without checking the soname dependencies, useful for related problem #978

### Automatic check before installing packages
Before any package gets installed (e.g. during `pmbootstrap install`, which [generates a full system image](https://wiki.postmarketos.org/wiki/Installation_guide#pmbootstrap_install)), each package and its dependencies are checked against being broken because of soname bumps. When that is the case, we abort and ask the user to run `pmbootstrap pkgrel_bump --auto`.

```
$ pmbootstrap install --add=testapp
*** (1/5) PREPARE NATIVE CHROOT ***
*** (2/5) CREATE DEVICE ROOTFS ("qemu-amd64") ***
NOTE: Run 'pmbootstrap pkgrel_bump --auto' to mark packages with outdated dependencies for rebuild. This will most likely fix this issue (soname bump?).
NOTE: More dependency calculation logging with 'pmbootstrap -v'.
ERROR: Could not find package 'so:libtestlib.so.1' in the aports folder and could not find it in any APKINDEX.
Run 'pmbootstrap log' for details.
See also: <https://postmarketos.org/troubleshooting>
```

### Testcases
* `test_soname_bump.py`: executes the dry run, fails if it finds broken packages (so we check for breakage in our binary packages repo whenever a new pull request comes in, usually at least once per day, [example output](https://travis-ci.org/postmarketOS/pmbootstrap/builds/328275911#L689)).
* `test_soname_bump.py`: builds the new test aports `testlib` and `testapp` from `test/testdata/pkgrel_bump/aports`, and creates the soname breakage on purpose to verify that `pkgrel_bump --auto` works as intended.

## How to test
After reviewing the code, please create the soname problem locally, and verify that `pmbootstrap` properly detects it at installation time:
```shell
# Copy testlib and testapp to aports, so we can build them
cp -r test/testdata/pkgrel_bump/aports aports/_temp

# Selet the qemu-amd64 device (because it uses x86_64 packages) and delete
# all chroots. Then build testapp and testlib (gets pulled in as dependency)
pmbootstrap zap
pmbootstrap config device qemu-amd64
pmbootstrap build testapp

# Bump the pkgrel of testlib. The package is written in a way that this will
# also increase the soname version.
pmbootstrap pkgrel_bump testlib

# This should rebuild testlib
pmbootstrap build testlib

# Delete the old testlib package to break testapp intentionally
sudo rm ~/.local/var/pmbootstrap/packages/x86_64/testlib-1.0-r0.apk
pmbootstrap index

# Perform a normal installation and ask pmbootstrap to add testapp. It should
# complain that the package depends on the missing "so:libtestlib.so.0" and
# abort the installation
pmbootstrap install --add=testapp

# This should fix the problem by increasing the testapp pkgrel
pmbootstrap pkgrel_bump --auto
pmbootstrap install --add=testapp

# Clean up
rm -r aports/_temp
sudo rm ~/.local/var/pmbootstrap/packages/x86_64/testlib*
sudo rm ~/.local/var/pmbootstrap/packages/x86_64/testapp*
pmbootstrap index
```


ollieparanoid:
Thanks for the review! Yeah that's how I've tried to do it first, but it didn't seem to be possible with `argparse` to make `--auto [--dry]` mutually exclusive with the `packages`.

Now that I think about it, we could add the `--dry` parameter in general, so it works with *both* `--auto` and `packages`. I'll update the PR accordingly.

MartijnBraam:
Yeah thats what I meant, don't make them exclusive but make `--dry` a modifier for the current command as a lot of people will expect.

ollieparanoid:
Updated. If you have time, please review again and test it 👍

MartijnBraam:
It seems to work good but I find this output not very nice:

```shell-session
$ ./pmbootstrap.py pkgrel_bump testlib --dry
[00:51:31] Increase 'testlib' pkgrel (3 -> 4)
[00:51:31] ERROR: pkgrels of package(s) would have been bumped!
[00:51:31] Run 'pmbootstrap log' for details.
[00:51:31] See also: <https://postmarketos.org/troubleshooting>
```

I also noticed I can't build `testapp` with `--force` after bumping the testlib but thats probably not an issue with the bumping code.

```
(004033) [00:55:39] Calculate depends of packages ['testlib'], arch: x86_64
(004033) [00:55:40] (native) build x86_64/testapp-1.0-r0.apk
(004033) [00:55:40] (native) % rm -rf /home/pmos/build
(004033) [00:55:40] % sudo cp -r /home/martijn/Projects/Other/pmbootstrap/aports/_temp/testapp/ /home/martijn/.local/var/pmbootstrap/chroot_native/home/pmos/build
(004033) [00:55:40] (native) % chown -R pmos:pmos /home/pmos/build
(004033) [00:55:40] (native) % cd /home/pmos/build && busybox su pmos -c 'CARCH=x86_64 abuild -d -f'
>>> testapp: abuild 3.1.0-r4
>>> testapp: Checking sanity of /home/pmos/build/APKBUILD...
>>> WARNING: testapp: No maintainer
>>> testapp: Cleaning temporary build dirs...
>>> testapp: Entering fakeroot...
/usr/bin/abuild: line 2554: ./testapp: not found
>>> ERROR: testapp*: check failed
>>> ERROR: testapp: all failed
```

ollieparanoid:
> It seems to work good but I find this output not very nice:

How would you do it?

> I also noticed I can't build `testapp` with `--force` after bumping the testlib but thats probably not an issue with the bumping code.

Good catch! That's a bug in the `APKBUILD` of `testapp` then, we probably need to `cd` in `check()` to fix this. I'll update the PR accordingly.

MartijnBraam:
I think it shouldn't trigger the normal error troubleshooting lines in the dry run mode so it becomes:

```shell-session
$ ./pmbootstrap.py pkgrel_bump testlib --dry
[00:51:31] Increase 'testlib' pkgrel (3 -> 4)
[00:51:31] pkgrels of package(s) would have been bumped!
```

So its probably as easy as just calling exit instead of throwing an exception to change the exit code.

ollieparanoid:
Fixed both issues. Please mark it as reviewed if you think it's ready to be shipped. Thanks!

PR #1120, opened 2018-01-13
Disallow running pmbootstrap as root

### LINKED ISSUE #1065: Don't allow pmbootstrap to run as root
We use sudo when we need to run commands as root. Two people have run pmbootstrap as root today and screwed up their work folder permissions that way, it would be better if we checked for this.

ollieparanoid:
Closes #1065.


pmb/__init__.py
```diff
@@ -37,8 +37,10 @@ def main():
 
     # Wrap everything to display nice error messages
     try:
-        # Sanity check
+        # Sanity checks
         other.check_grsec(args)
+        if not args.as_root and os.geteuid() == 0:
+            raise RuntimeError("Do not run pmbootstrap as root!")
```
NotKit: I wonder, does it makes sense to mention existence of "--as-root" option here?

ollieparanoid: I thought about that too and checked how `abuild` does it:
```
/ # abuild -h | grep root
 -F  Force run as root
```
```
/ # abuild
>>> ERROR: : Do not run abuild as root
```
After all we don't want to encourage the user running as root - and if the user really needs that, the option should be easy to find.

ollieparanoid:
Thanks a lot for reviewing and testing @magmastonealex and @NotKit!

PR #1125, opened 2018-01-15
pmbootstrap: add qcdt generation to the linux aportgen APKBUILDs

MayeulC:
This checks the next box in #688
When the device has bootimg_qcdt set to true, the following is done to
the linux APKBUILD:

 * Add dtbtool to makedepends
 * Call dtbTool during build() to generate dt.img
 * Add the generated dt.img in the package's boot/dt.img


MayeulC:
Merging this shouldn't be a problem, and should avoid the problem both @chin123 (I hope that's you, if not, my apologies) and I had when creating a new device port.

MayeulC:
OK, it seems that bootimg_qcdt is not set unless the flasher is heimdall-bootimg OR fastboot. So I changed this. There are some errors, still. For this patch, I changed the device parser so that it would change "true" (the string) to True (the boolean). Was it a bad idea?
I will publish the current revision shortly...

MayeulC:
OK, for whatever reason, I can't run the tests on my computer anymore, but it looks like they are passing here.

I would like to get more feedback on the True/False point here. And should I split it into a separate commit?

pmb/aportgen/linux.py
```diff
@@ -22,23 +22,62 @@
...
+
     content = """\
         # Kernel config based on: arch/""" + carch + """/configs/(CHANGEME!)
 
         pkgname=\"""" + pkgname + """\"
         pkgver=3.x.x
         pkgrel=0
-        pkgdesc=\"""" + name + """ kernel fork\"
-        arch=\"""" + arch + """\"
+        pkgdesc=\"""" + deviceinfo['name'] + """ kernel fork\"
```
ollieparanoid: For consistency with the rest of the code I recommend `deviceinfo["name"]` here.

pmb/aportgen/linux.py
```diff
@@ -22,23 +22,62 @@
...
+    build = """
+            unset LDFLAGS
+            make ARCH="$_carch" CC="${CC:-gcc}" \\
+            KBUILD_BUILD_VERSION="$((pkgrel + 1 ))-postmarketOS\""""
+
+    if deviceinfo.get("bootimg_qcdt") is True:
+        makedepends += " dtbtool"
+
+        build += """\n
+            # Generate master DTB
```
ollieparanoid: How about: `# Generate master DTB (deviceinfo_bootimg_qcdt)`

pmb/aportgen/linux.py
```diff
@@ -22,23 +22,62 @@
...
+            dtbTool -s 2048 -p "scripts/dtc/" -o "arch/arm/boot/dt.img" "arch/arm/boot/\""""
+
+        package += """\n
+            # Master DTB
```
ollieparanoid: How about: `# Master DTB (deviceinfo_bootimg_qcdt)`

MayeulC:
> Thank you very much for creating this PR @MayeulC! I've tested it, and it's working great. I like how you're simply passing the whole deviceinfo variable now. When the boolean-parsing is reverted, and bootimg_qcdt is added to the deviceinfo keys config, it's ready to be shipped in my opinion. 🚢

Well, thanks a lot to you for the lovely code and warm welcome :)
Passing the whole `deviceinfo` variable was first brought up by @drebrez on IRC.

> Could you please add bootimg_qcdt there?
I might as wall make a second, separate commit. This, plus a few other missing fields to carry over, a comment to warn to keep the other updated, and maybe a bit of wiki documentation. I also took the liberty of rearanging the lines and categories so that they would be the same there and in the aportgen.

> Right now, such a boolean variable in deviceinfo can actually be "true", "false" or an empty string.
And it still is True, False or "". But I agree that it's rather material for another PR, if ever needed. That's not too important anyway. But maybe add some checks somewhere in the future? We can discuss this further in the future if needed.

I also agree with the minor fixups. Let me finalize that, and I'll send the PR.

MayeulC:
Created issue #1128 for the tests problem, with my diagnostic. It might or might not be related to the previous error I had.

pmb/aportgen/linux.py
```diff
@@ -22,23 +22,62 @@
...
+            done
+            if ! [ -e "$_target" ]; then
+                error "Could not find zImage in $PWD!"
+                return 1
+            fi"""
+
+    build = """
+            unset LDFLAGS
+            make ARCH="$_carch" CC="${CC:-gcc}" \\
+            KBUILD_BUILD_VERSION="$((pkgrel + 1 ))-postmarketOS\""""
```
drebrez: @MayeulC could you please add 4 additional spaces before KBUILD_BUILD_VERSION? so that the resulting line will be generated like it was before
```
	KBUILD_BUILD_VERSION="$((pkgrel + 1 ))-postmarketOS"
```

pmb/aportgen/linux.py
```diff
@@ -22,23 +22,62 @@
...
+        makedepends += " dtbtool"
+
+        build += """\n
+            # Generate master DTB (deviceinfo_bootimg_qcdt)
+            dtbTool -s 2048 -p "scripts/dtc/" -o "arch/arm/boot/dt.img" "arch/arm/boot/\""""
+
+        package += """\n
+            # Master DTB (deviceinfo_bootimg_qcdt)
+            install -Dm644 "$builddir/arch/arm/boot/dt.img" \\
+            "$pkgdir/boot/dt.img\""""
```
drebrez: @MayeulC could you please add 4 additional spaces before `"$pkgdir/boot/....`? so that the resulting line will be generated with a proper indentation
```
# Master DTB (deviceinfo_bootimg_qcdt)
install -Dm644 "$builddir/arch/arm/boot/dt.img" \
	"$pkgdir/boot/dt.img"
```

pmb/aportgen/linux.py
```diff
@@ -22,23 +22,62 @@
...
+            unset LDFLAGS
+            make ARCH="$_carch" CC="${CC:-gcc}" \\
+            KBUILD_BUILD_VERSION="$((pkgrel + 1 ))-postmarketOS\""""
+
+    if deviceinfo["bootimg_qcdt"] == "true":
+        makedepends += " dtbtool"
+
+        build += """\n
+            # Generate master DTB (deviceinfo_bootimg_qcdt)
+            dtbTool -s 2048 -p "scripts/dtc/" -o "arch/arm/boot/dt.img" "arch/arm/boot/\""""
```
drebrez: @MayeulC @ollieparanoid for `aarch64` devices, the `arch/arm64/boot/` folder should be used, see https://github.com/postmarketOS/pmbootstrap/blob/8373294266c9ad19c17a9965ae03f0226f2b8019/aports/device/linux-motorola-potter/APKBUILD#L62

drebrez: same for the `install` in the package section

PR #1126, opened 2018-01-16
 Initial support for Moto G5 Plus (potter) 

chin123:
What works so far:
* Flash System through `pmbootstrap flasher flash_system`
* Boot Kernel through `pmbootstrap flasher boot`
* USB internet
* ssh  
However, the screen is just blank; it turns on but doesn't display anything.


ollieparanoid:
Also thanks for creating the wiki page and adding it to the overview. It would be super awesome if you could move it up to the booting devices now :+1:

Issue #933, opened 2017-11-23
Enable 32 bit x86 support - postmarket PC to build postmarket OS


dee-gomma:
Hi all,

For who like me is still running an old 32 bit PC and wants to play with postmarketOS, I've created a working fork [here](https://github.com/dee-gomma/pmbootstrap/tree/feature/enable-x86-32bit).
It should work for any x86 architecture (i386 etc., I have i686). Unfortunately it is not yet ready for a pull request (some temporary stuff to be cleaned up, see below), and I have very few spare time lately, so if anybody is interested in completing it that would be great.

Basically, what has been done:
- Generated needed aports for x86 architecture through "aportgen": binutils, busybox-static, gcc, musl
- Modified pmb/parse/arch.py adding x86 architecture mappings (possibly to be reviewed)
- Overwritten aports/cross/quemu-user-static-repack/APKBUILD: original arch x86_64 replaced with x86 and references updated accordingly (quick and dirty temporary solution, we should keep 2 versions for both x86_64 and x86 instead)

I was able to build kernel and pmOS on my i686 PC for a Samsung Galaxy GrandPrime (new potentially supported device, will push it as soon as I clean it up decently)
I could not compile Weston UI unfortunately, looks like a dependency ([libunwind-dev](https://pkgs.alpinelinux.org/packages?name=libunwind-dev)) has no x86 support.

Thanks to @drebrez for the guidance and you all for building up this great project!

ollieparanoid:
Great work! \o/

The problematic part is the `qemu-user-static-repack` package, and my favorite solution would be building that for Alpine instead of using the Debian package - then it would work with all architectures. But this is a really good start if someone is in the same situation with an `x86` PC!

I don't get what you mean with the `libunwind-dev` though, since we are cross-compiling you would only need it for the armhf/aarch64 chroot, where it is available. Could you pastebin the end of the log, where you got stuck?

dee-gomma:
Thank you for the feedback :)

Yes building `qemu` package independently would be the best solution in fact.

About `weston` and `libunwind-dev` dependency, it's a note I've added in my todo list to be investigated. I could not remember exactly the context and what I did to get there, so maybe I am wrong, I truly need to clean-up and try to reproduce it, then I'll ping you back again in case.
Anyway, being that considered, here below few inputs from the current status:
- on device, in `tmp/weston.log`, got the following: 
[...]

PR #1136, opened 2018-01-19
Add qemu-user-static-repack all architectures

drebrez:
This is the temporary update to the `qemu-user-static-repack` package mentioned in #933.
It simply download the debian package for all the architectures and while building it, during the `unpack` function it extracts only the `.deb` of the architecture which is going to build (thanks to $CARCH)

`$CARCH` variable explained here: https://github.com/alpinelinux/abuild/blob/master/abuild.in#L2404
script used to verify it
```
for arch in x86 x86_64 armhf aarch64; do
  pmbootstrap build_init -b $arch
  pmbootstrap chroot -b $arch -- abuild -A
done
```
This is a necessary step to make pmbootstrap be able to cross-compile from architectures other than `x86_64`

I've tested it on `x86` architecture and works fine, the `x86_64` shouldn't have changed but if someone can test that it still work would be great. I didn't managed yet to test it on `armhf` or `aarch64`.


drebrez:
Thanks for the fix, I just forgot to commit it in this PR

PR #1138, opened 2018-01-19
Google Nexus 10: Initial Support

BakaOsaka:
Adds initial support for Google Nexus 10 (samsung-manta), using the LineageOS kernel.
I am able to successfully boot postmarketOS and access the device over SSH, as well as display the Weston DE (albeit with no touch support).

I had to sideload the generated recovery image using ClockworkMod Recovery - If I tried to load using ```pmbootstrap flasher flash-system``` or TWRP, I would get a signature validation failure (footer). By using Clockwork Recovery I was able to disable signature validation and successfully sideload the image using ```pmbootstrap flasher --method adb sideload```.

Initial (low-quality) image:


BakaOsaka:
Whoops - Missed that one! Thanks for pointing it out - I've just updated it to refer to the correct defconfig.

Additionally, I've updated [the wiki](https://wiki.postmarketos.org/wiki/Google_Nexus_10_(samsung-manta)) to create a device page, adding it to the device overview page.

(PS: Apologies for the low quality photos!)

Issue #1124, opened 2018-01-15
"-Werrors" not removed by prepare() in APKBUILD


ashirviskas:
I'm trying to build a kernel with gcc6 for Nextbit Robin and I'm getting tons of Werrors, mostly with messages about bad indentation.

Here's an example below:

`drivers/staging/qcacld-2.0/CORE/HDD/src/wlan_hdd_cfg.c:4768:4: error: this 'for' clause does not guard... [-Werror=misleading-indentation]
    for (ptr = str; i_isspace(*ptr); ptr++);
    ^~~
`

I'm new to compiling kernels, so maybe I'm doing something wrong?

ollieparanoid:
In the `prepare()` of the `linux-` APKBUILD that has been generated should be the following:

```shell
# Remove -Werror from all makefiles
find . -type f -name Makefile -print0 | \
	xargs -0 sed -i 's/-Werror-/-W/g'
find . -type f -name Makefile -print0 | \
	xargs -0 sed -i 's/-Werror//g'
```

This changes `-Werror-*` to `-W*` and removes the `-Werror` flag completely for all `Makefiles`. If you have this, then for some reason it does not work with the kernel sources you have (they probably have specified the `-Werror` flag in another way which we don't catch).

To debug this, you could run the build once and let it fail, then enter the chroot and find the remaining Werror statement:
```shell
pmbootstrap chroot
cd /home/pmos/build/
grep . -r Werror
```

If you can't find it, we can also take a look if you could attach the `APKBUILD` to this issue.

ashirviskas:
Thank you for a quick response. Yes, those lines are present. 

After running those three lines I only get this:

`grep: Werror: No such file or directory`

ollieparanoid:
Sorry, I meant: `grep -r Werror .`
Making patches for the Werrors also kind of works, but usually there are many of them, and it's just much easier to disable all of them. (In my opinion it's not worth putting much energy in the kernel forks beyond of making them compile and work once, because they won't get updates from upstream anyway.)

ashirviskas:
It returns tons of text that is hardly readable for a newbie like me.

Here it is:

[errors.txt](https://github.com/postmarketOS/pmbootstrap/files/1637189/errors.txt)


I attached my archived linux-nextbit-robin folder below, with APKBUILD file included.
[linux-nextbit-robin.zip](https://github.com/postmarketOS/pmbootstrap/files/1637180/linux-nextbit-robin.zip)

[...]

PR #1139, opened 2018-01-19
aportgen linux: Also remove -Werror in Kbuild files

ollieparanoid:
When generating a new kernel fork package (`linux-my-device`) with either `pmbootstrap init` or `pmbootstrap aportgen linux-my-device`, we remove the -Werror flags from all Makefiles. Because usually the kernels were built with older GCC versions, that didn't throw as many errors as the current one does, and fixing all these warnings isn't helpful for a kernel that can never be updated properly anyway (real solution being mainlining as usually).

We discovered that one kernel had it in a Kbuild file in #1124, so this PR patches it for the Kbuild files as well. I've built the kernel from that issue successfully, without the PR it errors with a -Werror, and with the PR it does not throw an error at all.


ollieparanoid:
Thanks @drebrez! \o/

PR #1141, opened 2018-01-20
Odin export: make sure temp folder is empty

### LINKED ISSUE #1072: odin export fails, because it runs "rmdir /tmp/odin-flashable-tar" and the folder is not empty
Reported by tikh_av89 on IRC:
```
(005606) [12:05:37] (rootfs_samsung-lt013g) % rmdir /tmp/odin-flashable-tar 
rmdir: '/tmp/odin-flashable-tar': Directory not empty 
```

ollieparanoid:
Fixes #1072.


ollieparanoid:
That way it works for sure, even if the user created a git repository in there or did other strange stuff. Thanks for testing @drebrez!

PR #1145, opened 2018-01-20
Nextbit Robin booting. SSH over USB working

ashirviskas:
Got SSH over USB working. Nothing else tested.


aports/device/linux-nextbit-robin/APKBUILD
```diff
@@ -0,0 +1,83 @@
+# Kernel config based on: arch/arm64/configs/(CHANGEME!)
```
ollieparanoid: Please fill out the `(CHANGEME!)` here.

ashirviskas: Thanks! I am a bit busy with exams right now, but I'll try to do as much as I can.

ollieparanoid: Take your time, this is a free time project after all :+1: 

If you can't make it in the following weeks, please let us know here as comment which config you have used and we'll update the PR and the wiki.

ashirviskas: It didn't take too long, done!

Updated with config info and wiki too.

ashirviskas:
I got these:
image

image


By doing this:
while [ ! ]; do cat /sys/class/graphics/fb0/modes | sudo tee /sys/class/graphics/fb0/mode; done

ollieparanoid:
Thanks for making the change, and updating the wiki! \o/
I can confirm that compiling the packages works, shipping. :ship: 

> `while [ ! ]; do cat /sys/class/graphics/fb0/modes | sudo tee /sys/class/graphics/fb0/mode; done`
> Is it proper to have this dirty line included in the main repository? If so, what would be the best approach?

Can you try out the `msm-fb-refresher`? (enable it in your [`deviceinfo`](https://wiki.postmarketos.org/wiki/Deviceinfo_reference) and add the `msm-fb-refersher` to the depends of your device package, look at other packages where it is used if necessary (`grep -r msm-fb-refresher pmbootstrap/aports/`)).

The idea is, that the `msm-fb-refresher` is the proper workaround for the display refreshing (until we have the device mainlined and don't need such a hack at all). And if the `msm-fb-refresher` can't handle it, we could extend it (it's just a short C program).

ashirviskas:
@ollieparanoid Thank's! I was playing with it again and now it works! Kinda..

I still need to do "while [ ! ]; do msm-fb-refresher; done"

Running it once just refreshes a single frame.

ata2001:
@ashirviskas You're gonna need the `--loop` argument of msm-fb-refresher.

ollieparanoid:
So the `msm-fb-refersher` (enable it as described above) does not work?

ashirviskas:
Adding `--loop` argument works.

ashirviskas:
But why then is it not ran automatically on boot? Do I add a some manual script or do I need to add some other build parameters? I have `msm-fb-refresher` in depends and `deviceinfo_msm_refresher="true"`, but this doesn't make `msm-fb-refresher --loop` run on every boot, even though it should.  ([1](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/main/postmarketos-mkinitfs/init.sh.in#L19), [2](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/main/postmarketos-mkinitfs/init_functions.sh#L290))

Am I missing something?

drebrez:
@ashirviskas with the `deviceinfo_msm_refresher="true"` variable it should start automatically in the initramfs and it will be killed right before leaving it (chroot to the system partition). When the system loads, it should start again as a service (https://github.com/postmarketOS/pmbootstrap/blob/master/aports/main/msm-fb-refresher/msm-fb-refresher.init). Could you check the status of the service? (using `rc-status`)

ashirviskas:
It's not running. 

P.s. The display is refreshed once when it starts booting (Without it I'd still see bootloader screen when device boots and with it bootloader screen is replaced with a dark/empty screen when device boots/boots up)

ollieparanoid:
Thanks for checking! After you have enabled `msm-fb-refresher`, have you re-installed your system image? If you did not, you need to run the following line manually to enable the service:

https://github.com/postmarketOS/pmbootstrap/blob/b0869380ff1c03af3fb23401fc3376ce608825e0/aports/main/postmarketos-base/postmarketos-base.post-install#L23

ashirviskas:
My mistake, after making a clean clone it's now working perfectly!

[PR](https://github.com/postmarketOS/pmbootstrap/pull/1197)

PR #1146, opened 2018-01-20
Rename Nokia rx51->n900 and rm-696->n9

pavelmachek:
Rename packages, no other changes.


ollieparanoid:
(Heads up to @craftyguy: You need to reinstall your N900 installation.)

PR #1147, opened 2018-01-21
New device: Samsung Galaxy S7

Polynomdivision:
I wanted to give it a try and did my best (probably not good enough, though) at trying to create a device and a kernel package for the G930F (Galaxy S7) from Samsung.

While I got the kernel to compile with a rather hacky workaround, I did not test flashing or anything else besides compiling, as I do not want to risk breaking my phone.


aports/device/linux-samsung-g930f/APKBUILD
```diff
@@ -0,0 +1,81 @@
+# Kernel config based on: arch/arm64/configs/(CHANGEME!)
```
ollieparanoid: Please adjust this.

aports/device/linux-samsung-g930f/APKBUILD
```diff
@@ -0,0 +1,81 @@
...
+
+# Source
+_repository="android_kernel_samsung_universal8890"
+_commit="adb2bc5b68911a31b6b7da09b17a89faad577860"
+_config="config-${_flavor}.${arch}"
+source="
+	$pkgname-$_commit.tar.gz::https://github.com/LineageOS/${_repository}/archive/${_commit}.tar.gz
+	$_config
+	compiler-gcc6.h
+    00_commented_define.patch
```
ollieparanoid: Spaces used instead of tabs (we use spaces in Python scripts, but tabs in shell scripts and APKBUILDs like Alpine does it).

Polynomdivision:
I now replaced the CHANGEME with the appropriate path to the config and fixed the indentation of the one line in the APKBUILD.

@ollieparanoid Somehow I must've missed that one CHANGEME 😄
I really do hope that someone will continue this port or that I will get brave enough to risk my S7.

rrooij:
Thanks a lot. I have a S7 too but don't want to risk bricking it.

drebrez:
@Polynomdivision @rrooij alternatively you can flash the kernel in the recovery partition and boot it with the key combination, and the pmOS system on a SDcard. That shouldn't break the normal android os.

ollieparanoid:
@Polynomdivision: awesome work \o/

PR #1150, opened 2018-01-22
Allow to build musl/binutils/gcc for x86 architecture

drebrez:
This is now necessary for supporting the x86 architecture, due the recent changes in 9f674675eb582d063dcc8d72e9792c69319b2fec
Full pmbootstrap support for x86 architecture is almost done, only need to finalize the various functions in `pmb/parse/arch.py`


ollieparanoid:
I've added `x86` to the `pmb.config.build_device_architectures`, so it produces the same aports as we have with this PR, when you run `pmbootstrap aportgen gcc-armhf` etc.

Merging this as soon as Travis runs through, thanks for the contribution @drebrez!

drebrez:
@ollieparanoid are you sure that it's correct to add the `x86` architecture to the `pmb.config.build_device_architectures`? 
I'm checking all the places where `pmb.config.build_device_architectures` and there might be some issues.
For example `pmbootstrap chroot -b x86` opens a chroot where `uname -a` says `x86_64`, is it correct?

ollieparanoid:
The x86 chroot does not use QEMU if your host arch is `x86_64`, so it makes sense that `uname -a` says you are on `x86_64`. Please let me know if you find more places where this is questionable. From where I looked it did not seem to have any negative impact. But I'll wait for feedback next time I do such a change on top of your commits, so we can both have a look at it.

Thanks for noticing!

drebrez:
I theory shouldn't be a problem to have `x86` in `pmb.config.build_device_architectures`, that would be necessary when we will have an `x86` device.
For running `pmbootstrap` on an `x86` host is not necessary to have it, but the only solution I can think right now, would be to have 2 separated variables (host and device) and doesn't really make sense.

PR #1151, opened 2018-01-22
Implement installation using rsync (not cp). Should reduce wear of sd card.

pavelmachek:
This is _not_ ready to be merged, but it should show that approach is feasible. Speedup is huge on slow SD card (5 minutes for sd card copy -> 15 seconds).


ollieparanoid:
I think the fastest way to get this in would be a new `--rsync` argument, used like this:
```
pmbootstrap install --sdcard=/dev/mmcblk0 --no-fde --rsync
```
Having `--no-fde` would be mandatory for now, otherwise we would need to introduce another code path that handles encrypted devices. We could add that in the future if people really need it.
`--no-fde` would be mandatory with `--rsync` as well, as rsyncing to an image file does not make much sense.

Recommended TODOs:
- [ ] Add `--rsync` in `pmb/parse/arguments.py`
- [ ] Check for allowed combinations (`--rsync` only with `--no-fde` and `--sdcard`!) in `pmb/helpers/frontend.py`
- [ ] Use `args.sync` to trigger the `rsync` code.

Thank you very much for this fantastic proof of concept @pavelmachek, that really speeds up flashing a lot (though my sd card reader is not that slow, but still!).

pavelmachek:
I added option and checking that configuration is sane, but could not find suitable place in frontend.py, so I'm doing it in arguments.py.

pavelmachek:
Note: So far code is untested, and travis fails for reason I do not quite understand. Help welcome :-).

pavelmachek:
Ok, here's tested version, and I fixed trivial thing pointed by travis. Unfortunately, travis still complains, but I don't see _why_ it is complaining. Its logs grew too long and unreadable :-(.

drebrez:
@pavelmachek the error is this, try running `test/static_code_analysis.sh` locally before commit
```
./pmb/install/install.py:90:5: E115 expected an indented block (comment)
./pmb/install/install.py:96:1: W293 blank line contains whitespace
./pmb/install/install.py:99:1: E303 too many blank lines (3)
```

pavelmachek:
test/static_code_analysis does not work for me:

pavel@duo:/data/l/pmos/pmbootstrap$ test/static_code_analysis.sh
Test with shellcheck: ./test/static_code_analysis.sh
test/static_code_analysis.sh: 47: test/static_code_analysis.sh: shellcheck: not found
pavel@duo:/data/l/pmos/pmbootstrap$ sudo aptitude install shellcheck
...
Preparing to unpack .../shellcheck_0.3.4-3_i386.deb ...
Unpacking shellcheck (0.3.4-3) ...
Processing triggers for man-db (2.7.0.2-5) ...
Setting up shellcheck (0.3.4-3) ...
                                         
pavel@duo:/data/l/pmos/pmbootstrap$ test/static_code_analysis.sh
Test with shellcheck: ./test/static_code_analysis.sh
unrecognized option `-x'

Usage: shellcheck [OPTIONS...] FILES...
  -f FORMAT         --format=FORMAT          output format
  -e CODE1,CODE2..  --exclude=CODE1,CODE2..  exclude types of warnings
  -s SHELLNAME      --shell=SHELLNAME        Specify dialect (bash,sh,ksh,zsh)
  -V                --version                Print version information

pavel@duo:/data/l/pmos/pmbootstrap$

pavelmachek:
(And thanks, @drebrez !)

ollieparanoid:
Seems like your distribution ships an older version of shellcheck. You could also use the online version (although that's not so convenient): https://shellcheck.net/ ... or install with Haskell's package manager cabal (I think it works just like pip, I have not tested it).

[One idea](https://github.com/postmarketOS/pmbootstrap/issues/110#issuecomment-311197756) would be to package shellcheck for Alpine and use that version, so we don't depend on the host system.

pmb/config/__init__.py
```diff
@@ -267,7 +267,7 @@
 # Packages, that will be installed inside the native chroot to perform
 # the installation to the device.
 # util-linux: losetup, fallocate
-install_native_packages = ["cryptsetup", "util-linux", "e2fsprogs", "parted"]
+install_native_packages = ["cryptsetup", "util-linux", "e2fsprogs", "parted", "rsync"]
```
ollieparanoid: We can install that on demand (only when `--rsync` is specified) below, shortly before using it.

pmb/install/install.py
```diff
@@ -86,9 +86,13 @@ def copy_files_from_chroot(args):
             continue
         folders += [os.path.basename(path)]
 
-    # Run the copy command
-    pmb.chroot.root(args, ["cp", "-a"] + folders + ["/mnt/install/"],
-                    working_dir=mountpoint)
+    if not args.rsync:
```
ollieparanoid: How about swapping the if/else parts, so we can write `if args.rsync ... else ...`  instead of `if not args.rsync ... else ...`?

pmb/install/install.py
```diff
@@ -86,9 +86,13 @@ def copy_files_from_chroot(args):
...
+        # Run the copy command
+        pmb.chroot.root(args, ["cp", "-a"] + folders + ["/mnt/install/"],
+                        working_dir=mountpoint)
+    else:
+        pmb.chroot.root(args, ["rsync", "-avP", "--delete"] + folders + ["/mnt/install/"],
```
ollieparanoid: I would install `rsync` right here before using it with `pmb.chroot.apk.install(args, ["rsync"])`.

pmb/install/install.py
```diff
@@ -196,8 +202,11 @@ def install_system_image(args):
     logging.info("*** (3/5) PREPARE INSTALL BLOCKDEVICE ***")
     pmb.chroot.shutdown(args, True)
     (size_image, size_boot) = get_subpartitions_size(args)
-    pmb.install.blockdevice.create(args, size_image)
-    pmb.install.partition(args, size_boot)
+    if not args.rsync:
```
ollieparanoid: I would swap these as well (see above).

pmb/install/partition.py
```diff
@@ -63,24 +63,25 @@ def partition(args, size_boot):
 
     size_boot: size of the boot partition in bytes.
     """
-    # Convert to MB and print info
-    mb_boot = str(round(size_boot / 1024 / 1024)) + "M"
-    logging.info("(native) partition /dev/install (boot: " + mb_boot +
-                 ", root: the rest)")
+    if not args.rsync:
```
ollieparanoid: Having the `partition()` function *not* doing any partitioning, but only using it for the mounting seems to be counter intuitive. How about the following:
* Remove `partitions_mount(args)` from the end of this `partition()` function
* Revert the `if not args.rsync` change here
* Change the following line:
https://github.com/postmarketOS/pmbootstrap/blob/3a01537332e19a30d402010306f167c92b8e9a87/pmb/install/install.py#L200
To:
```python
if not args.rsync:
    pmb.install.partition(args, size_boot)
pmb.install.partitions_mount(args)
```
* In `pmb/install/__init.py`, add: `from pmb.install.partition import partitions_mount`

pmb/parse/arguments.py
```diff
@@ -396,4 +398,11 @@ def arguments():
                              " (otherwise you will have issues with noarch packages)"
                              " and try again.")
 
+    if args.action == "install":
```
ollieparanoid: Since we already have one huge function here for the arguments parsing (I'm slowly splitting it up for new actions at least, see the top of the file), I recommend to put this check in `pmb/helpers/frontend.py` here:
https://github.com/postmarketOS/pmbootstrap/blob/3a01537332e19a30d402010306f167c92b8e9a87/pmb/helpers/frontend.py#L201-L202

pavelmachek:
I fixed the review comments as proposed and did some light testing. Have to run now.

pmb/install/install.py
```diff
@@ -102,8 +107,9 @@ def copy_files_other(args):
 
     # Create /home/{user}
     homedir = rootfs + "/home/" + args.user
-    pmb.helpers.run.root(args, ["mkdir", rootfs + "/home"])
-    pmb.helpers.run.root(args, ["cp", "-a", rootfs + "/etc/skel", homedir])
+    if not args.rsync:
+        pmb.helpers.run.root(args, ["mkdir", rootfs + "/home"])
+    pmb.helpers.run.root(args, ["rsync", "-zavP", "--delete", rootfs + "/etc/skel", homedir])
```
ollieparanoid: As I understand this, you're using `rsync` *instead* of the `cp` here, for both the `--rsync` case and for the normal installation.

For better understanding:
* Why did you do this, to save lines of code and have less code paths? (good reasons imho)
* Or are there more advantages to using `rsync` here?

Two problems I see here:
* We only install `rsync` when `args.rsync` is specified
* It might be confusing that we use the ´rsync` command even if `--rsync` is not specified (maybe rename the parameter then?)

Just like you I want to get this PR in as quickly as possible. But to do that I really need to understand it.

pavelmachek:
Yeah, I wanted to have less codepaths, and originally installed rsync unconditionally (remember?), so this was ok. Then you asked for only installing rsync on demand, and I did that, which would break it. Corrected now, I switched to the obvious code.

pavelmachek:
I pushed updated version.

ollieparanoid:
> Yeah, I wanted to have less codepaths, and originally installed rsync unconditionally (remember?)

Yes, I remember :wink: The problem was, that I did not understand at this point that you meant to use `rsync` even for the normal copying. Anyway, I think this is a good idea, but at the same time this PR is ready to ship in the form it has now if you agree.

I have tested the following on my `samsung-i9100`:
* Installed Weston normally (without this PR), confirmed it boots
* Switch to xfce4 and run the `--rsync` installation. Confirm that it boots again.
* Install Weston again without `--rsync`, confirm that it boots.
* Install Weston again, without `--rsync` and without `--no-fde`, confirm that it boots.
* Check that `--rsync` properly aborts when `--no-fde` or `--sdcard` are missing

With the changes applied in the last commit, everything is working.

My proposal:
* Please review my changes
* When it's all good, we merge it so we have a real benefit right now
* Then we make a new issue with possible enchantments that could be done (use rsync even for normal copying, don't require the `--rsync` always run the rsync version when a previous pmOS installation was found, support FDE, ...)

What do you think?

pavelmachek:
Your changes look good to me. Thanks for fixes (and for testing). I believe merge is good next step.

pavelmachek:
Thanks! I have realized that there's no "are you sure" question in rsync case. It looks like a feature to me, but I guess I should mention it.

ollieparanoid:
Oh, I've missed that, thanks! Noted it down in the new improvements issue. I think it is fine for now, as `--rsync` is used by default yet.

PR #1153, opened 2018-01-23
Update deviceinfo for desirez

ajtag:
when running pmbootstrap init, complaint that deviceinfo_flash_methods needs changing to deviceinfo_flash_method


PureTryOut:
@ajtag could you update the checksum and increase the `$pkgver` like @ollieparanoid asked? It's been quite a while since any activity in this PR.

PR #1154, opened 2018-01-23
Enable native/distcc for packages ending with "-repack"

drebrez:


ollieparanoid:
Changed the title in the commit, because this is more about deleting two obsolete and hackish lines of code, rather than enabling cross-compilers (because repack packages don't compile anything).
Thanks!

PR #1155, opened 2018-01-23
Enable running pmbootstrap on x86 architecture

drebrez:
The mapping in `alpine_native` has been tested with 2 different VMs running AlpineLinux x86 and Ubuntu 32bit. 
Both the operating system were returning `i686`, we can also consider to already add other 32bit architectures there, like `i386` and `x86`, @ollieparanoid ?

The mapping in `alpine_to_debian`, which is used only for the `qemu-user-static-repack` and the `binfmt` has been validated with the content of the `qemu-user-static-repack` package ([qemu-user-binfmt.txt](https://bpaste.net/show/2238c6d753fe)). And the binaries `qemu-$arch-static` matches the names in the txt file.

Close: #933


drebrez:
@ollieparanoid I agree with you to add mappings only when someone needs them. 🚢

pmb/parse/arch.py
```diff
@@ -55,7 +56,8 @@ def alpine_to_debian(arch):
     """
 
     mapping = {
-        "x86_64": "amd64",
+        "x86": "i386",
+        "x86_64": "x86_64",
```
ollieparanoid: Actually... I've missed that earlier. Is the change from `x86_64` to `amd64` on purpose?

drebrez: @ollieparanoid yes, I've also explained in the initial message. In `qemu-user-static-repack` and his [qemu-user-binfmt.txt](https://bpaste.net/show/2238c6d753fe) file, the arch is `x86_64` and not `amd64`.

ollieparanoid: Now I get it, thanks for explaining!

PR #1157, opened 2018-01-23
Update Maemo packages (fixes #1148)

### LINKED ISSUE #1148: Hildon: Bad patch for missing titlebar on x86_64
https://github.com/postmarketOS/pmbootstrap/blob/0c3ce04a6209fd454b9c28798cea252d5ade1b55/aports/maemo/hildon-desktop/APKBUILD#L26 is not a correct fix, see https://github.com/maemo-leste/libmatchbox2/commit/50d64e6f204b00a7f13386febb62701511032d99, which is the reason for missing title bar.

Also, we'd appreciate if you open an issue on leste bugtracker next time you hit a bug in some of its packages.

NotKit:
hildon-desktop is now compiled against gnome-menus-3. For some reason it's missing from Alpine, so I renamed and updated old gnome-menus2 package. It wasn't used by anything else in aports.

I didn't pick up the latest hildon-desktop 2.2.157.6 version yet, as I suppose maemo-leste/hildon-desktop@220b3e6f is going to cause breakage with current cogl package, which is compiled against desktop OpenGL (GLX) without EGL Xlib platform enabled.


NotKit:
Hm, that's good then. But cogl misses --enable-xlib-egl-platform, which is needed for GLESv2 (and, say, to get it working with libhybris or proprietary drivers), so it probably still manages to use libGL somehow.

PR #1160, opened 2018-01-23
version.py: Fix validation of versions like "6.0_0002" (it's invalid!)

### LINKED ISSUE #1144: pmbootstrap did not detect, that the `libsamsung-ipc` pkgver was invalid
#854 implemented validating `pkgver` after parsing `APKBUILD` files. But for this version...
```shell-session
$ grep '^pkgver=' aports/main/libsamsung-ipc/APKBUILD
pkgver=6.0_0002
```
...`pmbootstrap` does *not* complain, but `apk` does (when you try to build in `--strict` mode).

I will fix the pkgver for that package, but this issue needs to be fixed in `pmbootstrap` as well.

ollieparanoid:
`pmb.parse.version.validate()` said `6.0_0002` was valid, but when asking `apk` it said it was invalid (because `0002` is not a valid version suffix).

This PR fixes it, and adds test cases for this specific case. We already have a great deal of test cases (all from `apk`), but these only check if comparing one version with another one works correctly.

@ncopa, @kaniini: From what I know, you're using `version.py` as well. So this is a heads up. Also since you guys know the `apk` code base: If you have time, maybe one of you could review this fix.

(Fixes #1144.)

### DIFF
--- a/pmb/parse/version.py
+++ b/pmb/parse/version.py
@@ -113,9 +113,12 @@ def parse_suffix(rest):
     :param rest: what is left of the version string, that we are
                  currently parsing, starts with a "suffix" value
                  (see below for valid suffixes).
-    :returns: (rest, value) rest is the input "rest" string without the
-              suffix, value is a signed integer (negative for pre-,
-              positive for post-suffixes).
+    :returns: (rest, value, invalid_suffix)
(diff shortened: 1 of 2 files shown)


kaniini:
:+1:

ollieparanoid:
Thanks @kaniini!

PR #1161, opened 2018-01-23
static_code_analysis.sh: Find CHANGEMEs in APKBUILDs

ollieparanoid:
When the device wizard generates a `linux-*` `APKBUILD`, it puts the following line at the beginning:
> `# Kernel config based on: arch/arm/configs/(CHANGEME!)`

With this PR, Travis will check whether there are any `aports/devices/*/*` files, where the `(CHANGEME!)` has not been replaced.

### How to test
* run `test/static_code_analsysis.sh`, it should not fail (you can abort it after the APKBUILD check, when it starts checking shell scripts)
* create a new device in `pmbootstrap init`
* run the script again, it should fail and print where you need to replace something.
* clean up: delete the aports/device/*-your-new-device folders again (see `git status`)


test/static_code_analysis.sh
```diff
@@ -19,6 +19,14 @@
 set -e
 DIR="$(cd "$(dirname "$0")" && pwd -P)"
 
+# Find CHANGEMEs in APKBUILDs
+if grep -qr '(CHANGEME!)' "$DIR/../aports/device"; then
+	echo "ERROR: Please replace '(CHANGEME!)' in the following files:"
+	cd "$DIR/.."
+	grep --color=always -r '(CHANGEME!)' aports/device
```
z3ntu: The `cd "$DIR/.."` line is to make the output sane, right?

drebrez: @ollieparanoid in busybox grep version there's no `--color=always` option, but it works installing the `grep` package in Alpine

drebrez: with `GREP_OPTIONS='--color=always' grep -r '(CHANGEME!)' aports/device` it works for both, but with the gnu grep it shows an annoying warning: `grep: warning: GREP_OPTIONS is deprecated; please use an alias or script`

drebrez: @ollieparanoid why not moving the `cd "$DIR"/..` before the `if`, and remove it also from below (shellcheck part)?

z3ntu: Yes makes sense

ollieparanoid: Since `shellcheck` is not even packaged for Alpine yet ([feature request](https://bugs.alpinelinux.org/issues/7105)), the `static_code_analysis.sh` script as a whole does not run on Alpine. So I'd keep the GNU `grep` syntax here if that is acceptable for everyone else.

drebrez: @ollieparanoid yes I also noticed that shellsheck is missing. So for me it's fine to keep it like this (I'm not using Alpine, only for some tests sometime), and let's wait that someone actually request it (And  in Alpine gets solved just by `apk add grep`).

ollieparanoid:
@z3ntu, @drebrez: thanks for reviewing! \o/

Issue #1122, opened 2018-01-14
pmbootstrap picks wrong package when two packages provide the same library


ollieparanoid:
@NotKit reported in #postmarketOS:
> pmbootstrap dependency resolver is a bit weird
> for some reason it tries to install libhybris-gles in buildroot, which conflicts with mesa-gles, required by mesa-dev, instead of picking mesa-gles
> https://pastebin.com/MvT6szqa - I suppose both mesa-egl and libhybris-egl provide libEGL.so.1, but I wonder why it goes for the first one even when mesa-gles is manually installed in buildroot 

Follow up questions:
* Is [this](https://github.com/postmarketOS/pmbootstrap/blob/feature/hybris/aports/hybris/libhybris/APKBUILD) `libhybris` APKBUILD the one you have used? If not, please upload it somewhere.
* Could you provide the same log, but with `pmbootstrap -v`, so it prints more information while resolving dependencies?

Thanks for reporting this!

NotKit:
Yes, APKBUILD is the same with slight changes for aarch64. The verbose log helped to understand what's going on:

(020035) [15:47:29] so:libEGL.so.1: 1263.013cf7e-r1 found in /mnt/HDD/pmOS/work/packages/aarch64/APKINDEX.tar.gz
(020035) [15:47:29] so:libEGL.so.1: 17.2.4-r1 found in /mnt/HDD/pmOS/work/cache_apk_aarch64/APKINDEX.066df28d.tar.gz (but 1263.013cf7e-r1 is bigger)
(020035) [15:47:29] so:libEGL.so.1: provided by 'libhybris-egl'

[pmbootstrap.log](https://github.com/postmarketOS/pmbootstrap/files/1629331/pmbootstrap.log)

PR #1162, opened 2018-01-23
Remove pmb.parse.apkindex.parse(): strict param / "challenge" action

ollieparanoid:
Preparation for #1122.

* `pmb.parse.apkindex.parse()`, removed strict parameter: This used to raise an exception when two entries in the apkindex provided the same package. Turns out this is *not* invalid after all, two packages can provide the same soname for example (e.g. libhybris, mesa-egl). In an APKINDEX, sonames are listed as they were packages ("so:libjpeg.so.8" etc.).
* Remove `pmbootstrap challenge` leftover code from reproducible builds effort, which was a dead end. This code uses the broken strict feature.


ollieparanoid:
Yeah, there is not much to test in this one. Thanks for reviewing! :tada:

PR #1164, opened 2018-01-24
Update KDE Frameworks to 5.42.0

PureTryOut:
Removed a now unnecessary patch to kdelibs4support and added a patch required to get kwin building with the newer framework version. All `$pkgrel` are updated for packages with deps on the framework.


aports/kde/breeze/APKBUILD
```diff
@@ -2,7 +2,7 @@
 # Maintainer: Bart Ribbers <bribbers@disroot.org>
 pkgname=breeze
 pkgver=5.11.5
-pkgrel=0
+pkgrel=1
```
z3ntu: Does this really need to get rebuilt?

ollieparanoid: @bhush9, if I quoted you right [here](https://github.com/postmarketOS/pmbootstrap/issues/978#issue-279156796) it is only necessary for the following:
>* kwin
>* plasma-integration
>* kwayland

So it is not needed, right?

z3ntu: Arch Linux also doesn't rebuild e.g. breeze on Framework releases.

PureTryOut: I didn't realize only a few packages needed rebuilding, I just updated the `$pkgrel` of every package having a dependency on the framework. I downgraded the `$pkgrel` again, and I'll write that short list down for next time 😉

PR #1165, opened 2018-01-24
Add initial support for the Moto G 2013 (falcon)

rymate1234:
This pull requests adds support for the 2013 Moto G, utilising the LineageOS kernel

The device has a lot of similariites with the 2014 Moto G, so that was used as a reference for any broken things

Things that work include successfully booting into a weston session with touch, running a bunch of the demos, and can be ssh'd into when connected via USB.

d6bekvz

Further details on the wiki page I've created: https://wiki.postmarketos.org/wiki/Moto_G_2013_(falcon)


aports/device/linux-motorola-falcon/APKBUILD
```diff
@@ -0,0 +1,93 @@
+# Kernel config based on: arch/arm/configs/(CHANGEME!)
```
ollieparanoid: Please adjust this (there will be an automated test for this in the future, the code for that has just been merged :+1:).

aports/device/linux-motorola-falcon/APKBUILD
```diff
@@ -0,0 +1,93 @@
...
+		msg "zImage found: $_zimg"
+		install -Dm644 "$_zimg" "$_target"
+		break
+	done
+	if ! [ -e "$_target" ]; then
+		error "Could not find zImage in $PWD!"
+		return 1
+	fi
+
+	## we need a dtb file
```
ollieparanoid: How about `# We need a dtb file` instead? (one `#` and starting the comment with an upper letter for consistency)

rymate1234:
Alright, added those two changes

ollieparanoid:
Confirming that it builds, shipping :ship:

Issue #970, opened 2017-12-02
Streamline repo scripts


ollieparanoid:
The repo is somewhat usable now and it's a big improvement over having no binary packages at all. But it still needs improvements.

- [ ] Change `pmbootstrap aportgen` to generate `gcc-aarch64` not for `arch=all` but only for specific ones (`x86_64` only for now?), otherwise it will build versions for `aarch64` and `armhf`
- [ ] use `--strict` for building
- [ ] improve the building script: don't use `pmbootstrap` directly, wrap it with a different config and work path, so the same PC can be used for quickly testing a build without messing with the repo
- [ ] polish repo build scripts and publish them
- [ ] properly delete old packages (right now we keep them all!)
- [ ] use proper repo description with pmbootstrap commit hash
- [ ] use -9 compression for gzip packages
- [ ] #797 Firmware packages: download blobs after installation, so we can build firmware packages, too (right now they don't get built)
- [ ] add `pmbootstrap build --all` to simply build all packages?

and finally (future!):
- [ ] automatize it completely: whenever a new commit comes in to master, run the build

I would tackle this some time after the 200d blog post, if that is okay @MartijnBraam (you host the mirrors - do we have enough space?).

PR #1166, opened 2018-01-24
Fix pmbootstrap zap -m / various zap improvements

ollieparanoid:
zap -m:
* APKINDEX parsing: parse the "origin" field as well, so we know
  where a subpackage comes from
* pmbootstrap zap -m: properly delete all packages, that do not
  have an aport or where the aport has another version. This also
  works with subpackages now,
  we use the origin field to resolve it.
* Only reindex when packages have been deleted in "zap -m"

zap in general:
* Show the amount of cleared up space after the deletion instead
  of "Done"
* Print "Shutdown complete" to "pmbootstrap log" instead of stdout
  (we need to call it twice during zap now to get the space
  calculation right)
* Add `--dry` argument to `pmbootstrap zap` (this was very useful
  for debugging) to list the packages/chroots that would get
  deleted
* Roughly output the command that would get executed to delete
  files, so it's obvious what's going on in --dry mode. (% rm ...)

This PR relates to streamlining the binary repository (#970),
because it makes it possible to delete outdated packages properly.

### DIFF
--- a/pmb/chroot/zap.py
+++ b/pmb/chroot/zap.py
@@ -45,9 +45,10 @@ def zap(args, confirm=True, dry=False, packages=False, http=False,
     and args.device set!
     """
     # Get current work folder size
-    pmb.chroot.shutdown(args)
-    logging.debug("Calculate work folder size")
-    size_old = pmb.helpers.other.folder_size(args, args.work)
+    if not dry:
+        pmb.chroot.shutdown(args)
+        logging.debug("Calculate work folder size")
+        size_old = pmb.helpers.other.folder_size(args, args.work)
 
     # Delete packages with a different version compared to aports, then re-index
     if mismatch_bins:
--- a/pmb/chroot/zap.py
+++ b/pmb/chroot/zap.py
@@ -86,9 +86,12 @@ def zap(args, confirm=True, dry=False, packages=False, http=False,
     args.cache["apk_repository_list_updated"].clear()
 
     # Override "done" message with saved MB
-    size_new = pmb.helpers.other.folder_size(args, args.work)
-    mb = (size_old - size_new) / 1024 / 1024
-    logging.info("Cleared up ~" + str(math.ceil(mb)) + " MB of space")
+    if dry:
+        logging.info("Dry run: nothing has been deleted")
+    else:
+        size_new = pmb.helpers.other.folder_size(args, args.work)
+        mb = (size_old - size_new) / 1024 / 1024
+        logging.info("Cleared up ~" + str(math.ceil(mb)) + " MB of space")
     sys.exit(0)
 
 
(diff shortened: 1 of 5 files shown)


pmb/chroot/zap.py
```diff
@@ -71,52 +78,79 @@ def zap(args, confirm=True, packages=False, http=False, mismatch_bins=False,
...
+                if not dry:
+                    pmb.helpers.run.root(args, ["rm", "-rf", match])
 
     # Chroots were zapped, so no repo lists exist anymore
     args.cache["apk_repository_list_updated"].clear()
 
+    # Override "done" message with saved MB
+    size_new = pmb.helpers.other.folder_size(args, args.work)
+    mb = (size_old - size_new) / 1024 / 1024
+    logging.info("Cleared up ~" + str(math.ceil(mb)) + " MB of space")
```
drebrez: @ollieparanoid when running with `--dry` this info will always be `0`, better would be to estimate how much will be cleared but probably it's quite hard to know.
What about not showing the line and log something that say `Dry run: nothing has been deleted.` or a better phrase?

pmb/chroot/zap.py
```diff
@@ -71,52 +78,79 @@ def zap(args, confirm=True, packages=False, http=False, mismatch_bins=False,
...
+        pmb.build.other.index_repo(args)
 
 
-def zap_old_bins(args, confirm=True):
+def zap_old_bins(args, confirm=True, dry=False):
     if confirm and not pmb.helpers.cli.confirm(args, "Remove outdated binary packages?"):
         return
     arch_native = pmb.parse.arch.alpine_native()
     if os.path.exists(args.work + "/cache_apk_" + arch_native):
         pmb.chroot.root(args, ["apk", "-v", "cache", "clean"])
```
drebrez: @ollieparanoid here you should add this:
```
logging.info("(buildroot_" + arch_native + ") apk -v cache clean")
if not dry:
    ....
```

ollieparanoid:
Great review, thanks! Updated it.

pmb/chroot/zap.py
```diff
@@ -16,39 +16,46 @@
...
+                     mirrors (e.g. from Alpine)
+    :param distfiles: Clear the downloaded files cache
 
     NOTE: This function gets called in pmb/config/init.py, with only args.work
     and args.device set!
     """
+    # Get current work folder size
+    pmb.chroot.shutdown(args)
+    logging.debug("Calculate work folder size")
+    size_old = pmb.helpers.other.folder_size(args, args.work)
```
drebrez: @ollieparanoid now that you also included the size computation at the end to the `not dry` case, also here is not necessary anymore

pmb/chroot/zap.py
```diff
@@ -71,52 +78,88 @@ def zap(args, confirm=True, packages=False, http=False, mismatch_bins=False,
...
     args.cache["apk_repository_list_updated"].clear()
 
+    # Override "done" message with saved MB
+    if dry:
+        logging.info("Dry run: nothing has been deleted")
+    else:
+        size_new = pmb.helpers.other.folder_size(args, args.work)
+        mb = (size_old - size_new) / 1024 / 1024
+        logging.info("Cleared up ~" + str(math.ceil(mb)) + " MB of space")
+    sys.exit(0)
```
drebrez: @ollieparanoid what's the purpose of this `sys.exit(0)`?

ollieparanoid: That way it does not print out "Done" at the end.

pmb/chroot/zap.py
ollieparanoid: True, I'll update it.

ollieparanoid:
It only calculates the initial size when running without dry now, as suggested. Do you think this is ready?

pmb/chroot/zap.py
ollieparanoid: I just noticed: we run this code during `pmbootstrap init` as well, when there are chroots. This means, it exits the `pmbootstrap init` code early!
I'll rewrite it to run the `sys.exit()` in `pmb/helpers/frontend.py` instead, so it only gets executed for `pmbootstrap zap`.

ollieparanoid:
Thanks @drebrez :tada:

PR #1168, opened 2018-01-25
Enable xwayland on t2m-flame

Azelphur:
postmarketos-demos runs fine with xwayland enabled. Ollieparanoid requested I make a PR for this change.

### DIFF
--- a/aports/device/device-t2m-flame/weston.ini
+++ b/aports/device/device-t2m-flame/weston.ini
@@ -1,9 +1,0 @@
-# XWayland seems to be broken for this device.
-# NOTE: This information might be outdated! simply delete /etc/xdg/weston/weston.ini
-# after installation to check if it is still broken, and please report if it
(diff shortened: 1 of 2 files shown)


aports/device/device-t2m-flame/weston.ini
```diff
@@ -5,5 +5,5 @@
 # config file).
 
 [core]
-xwayland=false
+xwayland=true
```
drebrez: instead of changing the variable to `true`, which is the default, please delete the file and remove it from the APKBUILD.
(ps: read the note few lines above in this file 😉 )

drebrez: @Azelphur and please also increate the APKBUILD version and run `pmbootstrap checksum device-t2m-flame` to update the hash of the files

ollieparanoid:
Thanks @Azelphur and @drebrez!

PR #1170, opened 2018-01-25
Update requirements for Alpine Linux distribution

drebrez:
As discovered few days ago, during `pmbootstrap install` we compute the folder size which uses GNU grep, on Alpine Linux it's not installed by default.


drebrez:
@ollieparanoid it's used during `pmbootstrap install` also, in `get_subpartitions_size` here https://github.com/postmarketOS/pmbootstrap/blob/12340fe5f60ced7d8917ad8fc8d24100f511a37f/pmb/install/install.py#L54-L56

ollieparanoid:
Sounds good to me, documented them [here](https://wiki.postmarketos.org/wiki/Testsuite).

PR #1174, opened 2018-01-27
Remove timestamp based rebuilds

### LINKED ISSUE #985: Binary repo hinders / confuses
I changed something my deviceinfo in aports/device/device-fairphone-fp2/
After that I rebuilt all my local packages (after pmb zap -p) because the changes didn't apply -> didn't help
After `sudo find ~/.local/var/pmbootstrap/ -name "device-fairphone*" -delete` apparently pmbootstrap noticed, that I have a changed version locally and didn't use the version from the repo anymore.
I ran zap multiple times during this process and also rebuilt device-fairphone-fp2 multiple times (`./pmbootstrap.py build device-fairphone-fp2 --arch=armhf --force`)

So apparently the downloaded, cached packages were prioritized over my changed local ones (yes, I didn't bump the pkgrel locally but you shouldn't have to do that, but that would have probably worked)

edit: I hope this comment isn't too confusing the way I wrote it ^^

### LINKED ISSUE #1023: timestamp based rebuilds thinks everything needs to be rebuilt after "git merge" instead of "git rebase"
@MoreRobustThanYou wrote:
> here's the verbose output for ```pmbootstrap -v build binutils-armhf```: https://pastebin.com/KaSUiQuz

You have the timestamp based rebuilds feature enabled, in which we ask git how much has changed in your aports folder when compared to `origin`. Usually that should only be what you are working on, e.g. a few APKBUILDs or python files at most. But in your case, there are tons of files.

This is how I think most people use git with `pmbootstrap`.

Initial clone:
```shell
cd pmbootstrap
git remote add git remote add -f MoreRobustThanYou git@github.com:MoreRobustThanYou/pmbootstrap
```

Updating `master` (which points to `postmarketOS/pmbootstrap`):
```shell
git checkout master
git pull
```

Updating your repo:
```shell
git fetch MoreRobustThanYou
```

Workarounds:
* Disable timestamp based rebuilds
* Clone the repo as described above

Some questions to help debugging this:
* How did you setup your repo?
* How do you integrate new changes? (you mentioned merging them?)

### LINKED ISSUE #1156: gtk+2.0-maemo-2.24.31-r3.apk is rebuilt in each and every build
Configure for nokia n900 with xfce4. Then, gtk+-maemo is rebuilt, in each and every build, and even two times doing one build?! It takes 40 minutes to build, so ... it is a bit of a problem. First build:

[12:51:31] (buildroot_armhf) build armhf/gtk+2.0-maemo-2.24.31-r3.apk
[13:28:49] (buildroot_armhf) install mesa-egl xfce4 xorg-server
[13:29:00] (buildroot_armhf) build armhf/postmarketos-ui-xfce4-0.1-r0.apk
[13:29:07] (buildroot_armhf) build armhf/gtk+2.0-maemo-2.24.31-r3.apk

(I'll post second build when this finishes).

I get these warnings, but they may not be related:

>>> gtk+2.0-maemo*: Script found. /bin/sh added as a dependency for gtk+2.0-maemo-2.24.31-r3.apk
/usr/bin/abuild: line 2554: git: not found

### LINKED ISSUE #1167: What to do about timestamp based rebuilds?
### Timestamp based rebuilds recap
Quoting the [50 days post](https://ollieparanoid.github.io/post/50-days-of-postmarketOS/#timestamp-based-rebuilds):
> `abuild` is the package building program from Alpine Linux, that `pmbootstrap` wraps around. Just like in most Linux distributions it considers packages up to date, when the [binary] package version is equal with the version of the package recipe. This has the side-effect, that package developers must pass `--force` to rebuild a package if they changed anything, but do not want to increase the version yet.
> [@mmaret [expected](https://github.com/postmarketOS/pmbootstrap/issues/69#issuecomment-307321715)] it to work like the classic [`make`](https://en.wikipedia.org/wiki/Make_(software)) build automation program, which means that **after modifying one file anywhere in the source tree, it automatically detects which packages are out of date and rebuilds them** [...]

### Three open bugs caused by this feature
As you know, I'm trying to fix all pmbootstrap bugs before working on new features. Of the [6 bugs](https://github.com/postmarketOS/pmbootstrap/issues?utf8=%E2%9C%93&q=label%3Abug+label%3Apmbootstrap+is%3Aopen+) we have right now, 3 bugs are in the timestamp based rebuilds feature:

* #1156 gtk+2.0-maemo-2.24.31-r3.apk is rebuilt in each and every build
* #1023: timestamp based rebuilds thinks everything needs to be rebuilt after "git merge" instead of "git rebase"
* #985: Binary repo hinders / confuses

(The other 3 bugs have PRs waiting for reviews, which directly resolve them or at least work towards resolving them.)

### Why is it so complicated?
Quote from https://github.com/postmarketOS/pmbootstrap/issues/1156#issuecomment-359954250:
> This seems easy at first, but in combination with `git`, when you switch branches or just cloned a repo, then the files will have newer timestamps than the binary packages. So pmbootstrap would think they need to be rebuilt.

> To work around this, we ask git to diff against current master, and only look at the files that have changed in regard to `master`. But this only seems to work correctly when using `git rebase` instead of `git merge` (and I'm not 100% sure of that even, need to investigate this more). This feature got even more complicated as the binary repo was introduced.

### Drop it?
Unfortunatelly I do not see a straight forward way to fix the issues. And because it is so complicated, it makes the code hard to maintain and PRs get really hard to review and understand.

I'm not saying this lightheartedly, because I spent a great deal of time to implement this in the first place. But what if we just dropped this feature and documented instead, that `--force` is necessary when not changing the version?

ollieparanoid:
Discussion leading up to this PR: #1167

If you want to build a package without changing the version number,
please use `--force` from now on. For example:

    pmbootstrap build --force hello-world

Prior to this commit, changes were detected automatically (timestamp
based rebuilds). However, that feature does not work as expected with
the binary package repository we have now, and depending on how you use
git, it has never worked. Close #1167, close #1156, close #1023 and
close #985.


ollieparanoid:
@z3ntu had a good idea, which I want to add to this PR to make the transition easier:
> ollieparanoid: maybe when running pmb build and it doesn't need a rebuild according to the logic, maybe a message should be printed like "Package is already built. Use --force to force a rebuild."

ollieparanoid:
It is implemented now. I have kept the "Done." message, because when building multiple packages it might be that some are up to date, and others are not. So it makes sense to have a message at the end when it's through in my opinion.

Example:
```shell-session
$ pmbootstrap build hello-world heimdall
[18:33:43] NOTE: Package 'hello-world' is up to date. Use 'pmbootstrap build hello-world --force' if needed.
[18:33:44] NOTE: Package 'heimdall' is up to date. Use 'pmbootstrap build heimdall --force' if needed.
[18:33:44] Done
```

@z3ntu: Is this what you had in mind? If you have a few minutes, maybe you could review the PR.

ollieparanoid:
Thanks!

PR #1176, opened 2018-01-28
Add initial support for Zenfone 5 (t00f/t00j)

yxist:
Current status:
  - flashing only manually (intel phones have different bootimg format)
  - usb networking works
  - xorg works using fbdev
  - packages needed for wayland are not available for x86
  - touch works in xorg and osk-sdl


aports/device/device-asus-t00f/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+pkgver=0.1
+pkgrel=0
+url="https://postmarketos.org"
+license="MIT"
+arch="noarch"
+options="!check"
+depends="linux-asus-t00f mkbootimg"
+subpackages="$pkgname-x11"
+source="
+    deviceinfo
```
ollieparanoid: We use tabs in APKBUILDs/shell scripts (just like Alpine) and spaces in Python. Could you adjust it in this APKBUILD?

aports/device/device-asus-t00f/APKBUILD
```diff
@@ -0,0 +1,36 @@
...
+    90-android-touch-dev.rules
+    xorg.conf
+"
+
+package() {
+	install -Dm644 "$srcdir"/deviceinfo \
+		"$pkgdir"/etc/deviceinfo
+    install -Dm644 "$srcdir"/90-android-touch-dev.rules \
```
ollieparanoid: In newer device packages, we put the touch device in the `deviceinfo` and let `devicepkg-dev` generate the udev rule: https://wiki.postmarketos.org/wiki/Device_specific_package#devicepkg-dev
Could you please check if that works as well?

yxist:
Udev rule works.  Everything builds and runs for me. I haven't been working on it for a long time, it was my main device back then, didn't want to brick it or something.

PR #1177, opened 2018-01-29
Nokia N900: add right keymap for X11.

pavelmachek:
Use subpackage so that configuration file is not used in configurations that do not need it.

### DIFF
--- a/aports/device/device-nokia-n900/APKBUILD
+++ b/aports/device/device-nokia-n900/APKBUILD
@@ -1,6 +1,6 @@
 pkgname=device-nokia-n900
 pkgver=1
-pkgrel=0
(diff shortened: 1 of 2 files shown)


ollieparanoid:
Unrelated note from something I was reminded while testing this PR:
`linux-firmware` has been split up in Alpine. I guess you know which subpackages are actually needed (there's one subpackage per `/lib/firmware/*` subfolder), so if you have time for it and want to do it, you could make a PR that lets `device-nokia-n900` depend on the subpackages instead of the whole `linux-firmware`, and therefore reduces the installation size by 200 MB or so.

pavelmachek:
Let me create separate PR for firmware.

PR #1180, opened 2018-01-29
Add initial device support for Sony Xperia T3 (sony-seagull)

MayeulC:
This pull request should be relatively straightforward, and includes initial support for a new device. Nothing fancy here, except maybe some of the patch headers :)

The kernel is fetched from a sony copyleft archive, version 3.4.0.
Display, touch and USB networking are working.


MayeulC:
> Is there a public backup (e.g. on GitHub) of that source code? If not, I would recommend to upload that .tar.bz2 as a whole into your personal GitHub account (so the checksum matches) just in case this one goes offline at some point.

Yeah, you are right, I thought about this as well. There is no backup AFAIK ( https://github.com/sonyxperiadev/kernel-copyleft/issues/13 ).

Well, I think I will just create a depot with the sources then (as there are multiple versions, it could actually end up being handy). Uploading https://github.com/MayeulC/sony-seagull-kernel as I speak. The issue here is that checksums won't match, but I am not sure whether it is really a problem or not.

> How about: $pkgname-$pkgver.tar.bz2::${_srcurl}

I will fix this, thanks :) I was wondering about this as well. I should just mention those points I am not sure about in my PRs, it probably makes the reviews easier. But I think that's about it.

> The Travis failure is unrelated to your PR

I noticed the travis failure was unrelated, don't worry, I will perform the changes, rebase and force push.

mirh:
~~So.. I'm like the first to praise official sony's repos.. 
Though, the basis of this being a static tarball it's kinda crappy.~~

Seriously man, you are a lucky one. 
Fuck even XDA, check [this repo](https://github.com/sonyxperiadev/kernel/tree/aosp/LA.BF64.1.2.2_rb4.7)
Many of the older devices devs would kill to be in your position.

ollieparanoid:
> The issue here is that checksums won't match, but I am not sure whether it is really a problem or not.

You could just upload the whole zip archive without extracting it, then the checksums would match. It is useful for security reasons - no one will look through the entire kernel source code for modifications, but if the hash of the archive matches with the one from the official mirror, then it's easy to see that the sources were not modified.

@mirh: Sorry, but I did not understand what you were saying. Could you rephrase that please?

MayeulC:
> You could just upload the whole zip archive without extracting it, then the checksums would match. It is useful for security reasons - no one will look through the entire kernel source code for modifications, but if the hash of the archive matches with the one from the official mirror, then it's easy to see that the sources were not modified.

~Yeah, that was my guess as well (hum, I am not sure anybody would actually double check this, but well...)~ Strike that, I just got it. You mean, as a backup mirror, not a CDN. I'm fine with that, I'll just upload the archive as a release in the aforementioned repository.

mirh:
> Sorry, but I did not understand what you were saying. Could you rephrase that please?

He was complaining on IRC (should I call it matrix? w/e) about sony only offering tarballs for his phone, the other day. 
And I might or might not almost be suggesting my usual: "did you check XDA". 

But then I found out that *ofc*, no way some OEM hadn't to impress with their FOSS department. Ie that repo I linked, which is not only what they employed in their stock rom, but much much better (too bad they use codeaurora branch names, instead of something.. A bit clearer)

Also, I hypothesize that (as long as they don't break android compatibility I guess) they are still also accepting contributions. So, instead of keeping that bunch of patches here, you could even upstream it.

MayeulC:
Yeah, I have closed the PR while pondering whether to try to use use a more recent kernel. This could go almost as-is, but I have to find out whether it is worth trying to make use of a more recent kernel first.

ollieparanoid:
Thanks for explaining @mirh and @MayeulC!

mirh:
No, indeed T3 section is very meager (really oddly imo)
Still, that's the usual first tip I tend to give to almost all developers.

PR #1189, opened 2018-01-31
Fix invalid /etc/apk/repositories entry on device

### LINKED ISSUE #1186: /mnt/pmbootstrap-packages repository still configured in installation
That repository only exists while working with the chroots with `pmbootstrap`, not afterwards. So it should be removed after copying everything to the installation partition (image or sd card). This makes the following warning go away on the device:
```
WARNING: Ignoring /mnt/pmbootstrap-packages/armhf/APKINDEX.tar.gz: No such file or directory
```

ollieparanoid:
It used to have an entry for /mnt/pmbootstrap-packages, which only makes sense while working on the chroot with pmbootstrap. After the installation on the device, there's no repo in that path.

This changes `pmbootstrap install`, as well as the Android recovery zip installation (therefore it would be cool if @ata2001 could take a look at it). I've also added four empty lines in the latter to make it more readable, is that okay?

Fixes #1186.

### How to test
* Do the normal installation
* Check if `/etc/apk/repositories` contains the `/mnt/pmbootstrap-packages` entry
* Do the Android recovery zip installation
* Check if `/etc/apk/repositories` contains the `/mnt/pmbootstrap-packages` entry


ata2001:
@ollieparanoid the recovery part looks good to me. I tested it on sony-amami, flashed fine and the entry was gone. (however I don't quite understand why the blank line between formatting and mounting is needed)

ollieparanoid:
Do you mean all blank lines, or a specific one? It felt more readable to me with them inserted, but I don't have a strong opinion on these. Feel free to add a commit that fomats it back to how you prefer it, or tell me which empty lines to remove.

Thanks a lot for reviewing and testing!

Issue #441, opened 2017-08-22
Package alternative mkbootimg/unpackbootimg (osm0sis' fork of the AOSP/CM C code)?


ollieparanoid:
In #145 we [were](https://github.com/postmarketOS/pmbootstrap/issues/145#issuecomment-314894061) [discussing](https://github.com/postmarketOS/pmbootstrap/issues/145#issuecomment-315189047) using [this fork](https://github.com/osm0sis/mkbootimg) as alternative for both `mkbootimg` and `unpackbootimg`. I have even mailed the author if he could add tags to the repository, so we could download properly versioned archives instead of a random git hash.

In the end, we didn't even need to change it, because the tool we use already outputs all needed information. But the author has responded and started to tag releases now:
https://github.com/osm0sis/mkbootimg/releases

...so I feel like I should at least let everyone know that we have the option to do this!

Some properties of this tool:
* Based on AOSP's and CyanogenMod's C code for mkbootimg/unpackbootimg (the current code is based on Python, since `mkbootimg` in AOSP was rewritten in Python)
* Has proper version tags
* Has as goal to support all kinds of devices (also has the `qtdb` feature, we recently switched the `mkbootimg` source for that)
* Prints out more than our current unpackbootimg
* Seems to be well maintained

So what we have right now works for us and I am not *that* interested anymore in changing it. But if you want to have this version of `mkbootimg` or if you are against it with a good reason, please comment here.

drebrez:
An advantage to have this one instead of what we have now (python based) is that we will not install anymore the python package in the device chroot and reduce the system image.
Is it correct @ollieparanoid ?

ollieparanoid:
We need Python for `postmarketos-splash`. If we replaced that with C/C++ code, then yes. But until then we need Python anyway.

drebrez:
@ollieparanoid why don't we build both (splashscreens and bootimg) in `native` chroot?

MartijnBraam:
In pmbootstrap yes, but if you update packages on your phone after the initial build you run mkinitfs and thus make-splash on the phone directly

ollieparanoid:
*Pinging @osm0sis so he does not miss the discussion. We still did not decide whether we actually use it though. Thanks for making a release which we could package!*

osm0sis:
Hey, just for reference, my fork is definitely more updated than others, including proper unpackbootimg support for extended cmdline which I don't think any others have right now, and sha256 support.

[...]

PR #1193, opened 2018-02-01
mkbootimg: replace python version with c version.

lramage94:
See https://github.com/postmarketOS/pmbootstrap/issues/441


drebrez:
@lramage94 @ollieparanoid this new package also provides `unpackbootimg` executable, what about removing the other `unpackbootimg` package that we have? (aports/main/unpackbootimg)

ollieparanoid:
Good idea! We could add `provides="unpackbootimg"` to this package to make the transition easier.

drebrez:
@ollieparanoid I'm not sure how exactly works that `provides` variable in APKBUILD, it's something like an alternative name for the same package? If there is only `mkbootimg` package and not the `unpackbootimg`, if I run `apk add unpackbootimg` it will install `mkbootimg`?

ollieparanoid:
Yes, it would work exactly like that.

ollieparanoid:
So I wrote a script to compare the boot.img files created by both tools, and ran it on 19 devices so far. The result is, that it is not identical for any of them. I think the order of the appended `dtb` files is different, I'll need to investigate this further. (CC @osm0sis, this might be interesting to you.)

[Here's the script.](https://gist.github.com/ollieparanoid/eb58ae612368111f2621d082363eaa1b) (If anyone wants to run this, please read it before executing it.)

drebrez:
Does the order of `dtb` files matters?

ollieparanoid:
I guess it does not matter for booting. But it makes it harder to tell if both files produce the same output.

osm0sis:
It'll always be different; Google's latest mkbootimg added os_version and os_patch_level so no chance of them being the same because of the updated header. dtb files aren't added any differently than CM's source before mine.

So unfortunately it's not really a valid automated check. What matters is that they boot.

ollieparanoid:
Right, that makes sense. Then I suggest we boot it on a hand full of devices at least, before merging it.

(I can do that later/tomorrow on my `lg-mako` (Nexus 4), would be nice if someone else could test on their devices as well.)

ollieparanoid:
Release `0.5.0` has been tagged to easily go back:
https://github.com/postmarketOS/pmbootstrap/releases/tag/0.5.0

Issue #1188, opened 2018-01-31
Add framebuffer debugging tools to debug-shell initramfs hook


ollieparanoid:
We would need to support adding files in initramfs hooks first. When that is done, we could add the following to the `debug-shell` hook:
* msm-fb-refresher
* [@drebrez' fb debug script](https://bpaste.net/show/a5fb301b21c5)
* some test scripts that draw to the framebuffer?

MayeulC:
Maybe a script that finds every `brightness` in `/sys`, and set it to `brightness_max`? I don't know whether it would be useful to have that, as compared to manual testing. Plus it could turn on the flashlight. But maybe we could add an interactive approach (or even a menu) that tries different things?

ollieparanoid:
> maybe we could add an interactive approach (or even a menu) that tries different things?
Just in case you have a UI on the screen in mind: we assume that the screen may not be working at this point. The user is connected via telnet and can execute debugging tools that way.

So for now I would recommend to keep the commands separate without any menu), and once we have a few tools, and good workflows for debugging, we could put that into scripts which do more hand-holding to make it easier for users.

drebrez:
I think adding the `msm-fb-refresher` into the debug-shell hook is problematic because we will need to install the package in the chroot and it will enable the service. 
Instead I suggest to add a simply flag in the `fbdebug` tool to perform the same ioctl pan operation, it will be just few lines of code and no additional package dependencies

ollieparanoid:
Actually the service gets enabled in `postmarketos-base` right now (which means, `postmarketos-base` depends on the `deviceinfo` to be present for that other issue, we can't simply make the device package depend on the pmOS-base package that way, unless we change this!).

But I agree with you, we could add these few lines into `fbdebug` to make packaging simpler.

drebrez:
Ah ok, now I see it, the package only creates the service.
But I'll add the ioctl pan in `fbdebug`, so it will be also easier to explain in the wiki:
> if `fbdebug -pan`(or something like that) refreshes the drawed image, then add `msm-fb-refresher=true` in deviceinfo

drebrez:
@ollieparanoid I've tried adding the `fbdebug` package to the dependencies of the `postmarketos-mkinitfs-hook-debug-shell`, but when I run `pmbootstrap initfs hook_del debug-shell` it doesn't uninstall the `fbdebug` package from the chroot.
[...]

PR #1196, opened 2018-02-01
Add initial framebuffer debug tool

drebrez:
I've started with a basic debug tool that performs the `FBIOGET_FSCREENINFO` and `FBIOGET_VSCREENINFO` IOCTLs calls and print out all the values in the struct retrieved.

I've chose the name `fbdebug` to follow the busybox naming for framebuffer tools (`fbset`, `fbsplash`,...)

Please review and test it, thanks.

#1188

### DIFF
--- a/aports/main/fbdebug/fbdebug.c
+++ b/aports/main/fbdebug/fbdebug.c
@@ -1,5 +1,5 @@
 /*
-Copyright (C) 2017 drebrez <drebrez@gmail.com>
+Copyright (C) 2018 drebrez <drebrez@gmail.com>
 
 This program is free software; you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
@@ -27,8 +27,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 void usage(char* appname)
 {
     printf("Usage: %s [-d DEV] [-h]\n\
-    -d	Framebuffer device (default /dev/fb0)\n\
-    -h	Show this help\n", appname);
+    -d  Framebuffer device (default /dev/fb0)\n\
+    -h  Show this help\n", appname);
 }
 
 void print_fix_screeninfo(struct fb_fix_screeninfo* scr_fix)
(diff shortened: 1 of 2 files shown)


aports/main/fbdebug/fbdebug.c
```diff
@@ -0,0 +1,160 @@
+/*
+Copyright (C) 2017 drebrez <drebrez@gmail.com>
```
ollieparanoid: We have 2018 now :smiley:

aports/main/fbdebug/fbdebug.c
```diff
@@ -0,0 +1,160 @@
...
+    printf("\treserved[0]      = %u\n", scr_var->reserved[0]);
+    printf("\treserved[1]      = %u\n", scr_var->reserved[1]);
+    printf("\treserved[2]      = %u\n", scr_var->reserved[2]);
+    printf("\treserved[3]      = %u\n", scr_var->reserved[3]);
+}
+
+int main(int argc, char** argv)
+{
+    const char *fb_device;
+	struct fb_fix_screeninfo scr_fix;
```
ollieparanoid: Tabs and spaces mixed, I'd recommend to decide on one throughout the file.

ollieparanoid:
It might make sense to create a wiki page, that explains how to use the tool (how you can use it to fix your framebuffer drivers).

aports/main/fbdebug/fbdebug.c
drebrez: thanks, I will fix it.
I wrote it using `vi` in the virtualmachine with alpine linux, I'm not so expert with `vi`.
I'll look for some way to display the special characters next time I have to use `vi`

drebrez:
@ollieparanoid I think we can explain the various tools that we will add to the debug-shell hook once we have them.
Next week I will probably extend this tool to add additional tests, like the ioctl pan and draw something on the screen. I was thinking about something like a simplified version of a TV test pattern 😄

eMPee584:
🤔 hmm why is it not called fbinfo 😃

drebrez:
@eMPee584 `fbinfo` sounds more like a tool to only get information about it, the idea is to add additional functionalities to test various IOCTLs and writing into it.
Maybe we can call it `fbtest`? @ollieparanoid what you prefer?

eMPee584:
maybe put the info part in a fbinfo executable, and the testing in a fbtest : )
or, rather go wild 🤠

drebrez:
As you prefer, my plan is to add more features next week, but I can do it
in another PR

ollieparanoid:
Alright, merged it. Smaller diffs are easier to review, and who knows, maybe it could already be useful :+1:

PR #1197, opened 2018-02-01
[nextbit-robin] Added msm-fb-refresher so display is now working properly

ashirviskas:


drebrez:
@ashirviskas there's no need to close the PR, just push another commit to the same branch

ollieparanoid:
Actually you can use the same branch and push more commits to it. The Travis check will run again on the newest commit then.

ashirviskas:
Gotcha!

PR #1198, opened 2018-02-01
Package GPS support for N900

pavelmachek:
So far I was able to get the binary to work on N900, but more work is needed. Anyway, I'd not mind getting it upstream so that others can help, and it would be good to know if there's something to be fixed in the packaging.

### DIFF
--- a/aports/device/gps-nokia-n900/APKBUILD
+++ b/aports/device/gps-nokia-n900/APKBUILD
@@ -1,0 +1,26 @@
+pkgname=gps-nokia-n900
+pkgver=0.0
+pkgrel=0
+pkgdesc="GPS support for Nokia N900"
+url="https://github.com/postmarketOS/gps-nokia-n900"
+arch="all"
+license="GPL3"
+depends=""
+makedepends="linux-headers"
+subpackages=""
+source="${pkgname}-${pkgver}.tar.gz::https://github.com/postmarketOS/gps-nokia-n900/archive/v${pkgver}.tar.gz"
+options="!check"
+
+builddir="$srcdir"/"$pkgname"-"$pkgver"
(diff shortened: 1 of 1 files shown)


aports/device/gps-nokia-n900/APKBUILD
```diff
@@ -0,0 +1,26 @@
+pkgname=gps-nokia-n900
+pkgver=0.0
+pkgrel=0
+pkgdesc="GPS support for Nokia N900"
+url="https://github.com/pali/0xFFFF"
```
ollieparanoid: I'd put `url="https://github.com/postmarketOS/gps-nokia-n900"` here.

aports/device/gps-nokia-n900/APKBUILD
```diff
@@ -0,0 +1,26 @@
...
+arch="all"
+license="GPL2"
```
ollieparanoid: You've chosen the GPLv3 in the project.

pavelmachek:
I rebased the patch, and updated it as requested.

pavelmachek:
Huh. Checks fail, but somehow I don't think I introduced this problem...?

>       assert [] == failed
E       AssertionError: assert [] == ['musl-armhf: 1.1.18-r5 != 1...5 != 1.1.18-r6 (from musl)']
E         Right contains more items, first extra item: 'musl-armhf: 1.1.18-r5 != 1.1.18-r6 (from musl)'
E         Use -v to get the full diff
test/test_upstream_compatibility.py:124: AssertionError
----------------------------- Captured stdout call -----------------------------

pavelmachek:
Ah, thanks, sorry for the noise!

PR #1200, opened 2018-02-03
Add device: Xperia Z3 Tablet Compact

kiljacken:
This provide basic support giving:
* Flashing
* Booting (from flashed kernel or `pmbootstrap flasher boot`
* Usb Net
* Touch input
* Screen


kiljacken:
Awesome! I'll be looking into getting more working in ~two weeks, as my phone is in for repair so my tablet has to do as 2FA device for a while :)

PR #1201, opened 2018-02-03
Only download APKINDEX for relevant architectures

ollieparanoid:
We're downloading the APKINDEX files for all architectures supported by
postmarketOS currently (x86, x86_64, armhf, aarch64). Most of the time,
we only need it for the native and device arch, so this PR reduces the
downloaded files to what is really necessary.

This PR gives `pmb.helpers.repo.update()` a new optional `arch` parameter, and uses that parameter wherever possible.

### DIFF
--- a/pmb/build/_package.py
+++ b/pmb/build/_package.py
@@ -214,7 +214,7 @@ def get_gcc_version(args, arch):
     <https://linux.die.net/man/1/ccache>
     :returns: a string like "6.4.0-r5"
     """
-    return pmb.parse.apkindex.package(args, "gcc-"+arch,
+    return pmb.parse.apkindex.package(args, "gcc-" + arch,
                                       args.arch_native)["version"]
 
 
--- a/pmb/build/_package.py
+++ b/pmb/build/_package.py
@@ -214,7 +214,8 @@ def get_gcc_version(args, arch):
     <https://linux.die.net/man/1/ccache>
     :returns: a string like "6.4.0-r5"
     """
-    return pmb.parse.apkindex.package(args, "gcc", arch)["version"]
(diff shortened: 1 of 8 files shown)


ollieparanoid:
Rebased on master, ready to get reviewed.

MartijnBraam:
If I have an armhf device selected and run `pmbootstrap update` it only updates x86_64, is this correct?

ollieparanoid:
Thanks for testing!

> If I have an armhf device selected and run pmbootstrap update it only updates x86_64, is this orrect?

This was expected with the code you've tested. But yeah, when thinking about this, that's not really intuitive. As a user, I would expect that such an update command takes care of all *relevant* arches automatically.

I rewrote the logic, now it does the following:
* update all existing APKBUILD files (because the existing files are the relevant ones, that pmbootstrap will only update if they are older than 4 hours. pmbootstrap will always download APKINDEX files that do not exist, hence non-existing files are not relevant for the update command).
* it is still possible to specify a certain arch with `--arch`
* it is possible to override the behavior of only updating existing files with `--non-existing`.

```
$ pmbootstrap update --help
usage: pmbootstrap update [-h] [--arch {aarch64,x86,x86_64,armhf}]
                          [--non-existing]

optional arguments:
  -h, --help            show this help message and exit
  --arch {aarch64,x86,x86_64,armhf}
                        only update a specific architecture
  --non-existing        do not only update the existing APKINDEX files, but
                        all of them
```

To test this, all existing APKINDEX files can be deleted like this:
```
sudo rm -v ~/.local/var/pmbootstrap/cache_apk_*/APKINDEX*
```

What do you think, does this make more sense and is it ready to go?

ollieparanoid:
(I have a good fix for #1272 ready, but it depends on this PR.)

PR #1202, opened 2018-02-03
Index parser: support multiple package providers

### LINKED ISSUE #1122: pmbootstrap picks wrong package when two packages provide the same library

ollieparanoid:
* The APKINDEX parser used to return a dictionary with one package for
  a given package name. This works for the installed packages database,
  because there can only be one provider for a package. But when
  parsing packages from binary repositories, we need to support
  multiple providers for one package. It is now possible to get a
  dictionary with either multiple providers, or just a single provider
  for each package.
* Dependency parsing logic has been adjusted, to support multiple
  providers. For multiple providers, the one with the same package
  name as the package we are looking up is prefered. If there is none
  (eg. "so:libEGL.so.1" is provided by "mesa-egl"), it prefers packages
  that will be installed anyway, and after that packages that are
  already installed. When all else fails, it just picks the first one
  and prints a note in the "pmbootstrap log".
* Added testcases for all functions in pmb.parse.apkindex and
  pmb.parse.depends
* pmbootstrap chroot has a new "--add" parameter to specify packages
  that pmbootstrap should build if neccessary, and install in the
  chroot. This can be used to quickly test the depencency resolution
  of pmbootstrap without doing a full "pmbootstrap install".

Fixes #1122.

@NotKit: Could you review/test this?

### DIFF
--- a/pmb/aportgen/busybox_static.py
+++ b/pmb/aportgen/busybox_static.py
@@ -26,14 +26,12 @@ import pmb.chroot.apk_static
 
 
 def generate(args, pkgname):
-    # Install busybox-static in chroot (so we have the APKINDEX and verified
-    # apks)
+    # Install busybox-static in chroot to get verified apks
     arch = pkgname.split("-")[2]
-    apkindex = pmb.chroot.apk_static.download(args, "APKINDEX.tar.gz")
     pmb.chroot.apk.install(args, ["busybox-static"], "buildroot_" + arch)
 
     # Parse version from APKINDEX
-    package_data = pmb.parse.apkindex.read(args, "busybox", apkindex)
+    package_data = pmb.parse.apkindex.package(args, "busybox")
     version = package_data["version"]
     pkgver = version.split("-r")[0]
     pkgrel = version.split("-r")[1]
--- a/pmb/aportgen/musl.py
+++ b/pmb/aportgen/musl.py
@@ -26,13 +26,12 @@ import pmb.chroot.apk_static
 
 
 def generate(args, pkgname):
-    # Install musl in chroot (so we have the APKINDEX and verified musl apks)
+    # Install musl in chroot to get verified apks
     arch = pkgname.split("-")[1]
-    apkindex = pmb.chroot.apk_static.download(args, "APKINDEX.tar.gz")
     pmb.chroot.apk.install(args, ["musl-dev"], "buildroot_" + arch)
 
     # Parse musl version from APKINDEX
-    package_data = pmb.parse.apkindex.read(args, "musl", apkindex)
+    package_data = pmb.parse.apkindex.package(args, "musl")
     version = package_data["version"]
     pkgver = version.split("-r")[0]
     pkgrel = version.split("-r")[1]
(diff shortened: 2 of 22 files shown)


NotKit:
I tested the changes on top of feature/libhybris branch, building rootfs for Droid 4 (motorola-maserati).

extra_packages = xorg-server,xf86-input-evdev,xf86-input-libinput,musl-utils,libhybris-gles,libhybris-egl,libhybris-4.4,postmarketos-ui-hildon

It tries to install both mesa-egl and libhybris-egl. I guess there is a problem with libhybris package (I had to add !tracedeps to options, as otherwise libhybris package would depend on both libhybris-7.1 and libhybris-7.1-caf, which can't be installed simultaneously).

[install.log](https://github.com/postmarketOS/pmbootstrap/files/1692659/install.log)

ollieparanoid:
I would like to reproduce this. The sources of `android-headers-4.4` and `droid-hal-motorola-maserati` give 404s, so I've updated them. But after doing that, building `hybris` fails with:
```shell
>>> libhybris: Entering fakeroot...
cp: can't stat '/home/pmos/build/src/tmpinstall/libhybris-7.1/*': No such file or directory
>>> ERROR: libhybris*: package failed
>>> ERROR: libhybris: all failed
```
Could you make sure that it builds again (with `--strict` to make sure, and with updated sources)?

ollieparanoid:
@NotKit: Regarding `libhybris` integration: I think the easiest solution would be if we use the `/etc/ld.so.conf.d/hybris` approach you mentioned in the chat. That in combination with this PR should make it work, right?

NotKit:
Sirry for the last reply, I somehow missed your last comment. I guess it should work, but can MUSL support ld.so.conf.d format? From a quick search, it looks like it doesn't support “include" directive used in /etc/ld.so.conf in Ubuntu.

ollieparanoid:
It [seems to provide](http://www.openwall.com/lists/musl/2015/12/24/8) a `/etc/ld-musl-$ARCH.path` file instead:
>  musl uses /etc/ld-musl-$ARCH.path instead; this is a simple newline or colon separated list of directories.

Regarding this PR, with it I've hit a situation once where pmbootstrap hang forever resolving dependencies... I'll try to reproduce and fix it before this is ready to go.

ollieparanoid:
I have compiled all sorts of packages, and I could not reproduce the hanging. Maybe it was just because there was a distfile being downloaded, and I did not notice it.

This PR has numerous benefits, besides multiple package providers, it refactored the whole APKINDEX parsing logic and added testcases for each of its functions.

So all in all I think this is good to go,  and just to be sure I've tagged version [0.6.0](https://github.com/postmarketOS/pmbootstrap/releases/tag/0.6.0) for this.

PR #1211, opened 2018-02-07
main: add triggerhappy

alive4ever:
Add triggerhappy package.

Note that triggerhappy build depends on linux-headers package,
which probably differs between kernel version.

Build tested with alpine armhf linux-headers 4.4.6-r2 package. Run
tested on armani, with the following result:

/dev/input/event0 generates ABS_MT_TRACKING_ID, ABS_MT_POSITION_X,
ABS_MT_POSITION_Y, ABS_MT_TOUCH_MAJOR
/dev/input/event1 handles KEY_VOLUMEUP.
/dev/input/event2 handles KEY_VOLUMEDOWN and KEY_POWER.


drebrez:
I don't know much about `triggerhappy` but it seems something similar to the `busybox acpid` which has been used already in the `device-nokia-n900` package.
See [aports/device/device-nokia-n900/device-nokia-n900.post-install#L3-L9](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/device/device-nokia-n900/device-nokia-n900.post-install#L3-L9) with [acpi.map](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/device/device-nokia-n900/acpi.map) and [acpi_handler.sh](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/device/device-nokia-n900/acpi_handler.sh).
Or ACPI events are something completely different?

alive4ever:
``triggerhappy`` doesn't handle ACPI.

It only listens to events from specified input interface - especially HID interface (keyboard, mouse, touchpad, touchscreen, joysticks).

This is useful in debugging touchscreen or button tagged by udev as keyboards, since keyboard key-presses don't generate ACPI events.
```
# thd --dump /dev/input/event*
```

When the service is started, it can be configured to launch specific command when the corresponding button is pressed.

```
/etc/triggerhappy/triggers.d/buttons.conf

KEY_VOLUMEUP 1 logger "VOLUMEUP is pressed"
KEY_VOLUMEUP 0 logger "VOLUMEUP is released"
KEY_VOLUMEDOWN 1 logger "VOLUMEDOWN is pressed"
KEY_VOLUMEDOWN 0 logger "VOLUMEDOWN is released"
KEY_POWER 1 logger "POWER is pressed"
KEY_POWER 0 logger "POWER is released"
```

alive4ever:
I've created a [wiki page](https://wiki.postmarketos.org/wiki/Troubleshooting:HID_buttons) explaining use cases of triggerhappy.

aports/main/triggerhappy/APKBUILD
```diff
@@ -0,0 +1,30 @@
+# Contributor: alive4ever
+pkgname=triggerhappy
+pkgver=0.5.0
+pkgrel=0
+pkgdesc="A lightweight hotkey daemon"
+url="https://github.com/wertarbyte/triggerhappy"
+arch="all"
+license="GPL"
```
ollieparanoid: The correct license seems to be `"GPL-3.0-or-later"` (See [APKBUILD_Reference#license](https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#license)).

aports/main/triggerhappy/APKBUILD
```diff
@@ -0,0 +1,30 @@
...
+makedepends="git linux-headers perl"
```
ollieparanoid: `git` does not seem to be required, it builds with `--strict` fine without `git` as dependency. Maybe you have added it because of the `/usr/bin/abuild: line 2554: git: not found` message? This is an upstream bug, with a PR pending to fix it: https://github.com/alpinelinux/abuild/pull/32

alive4ever:
> git does not seem to be required

I just added it without much thinking, since the source code is hosted on github so git is required. Thanks for pointing this out.

> The correct license seems to be "GPL-3.0-or-later"

Corrected. I've also installed the ``COPYING`` file during packaging.

ollieparanoid:
> I've also installed the ``COPYING`` file during packaging.

Actually installing OSI-approved licenses is [not necessary in Alpine](https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package#license), as they are provided by the [spdx-licenses](https://pkgs.alpinelinux.org/contents?file=GPL-3.0-or-later*&path=&name=spdx-licenses-*&branch=edge) packages:
> If a package has a special/custom license or is not listed as [OSI approved](https://opensource.org/licenses/alphabetical) we need to provide it with the release.

Could you remove it again? After doing that it should be ready to ship in my opinion :ship:

alive4ever:
>Actually installing OSI-approved licenses is not necessary in Alpine

I see. I have removed the installation of COPYING file in favor of ``spdx-licenses``.

PR #1214, opened 2018-02-08
New gp-peak device (Geeksphone Peak)

Rudloff:
Hello,

The Geeksphone Peak is one of the very first Firefox OS phones. Now that Firefox OS is discontinued, it is not really useful anymore so postmarketOS is a great oppotunity to use it again.

This is my first port so I'm learning as I go.
It currently boots and SSH over USB works but the screen stays black.
(weston fails with this error: `Failed to open frame buffer device ‘/dev/fb0’: Permission denied`)


aports/device/linux-gp-peak/config-gp-peak.armhf
```diff
@@ -0,0 +1,2808 @@
...
+CONFIG_SECCOMP_FILTER=y
+
+#
+# GCOV-based kernel profiling
+#
+# CONFIG_GCOV_KERNEL is not set
+CONFIG_HAVE_GENERIC_DMA_COHERENT=y
+CONFIG_RT_MUTEXES=y
+CONFIG_BASE_SMALL=0
+CONFIG_MODULES=y
```
alive4ever: CONFIG_MODULES is enabled. There should be ``make modules_install`` action during kernel packaging.

aports/device/linux-gp-peak/APKBUILD
```diff
@@ -0,0 +1,87 @@
...
+	for _zimg in zImage-dtb Image.gz-dtb *zImage Image; do
+		[ -e "$_zimg" ] || continue
+		msg "zImage found: $_zimg"
+		install -Dm644 "$_zimg" "$_target"
+		break
+	done
+	if ! [ -e "$_target" ]; then
+		error "Could not find zImage in $PWD!"
+		return 1
+	fi
```
alive4ever: Since you have CONFIG_MODULES enabled, there should be ``make modules_install`` during packaging. This is an example on how to install kernel modules.

```
	# External modules install
	cd "$builddir"
		KBUILD_BUILD_VERSION="$((pkgrel + 1 ))-postmarketOS" \
		INSTALL_MOD_PATH="$pkgdir" modules_install \
		|| return 1
```

Rudloff:
I added your code to `APKBUILD` and now I see the modules in `~/.local/var/pmbootstrap/chroot_native/home/pmos/build/pkg/linux-gp-peak/lib/modules/3.0.21/kernel`.

However, after I flash this new image on my device, I don't have any `/lib/modules/` folder and the `linux-gp-peak` package does not seem to contain any module:
```
linux-gp-peak-3.0.21-r0 contains:
boot/vmlinuz-gp-peak
usr/share/kernel/gp-peak/kernel.release
```

Am I missing something?

MartijnBraam:
Maybe you need to zap the device chroot for it to work

ollieparanoid:
@alive4ever: The `|| return 1` is not needed in `abuild` anymore (it was required earlier, thus it is still present in some `APKBUILD` files). Thanks for pointing out the missing external modules install. So far we have not paid attention to it, and only installed missing modules when we were actually using them.

@Rudloff: thanks for making the port and creating the PR! It would be nice if you could create a [device specific wiki page](https://wiki.postmarketos.org/wiki/Help:Device_Page) and add your device to the [device overview](https://wiki.postmarketos.org/wiki/Devices), then others could see your work and build upon it. I'll look into why the modules installation did not work for you.

aports/device/device-gp-peak/APKBUILD
```diff
@@ -0,0 +1,22 @@
+# Reference: <https://postmarketos.org/devicepkg>
+pkgname="device-gp-peak"
+pkgdesc="Peak"
```
ollieparanoid: How about: `Geeksphone Peak`

aports/device/device-gp-peak/deviceinfo
```diff
@@ -0,0 +1,32 @@
+# Reference: <https://postmarketos.org/deviceinfo>
+# Please use double quotes only. You can source this file in shell scripts.
+
+deviceinfo_format_version="0"
+deviceinfo_name="Peak"
```
ollieparanoid: How about: `Geeksphone Peak` (the full name, as in [the reference](https://wiki.postmarketos.org/wiki/Deviceinfo_reference))

aports/device/linux-gp-peak/APKBUILD
```diff
@@ -0,0 +1,94 @@
...
+
+	# External modules install
+	cd "$builddir"
+	make ARCH="$_carch" CC="${CC:-gcc}" \
+		KBUILD_BUILD_VERSION="$((pkgrel + 1 ))-postmarketOS" \
+		INSTALL_MOD_PATH="$pkgdir" modules_install \
+		|| return 1
```
ollieparanoid: The `\ || return 1` is not necessary anymore.

Rudloff:
I used `zap` and now the modules are correctly installed, thanks!
I was only running this:
```bash
pmbootstrap build --force linux-gp-peak
pmbootstrap install --no-fde
pmbootstrap flasher flash_system
pmbootstrap flasher flash_kernel
```

Maybe you could add a section explaining what to run after each change here: https://wiki.postmarketos.org/wiki/Porting_to_a_new_device

Rudloff:
@ollieparanoid I made the changes you requested.

PR #1215, opened 2018-02-09
Update unicsy-demo

pavelmachek:
This updates unicsy-demo package. New version starts up GSM signal monitoring application and battery monitoring panel. Sending/receiving SMSes should work.  Setting up GPRSworks .. sometimes.


pavelmachek:
unicsy_demo has bash hardcoded in few places. Should not be too hard to fix, I'll add it to my todo list.

Yes, your updates look good to me. Thanks!

PR #1219, opened 2018-02-10
New device: Samsung Galaxy Note 8.0 (WiFi) n5110

michitux:
The device boots, Weston works including touch. With #1205, Xfce works including touch and is quite usable. WiFi works when `firmware-samsung-i9305` is installed. Audio works with the right mixer settings, see https://wiki.postmarketos.org/wiki/Audio#WM1811. The kernel can be flashed using heimdall, I did not try flashing system using heimdall but flashed everything using TWRP and adb sideload.

The kernel is just the Lineage OS kernel without any further patches, only the required settings and the firmware path for WiFi have been changed.

Would it be a good idea to just include the audio configuration files in the device package? If yes, can/should there be a script that automatically sets the mixer settings on boot?


michitux:
I have added an initial sound configuration where speaker and internal mic are enabled. As far as I can see, the `alsa` service is not enabled by default though, so maybe it should be enabled after the installation? I have also been able to get the pen to work under x11/xfce4. I have added the udev rules directly now as both the pen and the touch panel need to be added to group for libinput to disable the touch screen when the pen is detected. I have also added a touchscreen calibration file for osk-sdl. I have not tested it with an actually encrypted system yet but I have tested osk-sdl as described in the wiki with this calibration file.

bacara:
Hi there,

I discovered postmarketOS today and tried to install it on my GT-N5110. After long readings and tries (first custom ROM installation), I managed to get everything installed through TWRP with a recovery zip. However, when I try to reboot using TWRP, I have the message "No OS installed! Are you sure you want to reboot?" that pops up. If I reboot the device anyway, the postmarketOS loader screen appears with the "Loading..." message and then, the device reboots by itself. Again and again of course. May I have some assistance about this? Thanks a lot.

michitux:
This message in TWRP is normal, I have the same, this is because postmarketOS creates two sub-partitions under the system partition for boot and system. Did you use ADB sideload? I'm asking because this is the only way I have tested it. Further, I have not tested encryption so it might not work (and afaik there are also problems with encryption and recovery zip files). If you have enabled it, try disabling it. If all that doesn't help, you could also try to install postmarketOS to an SD card. Flashing just the kernel works in download mode works using heimdall by using `pmbootstrap flasher flash_kernel`, I have used that several times. Feel free to update the wiki page with whatever you found works.

herbsmn:
i saw that a section was added in march to the wiki related to the mainlining of the kernel on the very similar midas devices. i help out at the replicant project https://replicant.us and just wanted to mention that if you guys at postmarket ever want to collaborate with us on mainlining the Note 8.0 kernel or replacing its bootloader with uboot, we could probably find some developers interested in helping with such a project.

PR #1222, opened 2018-02-12
New device: Nexus 5X (LG Bullhead)

PureTryOut:
As of right now it seems to boot, as a network interface is presented to which you can connect, but that's about it. Display doesn't work (it's stuck on the Google Logo) and you can not connect to it using SSH. Telnet works if the debug hook is added though.


ollieparanoid:
Thanks for making the PR, the code looks good!

However, the Travis results are weird. One of two failed, and for the succeeding one, building the kernel has actually failed (this might be a bug in the check script? :-\)
https://travis-ci.org/postmarketOS/pmbootstrap/builds/340410284#L7452

I'll rebase this to master and let Travis check it again.

Regarding the network problem, this should help:
https://wiki.postmarketos.org/wiki/Troubleshooting:boot#Kernel_boots_without_network_interface

ollieparanoid:
@PureTryOut: [it fails to build on Travis](https://travis-ci.org/postmarketOS/pmbootstrap/builds/340711590#L7691):
```
drivers/built-in.o: In function `vgacon_init':
/home/pmos/build/src/android_kernel_lge_bullhead-e577f272b6b17b8cd263f8a3cdc7581f7086aef2/drivers/video/console/vgacon.c:588: undefined reference to `screen_info'
/home/pmos/build/src/android_kernel_lge_bullhead-e577f272b6b17b8cd263f8a3cdc7581f7086aef2/drivers/video/console/vgacon.c:588: undefined reference to `screen_info'
drivers/built-in.o: In function `vgacon_startup':
/home/pmos/build/src/android_kernel_lge_bullhead-e577f272b6b17b8cd263f8a3cdc7581f7086aef2/drivers/video/console/vgacon.c:374: undefined reference to `screen_info'
/home/pmos/build/src/android_kernel_lge_bullhead-e577f272b6b17b8cd263f8a3cdc7581f7086aef2/drivers/video/console/vgacon.c:374: undefined reference to `screen_info'
/home/pmos/build/src/android_kernel_lge_bullhead-e577f272b6b17b8cd263f8a3cdc7581f7086aef2/drivers/video/console/vgacon.c:399: undefined reference to `screen_info'
drivers/built-in.o:/home/pmos/build/src/android_kernel_lge_bullhead-e577f272b6b17b8cd263f8a3cdc7581f7086aef2/drivers/video/console/vgacon.c:535: more undefined references to `screen_info' follow
make: *** [Makefile:806: vmlinux] Error 1
>>> ERROR: linux-lg-bullhead: all failed
>>> linux-lg-bullhead: Uninstalling dependencies...
```
(and I really need to fix the check script to not say everything's good when the build fails ._. good that we've catched this pretty early.)

PureTryOut:
I managed to find a fix for the kernel compilation [here](http://linux-kernel.2935.n7.nabble.com/PATCH-arm64-kernel-need-extern-variable-screen-info-for-related-driver-using-td652444.html), however the phone still doesn't boot further than the Google logo. Actually, it's even worse now, no network interface appears any more on my PC.

ollieparanoid:
How can it get worse than not compiling? :smiley: 

But seriously, that looks like a rebase has gone wrong or something?
Unfortunately, the only place I can find this error message is here: https://pastebin.com/ZxGBANdR
And that pastebin is not referenced anywhere on the indexed web from what I can tell (well, now it is here).

Some ideas:
* Use another kernel. I wonder how well the one from [CopperheadOS](https://github.com/CopperheadOS/kernel_lge_bullhead) works for example.
* Try to revert that commit
* Try an older branch that does not have much fixes rebased, which possibly does not have this regression.

PureTryOut:
The CopperheadOS kernel gives the exact same problems sadly. I'll try an older commit next.

ollieparanoid:
Okay, thanks for trying out both of them. Regarding userland, do you mean the initramfs code? That could be, although it (the usb networking part specifically) seems to be working for the other devices we have.

Since it does not work with the LineageOS kernel either, my guess is that you'd need to write something to sysfs to enable the usb networking. @drebrez had this with the HTC desire and invented some creative debugging methods in order to find this one.

https://github.com/postmarketOS/pmbootstrap/blob/bd4ebb3473764108daa92b308511be7e0a1b6948/aports/device/device-htc-bravo/initfs-hook.sh#L7-L8

What I would try next:
* create an initfs hook with exactly that and see if it works (see `device-htc-bravo`)
* if it does not work, try to get a blinking LED and/or dumping your sysfs to a file, then comparing that with the sysfs of a working Android system. [initfs hook examples](https://wiki.postmarketos.org/wiki/Troubleshooting:boot:initfshooks)

PureTryOut:
Well... Something wen't wrong rebasing... But, it boots! Weston launched, touchscreen works, and even Plasma worked (al be-it slow). Haven't managed to get osk-sdl working though. When using FDE, it's just stuck at the pmOS splash screen. Using the debug hook and using telnet, it seems `cryptsetup` is not available at all. However, this can be fixed later and this seems to be good to merge!

aports/device/device-lg-hammerhead/deviceinfo
```diff
@@ -2,7 +2,7 @@
 # Please use double quotes only. You can source this file in shell scripts.
 
 deviceinfo_format_version="0"
-deviceinfo_name="Google Nexus 5X"
+deviceinfo_name="Google Nexus 5"
```
ollieparanoid: Thanks for fixing that, I always confuse these two! This will need a checksum update (and pkgrel bump to be consequent) in `device-lg-hammerhead` as well.

PureTryOut: It does not actually. Originally I only (accidentally) changed this variable, but I didn't actually change the checksum of the package. Running `pmbootstrap checksum` just now didn't cause any change :wink:

PR #1225, opened 2018-02-13
postmarketos-ui-i3wm

MartijnBraam:
This is a work in progress to add i3wm to the desktop environment list in pmbootstrap.

Since i3 is only usable if you have a hardware keyboard I also added a dependency on lxdm instead of starting it through an /etc/profile.d script since lxdm works very well on the n900 and i3 doesn't have a launch script that also sets up xorg like xfce4 has.


aports/device/device-nokia-n900/i3wm/i3wm.conf
```diff
@@ -0,0 +1,49 @@
...
+        bindsym h split h
+        bindsym v split v
+}
+
+# Enter command mode
+bindsym Shift+space mode "command"
+
+# This is the slidey button on the side
+bindsym XF86ScreenSaver exec 
+--no-startup-id lock.sh
```
ollieparanoid: new line?

MartijnBraam: Yep, shouldn't make commits on a 47 column wide terminal

ollieparanoid:
LGTM. Should we merge this as initial version?
(Also I'd be interested on your opinion about shipping a custom wallpaper, see above.)

ollieparanoid:
@MartijnBraam: I've tested this and it works with Qemu. [Added a wiki article.](https://wiki.postmarketos.org/wiki/I3wm)
However, the `device-nokia-n900` package doesn't build quite yet:
```
install: can't stat '/home/pmos/build/src/i3wm/i3wm.conf': No such file or directory
```

https://travis-ci.org/postmarketOS/pmbootstrap/builds/343456822#L1208

Could you please fix that? Apart from that I'd say it's ready to go.

PR #1226, opened 2018-02-13
[WIP] gp-peak display

Rudloff:
This PR makes the display work correctly on gp-peak.

But it still misses one thing: `/dev/fb0` has restrictive permissions that need to be changed.
If I manually run `sudo chmod 666 /dev/fb0` before weston starts, the screen works correctly.

What is the cleanest way to do this ?
I tried adding this udev rule:
```udev
KERNEL=="fb0", GROUP="video",  MODE="0660"
```
But it seems to be ignored when I reboot. However the rule is correctly applied when I run `sudo udevadm test /devices/virtual/graphics/fb0`. Am I doing something wrong?


ollieparanoid:
I guess a good way to fix this would be patching the kernel to give it the correct permissions for `/dev/fb0` by default. @pavelmachek: does this make sense, and do you happen to have an idea where we would need to look for that?

@Rudloff: #706 could fix the udev problem, if would be awesome if you could try that. Thanks for making the PR! Added the `dont_merge_pr` label, because you stated that this PR is WIP.

Rudloff:
I tried #706 and the udev rules still seem to be ignored.
However I'm not used to writing udev rules so I might be doing something wrong.

ollieparanoid:
I don't know much about udev either, but from reading the docs I would try this:
* `udevadm trigger`: does this execute the rule properly as well (when ran manually after boot)?
* what does the syslog say regarding udev? I think it logs there. You can read it with `logread`.
* it might help to crank up the [log level](https://wiki.archlinux.org/index.php/Udev#Debug_output) in the `udef.conf` file.

Rudloff:
`udevadm trigger` does not seem to actually trigger the rule.

And even with `udev_log="info"`, I can't find anything related to udev in the logs.
(Here are the full logs if you want to see for yourself: https://gist.github.com/Rudloff/c3a3e18db46d9b18c77b9de7c33b94f3)

pavelmachek:
Permissions stuff... take a look how it works on other boards, but I believe this should stay in udev and out of kernel...

ollieparanoid:
Thanks for trying it out! There seems to be nothing in the logfile related to udev, and after reading [some more docs](https://wiki.gentoo.org/wiki/Udev#Log_monitor_messages), it seems like the proper way to monitor udev activities is `udevadm monitor`:
* set `udev_monitor="YES"` in `/etc/conf.d/udev`
* `service udev restart`
* `udevadm monitor`
* second terminal: `service udev-trigger restart`

What does that give you?

Rudloff:
According to `udevadm monitor`, the devices are correctly detected:
```
KERNEL[134.287764] add      /devices/virtual/vc/vcs (vc)
KERNEL[134.287834] add      /devices/virtual/vc/vcs1 (vc)
...
KERNEL[180.981926] add      /devices/virtual/graphics/fb0 (graphics)
```
And when I run `udevadm trigger` it is correctly detected:
```
KERNEL[220.477148] change   /devices/virtual/graphics/fb0 (graphics)
```
But it does not output anything else than add/change events.

I also tried to set `udev_debug="YES"` but `/run/udevdebug.log` stays empty.

ollieparanoid:
Okay, thanks again for trying that as well. From what I can tell, udev not working seems to be an issue with your kernel. I've just booted my `i9100`, and there the permissions are set right by default by the `50-udev-default.rules` you've mentioned yourself (that's not really new information, but I wanted to check just to make sure).

```
localhost:~$ ls /dev/fb0 -lah
crw-rw----    1 root     video      29,   0 Jan  1 01:04 /dev/fb0
```

Since we don't want to invest too much energy into fixing downstream kernels (they are outdated anyway, ideally we want to the mainline kernel for all devices), I would recommend a simple workaround for this issue:
* [install an `/etc/rc.local.d/` script](https://wiki.gentoo.org/wiki//etc/local.d) in the device package, which sets the correct permissions for `/dev/fb0`
* enable the `local` rc-service (I'd modify the post-install script of `postmarketos-base` to do that, so other devices could add `local.d` scripts if they need them as well)

What do you think about that?

Rudloff:
Thanks for your help.

I also thought about doing that but I'm afraid we won't be able to make the touchscreen work without udev 😞

ollieparanoid:
Right ... that would be a bummer. Have you tried it though? Just to make sure that we're not worrying about something that is actually working.

For this PR I would suggest to do the `/etc/rc.local.d` hack unless you have an alternative.

The real solution to this problem would of course be [mainlining](https://wiki.postmarketos.org/wiki/The_Mainline_Kernel) if you're down to that (but we have not streamlined the process much yet, so that's a lot of work and probably only feasible if the peripherals of your device are already mainlined, so you would "only" need to enable them with a device tree file). That is obviously out of scope for such a PR, so I'd recommend we finish and merge this first.

Rudloff:
Yes, I tried the touchscreen and it does not work: 
```
Jan  1 01:59:59 localhost user.notice user:weston: [17:51:55.499] event2  - himax-touchscreen: not tagged as supported input device
Jan  1 01:59:59 localhost user.notice user:weston: [17:51:55.499] event2  - not using input device '/dev/input/event2'
```

But yes, you're right, having at least the display working is better than nothing. I will update my PR with this hack.

I will also give the mainline kernel a chance but I'm not sure I know enough about the linux kernel to make it work.

Rudloff:
@ollieparanoid Since `udevadm test` manually triggers the rules, I used it in the local script and now display and touchscreen work!

PR #1231, opened 2018-02-15
Wireguard for postmarketos-stable

MartijnBraam:
Packaging the wireguard out-of-tree module for linux-postmarketos-stable.

- [x]  Rewrite the upstream alpine APKBUILD enough for it to build in pmbootstrap
- [x]  Add required dependencies to the `linux-postmarketos-stable` kernel config

From the wireguard makefile:
```
DEPENDS:=+IPV6:kmod-udptunnel6 +kmod-udptunnel4 +kmod-ipt-hashlimit
```


MartijnBraam:
The kernel side works with this patch, the userspace side also works but  the wg-quick script in the alpine repo's is broken (but that's not really related to this PR)

aports/main/wireguard-postmarketos-stable/APKBUILD
```diff
@@ -0,0 +1,57 @@
...
+_kpkg=linux-$_flavor
+_kver=4.14.9
+
+# when changing _ver we *must* bump _mypkgrel
+# we must also match up _toolsrel with wireguard-tools
+_ver=0.0.20180202
+_toolsrel=0
+_name=wireguard
+
+# verify the kernel version before entering chroot
```
ollieparanoid: Looks like this check comes from [here](https://git.alpinelinux.org/cgit/aports/tree/testing/wireguard-vanilla/APKBUILD). With `pmbootstrap` we only copy the aport into the chroot, which we are currently building. This means, that the condition always returns false, and this is dead code for us.

As I understand it, the purpose of this check is not having an inconsistent aports folder, where the kernel module has another version as the kernel. How about we add a Python test for that (similar to `test/test_upstream_compatibility.py`) instead? That way we also notice it when only modifying the kernel APKBUILD.

MartijnBraam: I've removed the check from the APKBUILD

I think that a consistency check is a great idea but out of scope for this PR.
I've created #1235 for this.

aports/main/wireguard-postmarketos-stable/APKBUILD
```diff
@@ -0,0 +1,47 @@
+_flavor=postmarketos-stable
+_kpkg=linux-$_flavor
+_kver=4.14.9
+
+# we must match up _toolsrel with wireguard-tools
+_ver=0.0.20180202
+_toolsrel=0
+_name=wireguard
```
ollieparanoid: This variable is unused.

PR #1232, opened 2018-02-16
Rudimentary ZSH autocompletion helper

V13Axel:
For our users with ZSH, I wrote a (rather barebones) autocompletion script that will autocomplete `pmbootstrap` commands.

`pmbootstrap [tab]` will allow autocompletion selection of all the actions reported by `pmbootstrap -h`:

selection_187

Then, the directories in `aports/device/` will be autocompletions as a second argument, but only if the user enters menuconfig, checksum, aportgen, build, or kconfig_check:

selection_189

Lastly, `pmbootstrap flasher` will currently suggest `flash_kernel` and `flash_system`:

selection_190

There is still a TON of stuff that `pmbootstrap` does that isn't covered in the script, and ~it~ target device/kernel completion currently only works while $PWD is the pmbootstrap directory, but it's a start. Since it's zsh-only, it's not `ash` as the contribution guidelines mention.

I also wasn't sure where to put it, so I went with `/helpers`, because it's not part of the main code in `/pmb`, and none of the other directories seemed appropriate either.

### DIFF
--- a/helpers/pmbootstrap-autocompletion.zsh
+++ b/helpers/pmbootstrap-autocompletion.zsh
@@ -1,38 +1,39 @@
 #!zsh
-#	Installation: 
+# Installation:
 #
-#	Copy this file to ~/.zsh/ (create it, if it doesn't exist, or put it 
-#	somewhere that makes sense to you) Then, insert the following line 
-#	in your ~/.zshrc (making sure to use the right folder name, if changed):
+# Copy this file to ~/.zsh/ (create it, if it doesn't exist, or put it
+# somewhere that makes sense to you). Then, insert the following line
+# in your ~/.zshrc (making sure to use the right folder name, if changed):
 #
-#		source ~/.zsh/pmbootstrap-auto-completion.zsh
+#	source ~/.zsh/pmbootstrap-auto-completion.zsh
 #
-#	Then, set the variable PMBOOTSTRAP_DIR to your `pmbootstrap` root.
-#	Example:
+# Then, set the variable PMBOOTSTRAP_DIR to your `pmbootstrap` root.
+# Example:
 #
-#		PMBOOTSTRAP_DIR=/home/axel/Git/pmbootstrap
-#	
-#	This file is VERY incomplete, and should really only be trusted to
-#	autocomplete directory names at this time.
+#	PMBOOTSTRAP_DIR=/home/axel/Git/pmbootstrap
+#
+# This file is rudimentary, pmbootstrap actions and packages are autocompleted
+# so far. Further ideas for improvements are here:
+# <https://github.com/postmarketOS/pmbootstrap/pull/1232>
 
 PMBOOTSTRAP_DIR=
 
-_pmbootstrap_commands () 
+_pmbootstrap_commands()
 {
-	grep '^def ' $PMBOOTSTRAP_DIR/pmb/helpers/frontend.py | cut -d ' ' -f 2 | cut -d '(' -f 1 | grep -v '^_'
+	grep '^def ' $PMBOOTSTRAP_DIR/pmb/helpers/frontend.py | cut -d ' ' -f 2 \
+		| cut -d '(' -f 1 | grep -v '^_'
 }
 
-_pmbootstrap_targets ()
+_pmbootstrap_targets()
 {
 	case $1 in
-		checksum|aportgen|build)
-			find $PMBOOTSTRAP_DIR/aports/ -mindepth 2 -type d -printf  '%f\n' | sed "s|$PMBOOTSTRAP_DIR/aports/||g"
+		build|checksum|pkgrel_bump)
+			find $PMBOOTSTRAP_DIR/aports/ -mindepth 2 -maxdepth 2 -type d \
+				-printf '%f\n' | sed "s|$PMBOOTSTRAP_DIR/aports/||g"
 			;;
-		kconfig_check)
-			ls -1 $PMBOOTSTRAP_DIR/aports/device/
-			;;
-		menuconfig)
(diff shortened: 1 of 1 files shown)


helpers/pmbootstrap-autocompletion.zsh
```diff
@@ -0,0 +1,50 @@
...
+_pmbootstrap_commands() 
+{
+    pmbootstrap -h | awk 'c&&!--c;/action:/{c=1}' | sed -e 's/{//g;s/}//g;s/,/ /g'
+}
+
+_pmbootstrap_targets()
+{
+    case $1 in
+        menuconfig|checksum|aportgen|build|kconfig_check)
+            \ls -1 $PWD/aports/device/
```
MartijnBraam: For the `checksum`, `build` and `aportgen` this should probably include all the subdirs in `$PWD/aports`

helpers/pmbootstrap-autocompletion.zsh
```diff
@@ -0,0 +1,50 @@
...
+#
+#       source ~/pmbootstrap-auto-completion.zsh
+# 
+# This file is VERY incomplete, and should really only be trusted to
+# autocomplete aports/ directory names.
+
+
+_pmbootstrap_commands() 
+{
+    pmbootstrap -h | awk 'c&&!--c;/action:/{c=1}' | sed -e 's/{//g;s/}//g;s/,/ /g'
```
ollieparanoid: All actions that can be specified as `pmbootstrap` commands have functions with the same name in `pmb/helpers/frontend.py` (this is literately how the program works - for all actions we call a function from that file). We could use this fact to get the list of actions without calling `pmbootstrap` and thus making it faster (you may not notice this on a fast machine, but `pmbootstrap -h` does not feel like it's instantly returning the help output on my laptop).

So how about we use something like that?
```shell
cat frontend.py | grep '^def ' | cut -d ' ' -f 2 | cut -d '(' -f 1 | grep -v '^_'
```

helpers/pmbootstrap-autocompletion.zsh
```diff
@@ -0,0 +1,50 @@
...
+    _arguments -C \
+        '1: :->command'\
+        '2: :->target'
+    
+        case $state in
+            command)
+                compadd `_pmbootstrap_commands`
+                ;;
+            target)
+                if [ -f $PWD/pmbootstrap.py ]; then #We must be in the pmbootstrap dir
```
ollieparanoid: What do you think about putting the `pmbootstrap` folder into a variable, so it works even if the user is not in that directory? (Is setting environment variables bad practice when writing shell completions?)

helpers/pmbootstrap-autocompletion.zsh
```diff
@@ -0,0 +1,50 @@
...
+            ;;
+    esac
+}
+
+_pmbootstrap ()
+{
+	local curcontext="$curcontext" state line
+	typeset -A opt_args
+
+    _arguments -C \
```
ollieparanoid: Tabs and spaces mixed, please be consistent throughout the file. We use tabs in most shell scripts (especially in all `APKBUILD` files since that is how Alpine does it), so I'd recommend tabs for this file. We use spaces for Python files (to follow PEP8 guidelines).

Also there are some minor syntax inconsistencies in this file: `_pmbootstrap ()` vs `_pmbootstrap()`, sometimes there are spaces at the end of the line, and there's no new line at the end of the file yet.

helpers/pmbootstrap-autocompletion.zsh
```diff
@@ -0,0 +1,50 @@
+#!zsh
+# Installation: 
+#   Copy this file to ~/
```
ollieparanoid: How about recommending to copy into `~/.zsh/` instead, so the home folder does not get cluttered?

V13Axel: Great idea, though `~/.zsh` doesn't always exist (such as on my machine). I'll have it say `Copy this file to ~/.zsh/ (create it, if it doesn't exist, or put it somewhere that makes sense to you)` instead. Cool?

helpers/pmbootstrap-autocompletion.zsh
V13Axel: Parsing the frontend file sounds great! I've implemented it on my machine and will update the PR soon.

helpers/pmbootstrap-autocompletion.zsh
V13Axel: Thanks for pointing that out! Still learning what everything needs.

helpers/pmbootstrap-autocompletion.zsh
V13Axel: My apologies! I was switching between VSCode and vim while developing this, a couple of times. I'll make sure that's cleaned up.

helpers/pmbootstrap-autocompletion.zsh
V13Axel: It doesn't look like there are any major reasons not to put it in a variable, and I don't see a whole lot of "best practices" written anywhere. While I don't necessarily like cluttering up the env, a single variable shouldn't be an issue.

Especially if the goal is to, someday, have this autocompletion be generated by `pmbootstrap` itself. At that point, it wouldn't need to be in the env at all, so using an env var for right now isn't a big deal IMO.

V13Axel:
As mentioned in the commit message, I've pushed a change that includes:

PMBOOTSTRAP_DIR variable, so autocomplete can work from anywhere
Parse frontend.py instead of `pmbootstrap -h` for action autocompletion
Consistency between tabs/spaces and command definition syntax
Better suggestions for checksum, aportgen, build, and menuconfig

V13Axel:
Looks great to me!

Issue #1218, opened 2018-02-10
make kconfig_check arch-specific to support CONFIG_LBDAF


ollieparanoid:
The [kconfig option `LBDAF`](https://wiki.postmarketos.org/wiki/Kernel_configuration#CONFIG_LBDAF) is required to mount `ext4` partitions (unless they have been created with non-default settings). But that option is only present in 32 bit (armhf, x86), not 64 bit (aarch64, x86_64) kernels. So if we added it to `kconfig_check` right now, it would incorrectly report the 64 bit device's kernels to be missing that option.

@MartijnBraam pointed out in #postmarketOS, that we could make the options we check for arch specific to resolve this.

The options we are checking for are defined in `pmb/config/__init__.py`:
```py
necessary_kconfig_options = {
    "ANDROID_PARANOID_NETWORK": False,
    "DEVTMPFS": True,
    ...
}
```

One approach would be to group these options:
```py
necessary_kconfig_options = {
    "all": {"ANDROID_PARANOID_NETWORK": False,
            "DEVTMPFS": True, ...},
    "armhf x86": {"LBDAF": True}
}
```

Polynomdivision:
I am currently working on this. Will update this issue (and hopefully create a PR), when I am done.

Polynomdivision:
@ollieparanoid I have created a PR (#1233). However, as it is written in the code, only the first architecture
listed under `arch` in the APKBUILD will be used to check the cofig options.
For me, the question that arises is whether a `linux-<device>` package can have multiple architectures. My guess would be "Yes", as the project is aiming for having a single kernel for every device, but asking never hurts 😄

MartijnBraam:
Hmm that means that `kconfig_check` gets an `--arch` option since the linux packages can definitely contain multiple archs

ollieparanoid:
Yes, one kernel can have multiple architectures, as in `aports/main/linux-postmarketos-*`. But it should work without an extra `--arch` parameter, see https://github.com/postmarketOS/pmbootstrap/pull/1233#discussion_r168916828.

Polynomdivision:
I pushed a new commit, which, as written in the comment under the PR, now extracts the architecture from the name of the config file, instead of getting it from the APKBUILD.

PR #1233, opened 2018-02-16
Check kernel config options based on the architecture

Polynomdivision:
Do not merge (yet).

This PR is supposed to solve #1218, which allows pmbootstrap to decide on which kernel config
options are to be checked, based on the `arch` option in the `APKBUILD` of a `linux-<device>` package.
The options under the `all` key are applied to every architecture. Using a key like `"armhf x86"` would result in the kernel config options being checked for both armhf and x86.

### DIFF
--- a/pmb/parse/kconfig.py
+++ b/pmb/parse/kconfig.py
@@ -23,6 +23,7 @@ import os
 
 import pmb.build
 import pmb.config
+import pmb.parse
 
 
 def is_set(config, option):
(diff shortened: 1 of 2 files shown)


pmb/parse/kconfig.py
```diff
@@ -58,25 +59,36 @@ def check(args, pkgname, details=False):
...
-                    logging.info("WARNING: " + path + ": CONFIG_" + key + " " +
-                                 should + " be set. See <" + link +
-                                 "> for details.")
-                else:
-                    logging.warning("WARNING: " + path + " isn't configured"
-                                    " properly for postmarketOS, run"
-                                    " 'pmbootstrap kconfig_check' for"
-                                    " details!")
-                    break
+        apkbuild = pmb.parse.apkbuild(args, aport + "/APKBUILD")
```
ollieparanoid: In this loop, we are only looking at one config file inside the aport, see this line above:
```python
for config_path in glob.glob(aport + "/config-*"):
```
The `config-*` files already have the architecture as file extension, so we can simply look at that and get the right one. For example:
```shell
$ cd aports/main/linux-postmarketos-mainline
$ ls config-*
config-postmarketos-mainline.aarch64  config-postmarketos-mainline.armhf  config-postmarketos-mainline.x86_64
```

Issue #1223, opened 2018-02-12
Gracefully handle missing /proc/sys/fs/binfmt_misc


ollieparanoid:
From [the wiki](https://wiki.postmarketos.org/wiki/Troubleshooting#sh:_can.27t_create_.2Fproc.2Fsys.2Ffs.2Fbinfmt_misc.2Fregister:_nonexistent_directory):
> ##  sh: can't create /proc/sys/fs/binfmt_misc/register: nonexistent directory
> This issue is known to happen in Alpine Virthardened. It also happened in Void Linux for one user.  These commands should fix it:
>
>`sudo modprobe binfmt_misc`
>`sudo mount -t binfmt_misc none /proc/sys/fs/binfmt_misc`
>
>If it still does not work, it is probably not supported by your host Linux distribution's kernel (`CONFIG_BINFMT_MISC`). Try to use another kernel. 

While this is on the troubleshooting page, it isn't really obvious and people get stuck with this error. It would be nice if we checked if `/proc/sys/fs/binfmt_misc` existed, and if it did not, printed a link that points to the right location in the troubleshooting wiki.

Polynomdivision:
I created a PR (#1236) that adds a check at the start, so that the user will be warned (and pointed to the Wiki), if binfmt_misc is not loaded.
I am, however, not sure if the start is the right place for this check or if the phrasing of the message would be fitting.

PR #1236, opened 2018-02-17
Add a check for binfmt_misc

Polynomdivision:
Do not merge (yet).

To solve #1223, I added a check at the start of pmbootstrap, that looks for `/proc/sys/fs/binfmt_misc` and prints a message to the user pointing him to the Wiki, if the path does not exist.

### DIFF
--- a/pmb/__init__.py
+++ b/pmb/__init__.py
@@ -39,6 +39,7 @@ def main():
     try:
         # Sanity checks
         other.check_grsec(args)
+        other.check_binfmt_misc(args)
         if not args.as_root and os.geteuid() == 0:
             raise RuntimeError("Do not run pmbootstrap as root!")
 
--- a/pmb/__init__.py
+++ b/pmb/__init__.py
@@ -39,7 +39,6 @@ def main():
     try:
         # Sanity checks
         other.check_grsec(args)
-        other.check_binfmt_misc(args)
         if not args.as_root and os.geteuid() == 0:
             raise RuntimeError("Do not run pmbootstrap as root!")
 
--- a/pmb/helpers/other.py
+++ b/pmb/helpers/other.py
@@ -71,12 +71,11 @@ def check_binfmt_misc(args):
     if os.path.exists(path):
         return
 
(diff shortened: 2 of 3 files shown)


pmb/__init__.py
```diff
@@ -39,6 +39,7 @@ def main():
     try:
         # Sanity checks
         other.check_grsec(args)
+        other.check_binfmt_misc(args)
```
ollieparanoid: We could move it here instead (between the `return` and install command):
https://github.com/postmarketOS/pmbootstrap/blob/3d35b45835c961e64b6399c69a9aab74f5cc0093/pmb/chroot/binfmt.py#L36-L38

That way it only runs when the `binfmt_misc` feature is used (e.g. when cross-compiling most packages, or when doing a installation to a foreign arch). There's a case when the `binfmt_misc` module is not needed: when building packages for the host architecture, or running the QEMU VM with the host arch.

pmb/helpers/other.py
```diff
@@ -61,6 +61,25 @@ def check_grsec(args):
...
+    Check if the 'binfmt_misc' module is loaded by checking, if
+    /proc/sys/fs/binfmt_misc/ exists. If it exists, then do nothing.
+    Otherwise, raise an exception pointing to user to the Wiki.
+    """
+    path = "/proc/sys/fs/binfmt_misc/"
+    if os.path.exists(path):
+        return
+
+    link = "https://wiki.postmarketos.org/wiki/Troubleshooting#sh:_can.27t_create_.2Fproc.2Fsys.2Ffs.2Fbinfmt_misc.2Fregister:_nonexistent_directory"
+    raise RuntimeError("It appears that your system has not loaded the"
```
ollieparanoid: How about:
> It appears that your system has not loaded the module 'binfmt_misc'. This is required to run foreign arch programs with QEMU (e.g. armhf on x86_64): `<https://postmarketos.org/binfmt_misc>`

(Just created that shorter URL with https://github.com/postmarketOS/postmarketos.org/commit/9d2769986cf3b12206d1729ea24280c56441958b)

Polynomdivision:
@ollieparanoid Well, I think that pmOS is an extremely cool project that I would love to see succeed, so I try to help out, where I (with my mediocre programming skills) can!

In the new commits, I changed up the error message of the exception raised by `check_binfmt_misc(...)` and moved the check away from `pmb/__init__.py` to `pmb/chroot/binfmt.py`.

Issue #1217, opened 2018-02-10
New initramfs hook that uses LEDs etc. to let us know pmOS booted (when networking and screen don't work)


ollieparanoid:
This idea was brought up the other day in #postmarketOS: An initramfs hook that tries to notify the user with alternative means that the kernel is actually booted:
* let the LEDs blink
* use the vibration motor
* use the flashlight
* ...

When the device does not respond over USB, and the screen does not work either, this would be an easy way to verify that postmarketOS has booted on the device.

See also: [Troubleshooting:boot - Does the kernel boot?](https://wiki.postmarketos.org/wiki/Troubleshooting:boot#Does_the_kernel_boot.3F)

MayeulC:
I can give it a go, I would need it right about now.

eMPee584:
Just for inspiration, I have something similar on my debianized qwerty android phone.. from my `/system/etc/init.d/99debian` (a lil' cleaned):

```
# blink the sh*t until boot complete
echo 1 > /sys/class/misc/backlightnotification/enabled
{ 
  while [ ! "$(getprop sys.boot_completed)" == "1" ] || [ ! "$(getprop init.svc.bootanim)" == "stopped" ] ; do
    echo 1 > /sys/class/misc/backlightnotification/notification_led
    sleep 0.05
    echo 0 > /sys/class/misc/backlightnotification/notification_led
    sleep 1
  done
  vibrate 5 200 0.6
  echo 0 > /sys/class/misc/backlightnotification/enabled
} &

```

MayeulC:
Thanks @eMPee584 for your input. It looks like the values you have here would be incompatible with the script I wrote. Are you running a mainline kernel? I thought every LED would be accessed trough the standard LED API (although I haven't been able to turn on the notification led again on my phone). This one includes a `brightness` file descriptor, as well as a `max_brightness` one, as far as I know.

The vibrate command is interesting as well... To be fair, I would be interested in peering at your sysfs tree :blush:

eMPee584:
yeah, the actual LED access is definitely all non-standard and OEM dependent.. what I deemed useful to show is the outer backgrounding loop.. but don't let it get in the way of spinning it your own preferred style :+1: 
my `/system/xbin/vibrate` very simply contains

```
#!/system/bin/sh
for duration in $@; do
  echo ${duration} > /sys/devices/virtual/timed_output/vibrator/enable
  sleep 0.1
done
```

eMPee584:
[...]

PR #1238, opened 2018-02-19
Add a maximum-attention hook to help with debugging boot issues

MayeulC:
The hook flashes every LED it can find, as well as the vibration motor.
More paths are probably necessary to support more devices and kernels.

This is intended as a first approach to #1217
Also somewhat related to #88 

The name comes from ollieparanoid on Matrix, IIRC.

### DIFF
--- a/aports/main/postmarketos-mkinitfs-hook-maximum-attention/00-maximum-attention.sh
+++ b/aports/main/postmarketos-mkinitfs-hook-maximum-attention/00-maximum-attention.sh
@@ -1,0 +1,56 @@
+#!/bin/sh
+. ./init_functions.sh
+
+BLINK_INTERVAL=2 # seconds
+VIBRATION_DURATION=400 #ms
+VIBRATION_INTERVAL=2 #s
+
+find_leds() {
+	find /sys -name "max_brightness" | xargs -I{} dirname {}
+}
+
+find_vibrator() {
+	echo /sys/class/timed_output/vibrator
+}
+
+
+# blink_leds takes a list of LEDs as parameters,
+# it iterates over every LED, and changes their value,
+# alternating between max_brightness and 0 every BLINK_INTERVAL
+blink_leds() {
+	state=false # false = off, true=on
+	while true; do
+		for led in $@; do
+			if [ "$state" = true ]; then
+				cat $led/max_brightness > $led/brightness
+			else
+				echo 0 > $led/brightness
+			fi
+			echo blinking LED: $led
+		done
+		sleep ${BLINK_INTERVAL}s
+		if [ "$state" = true ]; then
+			state=false
+		else
+			state=true
+		fi
+	done
+}
+
+# vibrate_loop vibrates each VIBRATION_INTERVAL for VIBRATION_DURATION
+# it takes a timed_device path to the vibrator as $1
+vibrate_loop() {
+	if [ ! -f $1/enable ]; then
+		return;
+	fi
+
+	while true; do
+		echo $VIBRATION_DURATION > $1/enable
+		sleep ${VIBRATION_INTERVAL}s
+	done
+}
+
+blink_leds $(find_leds) &
+vibrate_loop $(find_vibrator) &
+
+loop_forever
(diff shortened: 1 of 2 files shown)


aports/main/postmarketos-mkinitfs-hook-maximum-attention/00-maximum-attention.sh
```diff
@@ -0,0 +1,62 @@
...
+
+
+# blink_leds takes a list of LEDs as parameters,
+# it iterates over every LED, and changes their value,
+# alternating between max_brightness and 0 every BLINK_INTERVAL
+blink_leds() {
+        state=false # false = off, true=on
+        while :
+        do
+
```
MayeulC: OK, there might be a stray whiteline here

MartijnBraam:
Hmm it doesn't seem to work with the notification led on my hammerhead, but that led has a slightly more complicated sysfs interface. To turn that one do:

```bash
cat max_brightness > brightness
echo 100 0 > on_off_ms
# You have to write a different value to rgb_start for every change
echo 1 > rgb_start
```
You can also make the kernel blink the led instead of the script by setting on_off_ms to the blinking times

```bash
echo 2000 2000 > on_off_ms
```

I guess you can just check if the current led has a rgb_start file to do this procedure before entering the blinking loop

MayeulC:
Thanks a lot for your review, @MartijnBraam. THere are a few things there (hence why I called it a first approach):
* The LCD backlight should flash as well (if applicable), but I am not sure whether I should add a new splash to make it more visible
* currently, the hook loops forever. I am not sure if it should continue the boot process or not, I think it is better this way, as it is just intended as a one-off to diagnose boot issues
* There seem to be a lot of different vendor interfaces out there. I didn't really want to support non-standard ones, but I guess it would make the porting process easier.

I didn't really understand the purpose of `rgb_start` here, though.

drebrez:
@MayeulC when I get home I will try on the various devices I have and let you know the results

MartijnBraam:
The backlight goes to full brightness and then turns of, that is probably a sideeffect of some fbdev code in the kernel but the led does get controlled once.

Also the rgb led is slightly weird but the `rgb_start` file is to make the new settings apply all at once since you have to change 2 (or more) settings and accros 3 led devices usually if you wan't to flash a non-primary color. 

I think this hook should support as many leds as possible since it is for debugging booting issues.

drebrez:
Tested and here are the results:

| device         | display backlight | button backlight | front led | camera led | vibrator |
|----------------|-------------------|------------------|-----------|------------|----------|
| samsung-i9070  | NO                | YES              | -         | NO         | YES      |
| samsung-maguro | YES               | NO               | NO        | NO         | YES      |
| htc-bravo      | YES               | YES              | YES       | YES        | YES      |
| samsung-klte   | YES               | YES              | YES       | YES        | YES      |
| huawei-y530    | NO                | YES              | YES       | NO         | YES      |

aports/main/postmarketos-mkinitfs-hook-maximum-attention/00-maximum-attention.sh
```diff
@@ -0,0 +1,58 @@
...
+BLINK_INTERVAL=2 # seconds
+VIBRATION_DURATION=400 #ms
+VIBRATION_INTERVAL=2 #s
+
+find_leds() {
+	find /sys -name "max_brightness" | xargs -I{} dirname {}
+}
+
+find_vibrator()
+{
```
ollieparanoid: How about `find_vibrator() {` for consistency?

aports/main/postmarketos-mkinitfs-hook-maximum-attention/00-maximum-attention.sh
```diff
@@ -0,0 +1,58 @@
...
+			state=false
+		else
+			state=true
+		fi
+	done
+}
+
+# vibrate_loop vibrates each VIBRATION_INTERVAL for VIBRATION_DURATION
+# it takes a timed_device path to the vibrator as $1
+vibrate_loop()
```
ollieparanoid: same here

aports/main/postmarketos-mkinitfs-hook-maximum-attention/00-maximum-attention.sh
MayeulC: Yes, you are right, I am just *soo* used to my usual C syntax that I tend to mix things up. Thanks for noticing!

MayeulC:
@ollieparanoid, I fixed the formatting issue (thanks for noticing! I have C-style formatting hardcoded in my fingers).

I think this should get merged, I will (maybe with some others) think about a way to add more device in a way that's as generic as possible, but will probably be busy for a few days.

PR #1242, opened 2018-02-21
Patch KWIN to support amazon-thor

MoreRobustThanYou:
```amazon-thor```'s framebuffer uses some weird framebuffer format, so I updated the kwin patch to support it.

Touch doesn't work?


MoreRobustThanYou:
Logread: https://pastebin.com/rVdvJxzg

PureTryOut:
Not sure if I'm a fan of editing the original patch. Could you try to add it as a separate patch instead? Because I'm assuming your patch doesn't add support for "RGB32 framebuffer with BGR order" but rather some other format.

MoreRobustThanYou:
K

MoreRobustThanYou:
Didn't mean to close this, woops.

bhush9:
@MoreRobustThanYou have you confirmed that it works on your device and colors are correct?

MoreRobustThanYou:
yes

MartijnBraam:
Its very weird that we have two Format_RGBA8888 blocks in the fbdev driver, with one being RGBA and one being actually ABGR, and you don't check for the alpha channel offset.

If the colors are correct with this patch it means that the driver expects RGBA but advertises ABGR in which case the fbdev driver needs a patch (like hammerhead)

ollieparanoid:
@MoreRobustThanYou: thanks for making the PR!

What @MartijnBraam sounds logical though, what happens when you use the [06_fix_mdss_fb_rgb_mode.patch](https://wiki.postmarketos.org/wiki/Troubleshooting:display#Kernel_patch) from hammerhead? (I think you've mentioned at some point in IRC that you tested the patch already to fix X11?)

MoreRobustThanYou:
img_20180223_122439_1

MartijnBraam:
Yep the colors are incorrect, see the settings icon in this image: https://wiki.postmarketos.org/wiki/File:Galaxy-nexus-plasma.jpg

astralloop:
@MartijnBraam I will note that pic was taken with the kernel patch, not the kwin patch

MoreRobustThanYou:
@MartijnBraam I will note that pic was taken with the kernel patch, not the kwin patch

ollieparanoid:
Closing, as you've fixed the kernel in #1252.
@MoreRobustThanYou: thanks for your work on this!

PR #1243, opened 2018-02-21
gp-peak: Add support for osk-sdl and audio

Rudloff:
I used the mode from [the wiki](https://wiki.postmarketos.org/wiki/Osk-sdl#Create_fb.modes) and only adjusted the resolution.


Rudloff:
Sorry about that. I admit I find it a bit confusing to have two different folders for each device (even if I understand the reasons behind it).
I moved the files and merged the two PRs.

ollieparanoid:
> I find it a bit confusing to have two different folders for each device

Yeah I agree, it is confusing. When more and more devices will be able to use the mainline kernel, then the extra kernel package per device is gone and this gets easier ;)


Code looks good, I've rebased the PR on master and will merge this when Travis runs through.

Thanks a lot!

PR #1245, opened 2018-02-21
New device (Jolla Phone)

PureTryOut:
Kernel compiles, it flashes (kernel and sd card, @MartijnBraam reported flashing the system doesn't work) and boots. However, no mac address gets assigned so you have to do that manually, and the screen doesn't work. The touchscreen does detect presses though.


PureTryOut:
Regarding network: no, as the commands I use to connect are used on the PC, not on the phone.

Regarding display: I used the method you described, seems useful. I found a few screen related differences (mainly in `/sys/devices/virtual/graphics/fb0/pan` and `/sys/devices/virtual/graphics/fb0/virtual_size`). However, applying the settings in a `initfs-hook.sh` script placed in `/etc/postmarketos-initfs/` didn't seem to do much useful. Interestingly, starting debugging today, the postmarketOS splash appeared a few times. It now stopped doing that for some reason though, I'm not at all sure what's happening.

ollieparanoid:
@sjamaan: back in July 2017 you've [documented in the wiki](https://github.com/postmarketOS/pmbootstrap/wiki/Boot-process/_compare/d94a4f369217d96a208a5dba06d55de128c62d49...40980ed4b9172a8caa97c8d645186d1e2ccad8eb) this information:
> If this fails, you may need to first assign an ethernet address to it [...]

Do you happen to remember if we had a fix for that, which did *not* require the MAC address to be set for the ethernet device?

PR #1246, opened 2018-02-21
gp-peak: Add audio support

Rudloff:
Hello,

This PR makes audio work on the gp-peak. (It works without this patch but only as root.)

The `postmarketos-base.post-install` change might help other devices as well.


Rudloff:
I'm merging this PR with #1243.

PR #1247, opened 2018-02-21
Don't mark all packages as explicitly installed

### LINKED ISSUE #1212: All packages installed by pmbootstrap have dependencies that can't get automatically removed
@drebrez wrote in https://github.com/postmarketOS/pmbootstrap/issues/1188#issuecomment-363605460:
> I've tried adding the `fbdebug` package to the dependencies of the `postmarketos-mkinitfs-hook-debug-shell`, but when I run `pmbootstrap initfs hook_del debug-shell` it doesn't uninstall the `fbdebug` package from the chroot.

The thing is, `pmbootstrap` resolves all dependencies on its own, and then explicitly lists all packages that should be installed in the `apk add` command. That means, when one wants to uninstall a certain package, all dependencies that have been installed by `pmbootstrap` are still marked as explicitly installed and do not get automatically removed.

ollieparanoid:
pmbootstrap does dependency resolving on its own, and passes the list
of resolved packages to apk when we want it to install something. The
reason was outlined in #129:

> fixing #120: packages do not get updated in "pmbootstrap install"
> after they have been rebuilt. For this to work, we specify all
> packages explicitly for abuild, instead of letting abuild do the
> resolving.

This new PR fixes #1212 (which noted that all of these dependencies
were explicitly marked for installation) by doing the following:
1. All packages and dependencies get attached to the virtual package
  .pmbootstrap instead of world
2. We install the packages (without depends) explcitly
3. .pmbootstrap gets removed, which means that all packages from 1.
  stay installed, but are no longer marked as explicitly installed.
  They will get removed automatically, when the depending packages get
  removed.

In addition, the mechanism for replacing the package of locally built
packages with their full path, was broken and has been fixed in this
commit. This is necessary to update packages of the same version with
apk.


ollieparanoid:
Alright, [Travis fails](https://travis-ci.org/postmarketOS/pmbootstrap/builds/344526696#L2132-L2133) because:

```
var/cache/misc/postmarketos-base-3-r12.post-install: .: line 3: can't open '/etc/deviceinfo'
ERROR: postmarketos-base-3-r12.post-install: script exited with error 2
```

Which means, `postmarketos-base` currently expects the device package to be installed before itself. I've thought about this, and it makes more sense to me if the device packages depend on `postmarketos-base`, not the other way around. Because that's the only way we can put the dependency information properly in the aports. Please tell me if someone sees this differently.

I'll prepare a new PR soon for that change.

z3ntu:
Couldn't we do that in the pmbootstrap code? Because imo as this dependency has to be applied to ALL packages and there isn't any "alternative" to this package (compared to the dri-swrast or so) I think it could be added/installed in the pmb code. (hope I'm clear :slightly_smiling_face: )

ollieparanoid:
`apk` doesn't only use this information during the initial installation (where the Travis error happened above), but also during updates to determine in which order post-update scripts etc get executed. So if we give it the right dependency information, we know for sure that it will execute those in the right order.

So besides the general feeling of this being a cleaner approach (giving apk the right information vs. hacking the installation order in pmbootstrap), this is a real reason where we can run into problems. Does that make sense?

z3ntu:
Yeah okay :)

ollieparanoid:
Thanks for testing @drebrez!

PR #1256, opened 2018-02-24
qemu: allow the user to specify the kernel flavor

z3ntu:
Suggestion for a better help text?


pmb/parse/arguments.py
```diff
@@ -111,6 +111,7 @@ def arguments_qemu(subparser):
                      help="guest RAM (default: 1024)")
     ret.add_argument("-p", "--port", type=int, default=2222,
                      help="SSH port (default: 2222)")
+    ret.add_argument("--flavor", help="kernel flavor to use for booting qemu")
```
ollieparanoid: You could use that one (also used by `pmbootstrap initfs build` etc.):
> name of the kernel flavor (run 'pmbootstrap flasher list_flavors' to get a list of all installed flavors)

z3ntu: updated it. I hope the indentation is correct :)

PR #1258, opened 2018-02-24
device packages: depend on postmarketos-base

ollieparanoid:
* Add `postmarketos-base` as dependency for all `device-*` packages (current and new ones generated with `aportgen`)
* `postmarketos-base`: don't source `/etc/deviceinfo` in `post-install` anymore (it was used to enable the `msm-fb-refresher` service depending on the `deviceinfo`)
* `msm-fb-refresher`: enable the service in `post-install`
* Add a test case to `test/test_aports_deviceinfo.py` and rename it to `test/test_aports.py`

This is required for #1247 "Don't mark all packages as explicitly installed" (see also: the discussion in that thread).


ollieparanoid:
Thanks for the reviews guys! :tada:

PR #1261, opened 2018-02-25
Add nonfree-firmware subpackage for fp2

z3ntu:


aports/device/device-fairphone-fp2/APKBUILD
```diff
@@ -8,7 +8,7 @@ license="MIT"
 depends="linux-fairphone-fp2 linux-firmware mkbootimg mesa-dri-swrast"
```
ollieparanoid: What is `linux-firmware` needed for, and why don't we move that to the `nonfree_firmware` subpackage as well?

Also the `linux-firmware` package has been split up in multiple packages, so you can specify which parts of it are needed instead of the full package. This reduces the installation size a lot.

https://pkgs.alpinelinux.org/packages?name=linux-firmware-*&branch=edge&repo=main&arch=armhf

z3ntu: I actually don't know why I have the `linux-firmware` package in there. I guess I copy-pasted it from another device... Do you know what it might be needed for on the fp2? :D Otherwise I'll remove it.

ollieparanoid: I don't, so I would remove it for now, and add it back on demand. :smile:

PR #1264, opened 2018-02-25
New device: Moto G 4G (2013)

pikpok:
This device is very similar to falcon and titan, so most of the work here was just copy-paste. 

As of now Weston with XWayland and WiFi, I've yet to try FDE.
XFCE and Hildon were two Xorg distros that I've tried and found not working (XFCE segfaults), but `06_fix_mdss_fb_rgb_mode.patch` doesn't seem to fix the issue.

I also wanted to reuse `firmware-motorola-titan` package because those devices seem to share firmware - however, I needed one more file included in `prima` folder and I added it in this PR. I don't know if that's a good practice, or maybe package should be renamed to be more common, or should I move it to separate firmware package?

I will create wiki page describing port later.


aports/device/linux-motorola-peregrine/02_gpu-msm-fix-gcc5-compile.1.patch
```diff
@@ -0,0 +1,33 @@
+source:
```
PureTryOut: You seem to have this file 2 times, please remove one of them.

PureTryOut:
Uhm, ignore that review. You already removed what I was reviewing :wink:

aports/device/linux-motorola-peregrine/02_gpu-msm-fix-gcc5-compile.1.patch
PureTryOut: You seem to have this file twice, please remove this one.

aports/firmware/firmware-motorola-titan/APKBUILD
```diff
@@ -1,6 +1,6 @@
 pkgname=firmware-motorola-titan
 pkgver=1
-pkgrel=0
+pkgrel=1
 pkgdesc="Firmware files for Motorola Moto G 2014 (titan)"
```
ollieparanoid: How about noting in the `pkgdesc` that this is also for the Moto G 4G now?

Other than that we don't have strict guidelines for the firmware packages yet, and I think it's fine to use the `motorola-titan` package in this case. Here it does not get blown up much by the additional file, and it saves us the redundancy of introducing another firmware package that has almost the same content.

Sposito:
Hey I am trying here on mine it stucks on a black screen(mine is uk model), anybody else got it successful on peregrine?

pikpok:
Hi @Sposito, for me it was working fine, maybe try to ping me on Matrix/IRC and I'll try to help, closed PR isn't a good place for this IMHO :slightly_smiling_face: https://wiki.postmarketos.org/wiki/Matrix_and_IRC

ollieparanoid:
@Sposito: this was probably caused by #1364 (new version of `mkfs.ext4` enabled a flag by default which some kernels couldn't handle), which is fixed now. If you pull the code from master and run the installation again, then it will probably work :+1:

PR #1268, opened 2018-02-26
Make nonfree depends optional (2/2): device pkgs

### LINKED ISSUE #756: Make proprietary firmware and userspace optional

ollieparanoid:
* Add nonfree_firmware subpackage to all devices, that depend on
  nonfree firmware.
* Some packages were depending on `linux-firmware`, but without having
  Wifi working. Removed that dependency as it was probably added by
  accident. If it was really necessary, chosing the appropriate
  split linux-firmware package (e.g. linux-firmware-brcm) is better
  anyway (that has been changed recently in Alpine and is possible
  now).
* Add a test case that makes sure we don't have firmware depends
  without subpackages in device aports anymore.

Almost finishes #756.


ollieparanoid:
Thanks a lot for the quick review!

I agree that it is some additional maintenance effort to keep that field updated. But it will be shown to the user, e.g. for the fairphone-fp2 `pmbootstrap init` looks like the following (this is already merged in `master`):
> Device [qemu-amd64]: fairphone-fp2
This device has proprietary components, which trade some of your freedom with making more peripherals work.
We would like to offer full functionality without hurting your freedom, but this is currently not possible for your device.
device-fairphone-fp2-nonfree-firmware: WCNSS firmware (WiFi, Bluetooth, FM)
Enable this package? (y/n) [y]:

I would prefer if we had one firmware package per peripheral, not one firmware package per device as we have now in some cases. That way it would be easier to keep track of security updates, and it would make it easier to fill out the `pkgdesc` variable (because we see directly which peripheral-firmware the `device-...-nonfree-firmware` subpackage depends on).

So with that in mind, do you think it is okay?

PR #1274, opened 2018-02-28
Move aports/firmware to aports/nonfree-firmware

### LINKED ISSUE #756: Make proprietary firmware and userspace optional

ollieparanoid:
Added a test case as well. When there's still an aport in the old
folder, it errors with something like:

>  The 'aports/firmware' folder has been renamed to
  'aports/nonfree-firmware'. Please move 'firmware-asus-flo',
  'firmware-adreno' to the 'aports/nonfree-firmware' folder.

Closes #756.


z3ntu:
What's the advantage of having a different folder name? Isn't all firmware proprietary normally?

ollieparanoid:
Thanks for asking! In https://github.com/postmarketOS/pmbootstrap/issues/756#issuecomment-354666117 was the plan outlined to rename `aports/firmware` to `aports/nonfree-firmware` and to create a new folder `aports/nonfree-userland` (when we actually have userland drivers packaged, e.g. for the Nokia N9).
So the name makes more sense when we will actually have the `nonfree-userland` folder.

But in theory it would be possible to have libre firmware as well, that's what @craigcomstock (aka unreasonable in #postmarketOS-lowlevel) is working in [for cellular modems](https://lists.osmocom.org/pipermail/baseband-devel/2017-December/005462.html).

I've created this PR as that was the last open TODO list entry of #756 so it can be closed. If you guys think this change doesn't make sense right now/at all, that's also fine with me, then we could scratch that.

z3ntu:
I think currently the directory structure makes a lot of sense for the firmware packages: They are in a folder called `firmware` and are all prefixed with `firmware-`. If we are moving them to a folder called `nonfree-firmware` but still calling the package only `firmware-` is a bit weird. And also the `linux-firmware` package is called `linux-firmware` and not `linux-nonfree-firmware` :wink: So I would vote for stalling the PR until we have something that prompts this change.

ollieparanoid:
Yeah, we won't be able to change the `linux-firmware` name in upstream.
Okay then, let's look at this again when we have a reason to.
I really appreciate your feedback, thanks again!

Issue #1272, opened 2018-02-27
Aportgen gcc-* requires manual updating of the git checkout of upstream aports


MartijnBraam:
I'm trying to do the aportgen process from the wiki since it seems like the upstream has updated. I can't do the last step with building GCC since it can't install because unsatisfiable constraints

```
ERROR: unsatisfiable constraints:
  libstdc++-6.4.0-r7:
    breaks: g++-aarch64-6.4.0-r6[libstdc++=6.4.0-r6]
    satisfies: gcc-6.4.0-r7[so:libstdc++.so.6]
  gcc-6.4.0-r7:
    breaks: g++-aarch64-6.4.0-r6[gcc=6.4.0-r6]
  .pmbootstrap-0:
    masked in: cache
    satisfies: world[.pmbootstrap]
```
This is after aportgen and after zapping the chroots.

(My current checkout is in https://github.com/postmarketOS/pmbootstrap/tree/fix/compiler-update)

ollieparanoid:
Thanks for reporting! I'm looking into this now.

ollieparanoid:
Okay, what happened here, is that `aportgen` actually downgraded the package, because the `git` checkout of Alpine's `aports` were not up to date. This is my faul, as this is not documented yet and `pmbootstrap` doesn't handle this case gracefully either.

So this works for the first time:
```
$ pmbootstrap aportgen gcc-armhf gcc-aarch64 gcc-x86_64
```

But when running it again, the git repo needs to be updated first:
```
$ pmbootstrap chroot
# apk add git
# su - pmos
$ cd ~/git/aports_upstream
$ git pull
$ exit
# exit
```
Then you can run:
```
$ pmbootstrap aportgen gcc-armhf gcc-aarch64 gcc-x86_64
```

I've updated the `gcc-*` packages in `master` and I'm building them right now for the repo. I'll leave this issue open to fix the workflow properly in `pmbootstrap`.

ollieparanoid:
The trick is to specify the arch, then building it is working as expected:
```
$ pmbootstrap build --arch=x86_64 gcc-armhf
```
Otherwise it build for the first arch it finds in the `APKBUILD`, which is `aarch64` in the case of `gcc-armhf` and `armhf` in the case of `gcc-aarch64` (which got me really confused at first TBH).
That sounds also like something we could improve from `pmbootstrap` side, at least warn the user that this will probably not work.

ollieparanoid:
[...]

PR #1277, opened 2018-02-28
Always prefer native/device arch when building

ollieparanoid:
In case the user does not specify for which arch packages should be
built with `pmbootstrap build`, we detect it automatically.

Previous logic was, that if the APKBUILD's arch is "all" or "noarch",
then prefer the native arch, and otherwise use the first one in the
list of available arches.

New behavior is, that we also check if the list of possible arches
contains the native arch (and if that fails, the device arch). If that
is the case, we return the native/device arch instead of the first one
in the list.

### Use case

The arch from `gcc-armhf` and similar packages (as generated by
`pmbootstrap aportgen`) used to be "all", but is nowadays a specific
list of arches. This means, that after updating the `gcc-armhf` and
`gcc-aarch64` packages, and calling `pmbootstrap build gcc-armhf`,
it will try to build `gcc-armhf` for `aarch64` instead of the native
architecture, because that is the first one listed.
And since compiling to `aarch64` requires `gcc-aarch64`, it will build
that for the native architecture first.

So you're asking for `gcc-armhf` and it compiles `gcc-aarch64`, which
is very confusing (see #1272).

### DIFF
--- a/pmb/build/autodetect.py
+++ b/pmb/build/autodetect.py
@@ -53,9 +53,11 @@ def arch(args, pkgname):
     Find a good default in case the user did not specify for which architecture
     a package should be built.
 
-    :returns: native architecture (x86_64 in most cases) when the APKBUILD has
-              "noarch" or "all". Otherwise the first architecture in the
-              APKBUILD.
+    :returns: arch string like "x86_64" or "armhf". Preferred order, depending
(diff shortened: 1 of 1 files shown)


ollieparanoid:
Thank for reviewing @z3ntu!

PR #1278, opened 2018-02-28
Skip virtual packages when parsing APKINDEX

### LINKED ISSUE #1273: Fix 'missing required key "timestamp" in block'
This happens, when a virtual package is still installed. We use the `.pmbootstrap` virtual package nowadays to install packages, that should not be explicitly marked as installed.
Anyway, when an error happens before the `.pmbootstrap` package gets uninstalled (which happens directly after installing it), and you do anything with `pmbootstrap`, it will fail to parse the package database. Because virtual packages like `.pmbootstrap` don't have the `"timestamp"` key in that database file.

Workaround: `pmbootstrap zap`

ollieparanoid:
Since PR #1247 we are using the virtual package `.pmbootstrap` to mark
packages as not explcitly installed. In case `pmbootstrap` doesn't
finish the installation because of a bug, the `.pmbootstrap` virtual
package may not get uninstalled.

As virtual packages don't have the "timestamp" attribute set in the
package database (which indicates when the package was built), the
APKINDEX parser fails to parse them.

Changes:
* pmb.parse.apkindex.parse_next_block(): don't require the "timestamp"
  attribute to be set (but the arch attribute instead, which is always
  present)
* pmb.parse.apkindex.parse(): when a block does not have a `timestamp`
  attribute, skip it, because it must be a virtual package.
* add test cases for both functions with a package database that
  contains a virtual package.

Fixes #1273.

### DIFF
--- a/pmb/parse/apkindex.py
+++ b/pmb/parse/apkindex.py
@@ -37,9 +37,16 @@ def parse_next_block(args, path, lines, start):
     :returns: a dictionary with the following structure:
               { "arch": "noarch",
                 "depends": ["busybox-extras", "lddtree", ... ],
+                "origin": "postmarketos-mkinitfs",
                 "pkgname": "postmarketos-mkinitfs",
                 "provides": ["mkinitfs=0.0.1"],
+                "timestamp": "1500000000",
                 "version": "0.0.4-r10" }
+              NOTE: "depends" is not set for packages without any dependencies,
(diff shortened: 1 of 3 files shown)


ollieparanoid:
Oh, I forgot to include the test instructions (I had written them down somewhere lol). Thanks for figuring it out, for testing and for reviewing, I highly appreciate it! \o/

PR #1279, opened 2018-03-01
New device: Samsung Galaxy Tab 10.1 (p4wifi)

Decatf:
What's working:
- Flashing with heimdall
- USB SSH access
- Touch
- DRM fbdev
- 2D acceleration (Grate driver)
- Wifi

It runs on a mainline-ish kernel 4.14. I've forward ported most of the device drivers. I've been trying to transition over to mainline drivers where possible. The kernel is a bunch of patches containing out of tree stuff which I rebase on to the mainline kernels.

The Weston demo is working with framebuffer only. Xfce4 and Mate desktop can run with 2D acceleration using the reverse engineer [Grate driver](https://github.com/grate-driver/). The 3D implementation is enough to show glxgears.

I've been trying to package the Grate driver but it is not included in this pull request. The Grate driver requires an out of tree `mesa` and `libdrm`. Plus a new `xf86-video-opentegra` package. There are some package conflicts when I compile mesa. It's kind of a mess which I haven't been able to sort out yet.


aports/firmware/firmware-samsung-p4wifi/APKBUILD
```diff
@@ -0,0 +1,34 @@
...
+source="$source $pkgname-$_commit_nvram-$_nvram_file::$_url_nvram";
+
+package() {
+	install -Dm644 "$srcdir/$pkgname-$_commit_fw-$_fw_file" \
+		"$pkgdir/lib/firmware/postmarketos/$_fw_file"
+	install -Dm644 "$srcdir/$pkgname-$_commit_nvram-$_nvram_file" \
+		"$pkgdir/lib/firmware/postmarketos/$_nvram_file"
+
+	install -Dm644 "$srcdir/$pkgname-$_commit_fw-$_fw_file" \
+		"$pkgdir/lib/firmware/brcm/brcmfmac4330-sdio.bin"
```
ollieparanoid: We put all firmware files installed by postmarketOS into `/lib/firmware/postmarketos`, and add a patch file that patches the path in the kernel sources if necessary. Could you also do that here, or is there a specific reason why you've kept it?

aports/firmware/firmware-samsung-p4wifi/APKBUILD
```diff
@@ -0,0 +1,34 @@
...
+	install -Dm644 "$srcdir/$pkgname-$_commit_nvram-$_nvram_file" \
+		"$pkgdir/lib/firmware/brcm/brcmfmac4330-sdio.txt"
```
ollieparanoid: (also here)

aports/firmware/firmware-samsung-p4wifi/APKBUILD
Decatf: I didn't want to put more hacks into the driver. The brcmfmac driver is less flexible in specifying firmware paths. The default directory `brcm` is hardcoded. The filenames are also predefined for the set of supported devices.

I've discovered that there's a module parameter `alternative_fw_path` which prefixes the firmware path. So the resulting path can be `postmarketos/brcm/brcmfmac4330-sdio.bin`. Is this acceptable?

aports/device/device-samsung-p4wifi/APKBUILD
```diff
@@ -0,0 +1,29 @@
...
+build() {
+	devicepkg_build $startdir $pkgname
+}
+
+package() {
+	devicepkg_package $startdir $pkgname
+}
+
+nonfree_firmware() {
+    pkgdesc="bcm4330 wifi firmware"
```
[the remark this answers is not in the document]
Decatf: Ok, I'll correct the spacing.

aports/firmware/firmware-samsung-p4wifi/APKBUILD
ollieparanoid: Yes, the solution with `alternative_fw_path` sounds good! Thanks for looking into this :tada:

PR #1281, opened 2018-03-02
Fix Nokia N9 build and update kernel

filippz:
Dtb entry in deviceinfo prevented building of image so it's removed since it's being appended while building kernel. Kernel wasn't applying patch from base version so that was fixed and kernel was updated to v4.16-rc3.


z3ntu:
Please update the checksums of `linux-nokia-n9` by running `./pmbootstrap.py checksum linux-nokia-n9`

aports/device/linux-nokia-n9/APKBUILD
```diff
@@ -51,8 +50,10 @@ prepare() {
 	local _patch_failed=
 	cd "$ksrcdir"
 
+	source_patches="linux-v$_basever-v$_pkgver.patch
+	$source"
 	# first apply patches in specified order
-	for i in $source; do
+	for i in $source_patches; do
```
ollieparanoid: This confused me a bit, but now I see what it does:
* it makes sure, the linux-v... patch gets applied first, and then all other patches in source.
* the patch does not get applied again, because the entry in $source has the download URL at the end, which does not end in `*.patch`.

filippz:
Thank you both. I've forgotten to remove the patch that belongs to WIP support for SGX GPU, but I see @ollieparanoid sorted it out!

PR #1282, opened 2018-03-03
Fix build --src for foreign-arch chroots

ollieparanoid:
The `suffix` argument was not specified in chroot commands executed in
`pmb.build._package.override_source()`. Because of that, it was not
possible to use "build --src" when compiling in a non-native chroot,
for example:

```
$ pmbootstrap build hello-world --arch=armhf --src=~/code/pmbootstrap/aports/main/hello-world
...
(native) % rm /tmp/APKBUILD.append
rm: can't remove '/tmp/APKBUILD.append': No such file or directory
```

Thanks to @scintill for reporting this in https://github.com/postmarketOS/pmbootstrap/pull/1210#issuecomment-369827265!

@scintill: do you want to test and review this?


ollieparanoid:
Thanks for the review @scintill!

PR #1283, opened 2018-03-03
linux-postmarketos-mainline: fix weston (x86_64)

ollieparanoid:
Since it works for linux-postmarketos-stable, I've simply copied over
the config, and ran "pmbootstrap menuconfig" on it again. Then it
worked again in Qemu.


z3ntu:
How should we handle the kernel configs for the different postmarketos flavors anyways? Currently (before this PR) we have 3 different configs that all have some options enabled. After this PR we still have the lts config and the mainline/stable config.

ollieparanoid:
I would recommend to keep them in sync as good as we can. The `lts` config is also almost the same as the `stable` config, I just ran a diff.

z3ntu:
Nothing to do directly with this PR but I sometimes hit the following error while running pmbootstrap:
```
 /m/h/p/pmbootstrap   fix/linux-pmos-mainline-doesnt-boot-weston $…  pmb install --add linux-postmarketos-lts,linux-postmarketos-stable,linux-postmarketos-mainline --no-fde
[10:20:08] *** (1/5) PREPARE NATIVE CHROOT ***
[sudo] password for luca: 
[10:20:12] (native) install
[10:20:13] *** (2/5) CREATE DEVICE ROOTFS ("qemu-amd64") ***
[10:20:13] Update package index (12x)
[10:39:47] (rootfs_qemu-amd64) install
[10:39:47] ERROR: can only concatenate list (not "odict_keys") to list
[10:39:47] Run 'pmbootstrap log' for details.
[10:39:47] See also: <https://postmarketos.org/troubleshooting>
```

z3ntu:
image
:+1:

Issue #1081, opened 2018-01-04
Implement charging mode


drebrez:
I've started with the packaging of the `charging-sdl` which can be tested in the branch:
https://github.com/postmarketOS/pmbootstrap/tree/feature/charging-sdl

### Building the package
`pmbootstrap checksum charging-sdl`
`pmbootstrap build charging-sdl --arch armhf`

### Installation on the device
`pmbootstrap install --add charging-sdl`
or copy it into the device:
`scp $work/packages/armhf/charging-sdl-0.1-r0.apk user@172.16.42.1:/home/user/`
`ssh user@172.16.42.1`
`apk add ./charging-sdl-0.1-r0.apk`

### Testing
`export DFBARGS="system=fbdev,no-cursor,disable-module=linux_input"`
`charging-sdl -pcf /usr/share/fonts/ttf-dejavu/DejaVuSans.ttf`

### Todo

- [ ] Package `charging-sdl` with proper version tag
- [ ] Include `charging-sdl` into the `initramfs-extra`
- [ ] Identify a common way to detect the charging mode on each device
- qualcomm lk adds `androidboot.mode=charger` to the kernel cmdline when it's supposed to go to charger mode (see [lk code](https://github.com/FairphoneMirrors/android_kernel_lk_fairphone_FP2/blob/87f6ac005aa39fd22983512a21b029ca4a376c42/target/FP2/init.c#L703-L714)) (thanks to @z3ntu)
- some Samsung devices have `lpm_boot=1` in the kernel cmdline for charge mode

| device         | normal                   | charge mode                                              |
|----------------|--------------------------|---------------------------------------------------|
| samsung-i9070  | lpm_boot=0               | lpm_boot=1                                        |
| samsung-klte   | androidboot.baseband=msm | androidboot.baseband=lpm androidboot.mode=charger |
| samsung-maguro | -                        | androidboot.mode=charger                          |
| huawei-y530    | -                        | androidboot.huawei_type=oem_rtc                   |
| sony-seagull | startup=0x00000001 warmboot=0x00000000 | startup=0x00010004 warmboot=0x77665501 | 

- [ ] Update [postmarketos-mkinitfs/init.sh.in](https://github.com/postmarketOS/pmbootstrap/blob/master/aports/main/postmarketos-mkinitfs/init.sh.in) to handle the charging mode
- [ ] Tweak the system to reduce performance and improve charging speed
[...]

PR #1285, opened 2018-03-04
Implement charging mode

drebrez:
This is the initial implementation of the charging mode.

- The  `charging-sdl` is packaged using the latest commit, a tagged version would be better.
- Added `charging-sdl` and `triggerhappy` to the `initramfs-extra`
- init script changes: 
  - Extracted common code to configure `directfb` and `tslib` in a function
  - Check charging boot mode against a list of cmdline options
  - Use the same font file configured for `osk-sdl`
  - Configure `triggerhappy` to start `charging-sdl` (if it's not already running) when the power key is pressed
  - Start `charging-sdl` once before starting `triggerhappy` (to show something instead of the loading splash screen)

Please review the code and test it on some devices.

Close: #1081


aports/main/charging-sdl/APKBUILD
```diff
@@ -0,0 +1,25 @@
...
+pkgrel=0
+pkgdesc="SDL2 charging application for the initramfs"
+url="https://github.com/postmarketOS/charging-sdl"
+arch="all"
+license="GPL3"
+depends="sdl2 sdl2_ttf mesa-gl directfb"
+makedepends="linux-headers sdl2-dev sdl2_ttf-dev"
+subpackages=""
+_commit="55aba2531e170807a3bc2fe6469345bc5d3443a2"
+source="${pkgname}-${_commit}.tar.gz::https://github.com/postmarketOS/charging-sdl/archive/${_commit}.tar.gz"
```
ollieparanoid: Here is the tagged release:
https://github.com/postmarketOS/charging-sdl/releases/tag/0.1

aports/main/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -267,20 +267,46 @@ start_udhcpd() {
...
+start_charging_mode(){
+	# Check cmdline for charging mode
+	chargingmodes="
+		androidboot.mode=charger
+		lpm_boot=1
+		androidboot.huawei_type=oem_rtc
+		startup=0x00010004
+	"
+	# shellcheck disable=SC2086
+	if ! grep -Eq "$(echo $chargingmodes | tr ' ' '|')" /proc/cmdline; then return; fi
```
ollieparanoid: How about:
```shell
grep -Eq "$(echo $chargingmodes | tr ' ' '|')" /proc/cmdline || return
```

aports/main/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -267,20 +267,46 @@ start_udhcpd() {
...
+	setup_directfb_tslib
+	# Get the font from osk-sdl config
+	fontpath=$(awk '/^keyboard-font/{print $3}' /etc/osk.conf)
+	# Set up triggerhappy config
+	{
+		echo "KEY_POWER 1 [ ! \$(pgrep -x charging-sdl) ] && charging-sdl -pcf $fontpath"
```
ollieparanoid: I think this would also work:
```shell
echo "KEY_POWER 1 \$(pgrep -x charging-sdl) || charging-sdl -pcf $fontpath"
```

aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -98,6 +98,8 @@ BINARIES_EXTRA="
 	/sbin/e2fsck
 	/usr/sbin/resize2fs
 	/usr/bin/osk-sdl
+	/usr/bin/charging-sdl
+	/usr/sbin/thd
```
ollieparanoid: How about we sort that list alphabetically?

drebrez: agree, but also mixing executables and libraries or keep the executables first (in alphabetic order), and the libraries after (also in alphabetic order)?

ollieparanoid: I would mix them, but I don't really have a preference.

PR #1291, opened 2018-03-04
aportgen: Gracefully handle old aports_upstream git checkout

### LINKED ISSUE #1272: Aportgen gcc-* requires manual updating of the git checkout of upstream aports

ollieparanoid:
### Introduction
In order to get cross-compilers, we generate a few aports (e.g.
binutils-armhf, gcc-armhf) automatically from Alpine's aports.
pmbootstrap was already able to perform a git checkout of Alpine's
aports repository. But it needed to be manually updated. Otherwise
the `pmbootstrap aportgen` command could actually downgrade the aport
instead of updating it to the current version.

### Git wrapper?
After thinking about adding a dedicated pmbootstrap command for
updating git repositories, I thought it would be better to not open
that can of worms (pmbootstrap as general git wrapper? no thanks).

### Comparing versions!
The solution implemented here compares the upstream aport version of
the git checkout of a certain package (e.g. gcc for gcc-armhf) with the
version in Alpine's binary package APKINDEX. When the aport version is
lower than the binary package version, it shows the user how to update
the git repository with just one command:

```
pmbootstrap chroot --add=git --user -- git -C /mnt/pmbootstrap-git/aports_upstream pull
```

Changes:
* `pmb.aportgen.core.get_upstream_aport()`: new function, that returns
  the absolute path to the upstream aport on disk, after checking the
  version of the aport against the binary package.
* Use that new function in pmb.aportgen.gcc and pmb.aportgen.binutils
* New function `pmb.helpers.repo.alpine_apkindex_path()`: updates the
  APKINDEX if necessary and returns the absolute path to the APKINDEX.
  This code was basically present already, but not as function, so now
  we have a bit less overhead there.
* `pmbootstrap chroot`: new `--user` argument
* `pmb.parse.apkbuild`: make pkgname check optional, as it fails with
  the official gcc APKBUILD before we modify it (the current APKBUILD
  parser is not meant to be perfect, as this would require a full shell
  parsing implementation).
* Extended `test_aportgen.py` and enabled it by default in
  `testcases_fast.sh`.  Previously it was disabled due to traffic
  concerns (cloning the aports repo, but then again we do a full KDE
  plasma mobile installation in Travis now, so that shouldn't matter
  too much).
* `testcases_fast.sh`: With "test_aport_in_sync_with_git" removed
  from the disabled-by-default list (left over from timestamp based
  rebuilds), there were no more test cases disabled by default. I've
  changed it, so now the qemu_running_processes test case is disabled,
  and added an `--all` parameter to the script to disable no test
  cases. Travis runs with the `--all` parameter while it's useful to
  do a quick local test without `--all` in roughly 2 minutes instead of
  10.
* `aports/cross/binutils-*`: Fix `_mirror` variable to point to current
  default Alpine mirror (so the aportgen testcase runs through).

Fixes #1272.

### DIFF
--- a/pmb/helpers/repo.py
+++ b/pmb/helpers/repo.py
@@ -170,3 +170,26 @@ def update(args, arch=None, force=False, existing_only=False):
         pmb.helpers.run.root(args, ["cp", temp, target])
 
     return True
+
+
+def alpine_apkindex_path(args, repo="main", arch=None):
+    """
+    Get the path to a specific Alpine APKINDEX file on disk and download it if
+    necessary.
+
+    :param repo: Alpine repository name (e.g. "main")
+    :param arch: Alpine architecture (e.g. "armhf"), defaults to native arch.
+    :returns: full path to the APKINDEX file
+    """
+    # Repo sanity check
+    if repo not in ["main", "community", "testing", "non-free"]:
+        raise RuntimeError("Invalid Alpine repository: " + repo)
+
+    # Download the file
+    update(args, arch)
+
+    # Find it on disk
+    arch = arch or args.arch_native
+    repo_link = args.mirror_alpine + args.alpine_version + "/" + repo
+    cache_folder = args.work + "/cache_apk_" + arch
+    return cache_folder + "/APKINDEX." + hash(repo_link) + ".tar.gz"
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ script:
   - yes "" | ./pmbootstrap.py init
   - ./pmbootstrap.py kconfig_check
   - test/check_checksums.py --build
-  - test/testcases_fast.sh
+  - test/testcases_fast.sh --all
 after_success:
   - coveralls
 after_failure:
--- a/aports/cross/busybox-static-aarch64/APKBUILD
+++ b/aports/cross/busybox-static-aarch64/APKBUILD
@@ -6,7 +6,7 @@ pkgver=1.27.2
 pkgrel=8
 
 _arch="aarch64"
-_mirror="https://nl.alpinelinux.org/alpine/"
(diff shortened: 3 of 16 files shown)


pmb/helpers/repo.py
```diff
@@ -170,3 +170,26 @@ def update(args, arch=None, force=False, existing_only=False):
...
+    """
+    Get the path to a specific Alpine APKINDEX file on disk and download it if
+    necessary.
+
+    :param repo: Alpine repository name (e.g. "main")
+    :param arch: Alpine architecture (e.g. "armhf"), defaults to native arch.
+    :returns: full path to the APKINDEX file
+    """
+    # Repo sanity check
+    if repo not in ["main", "community", "testing", "non-free"]:
```
drebrez: Looking at `http://dl-cdn.alpinelinux.org/alpine/edge/` it also contains a `releases` repository

ollieparanoid: Thanks for checking! Yeah, these are `iso` images of the Alpine releases (e.g. 3.7.0).

ollieparanoid:
@drebrez: Thank you very much! :tada:

PR #1301, opened 2018-03-07
deviceinfo: remove external_disk_install and external_disk, use external_storage instead

drebrez:
This checks the next box in #688

Removed `deviceinfo_external_disk_install` and `deviceinfo_external_disk` variables in `deviceinfo` files and use `deviceinfo_external_storage` instead.

Also adapted the related python code and tests to work with the new variable.

TODO when the PR will be closed: 

- [ ]  update the https://wiki.postmarketos.org/wiki/Deviceinfo_reference page


drebrez:
@ollieparanoid Wiki updated 👍

PR #1303, opened 2018-03-07
Fix building packages by provides name

### LINKED ISSUE #1220: pmb.chroot.apk.install() does not build package providers
This came up while testing #1193. The `mkbootimg` package provides `unpackbootimg` now. When we run...
```py
pmb.chroot.apk.install(args, ["unpackbootimg"])
```
...and there is no binary package of `mkbootimg`, it says that there's no APKBUILD for `unpackbootimg` and no binary package. Instead it should build `mkbootimg` and install it.

This might be fixed by pending PR #1202 already, I haven't tested it yet.

ollieparanoid:
Use case: `mkbootimg` provides the `unpackbootimg` package. When
running `pmb.chroot.apk.install(args,"unpackbootimg")`, it was not
able to properly build the package.

Reproducing the error:
```shell
sudo rm ~/.local/var/pmbootstrap/packages/x86_64/mkbootimg*
pmbootstrap index
pmbootstrap --mirror-pmOS="" chroot --add=unpackbootimg
```

Or alternatively (simpler but less illustrative):
```shell
pmbootstrap build unpackbootimg --force
```

Fixes #1220.

### DIFF
--- a/pmb/build/other.py
+++ b/pmb/build/other.py
@@ -58,7 +58,8 @@ def find_aport(args, package, must_exist=True):
             # Search in subpackages
             for path_current in glob.glob(args.aports + "/*/*/APKBUILD"):
                 apkbuild = pmb.parse.apkbuild(args, path_current)
-                if package in apkbuild["subpackages"]:
+                if (package in apkbuild["subpackages"] or
+                        package in apkbuild["provides"]):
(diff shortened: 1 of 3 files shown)


ollieparanoid:
Thanks for the review!

PR #1306, opened 2018-03-08
pmbootstrap zap -m: delete outdated packages too

ollieparanoid:
_This is important for streamlining the binary repository (#970), so we can clean old packages regularly and don't run out of space._

`-m` is for deleting local compiled packages, for which there is no
aport with the same version. Prior to this change, this only worked
for packages where no aport exists, or for packages that are newer
than the aports.

That is, because we used the usual `APKINDEX` parsing logic, which
ignores old packages in the `APKINDEX` and only returns the one with the
highest version (that makes sense during dependency resolution).

Changes:
* New `pmb.parse.apkindex.parse_blocks()` function that returns a raw
  list of blocks, instead of the dict with removed duplicates with
  lower version you get from the usual `.parse()` function.
* Renamed each of the zap flags and their descriptions to make clear
  what they are doing now.
  ```
  short  long (old)         long (new)
  -p     --packages         --pkgs-local
  -m     --mismatch-bins    --pkgs-local-mismatch
  -o,    --old-bins         --pkgs-online-mismatch
  ```

### DIFF
--- a/pmb/chroot/zap.py
+++ b/pmb/chroot/zap.py
@@ -23,20 +23,21 @@ import os
 
 import pmb.chroot
 import pmb.helpers.run
+import pmb.parse.apkindex
 
 
-def zap(args, confirm=True, dry=False, packages=False, http=False,
-        mismatch_bins=False, old_bins=False, distfiles=False):
+def zap(args, confirm=True, dry=False, pkgs_local=False, http=False,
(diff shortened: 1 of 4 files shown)


ollieparanoid:
Big thanks for reviewing and testing @drebrez! :tada:

PR #1313, opened 2018-03-10
i9195: add touchscreen to deviceinfo

scintill:


rrooij:
Nice! Does this fix the touch screen issue?

scintill:
Yeah, touch in the Weston demos seems to work now. I can't seem to drag windows though -- is that supposed to work?

Issue #1054, opened 2017-12-29
Get qualcomm modems working


lawl:
This issue is to split out the progress on getting qualcomm modems working from the general [cellular modem issue](https://github.com/postmarketOS/pmbootstrap/issues/598) since I like to document the progress also for myself, and for everyone else that wants to look into this. Last but not least to avoid duplicate work.

lawl:
Quick update:

I have definitely gotten `qmuxd` working on pmOS.

All of this work is currently on my dev phone and not packaged since all of this is explorative at this point.

My understanding currently is that we could maybe completely skip rild on qualcom modems since `ofono` can talk `qmi` and there is a project [here](https://github.com/scintill/qmiserial2qmuxd) that translates `qmi` to `qmuxd` socket messages. Not sure though.

Quick rundown of what I have done so far for documentation purposes and in case someone else also wants to look into this and as a memo for myself.

I run `qmuxd` by simply ripping `/vendor/lib64 /system/lib64` and `/system/bin/linker64` out of the `lineageOS` system and throw them on my phone, this allows `qmuxd` to run using the `lineageOS` bionic stuff. (For development purposes only, to be trimmed down/chrooted/namespaced/seccomp'd to smithereens)

The error creating the sockets above was from me not really paying attention. `qmuxd` calls `setuid(1001)` so even though i was running it as root and prepared the `/dev/socket/...` folders for it, it failed of course.

This is my current preparation script:

```
# cat makesockets.sh 
#!/bin/sh
mkdir /dev/socket

mkdir /dev/socket/qmux_bluetooth
chown 1001 /dev/socket/qmux_bluetooth

mkdir /dev/socket/qmux_radio
chown 1001 /dev/socket/qmux_radio

mkdir /dev/socket/qmux_audio
chown 1001 /dev/socket/qmux_audio

mkdir /dev/socket/qmux_gps
chown 1001 /dev/socket/qmux_gps

mkdir /dev/socket/qmux_nfc
chown 1001 /dev/socket/qmux_nfc
```
`qmuxd` now runs and listens for connections.

I now run `socat PTY,link=/tmp/qmiserial2qmuxd
.pty,cfmakeraw EXEC:"./qmiserial2qmuxd"
` which requires a small patch to add an [#ifdef ANDROID](https://github.com/scintill/qmiserial2qmuxd/blob/master/qmiserial2qmuxd.c#L39) guard for the `termio.h` include.

[...]

PR #1314, opened 2018-03-10
i9195/Qualcomm modem support

scintill:
I've tested with SMS.  See f5e8e1f and 
72bbfaf for an idea of the changes needed to incorporate this into a similar device. (I'd be willing to help someone port this to their device, if you're familiar with fetching source, compiling, and troubleshooting.) I've been testing with ofono like this:

```sh
sudo service ofono start

sudo dbus-send --print-reply --system --dest=org.ofono /gobi_0 org.ofono.Modem.SetProperty string:"Powered" variant:boolean:"true"
sudo dbus-send --print-reply --system --dest=org.ofono /gobi_0 org.ofono.Modem.GetProperties
sudo dbus-send --print-reply --system --dest=org.ofono /gobi_0 org.ofono.Modem.SetProperty string:"Online" variant:boolean:"true"
sudo dbus-send --system --print-reply --dest=org.ofono /gobi_0 org.ofono.MessageManager.SendMessage string:"$PHONE_NUMBER" string:"Hello world! -postmarketOS"

# receive SMS like this (might be spammy if you have other DBus apps):
sudo dbus-monitor --system
```

## What next?

ofono mainline doesn't have support for QMI device voicecalls last I saw, but Sysmocom and Alexander Couzens have patches [here](https://git.sysmocom.de/ofono/) that I've successfully used in my [Android oFono-based RIL](https://github.com/scintill/android_frameworks_opt_telephony_ril_ofono). I have a couple of relevant patches [here](https://github.com/scintill/android_external_ofono/commits/master) too. There might be some audio device initialization that needs to be done to route the mic to the phone or something, but it's probably not too hard to figure out since my fully open-source Android stack was working.

I also had data connections working on my Android system, so I anticipate that wouldn't be far away either. We'll need a patch to set something like `ofono_modem_set_string(modem, "NetworkInterface", iface_name);`, maybe from udev properties. After that, I imagine there's pre-existing things (NetworkManager?) that can take care of a network connection that ofono provides.

And of course, we need a UI to use the phone features... I tried getting Plasma Mobile running on software rendering, but it was extremely slow and crashing or something. Maybe I will try again when I get voicecalls working - I know it has a dialer but I'm not sure about SMS support.

Ping #1054.


scintill:
I think the Travis build failed because androidfilehost.com expired the download link for the firmware. Thought it might happen... any thoughts on where else to find/host it?

aports/main/libqipcrtr4msmipc/APKBUILD
```diff
@@ -0,0 +1,23 @@
+pkgname=libqipcrtr4msmipc
+pkgver=0.1_git20180304
+pkgrel=0
+pkgdesc="LD_PRELOADable library to emulate AF_QIPCRTR socket on devices with only AF_MSM_IPC"
+url="https://github.com/scintill/libqipcrtr4msmipc"
```
ollieparanoid: It's awesome that you wrote wrappers to get this working \o/

I'm playing with the idea of introducing a new `aports/modem` subfolder, where all modem specific aports (`qcom_rmtfs`, `libqipcrtr4msmipc`, `libsmdpkt_wrapper`, `qrtr`, `msmipc-dev` and the temporary `ofono` fork) could reside. The reason is, that from reading the package names it is not super obvious that this is all modem related. What do you think about that?

aports/main/libqipcrtr4msmipc/APKBUILD
```diff
@@ -0,0 +1,23 @@
+pkgname=libqipcrtr4msmipc
+pkgver=0.1_git20180304
```
ollieparanoid: Since this is your own repo, how about tagging releases and using them here?

aports/main/libqipcrtr4msmipc/APKBUILD
```diff
@@ -0,0 +1,23 @@
...
+options="!check"
+builddir="$srcdir/libqipcrtr4msmipc-$_commit"
+
+build() {
+	cd "$builddir"
+	make C_INCLUDE_PATH=/usr/src/msmipc-dev/include
+}
+
+package() {
+	install -Dm644 "$builddir"/libqipcrtr4msmipc.so "$pkgdir"/usr/lib/preload/libqipcrtr4msmipc.so
```
ollieparanoid: @NotKit: here's a good preloading example :smile:

aports/main/ofono/APKBUILD
```diff
@@ -0,0 +1,45 @@
+# Based on https://git.alpinelinux.org/cgit/aports/tree/testing/ofono/APKBUILD?id=014ae282b4a9152a5b64451f2815f34fcb53507c
+# Thanks to upstream maintainer, Francesco Colista <fcolista@alpinelinux.org>
```
ollieparanoid: We don't mention the upstream maintainers in our APKBUILDs, so they don't get annoyed with questions for postmarketOS packages. But they could be credited in the commit message. See also: https://wiki.postmarketos.org/wiki/Create_a_package#General_information

Ideally we would only fork `ofono` temporarily anyway, until the changes are upstreamed into `ofono` and then we could use the upstream package from Alpine again.

aports/device/device-samsung-i9195/APKBUILD
```diff
@@ -17,6 +18,18 @@ build() {
 
 package() {
 	devicepkg_package $startdir $pkgname
+
+	mkdir "$pkgdir"/boot
+	ln -s /dev/disk/by-partlabel/modemst1 "$pkgdir"/boot/modem_fs1
+	ln -s /dev/disk/by-partlabel/modemst2 "$pkgdir"/boot/modem_fs2
+	ln -s /dev/disk/by-partlabel/fsg      "$pkgdir"/boot/modem_fsg
```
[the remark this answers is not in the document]
scintill: qcom_rmtfs uses those paths [here](https://github.com/andersson/rmtfs/blob/0f800fa4334b03a47941ce1b577f51a69552e414/storage.c#L26). (One is a string the modem sends in the request, the other is the actual Linux filesystem path; in this case, the same.) They are a bit weird. What path would be better? I don't want to embed the by-partlabel path into rmtfs, because it might differ by device.

aports/firmware/firmware-samsung-i9195/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+subpackages="$pkgname-modem"
+url="https://forum.xda-developers.com/galaxy-s4-mini/general/modem-samsung-galaxy-s4-mini-gt-i9195-t3390780"
+arch="noarch"
+license="proprietary"
+# build system wants to strip binaries, so stop that, and thinks they are wrong arch, so ignore that
+options="!strip !archcheck"
+makedepends="p7zip"
+
+source="$_ver.tar::https://wa1.androidfilehost.com/dl/gMdb8sYfl8Zy2_daybpA-Q/1520163411/889964283620758806/I9195XXUCQG1.tar.md5
+SHA512SUMS.modem"
```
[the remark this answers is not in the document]
scintill: Ok. I have not done verification really. I understand the files are signed, so hopefully that's enough verification.

aports/main/libqipcrtr4msmipc/APKBUILD
scintill: aports/modem sounds good. I considered device too, but something separate would probably be better.

aports/main/libqipcrtr4msmipc/APKBUILD
scintill: Is the path prefix /usr/lib/preload good? I wanted something that wouldn't conflict with things, and would be clear this was not a normal lib.

I've gone back and forth on whether to use preload or to patch the packages. For now, I like preload because it can apply to several different programs and doesn't have maintenance overhead when upgrading package versions.

aports/device/device-samsung-i9195/APKBUILD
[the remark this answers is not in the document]
MartijnBraam: I guess this is for the modem firmware? The correct place is `/lib/firmware/qmi` which can be either provided by a firmware package or by a mount/symlink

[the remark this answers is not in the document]
ollieparanoid: > They are a bit weird. What path would be better?

From reading up on [FHS](http://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html), the best fit I can find is somewhere in [`/etc`](http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s07.html):
> The /etc hierarchy contains configuration files. A "configuration file" is a local file used to control the operation of a program; it must be static and cannot be an executable binary.

After all, the symlinks are there for configuring the device path for `qcom_rmtfs`. It isn't perfect, but it fits much better than `/boot` in my opinion (which is on a [different partition](https://wiki.postmarketos.org/wiki/Partition-layout) in postmarketOS actually).

So how about something like:
`/etc/qcom_rmtfs/partition-symlinks/modem_fs1`

(If someone has a better idea, don't be shy to post it here.)

aports/main/libqipcrtr4msmipc/APKBUILD
ollieparanoid: Just in case it's not obvious: there's no code change needed in `pmbootstrap`, just move the files to the other folder and it will work out of the box.

aports/main/libqipcrtr4msmipc/APKBUILD
ollieparanoid: > Is the path prefix /usr/lib/preload good? I wanted something that wouldn't conflict with things, and would be clear this was not a normal lib.

Yeah, that's a good name!

> I've gone back and forth on whether to use preload or to patch the packages. For now, I like preload because it can apply to several different programs and doesn't have maintenance overhead when upgrading package versions.

This sounds reasonable for now. From what I understand it probably makes sense to upstream these changes in the long run to ofono/qcom_rmtfs, so the wrappers are not needed and we can use the vanilla sources from upstream without carrying patches.

aports/device/device-samsung-i9195/APKBUILD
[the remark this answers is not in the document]
scintill: @MartijnBraam Those partitions aren't firmware. I don't know much about what specifically is in them, but they're partitions the modem can read or write to, to store persistent data. I don't know if they can be packaged (as files in the Linux filesystem rather than raw partitions on the MMC), since they might contain IMEI, and I don't know what a good initial value for them is.

@ollieparanoid That /etc path makes sense to me.

scintill:
Thanks for the feedback. I'm also really happy there are no blobs (other than baseband firmware.) The latest push should address all issues. I also made the qcom_rmtfs symlinks be a subpackage that only gets installed if qcom_rmtfs is installed.

@z3ntu Yes, I was able to send and receive SMS, albeit from the command line. Now someone package a UI for oFono SMS :wink: (or tell me if there is one that works on my device.) I've looked at [the messaging app for Ubuntu Touch](https://launchpad.net/messaging-app) a bit.

ollieparanoid:
> And of course, we need a UI to use the phone features... I tried getting Plasma Mobile running on software rendering, but it was extremely slow and crashing or something.

Yeah, that doesn't work properly without hardware acceleration. The best fix we have is running the mainline kernel on it, and using and open source userspace drivers (together with the GPU blobs, no way around that right now, just like cellular modem firmware). This works well for the Sony Xperia Z2 Tablet for example. But of course mainlining is not an easy task (I would like to streamline the process and put up a step by step guide for that in the wiki in the future).

But basically if your SoC is supported in mainline already, and the peripherals are as well, then you would "only" need to write a device tree file that enables the peripherals properly. If you want to go down that path, people in #postmarketOS know about the process, and we have [some](https://wiki.postmarketos.org/wiki/The_Mainline_Kernel) [information](https://wiki.postmarketos.org/wiki/Mainline_FAQ) in the wiki.

> Now someone package a UI for oFono SMS :wink: (or tell me if there is one that works on my device.) I've looked at [the messaging app for Ubuntu Touch](https://launchpad.net/messaging-app) a bit.

I don't know on which stack that is based. If it is also QML, I guess we would run into the same performance problems (@bhush9, you probably know this?). Launchpad is down right now, so I can't look it up easily. However, packaging it shouldn't be much of a problem in either case.

@pavelmachek worked on a general phone functionality demo interface called [unicsy_demo](https://github.com/pavelmachek/unicsy_demo/). It is not aimed at end users like the ones from Plasma Mobile, ubports, Hildon, LuneOS UI and I have not tested it myself (might be that it only works with the N900). But we have it packaged in postmarketOS (as `unicsy-demo`), and I'm pretty sure it works without hardware accelerated graphics. So you could simply try that one out if you like to (e.g. by running `apk install unicsy-demo` on the phone as root after installing postmarketOS). `apk info -L unicsy-demo` gives you the list of installed files afterwards.

z3ntu:
Yes, the messaging app is written in QML. See the ubports fork: https://github.com/ubports/messaging-app

pavelmachek:
Mainlining... is not really an easy task. Often code needs to be
significantly modified in the process.

I can provide rough instructions, but I don't think step by step guide
is easily doable. You don't have step by step guide for writing good
code, and that's subset of the mainlining task.

pavelmachek:
If you have ofono running, then yes, unicsy-demo should provide sms
functionality for you.

You want to install it, then run /usr/share/unicsy/ofone/ofone as
root. It should not be N900-specific, but will work better in
landscape mode. There's environment variable you can use to switch to
vertical mode.

I'd really like to get ofone to work on more than N900, so yes, I'm
willing to help. If ofono is already working ok (you can check -- it
includes test scripts, see
https://wiki.postmarketos.org/wiki/Nokia_N900) ofone should be easy to
get into usable state.

									Pavel
-- 

Issue #1302, opened 2018-03-07
Can't use the pmbootstrap newapkbuild


bhush9:
```
(015477) [21:25:43] % tail -f /home/bshah/src/PostmarketOS/cache/log.txt -n 60
(015477) [21:25:43] *** output passed to pmbootstrap stdout, not to this log ***
(015485) [21:25:46] (native) % rm -r /home/pmos/build
(015485) [21:25:46] (native) % busybox su pmos -c 'mkdir -p /home/pmos/build'
(015485) [21:25:46] (native) % cd /home/pmos/build && busybox su pmos -c 'newapkbuild -n kwayland-integration -d KWayland integration -C -f'
(015485) [21:25:46] *** output passed to pmbootstrap stdout, not to this log ***
(015485) [21:25:46] (native) % busybox su pmos -c 'mv /home/pmos/build/kwayland-integration/src /home/pmos/build'
(015485) [21:25:46] (native) % busybox su pmos -c 'mv /home/pmos/build/kwayland-integration/APKBUILD /home/pmos/build'
(015485) [21:25:46] (native) % busybox su pmos -c 'mv /home/pmos/build/kwayland-integration/kwayland-integration /home/pmos/build'
mv: can't rename '/home/pmos/build/kwayland-integration/kwayland-integration': Directory not empty
(015485) [21:25:46] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(015485) [21:25:46] NOTE: The failed command's output is above the ^^^ line in the logfile: /home/bshah/src/PostmarketOS/cache/log.txt
(015485) [21:25:46] ERROR: Command failed: (native) % busybox su pmos -c 'mv /home/pmos/build/kwayland-integration/kwayland-integration /home/pmos/build'
(015485) [21:25:46] Run 'pmbootstrap log' for details.
(015485) [21:25:46] See also: <https://postmarketos.org/troubleshooting>
(015485) [21:25:46] Traceback (most recent call last):
  File "/home/bshah/src/PostmarketOS/pmbootstrap/pmb/helpers/run.py", line 53, in core
    stderr=args.logfd)
  File "/usr/lib/python3.6/subprocess.py", line 291, in check_call
    raise CalledProcessError(retcode, cmd)
[...]

PR #1316, opened 2018-03-10
Properly escape commands in pmb.chroot.user()

ollieparanoid:
## Introduction
In #1302 we noticed that `pmb.chroot.user()` does not escape commands
properly: When passing one string with spaces, it would pass them as
two strings to the chroot. The use case is passing a description with
a space inside to `newapkbuild` with `pmboostrap newapkbuild`.

This is not a security issue, as we don't pass strings from untrusted
input to this function.

## Functions for running commands in pmbootstrap
To put the rest of the description in context: We have four high level
functions that run commands:
* `pmb.helpers.run.user()`
* `pmb.helpers.run.root()`
* `pmb.chroot.root()`
* `pmb.chroot.user()`

In addition, one low level function that the others invoke:
* `pmb.helpers.run.core()`

## Flawed test case
The issue described above did not get detected for so long, because we
have a test case in place since day one, which verifies that all of the
functions above escape everything properly:
* `test/test_shell_escape.py`

So the test case ran a given command through all these functions, and
compared the result each time. However, `pmb.chroot.root()`
modified the command variable (passed by reference) and did the
escaping already, which means `pmb.chroot.user()` running directly
afterwards only returns the right output when *not* doing any escaping.

Without questioning the accuracy of the test case, I've escaped
commands and environment variables with `shlex.quote()` *before*
passing them to `pmb.chroot.user()`. In retrospective this does not
make sense at all and is reverted with this commit.

## Environment variables
By coincidence, we have only passed custom environment variables to
`pmb.chroot.user()`, never to the other high level functions. This only
worked, because we did not do any escaping and the passed line gets
executed as shell command:
```
$ MYENV=test echo test2
test 2
```
If it was properly escaped as one shell command:
```
$ 'MYENV=test echo test2'
sh: MYENV=test echo test2: not found
```
So doing that clearly doesn't work anymore. I have added a new `env`
parameter to `pmb.chroot.user()` (and to all other high level functions
for consistency), where environment variables can be passed as a
dictionary. Then the function knows what to do and we end up with
properly escaped commands and environment variables.

## Details
* Add new `env` parameter to all high level command execution functions
* New `pmb.helpers.run.flat_cmd()` function, that takes a command as
  list and environment variables as dict, and creates a properly escaped
  flat string from the input.
* Use that function for proper escaping in all high level exec funcs
* Don't escape commands *before* passing them to `pmb.chroot.user()`
* Describe parameters of the command execution functions
* `pmbootstrap -v` writes the exact command to the log that was
  executed (in addition to the simplified form we always write down for
  readability)
* `test_shell_escape.py`: verify that the command passed by reference
  has not been modified, add a new test for strings with spaces, add
  tests for new function `pmb.helpers.run.flat_cmd()`
* Remove obsolete commend in `pmb.chroot.distccd` about environment
  variables, because we don't use any there anymore
* Add `TERM=xterm` to default environment variables in the chroot,
  so running ncurses applications like `menuconfig` and `nano` works out of the box


pmb/chroot/root.py
```diff
@@ -41,46 +40,54 @@ def executables_absolute_path():
...
-    log_message += " ".join(cmd)
+        msg += "cd " + working_dir + "; "
+    msg += " ".join(cmd)
+
+    # Merge env with defaults into env_all
+    env_all = {"CHARSET": "UTF-8",
+               "HISTFILE": "~/.ash_history",
+               "PATH": pmb.config.chroot_path,
+               "SHELL": "/bin/ash",
+               "TERM": "xterm"}
```
NotKit: Is it ok to hardcode it to "xterm"? Konsole seems to have this set to "xterm-256color" by default, but probably this makes sense only in interactive session like "pmboostrap chroot".

MartijnBraam: I think xterm is a pretty safe default, I don't think we need any special terminal features.

ollieparanoid: It is what we've used so far for `menuconfig`. I don't mind changing it when using another value gives us an advantage. Alpine ships [xterm-256color](https://pkgs.alpinelinux.org/contents?file=*xterm-256color*&path=&name=&branch=edge) as well.

I'm merging this version (which does not change the value, only move it to export it by default), but if you know of any advantage we could easily make a follow-up PR that changes the value.

ollieparanoid:
@NotKit: thanks a lot for testing and reviewing, highly appreciated! :tada:

PR #1317, opened 2018-03-10
Mount tmpfs in chroot's /dev folders

### LINKED ISSUE #1175: Mount tmpfs in /dev folder in chroots
We need to have device nodes inside the chroots (which are inside the work folder configured in `pmbootstrap init`). It is possible to create them when the work folder is on an ext4 partition for example, but with lots of other file systems it not possible to create device nodes: ecryptfs, encfs, ntfs, virtualbox shared folders, fat32 and even btrfs (subvolumes only?) as I have just learned in the chat from @WilliamO7.

@MartijnBraam proposed, that we simply mount a devtmpfs inside the chroot's dev-folder, which sounds like a great solution!

@MartijnBraam: Does devtmpfs have an advantage here over tmpfs?

Related: #472

ollieparanoid:
Mount tmpfs inside the chroot's dev folder to make sure we can create device nodes, even if the filesystem of the work folder does not support it. Closes #1175.

@ethanrjones97 wrote in https://github.com/postmarketOS/pmbootstrap/issues/1175#issuecomment-371958793:
> I managed to port a device over with the patch applied, that seems to be a pretty good test :man_shrugging: 

Thanks, that is some solid testing indeed!

I'm also putting this up for review as usually. Would you or someone else take a look at the code before it gets merged?

### DIFF
--- a/pmb/chroot/mount.py
+++ b/pmb/chroot/mount.py
@@ -22,10 +22,31 @@ import pmb.parse
 import pmb.helpers.mount
 
 
+def mount_dev_tmpfs(args, suffix="native"):
+    """
+    Mount tmpfs inside the chroot's dev folder to make sure we can create
+    device nodes, even if the filesystem of the work folder does not support
+    it.
+    """
+    # Do nothing when it is already mounted
+    dev = args.work + "/chroot_" + suffix + "/dev"
+    if pmb.helpers.mount.ismount(dev):
+        return
+
+    # Create the folder structure and mount it
+    if not os.path.exists(dev):
+        pmb.helpers.run.root(args, ["mkdir", "-p", dev])
+    pmb.helpers.run.root(args, ["mount", "-t", "tmpfs",
+                                "-o", "size=1M,noexec,dev",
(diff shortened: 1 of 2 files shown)


pmb/chroot/mount.py
```diff
@@ -22,10 +22,31 @@
...
+    it.
+    """
+    # Do nothing when it is already mounted
+    dev = args.work + "/chroot_" + suffix + "/dev"
+    if pmb.helpers.mount.ismount(dev):
+        return
+
+    # Create the folder structure and mount it
+    if not os.path.exists(dev):
+        pmb.helpers.run.root(args, ["mkdir", "-p", dev])
```
z3ntu: `mkdir -p` does not fail if the folder already exists but I guess it's cheaper to not run it every time and check before creating the directory.

ollieparanoid: Uh, that command does not run in a chroot. But still, running commands is more expensive than simply checking with Python if the folder exists.

ollieparanoid:
Thanks for reviewing!

PR #1320, opened 2018-03-11
pmbootstrap newapkbuild: Properly parse arguments

ollieparanoid:
### Introduction
The `pmbootstrap newapkbuild` action wraps Alpine's `newapkbuild`. We
used to directly pass all arguments to `newapkbuild` without verifying
in Python whether they make sense or not. However, as `newpakbuild`
doesn't do strict sanity checks on the arguments, it is easy to end up
with unexpected behavior when using the command for the first time.

For example, `newapkbuild` allows either specifying a PKGNAME or SRCURL
as last parameter, and also allows setting a PKGNAME with the `-n`
parameter. It only makes sense to use that option when passing a
SRCURL.

### Change
With this commit, we duplicate the options that should be passed through
to `newapkbuild` and use argparse to fully sanitize the options and
display a help page (`pmbootstrap newapkbuild -h`) that is consistent
with the other help pages.

### Details
* The `-f` (force) flag does not get passed through anymore. Instead we
  use it in Python to skip asking if an existing aport should be
  overwritten (the aports are outside of the chroot, so `newapkbuild`
  can't handle it in a way that makes sense for pmbootstrap).
* Output of `newapkbuild` gets redirected to the log file now, as we
  don't need it to display a help page.
* Don't verify the pkgver while creating the new APKBUILD. When passing
  a SRCURL, the pkgver gets extracted from the end of the URL and may
  not have a valid format yet (but we want the APKBUILD anyway).
* Stored options passed through in `pmb/config/__init__.py` and use it
  in both `pmb/parse/arguments.py` and `pmb/helpers/frontend.py`.
* Only allow `-n` with SRCURL
* The postmarketOS aports folder gets specified with `--folder` now.
  That way the generated help page is much closer to the original one
  from `newapkbuild`. The default is `main`.
* Made the package type flags (CMake, autotools, ...) exclusive so only
  one of them can be specified

### Original help page
```
$ pmbootstrap -q chroot -- newapkbuild -h
newapkbuild 3.1.0-r4 - generate a new APKBUILD
Usage: newapkbuild [-n NAME] [-d DESC] [-l LICENSE] [-u URL]
       [-aCpy] [-s] [-cfh]
       PKGNAME[-PKGVER]|SRCURL
Options:
  -n  Set package name to NAME
  -d  Set package description (pkgdesc) to DESC
  -l  Set package license to LICENSE
  -u  Set package URL
  -a  Create autotools package (use ./configure ...)
  -C  Create CMake pakckage (Assume cmake/ is there)
  -m  Create meson package (Assume meson.build is there)
  -p  Create perl package (Assume Makefile.PL is there)
  -y  Create python package (Assume setup.py is there)
  -s  Use sourceforge source URL
  -c  Copy a sample init.d, conf.d, and install script to new directory
  -f  Force even if directory already exist
  -h  Show this help
```

### New help page
```
$ pmbootstrap newapkbuild -h
usage: pmbootstrap newapkbuild [-h] [--folder FOLDER] [-n PKGNAME]
                               [-d PKGDESC] [-l LICENSE] [-u URL]
                               [-a | -C | -m | -p | -y] [-s] [-c] [-f]
                               PKGNAME[-PKGVER] | SRCURL

positional arguments:
  PKGNAME[-PKGVER] | SRCURL
                        set either the package name (optionally with the
                        PKGVER at the end, e.g. 'hello-world-1.0') or the
                        download link to the source archive

optional arguments:
  -h, --help            show this help message and exit
  --folder FOLDER       set postmarketOS aports folder (default: main)
  -n PKGNAME            set package name (only use with SRCURL)
  -d PKGDESC            set package description
  -l LICENSE            set package license identifier from
                        <https://spdx.org/licenses/>
  -u URL                set package URL
  -a                    create autotools package (use ./configure ...)
  -C                    create CMake package (assume cmake/ is there)
  -m                    create meson package (assume meson.build is there)
  -p                    create perl package (assume Makefile.PL is there)
  -y                    create python package (assume setup.py is there)
  -s                    use sourceforge source URL
  -c                    copy a sample init.d, conf.d and install script
  -f                    force even if directory already exists
```

### DIFF
--- a/pmb/build/newapkbuild.py
+++ b/pmb/build/newapkbuild.py
@@ -1,5 +1,5 @@
 """
-Copyright 2017 Oliver Smith
+Copyright 2018 Oliver Smith
 
 This file is part of pmbootstrap.
 
@@ -24,7 +24,7 @@ import pmb.helpers.cli
 import pmb.parse
 
 
-def newapkbuild(args, folder, args_passed):
+def newapkbuild(args, folder, args_passed, force=False):
     # Initialize build environment and build folder
     pmb.build.init(args)
     build = "/home/pmos/build"
@@ -34,15 +34,14 @@ def newapkbuild(args, folder, args_passed):
     pmb.chroot.user(args, ["mkdir", "-p", build])
 
     # Run newapkbuild
-    pmb.chroot.user(args, ["newapkbuild"] + args_passed, log=False,
-                    working_dir=build)
+    pmb.chroot.user(args, ["newapkbuild"] + args_passed, working_dir=build)
     glob_result = glob.glob(build_outside + "/*/APKBUILD")
     if not len(glob_result):
         return
 
     # Paths for copying
     source_apkbuild = glob_result[0]
-    pkgname = pmb.parse.apkbuild(args, source_apkbuild)["pkgname"]
+    pkgname = pmb.parse.apkbuild(args, source_apkbuild, False)["pkgname"]
     target = args.aports + "/" + folder + "/" + pkgname
 
     # Move /home/pmos/build/$pkgname/* to /home/pmos/build/*
@@ -54,8 +53,8 @@ def newapkbuild(args, folder, args_passed):
     # Overwrite confirmation
     if os.path.exists(target):
         logging.warning("WARNING: Folder already exists: " + target)
-        if not pmb.helpers.cli.confirm(args, "Continue and delete its"
-                                       " contents?"):
+        question = "Continue and delete its contents?"
+        if not force and not pmb.helpers.cli.confirm(args, question):
             raise RuntimeError("Aborted.")
         pmb.helpers.run.user(args, ["rm", "-r", target])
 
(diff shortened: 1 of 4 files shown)


z3ntu:
The last argument (pkgname/srcurl) should be mandatory, right? Because it isn't right now.
```
$ ./pmbootstrap.py newapkbuild kde -d 'KWayland integration' -C -f
[15:24:02] Create /mnt/hdd/postmarketOS/pmbootstrap/aports/main/kde
[15:24:02] Done
```

ollieparanoid:
Damn, looks like it is still not intuitive :p

The postmarketOS aports folder has been moved to the `--folder` argument. So it used `kde` as `pkgname` in your example.

We could add a check that throws an error when you specify `kde` or `main` as `pkgname`. (On the other hand, I don't think many people have used the old syntax yet, and if you look at the generated package it becomes clear that it used `kde` as `pkgname`).

Thoughts on that?

z3ntu:
I copied the command from #1302 and didn't check if the syntax was different :)
I don't think such a check would be needed as, as you said, you will see that the generation didn't work as expected and you'll find out why.

bhush9:
I testted this, and it seems to work quite nicely

PR #1324, opened 2018-03-11
i9195/WiFi support

j0ki:
This package is based on `firmware-fairphone-fp2`
I extracted the firmware blobs directly from a Samsung Galaxy S4 Mini.
Unfortunately NetworkManager is currently not working, so it should be disabled:
`rc-update del networkmanager`
Connecting to access points is done with
`wpa_passphrase YOUR_SSID YOUR_PASSPHRASE >> /etc/wpa_supplicant/wpa_supplicant.conf`
The connection is automatically established by the wpa_supplicant service.
There are two wireless interfaces: `wlan0` and `p2p0`
wpa_supplicant uses `p2p0`, but the connection is reliable (I did a few reboots and some hours of ssh sessions).


aports/firmware/firmware-samsung-i9195-wlan/APKBUILD
```diff
@@ -0,0 +1,46 @@
+pkgname=firmware-samsung-i9195-wlan
```
drebrez: @j0ki you can add it as a subpackage of the existing `firmware-samsung-i9195` package.
just edit the subpackages line of the APKBUILD like this:
`subpackages="$pkgname-modem $pkgname-wlan"`
and add the corresponding function:
```
wlan() {
	pkgdesc="Samsung Galaxy S 4 Mini Wireless"
...
install....
...
}
```

aports/firmware/firmware-samsung-i9195-wlan/APKBUILD
```diff
@@ -1,4 +1,4 @@
-pkgname=firmware-samsung-i9195
+pkgname=firmware-samsung-i9195-wlan
 pkgver=18.03.11
 pkgrel=0
```
drebrez: @j0ki you can also add the `firmware-samsung-i9195-wlan` package to the `depends` variable of the `nonfree-firmware` subpackage of the device, here:
https://github.com/postmarketOS/pmbootstrap/blob/263578ec43c3d6b1cbb24d34495a93a1e3a3f6be/aports/device/device-samsung-i9195/APKBUILD#L26

aports/firmware/firmware-samsung-i9195-wlan/APKBUILD
drebrez: @ollieparanoid what you think about this comment? better to keep a single package which gather all firmware files and split them in subpackages or to have separated firmware packages (wifi, modem, ...)?

ollieparanoid: Long term it would be nice to have separated packaged for the specific components, so they could be shared between devices. But we don't handle that strictly yet, and having all firmware related to one device in one package is probably easier.

j0ki:
I made the suggested changes.
* `firmware-samsung-i9195-wlan` is now a subpackage of `firmware-samsung-i9195`.
* and a dependency to `nonfree-firmware`.

aports/device/device-samsung-i9195/APKBUILD
```diff
@@ -23,7 +23,7 @@ package() {
 nonfree_firmware() {
 	pkgdesc="Modem firmware"
```
drebrez: @j0ki can you please change the subpackage description to something like `Wifi, Modem firmware`?

aports/firmware/firmware-samsung-i9195/APKBUILD
```diff
@@ -1,17 +1,34 @@
 pkgname=firmware-samsung-i9195
 # XDA post explains the meaning of the firmware version, which I've translated into pkgver
-pkgver=201707.1
+pkgver=201707.1.1
```
drebrez: @ollieparanoid this `pkgver` was related to the modem firmware, now that the package contains multiple firmware should we change it back to a normal sequential numbering?

drebrez:
@j0ki nice work, I've added a few more comments and I'm trying to test it on my samsung i9195.
But since the display was broken, now that I've dismantled it, removed the display and reassembled, the kernel crashes at boot (I think trying to initialize the touchscreen)

aports/firmware/firmware-samsung-i9195/APKBUILD
ollieparanoid: sounds good to me!

j0ki: is it ok to use a lower major version number? I'm not sure how package management would handle this.

ollieparanoid: Good that you're asking about that. it doesn't handle it gracefully, and there isn't something that makes the lower pkgver override the bigger one (some distros have an "epoch" field for that).

So this may cause people who already built the old firmware package to get the old one instead of the new one. `pmbootstrap` will print a warning though, and since postmarketOS is not that stable and there won't be that many users of that specific device, I think this breaking change is acceptable.

aports/device/device-samsung-i9195/APKBUILD
j0ki: Yes, sorry. It is changed now.

aports/firmware/firmware-samsung-i9195/APKBUILD
j0ki: The new version number is `1.1.0`.

drebrez:
@j0ki can you check which model exactly do you have? because in the wiki it says `Samsung Galaxy S4 Mini LTE` and it uses the LOS kernel of the `MSM8930` which is, from my research, the samsung-i9190, instead the i9195 has a `MSM8916`.
Under the battery of the one I have there is: `MODEL: GT-9195`, maybe that's the reason it doesn't boot

j0ki:
@drebrez the id under the battery says: `MODEL : GT-I9195`, `FCC ID : A3LGTI9195` and `SSN : -I9195GSMH`. I hope this helps.

drebrez:
@j0ki ok I have exactly the same values under the battery, at least I know that it's the correct kernel/device.

Also nice research, so they are all `MSM8930AB` except the `GT-I9195I` (with the leading **i**) which have the same case but a different board (`MSM8916`).

aports/device/device-samsung-i9195/APKBUILD
```diff
@@ -21,9 +21,9 @@ package() {
 }
 
 nonfree_firmware() {
-	pkgdesc="Modem firmware"
+	pkgdesc="Firmware for Modem and Wifi"
```
drebrez: @j0ki please increase the `pkgrel` of this APKBUILD

drebrez:
@j0ki @ollieparanoid Everything looks good to me, I've tested it before the last change of increasing the `pkgrel` (force building it) and I was able to scan the wifi networks 👍

PR #1327, opened 2018-03-12
Set the hostname as the device name

drebrez:
During the `pmbootstrap install`, also set the hostname as the device name. Also update the `/etc/hosts` file to resolve the new hostname.

It will change it only if the actual files contains the default value specified by Alpine here: 
https://git.alpinelinux.org/cgit/aports/tree/main/alpine-baselayout/APKBUILD#n122

If future we may consider to prompt the user for the hostname, like almost every distribution installer does.

This also solves an issue with X11 not starting properly due the NetworkManager doing something with an hostname not set.
See here: https://bbs.archlinux.org/viewtopic.php?id=108695

### DIFF
--- a/pmb/install/_install.py
+++ b/pmb/install/_install.py
@@ -236,21 +236,12 @@ def setup_keymap(args):
 
 def setup_hostname(args):
     """
-    Set the hostname as the device name unless not modified already
+    Set the hostname and update localhost address on /etc/hosts
     """
     suffix = "rootfs_" + args.device
-
-    alpine_hostname = "localhost"
-    new_hostname = args.device
-    hostname = pmb.chroot.root(args, ["cat", "/etc/hostname"], suffix, return_stdout=True)
-    if hostname.strip() == alpine_hostname:
-        pmb.chroot.root(args, ["sh", "-c", "echo '" + new_hostname + "' > /etc/hostname"], suffix)
-
-    alpine_hosts = "127.0.0.1\tlocalhost localhost.localdomain"
-    new_hosts = "127.0.0.1\t" + args.device + " " + args.device + ".localdomain"
-    hosts = pmb.chroot.root(args, ["cat", "/etc/hosts"], suffix, return_stdout=True)
-    if hosts.strip() == alpine_hosts:
-        pmb.chroot.root(args, ["sh", "-c", "echo '" + new_hosts + "' > /etc/hosts"], suffix)
+    pmb.chroot.root(args, ["sh", "-c", "echo '" + args.hostname + "' > /etc/hostname"], suffix)
+    regex = "s/^127\.0\.0\.1.*/127.0.0.1\t" + args.hostname + " localhost.localdomain localhost/"
+    pmb.chroot.root(args, ["sed", "-i", "-e", regex, "/etc/hosts"], suffix)
 
 
 def install_system_image(args):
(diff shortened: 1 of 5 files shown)


pmb/install/_install.py
```diff
@@ -234,6 +234,25 @@ def setup_keymap(args):
...
+    suffix = "rootfs_" + args.device
+
+    alpine_hostname = "localhost"
+    new_hostname = args.device
+    hostname = pmb.chroot.root(args, ["cat", "/etc/hostname"], suffix, return_stdout=True)
+    if hostname.strip() == alpine_hostname:
+        pmb.chroot.root(args, ["sh", "-c", "echo '" + new_hostname + "' > /etc/hostname"], suffix)
+
+    alpine_hosts = "127.0.0.1\tlocalhost localhost.localdomain"
+    new_hosts = "127.0.0.1\t" + args.device + " " + args.device + ".localdomain"
```
ollieparanoid: `localhost` should still be in that line, see:
https://github.com/alpinelinux/alpine-conf/blob/b1febd985fe1b5cd183b19ee5e491fceda0e0a9a/setup-alpine.in#L179

drebrez: @ollieparanoid nice find, we can reuse that `sed` command to set the `127.0.0.1` line only and leave the rest as it is.

drebrez:
@ollieparanoid Thanks for the review, I agree that we can already implement that we prompt the user.
I'll add it to the `pmbootstrap init` then 😉 

Should we move also the question about the SSH public keys there? So that everything is in `init` and during the install it will prompt only the password? (addressable in a new PR)

pmb/install/_install.py
```diff
@@ -234,6 +234,25 @@ def setup_keymap(args):
         logging.info("NOTE: No valid keymap specified for device")
 
 
+def setup_hostname(args):
+    """
+    Set the hostname as the device name unless not modified already
```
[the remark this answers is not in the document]
drebrez: Yes probably

ollieparanoid:
> Should we move also the question about the SSH public keys there? So that everything is in `init` and during the install it will prompt only the password? (addressable in a new PR)

Yes, that makes sense!

drebrez:
@ollieparanoid comments addressed, please review it again with the new logic

drebrez:
@ollieparanoid I'm fine with your changes, but there is still something wrong I think. In my last commit I changed to always suggest the device name in the `pmb.helpers.cli.ask` call, this mean that if someone chose some different name, the next time he calls init, it doesn't suggest anymore what he configured but it defaults again to the device name.
I was thinking how can we solve it?
It should be something like this maybe?

> if `<configured hostname>` is in `<list of all devices>` and `<configured hostname>` != `<current device>`
then ask with default = `<current device>`

ollieparanoid:
You are right, I've missed that.
> (args.hostname or device)

Yep, that should do it!

drebrez:
@ollieparanoid ok, i've updated it.
But considering again the `pmbootstrap config hostname "'; rm -rf /; #"`, maybe we should extract the hostname validation logic into a helper function and call it from `setup_hostname` during the install.
Because having that `... rm -rf ...` in the `/etc/hostname` of the device, it may also be a problem. No?

ollieparanoid:
I agree, let's do this properly.

drebrez:
@ollieparanoid `pmb.helpers.other.validate_hostname()`?

PR #1329, opened 2018-03-13
 Fix: /var/cache/distfiles writable by everyone 

ollieparanoid:
As noted in commit 255c715, `/var/cache/distfiles` is writable by everyone. It is supposed to be writable only by `root` and by the `abuild` group (in which we put the `pmos` user already for building packages).

Changes:
* `pmb.build.init()`: make `/var/cache/distfiles` writable only by members of the `abuild` group (and root)
* Increase workfolder version to 2
* Add migration code that fixes the permissions for existing work folders
* Refactor the migration code a bit to make this possible


pmb/helpers/other.py
```diff
@@ -118,12 +119,26 @@ def migrate_work_folder(args):
...
+    if current == 1:
+        # Ask for confirmation
+        logging.info("Changelog:")
+        logging.info("* Fix: cache_distfiles was writable for everyone")
+        logging.info("Migration will do the following:")
+        logging.info("* Fix permissions of '" + args.work +
+                     "/cache_distfiles'")
+        if not pmb.helpers.cli.confirm(args):
+            raise RuntimeError("Aborted.")
+        current = 2
```
drebrez: @ollieparanoid `current = 2` is not necessary here

pmb/helpers/other.py
```diff
@@ -78,29 +78,30 @@ def check_binfmt_misc(args):
...
+    current = 0
     path = args.work + "/version"
     if os.path.exists(path):
         with open(path, "r") as f:
-            current = f.read().rstrip()
+            current = int(f.read().rstrip())
 
     # Compare version, print warning or do nothing
     required = pmb.config.work_version
     if current == required:
```
drebrez: @ollieparanoid what you think about adding a check if the current version is greater than the required?
Do the versions are backward compatibles? otherwise it's better to raise an error

ollieparanoid: You're right, it's better to have such a check!

ollieparanoid:
Thanks for the review! Updated.

drebrez:
@ollieparanoid tested again and works good 👍

ollieparanoid:
Thanks \o/

PR #1330, opened 2018-03-13
Qemu test case: wait dynamically until booted

ollieparanoid:
#### Introduction
We have a test case, that performs full installations with with XFCE4
and Plasma Mobile. After each installation is through, we run
`pmbootstrap qemu` and connect to the VM via SSH to check for specific
running processes.

#### Change
Before this commit, we assumed that the boot takes 90 seconds at most
and then the processes should exist. But sometimes the boot is faster
(so Travis waits just doing nothing) and sometimes it is slower (which
causes the test case to fail), all depending on the current workload
of the Travis CI servers. Now we try to find the processes we're
interested in each second for 3 minutes until we either succeed or
reach the time limit.

### DIFF
--- a/test/test_qemu_running_processes.py
+++ b/test/test_qemu_running_processes.py
@@ -31,7 +31,6 @@ import sys
 import shutil
 import shlex
 import time
-import logging
 
 # Import from parent directory
 pmb_src = os.path.realpath(os.path.join(os.path.dirname(__file__) + "/.."))
@@ -113,8 +112,6 @@ class Qemu(object):
         pmbootstrap_yes(args, config, ["install", "--no-fde"])
         self.process = pmbootstrap_run(args, config, ["qemu", "--display",
                                                       "none"], background=True)
-        logging.info("(test) wait 90s until the VM booted up")
-        time.sleep(90)
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


test/test_qemu_running_processes.py
```diff
@@ -159,7 +177,7 @@ def test_xfce4(args, tmpdir, qemu):
                              "Thunar", "dbus-daemon", "xfwm4"])
 
     # self-test of is_running()
-    assert is_running(args, ["invalid-process"]) is False
+    assert is_running(args, ["invalid-process"], 1) is False
```
z3ntu: Is the try of 1 on purpose? (I'm not familiar with the tests so I might very well be wrong)

drebrez: I think it should be fine to try only once there, since few lines above just called `is_running` with the default values (180 retries), at this point qemu should be up and running

ollieparanoid: @drebrez is right, the above test waits until it is up and running, and then we check once if the "invalid-process" can *not* be found, to make sure that the process finding logic works and doesn't just always return `True`.

ollieparanoid:
Thanks guys for having a look at it! Is there anything else worth discussing (e.g. do you think we should comment the `is_running(args, ["invalid-process"], 1)` to make it more obvious) or is it ready to be shipped? :ship:

PR #1331, opened 2018-03-13
Travis CI: folded output/default tracebacks

### LINKED ISSUE #1255: Fold Travis output
Right now, the travis output is pretty unreadable. It would be nice if we had folding for at least every command in the `script` section of our `travis.yml`:
https://github.com/postmarketOS/pmbootstrap/blob/2f78dad318d5c28e466abff895f4ade9cf0aaccb/.travis.yml#L11-L16

Bonus points for also folding builds of individual packages in `test/check_checksums.py --build` (like [Alpine](https://github.com/alpinelinux/aports/tree/master/.travis) does it).

ollieparanoid:
### Changes
* Moved the `script` and `after_failure` sections of `.travis.yml`
  to extra files in a new `.travis/` folder
* Copy paste `.travis/common.sh` from Alpine's aports. They have neat
  formatting functions in there for folding
* Add a folding block around each script that gets called
* Use native tracebacks in pytest. These are the same as we see when
  pmbootstrap crashes usually, instead of printing out the entire
  function that causes an error. I think this makes the error output
  easier to read.

EDIT: Closes #1255.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,15 +9,10 @@ install:
   - sudo apt install shellcheck qemu-system-x86
   - pip install flake8 pytest-cov python-coveralls
 script:
-  - test/static_code_analysis.sh
-  - yes "" | ./pmbootstrap.py init
-  - ./pmbootstrap.py kconfig_check
-  - test/check_checksums.py --build
-  - test/testcases_fast.sh --all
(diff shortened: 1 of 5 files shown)


z3ntu:
Well you should mention #1255 somewhere :)

Issue #1333, opened 2018-03-14
busybox: tar applet fails to extract (some) symlinks


alive4ever:
After upgrading busybox to version 1.28, busybox tar applet fails to extract
(some) symlinks.

With my recycled device ([armani](https://wiki.postmarketos.org/wiki/Xiaomi_Redmi_1S_(xiaomi-armani))/armhf), the failed symlinks extraction occurs
with busybox applets inside /bin directory. There are no symlinks to busybox
applets at all here. Even important applet such as /bin/ash and /bin/sh are not
available, causing `chroot` inside the debug-shell to fail unless a specific
command is specified, such as `/bin/bash` or `/bin/busybox sh`.

```
/tmp/postmarketos/chroot/mnt/pmOS # ls -F bin/
bash*           findmnt*        lsmod@          mountpoint*     umount*
bbsuid*         groups*         mksh*           rc-status*      uniso*
busybox*        kmod*           modinfo@        su*             wdctl*
busybox-extras* login*          more*           tar*
dmesg*          lsblk*          mount*          udevadm*
```
As comparison, here are files present inside device chroot (`pmbootstrap chroot -r`)
```
/ # ls -F bin/
arch@            date@           fsync@     linux32@   more*           rc-status*  sync@
ash@             dd@             getopt@    linux64@   mount*          reformime@  tar*
base64@          df@             grep@      ln@        mountpoint*     rev@        touch@
bash*            dmesg*          groups*    login*     mpstat@         rm@         true@
bbconfig@        dnsdomainname@  gunzip@    ls@        mv@             rmdir@      udevadm*
bbsuid*          dumpkmap@       gzip@      lsblk*     netstat@        run-parts@  umount*
busybox*         echo@           hostname@  lsmod@     nice@           sed@        uname@
busybox-extras*  ed@             ionice@    lzop@      pidof@          setpriv@    uniso*
cat@             egrep@          iostat@    makemime@  ping@           setserial@  usleep@
chgrp@           false@          ipcalc@    mkdir@     ping6@          sh@         watch@
chmod@           fatattr@        kbd_mode@  mknod@     pipe_progress@  sleep@      wdctl*
chown@           fdflush@        kill@      mksh*      printenv@       stat@       zcat@
conspy@          fgrep@          kmod*      mktemp@    ps@             stty@
cp@              findmnt*        link@      modinfo@   pwd@            su*
```

Another noticeable failure is missing `/sbin/init` symlink, which causes
[...]

PR #1339, opened 2018-03-15
Fix regex in recovery installer

ata2001:
It was able to parse the fstab on sony-amami.


lduboeuf:
not sure if i'm doing right for testing. Anything to do on device ?

I fetched the pull request locally and switched to the PR branch,
pmbootstrap zap
pmbootsrap init
pmbootstrap install --no-fde --android-recovery-zip
install android-recovery.zip via TWRP
Same error occured: https://bpaste.net/show/36b2948edd95

ollieparanoid:
Thanks for making the PR @ata2001 and for testing @lduboeuf!
In your log, there's this line:
> tar: skipping unsafe symlink to '/bin/busybox' in archive, set EXTRACT_UNSAFE_SYMLINKS=1 to extract

Which is bug #1333, and the fix for it was just merged to `master` (but it is not included in this PR yet).

@lduboeuf: you could try if the version from `master` works if you want to, I'm not sure if the fix from this PR is still needed (but @ata2001 could probably answer this). If the version from `master` does not work, we need to rebase this PR on master and then try it again.

lduboeuf:
no more `dd: can't open '/dev/block/mmcblk0p5` with master but still have 

> tar: skipping unsafe symlink to '/bin/busybox' in archive, set EXTRACT_UNSAFE_SYMLINKS=1 to extract

but maybe i missed something because i have lines like this:

> Error: Partition(s) 1, 2, 3, ... on /dev/block/mmcblk0p9 have been written, but we have been unable to inform the kernel of the change, probably because it/they are in use.  As a result, the old partition(s) will remain in use.  You should reboot now before making further changes.
> + parted -s /dev/block/mmcblk0p9 'set 1 boot on'

drebrez:
@ollieparanoid I think that here it's not a `tar` related problem. It was discovering 2 partition for the `/boot`, see old log:
```
+ awk /^\/boot/ {print $3} /recovery.fstab
+ _BOOT_PARTITION=/dev/block/mmcblk0p5
/dev/block/mmcblk0p2
+ [ ! -z /dev/block/mmcblk0p5
/dev/block/mmcblk0p2 ]
+ echo boot partition found at /dev/block/mmcblk0p5
/dev/block/mmcblk0p2
boot partition found at /dev/block/mmcblk0p5
/dev/block/mmcblk0p2
+ export BOOT_PARTITION=/dev/block/mmcblk0p5
/dev/block/mmcblk0p2
```
@ata2001 rewrote the `awk` command and in the last log that @lduboeuf posted it seems fixed.
@lduboeuf the error about unable to inform the kernel is misleading, it is more a warning than an error. You can ignore it.

ata2001:
Now that I've merged master, the tar error should be gone too. As drebrez said, just ignore the parted error, it gets printed on most devices even if it successfully creates the partitions.

@ollieparanoid This patch is still necessary, the old regex matched everything that starts with /boot.

PR #1346, opened 2018-03-18
New device: Xiaomi Redmi 4X (xiaomi-santoni)

pinoaffe:
First of all: thank you to everyone who helped me during the porting,
it was quite a lot of fun to port pmos to my phone (xiaomi redmi 4x aka santoni).

What's working:
- screen
- usb otg
- flashing using fastboot

What does not work:
- usb networking
- wlan
- touch
- everything else

It runs on the recently released official android kernel for the device, with a few patches to make it compile.
Weston works, kde plasma works as well but is slow and seems buggy (the relative size of things on the screen often makes no sense at all, ui elements overlap, etc).
I'm going to try to get at the very least touch and usb networking to work, but have no clue as to whether I'll manage.


aports/device/linux-xiaomi-santoni/APKBUILD
```diff
@@ -0,0 +1,86 @@
...
+pkgver=3.18.31
+pkgrel=16
+pkgdesc="Xiaomi Redmi 4X kernel fork"
+arch="aarch64"
+_carch="arm64"
+_flavor="xiaomi-santoni"
+url="https://kernel.org"
+license="GPL2"
+options="!strip !check !tracedeps"
+makedepends="perl sed installkernel bash gmp-dev bc linux-headers elfutils-dev git"
```
ollieparanoid: git is usually not required for building the kernel - the message about git errors during the build is a bug in abuild which will be fixed some time soon (we already submitted a PR). Could you try if it works without git?

aports/device/device-xiaomi-santoni/APKBUILD
```diff
@@ -0,0 +1,22 @@
+# Reference: <https://postmarketos.org/devicepkg>
+pkgname="device-xiaomi-santoni"
+pkgdesc="Xiaomi Redmi 4X"
+pkgver=0.1
+pkgrel=3
```
ollieparanoid: could you reset that to 0? (If you want to force a build without increasing the `pkgrel`, you can use `pmbootstrap build --force`)

aports/device/linux-xiaomi-santoni/APKBUILD
```diff
@@ -0,0 +1,86 @@
+# Kernel config based on: arch/arm64/configs/S88536AA2-perf_defconfig
+
+pkgname="linux-xiaomi-santoni"
+pkgver=3.18.31
+pkgrel=16
```
ollieparanoid: `pkgrel=0`

aports/device/linux-xiaomi-santoni/APKBUILD
pinoaffe: yeah it does work without git, but because of the warnings I added it as a dependency, will remove again.

PR #1348, opened 2018-03-19
Kernel updates

z3ntu:
Test results with weston:
* mainline: doesn't work anymore - regression upstream?
* stable: works
* lts: works


ollieparanoid:
Thanks for making the PR and for testing everything. Now we know about the regression beforehand, that's good. My plan to fix this is in #1352.

Before I merge this, I recommend to revert the `linux-postmarketos-mainline` update for now and as usually I'll build it once on the repo PC to fill up ccache.

z3ntu:
I reverted the last commit and with 4.16_rc5 Weston starts without any problem.

Issue #785, opened 2017-10-17
pmbootstrap/pmOS should offer a 'debug mode'


craftyguy:
This debug mode would produce a pmOS install that
-  disables splash screens on boot
-  prints kernel and openrc messages to screen (pass `PMOS_NO_OUTPUT_REDIRECT` kernel cmdline)
-  disables fde
-  adds the usb hook to initramfs. 
-  pmbootstrap presents a big fat warning that pmos is running in 'debug mode' and should not be used for production
-  modify /etc/os-release to indicate that it's pmOS-debug or something
-  This would be a (non-default) parameter to `pmbootstrap install`, like `--debug-mode` or `-d` for short (assuming `d` isn't already taken.. I didn't look)
-  Install some extra tools to help with debugging:
   - evtest

Anything else that would be useful? Any additional tools that could be installed automatically when this is enabled to assist folks with porting/debugging their device?

drebrez:
@craftyguy nice idea, here some ideas that can be added to the list:

- do not install any UI
- add some kernel cmdline to have a more verbose kernel log 
- install all possible -dbg subpackages?
- possible extra tools:
   - gdb-server (not sure it works for armhf)
   - strace
   - sysstat (performance monitoring tools)
   - procps
   - nmap

craftyguy:
@drebrez great suggestions! I have combined your list with the original one in my comment to make it easier to parse out the 'specification' for this request!

ollieparanoid:
Cool idea!

> prints kernel and openrc messages to screen

I've tested that about half a year ago with my lg-mako and it was basically unusable. it took 5 or so? minutes until it printed anything to the screen (I randomly left it powered on but I thought the image didn't even boot). So I think we can't enable that for all devices, at least not with the standard stuff from the kernel. But that may also be `lg-mako` specific and it may work fine with other Androids.

craftyguy:
@ollieparanoid 
>     prints kernel and openrc messages to screen

Yea I was thinking this might actually just be setting the `PMOS_NO_OUTPUT_REDIRECT`, since that allows for quite a few useful messages to be printed to screen. Is that what you did in your test with Mako, or did you set things like `debug` on the kernel cmdline? That would slow down all but the beefiest systems.

Oh, maybe adding `earlyprintk` would be useful too!

ollieparanoid:
[...]

PR #1358, opened 2018-03-21
Add additional files into initramfs when debug-shell is installed

drebrez:
This PR uses the same approach as the Alpine's mkinitfs (also discussed in https://github.com/postmarketOS/pmbootstrap/issues/1188#issuecomment-363897670) to add the possibility to the `hooks` to include additional binaries to the `initramfs`.

The `mkinitfs` script search the `*.files` in the path `/etc/postmarketos-mkinitfs/files/` and each line of those files is a binary file that has to be added to the `initramfs`.

Right now only the `debug-shell` hook uses this approach to add some files mentioned in #785 that are useful for the kernel mainlining process.


aports/main/postmarketos-mkinitfs-hook-debug-shell/APKBUILD
```diff
@@ -1,16 +1,20 @@
...
 license="GPL2"
+options="!check"
 
 package() {
 	mkdir -p "$pkgdir"/etc/postmarketos-mkinitfs/hooks/
 	install -Dm644 "$srcdir"/20-debug-shell.sh \
-		"$pkgdir"/etc/postmarketos-mkinitfs/hooks/
+		"$pkgdir"/etc/postmarketos-mkinitfs/hooks/20-debug-shell.sh
+	install -Dm644 "$srcdir"/debug-shell.files \
+		"$pkgdir"/etc/postmarketos-mkinitfs/files/debug-shell.files
```
ollieparanoid: What do you think about naming the file list files like the hook, so this would be `20-debug-shell.files`?

aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -111,9 +111,13 @@ get_binaries()
 	if [ "${deviceinfo_msm_refresher}" == "true" ]; then
 		BINARIES="${BINARIES} /usr/sbin/msm-fb-refresher"
 	fi
-	if [ -f /usr/sbin/fbdebug ]; then
-		BINARIES="${BINARIES} /usr/sbin/fbdebug"
-	fi
+	for file in "/etc/postmarketos-mkinitfs/files"/*.files; do
+		[ -f "$file" ] || continue
+		while IFS= read -r line; do
+			[ -f "$line" ] || continue
```
ollieparanoid: The hooks should always have all files mentioned in their lists, because they depend on the packages providing these files, right?
That means if a file mentioned in a line does not exist, this is a bug in the packaging or in the file list. So I would `exit 1` with an error message here instead of silently ignoring it.

drebrez: I've tried like you suggested, to return an empty string and check the empty string in `copy_files`.
At the beginning it was solved, but a lot of problems started to raise when testing `pmbootstrap initfs hook_add debug-shell` with a non-existing file in the "list". 
Something related to the packages installed but the trigger failed and also the `.pmbootstrap` virtual package still present.
I've tried to add a try/catch in `pmb.chroot.apk.install` and call `apk del .pmbootstrap` in case of error but it keeps triggering the `postmarketos-mkinitfs`'s post-install script and raise some other errors.
Patch: https://pastebin.com/NeVcEvMF

ollieparanoid: Thanks for trying it out! Looks like I didn't think this through. I've pushed an alternative version, that puts the check into a new function (so we can simply `exit` from there) and runs it at the beginning. What do you think?

ollieparanoid:
@z3ntu and @opendata26:
`kmscube` pulls in quite a few dependencies (`mesa-egl`, `mesa-gbm`, `libdrm`, `libx11`, `libxcb`, `libxau`, `libxdmcp`, `libbsd`), and if it was added to the initramfs, it increases the size about 3 MB (tested with armhf). Since we've ran into size problems with some devices, I would not recommend to include it by default.

But now with the `*.files` mechanism in the initramfs, it would be relatively easy to add it on demand:
```shell
pmbootstrap chroot -- mkdir -p /etc/postmarketos-mkinitfs/files
pmbootstrap chroot -- sh -c "echo /usr/bin/kmscube > /etc/postmarketos-mkinitfs/files/custom.files"
pmbootstrap initfs
```
We could add this to the mainlining guide once this PR is merged.

drebrez:
@ollieparanoid I think you solution it's the only possible, checking it in the subshell it's too complicated.
I've tested it and I think there is a problem, because it fails in the trigger and the packages remains installed. Trying to remove the hook with `pmbootstrap initfs hook_del debug-shell` after it failed it gives these errors:
```
(012040) [17:40:26] (rootfs_samsung-klte) % apk del postmarketos-mkinitfs-hook-debug-shell
World updated, but the following packages are not removed due to:
  postmarketos-mkinitfs-hook-debug-shell:
                                          .pmbootstrap
```
Executing `pmbootstrap chroot -r -- apk del .pmbootstrap` seems to cleanup a bit

ollieparanoid:
If we wanted to catch that, I think we would need to wrap every installation of the hook with Python logic in `pmbootstrap` that uninstalls the package again if `apk` retuns `-1` (because the trigger failed). The hook could be installed in multiple ways, such as `pmbootstrap install`, `pmbootstrap initfs`, `pmbootstrap chroot --add=...`.

Honestly I don't think this is feasible. On the other hand, as soon as a trigger failed, `apk` stores this error, and you need to run `apk fix` to run the trigger again and to resolve it. Which means just about every `apk` command we run in the chroot will return an error, and that will people make `zap` the chroot.

So I would keep it that way and expect from users to run `zap` if their package made the trigger exit with error. Thoughts on that?

drebrez:
@ollieparanoid I agree that it's quite complicated to handle that case. So, for me it's ok to keep it like this 👍

PR #1360, opened 2018-03-24
Add support for GT-S756x (kylessopen, Galaxy Trend/S Duos/Ace II x)

racer123:
Hello!
Here's what I've done to port postmarketOS to kylessopen (also known as GT-S7560, GT-S7562, Galaxy Trend, S Duos, Ace II x).
Currently it's booting, ssh over usb and display is working (with weston, xfce4 didn't work.) Touchscreen works also, Wifi isn't working yet.

PR #1361, opened 2018-03-25
 Add support for GT-S756x (kylessopen, Galaxy Trend/S Duos/Ace II x) 

racer123:
(I accidentally closed #1360 by re-creating this branch, so I am making a new PR)
Hello!
Here's what I've done to port postmarketOS to kylessopen (also known as GT-S7560, GT-S7562, Galaxy Trend, S Duos, Ace II x).
Currently it's booting, ssh over usb and display is working (with weston, xfce4 didn't work.) Touchscreen works also, Wifi needs additional firmware but seems to work after adding those.


aports/device/device-samsung-kylessopen/deviceinfo
```diff
@@ -0,0 +1,37 @@
...
+deviceinfo_modules_initfs=""
+deviceinfo_external_storage="true"
+deviceinfo_arch="armhf"
+
+# Device related
+deviceinfo_keyboard="false"
+#deviceinfo_external_disk="true"
+deviceinfo_screen_width="480"
+deviceinfo_screen_height="800"
+deviceinfo_dev_touchscreen="/dev/input/event0"
```
PureTryOut: So this works after all?

racer123: Yes, I'm not sure why it didn't before but now it works. Though, I suspect that msm-fb-refresher is now causing random reboots.

PR #1363, opened 2018-03-26
pmbootstrap init: kernel selection

### LINKED ISSUE #1263: pmbootstrap init: Allow switching between mainline and downstream kernel (for devices that have support for both)
@MayeulC wrote in https://github.com/postmarketOS/pmbootstrap/issues/756#issuecomment-368332143:
> [...] idea regarding multiple kernels: maybe pmbootstrap will be able to offer the user a choice at some point; I think the way pacman handles multiple providers is nice, and such a mechanism could be unified for choosing between DE, kernel and other components.

While we can't implement that at the package manager level (`apk`) that easily, we can do it in `pmbootstrap init`. We could use the same mechanism with subpackages to choose the different kernels like we do from now on with the `nonfree` packages (#1254).

ollieparanoid:
### Example output
```
[17:12:27] Device [qemu-amd64]: sony-amami
[17:12:29] Which kernel do you want to use with your device?
[17:12:29] Upstream kernels (mainline, stable, lts) get security updates, but may have less working features than downstream kernels.
[17:12:29] Available kernels (2):
[17:12:29] * downstream: Display and wifi works (see device table for details)
[17:12:29] * mainline: For kernel development only (most features aren't working)
[17:12:29] Kernel [downstream]:
```

### Details
* new config option `"kernel"` with possible values:
  `"downstream", "mainline", "stable", "lts"` (downstream is always
  `linux-$devicename`)
* ask for the kernel during `pmbootstrap init` if the device package
  has kernel subpackages and install it in `_install.py`
* postmarketos-mkinitfs: display note instead of exit with error when
  the `deviceinfo_dtb` file is missing (because we expect it to be
  missing for downstream kernels)
* device-sony-amami:
  * add kernel subpackages for downstream, mainline
  * set `deviceinfo_dtb`
* device-qemu-amd64: add kernel subpackages for stable, lts, mainline
* test cases and test data for new functions
* test case that checks all aports for right usage of the feature:
  * don't mix specifying kernels in depends *and* subpackages
  * 1 kernel in depends is maximum
  * kernel subpackages must have a valid name

Closes #1263. When this is merged, I will update the [device specific package wiki page](https://wiki.postmarketos.org/wiki/Device_specific_package#devicepkg_build).


pmb/parse/_apkbuild.py
```diff
@@ -209,3 +209,37 @@ def subpkgdesc(path, function):
...
+    subpackages = apkbuild(args, apkbuild_path)["subpackages"]
+
+    # Read kernels from subpackages
+    ret = {}
+    subpackage_prefix = "device-" + device + "-kernel-"
+    for subpackage in subpackages:
+        if not subpackage.startswith(subpackage_prefix):
+            continue
+        type = subpackage[len(subpackage_prefix):]
+        func = "kernel_" + type
```
z3ntu: not really a problem but type is a built in function (see https://stackoverflow.com/a/10568115/3527128)
might be better to call it ktype or so :)

z3ntu:
also maybe add a testcase to check if one kernel is added at all if we're at it?

ollieparanoid:
> also maybe add a testcase to check if one kernel is added at all if we're at it?

Good idea!

> Also I remembered, potentially different cmdlines are needed for different kernels (e.g. on mainline I use `cma=64m msm.vram=16m` and downstream `console=ttyHSL0,115200,n8`)

I have two solutions for this, and I recommend to pick the appropriate one when we hit this case:
1. package the `deviceinfo` in the kernel subpackage, so we can have 2 versions (one for upstream kernels and one where we replace the kernel commandline with sed or something)
2. add two kernel cmdline variables to the deviceinfo, one for upstream and one for downstream kernels.

craftyguy:
the term 'downstream' may be confusing to new folks, though I'm not sure what else to call it. Maybe have a short description in init.py stating that this is "typically an older kernel from Android" or something?

ollieparanoid:
@z3ntu: See my answer above to the `deviceinfo` problem. Thoughts?

@craftyguy: Good point. I've adopted the term from various talks on mainlining, which talk about upstream and downstream kernels. We could explain the term right above the "available kernels", would that work?

craftyguy:
@ollieparanoid sound good :)

drebrez:
@ollieparanoid FYI, also for the Samsung S5 I have different `cmdline` and QCDT should be set only for downstream kernel.
Solution 1.) simple but the actual cmdline will be stored in the APKBUILD instead of the `deviceinfo` file.
Solution 2.) Requires additional changes in the various commands using it
What about a solution similar to 2. by having both lines in `deviceinfo`, e.g.:
```
deviceinfo_kernel_cmdline_mainline="cma=64m msm.vram=16m"
deviceinfo_kernel_cmdline_downstream="console=ttyHSL0,115200,n8"
```
but in the subpackage installation just remove and rename all the other lines? That way we can have a common logic between all the APKBUILDs

ollieparanoid:
@drebrez: Wow, that sounds like one really nice solution :tada:

NotKit:
The change is going to be useful for Droid 4 as well, since currently it has downstream kernel packages, but there is mostly working mainline kernel for it.

ollieparanoid:
I've put @drebrez proposal into a new issue #1368 and addressed the review comments, as well as deleted the `lts` kernel (as discussed in IRC yesterday - we don't really use it, and updates come in about once a week which is a lot maintenance extra effort). If someone has time, it would be nice to review it again and test it. Thanks!

aports/main/linux-postmarketos-lts/APKBUILD
```diff
@@ -1,198 +0,0 @@
-_flavor=postmarketos-lts
```
drebrez: @ollieparanoid `linux-postmarketos-lts` is still used by the `device-qemu-vexpress` package

aports/device/device-qemu-amd64/APKBUILD
```diff
@@ -29,5 +35,17 @@ weston() {
 		"$subpkgdir"/etc/xdg/weston/weston.ini
 }
 
+kernel_stable() {
+	pkgdesc="Stable for everyday usage (recommended)"
+	depends="linux-postmarketos-mainline"
```
drebrez: @ollieparanoid depends on the wrong `linux-postmarketos` package, should be `linux-postmarketos-stable`

ollieparanoid:
@drebrez: thanks for finding these bugs, updated the PR!

drebrez:
@ollieparanoid I get an error:
(device: soni-amami, kernel: mainline)
```
drebrez@drebrez-pc:~/Development/pmbootstrap$ pmbootstrap initfs build
[00:38:53] ERROR: list index out of range
```
```
  File "/home/drebrez/Development/pmbootstrap/pmb/__init__.py", line 61, in main
  File "/home/drebrez/Development/pmbootstrap/pmb/helpers/frontend.py", line 170, in initfs
    pmb.chroot.initfs.frontend(args)
  File "/home/drebrez/Development/pmbootstrap/pmb/chroot/initfs.py", line 98, in frontend
    flavor = pmb.chroot.other.kernel_flavor_autodetect(args, suffix)
  File "/home/drebrez/Development/pmbootstrap/pmb/chroot/other.py", line 43, in kernel_flavor_autodetect
    return kernel_flavors_installed(args, suffix)[0]
IndexError: list index out of range
```

ollieparanoid:
@drebrez: Thanks for your incredible patience and for all the testing! \o/

PR #1365, opened 2018-03-27
New device: hTC Incredible S (htc-vivo)

Aerotinge:
What's working:

screen
flashing using fastboot
USB networking

What does not work:

boot (something broken in the kernel I guess, [dmesg here](https://pastebin.com/pWNbBPvq))
everything else

I'll try to switch a fork based on OpenDesireProject msm7x30 kernel. If this kernel is helpless...
Or should I use legacy stock kernel releases?


Aerotinge:
> What exactly does not work while booting? I guess you've obtained the dmesg by booting a known working ROM afterwards and reading /proc/last_kmesg or something?
The kernel boots, and I can ssh into the device, getting dmesg.
But Weston never comes up. All I have is a pmOS loading splash.
Not knowing what to do next...

mirh:
https://github.com/rqmok/android_kernel_htc_msm7x30 maybe this could be better than opendesire

Aerotinge:
> https://github.com/rqmok/android_kernel_htc_msm7x30 maybe this could be better than opendesire

I used to build LineageOS 13 with this. yap this is working but it's my last choice.
Thanks anyway.

> Would it be possible that you move it to the upper table, too?
> Regarding the weston-doesn't-show-up-issue: what does the syslog say? (run logread)

OK,I'll move it to the upper table right after pr has been accepted.
[logread(Weston)](https://bpaste.net/show/5d4a55a8574a) & [logread(none)](https://bpaste.net/show/0d24c07e3fd8) here.
Neither of them works. 💀

ollieparanoid:
Thanks! Merging as soon as Travis runs through.

The "none" UI will just hang at the Loading... screen, that is expected (since there's no UI that will start up). However, for `weston`, this seems to be the problem:
```
Jan  1 15:59:59 vivo user.notice user:weston: [02:58:38.241] Calculating pixman format from:
Jan  1 15:59:59 vivo user.notice user:weston:                 - type: 0 (aux: 0)
Jan  1 15:59:59 vivo user.notice user:weston:                 - visual: 2
Jan  1 15:59:59 vivo user.notice user:weston:                 - bpp: 32 (grayscale: 0)
Jan  1 15:59:59 vivo user.notice user:weston:                 - red: offset: 0, length: 8, MSB: 0
Jan  1 15:59:59 vivo user.notice user:weston:                 - green: offset: 8, length: 8, MSB: 0
Jan  1 15:59:59 vivo user.notice user:weston:                 - blue: offset: 16, length: 8, MSB: 0
Jan  1 15:59:59 vivo user.notice user:weston:                 - transp: offset: 0, length: 0, MSB: 0
Jan  1 15:59:59 vivo user.notice user:weston:                 - pixman_type: 0
Jan  1 15:59:59 vivo user.notice user:weston: [02:58:38.246] Type guessed: 0:
Jan  1 15:59:59 vivo user.notice user:weston: [02:58:38.247] Frame buffer uses an unsupported format.
Jan  1 15:59:59 vivo user.notice user:weston: [02:58:38.248] Failed to get frame buffer info: Invalid argument
Jan  1 15:59:59 vivo user.notice user:weston: [02:58:38.265] Creating frame buffer failed.
Jan  1 15:59:59 vivo user.notice user:weston: [02:58:38.273] fatal: failed to create compositor backend
```

I think your kernel is reporting the wrong framebuffer information. To work around this, you could try to force a "pixman mode" (see `deviceinfo_weston_pixman_type` in [deviceinfo reference](https://wiki.postmarketos.org/wiki/Deviceinfo_reference#weston)) or you could [patch your kernel](https://wiki.postmarketos.org/wiki/Troubleshooting:display#Screen_output_messed_up) (when done right, this should make it work for all interfaces, not only for weston). @drebrez even made a framebuffer debug program, so he might be able to tell you more.

PR #1366, opened 2018-03-27
New device: oneplus oneplus2

lduboeuf:
device boot (don't see splash screen yet), network is up


lduboeuf:
humm, since i've added "postmarketos-base" tried a fresh install zap->install->flash and can't connect ssh anymore ( not a ssh key pb)

lduboeuf:
ok, i can ssh again , did via recovery installation using twrp... i really need to learn more about all that :)

lduboeuf:
now, device boot, see spash screen and weston is loaded. no touch atm

lduboeuf:
travis complained : "ERROR: [Errno 104] Connection reset by peer", can't reload build job

NotKit:
Please remove aports/device/linux-oneplus-oneplus2/fix-qcacld-ident.patch, this patch is unused and isn't listed in APKBUILD now.

lduboeuf:
well seems that this [patch](https://github.com/postmarketOS/pmbootstrap/blob/7ff0568d6fec60ff9cb8f84d7376a12e4f147cf9/aports/device/linux-oneplus-onyx/00_mt_width_major_params.patch) get the idea but don't know which driver is used by this device

NotKit:
This patch should help, although I removed ABS_MT_WIDTH_MAJOR bit instead since no value gets provided: https://gist.github.com/NotKit/d0acce00eea9d80512d541156278eda4

Regarding your udev rule, maybe it would be better to use driver name instead of device path?

drebrez:
@NotKit FYI since we have the device path of the touchscreen in the `deviceinfo` (necessary for osk-sdl), we are generating the udev rule automatically using the `devicepkg-dev` package.
This simplify a lot the porting of new devices and most of the times the touchscreen works out of the box with the generated udev rule.

lduboeuf:
ok added device to wiki

ollieparanoid:
Thanks again!

lduboeuf:
is it fine to delete branch ?

ollieparanoid:
yep :smiley:

PR #1369, opened 2018-03-27
Travis: Check for device documentation

ollieparanoid:
Add a simple script, that checks whether all device's codenames can be found in the device overview in the wiki. This way we're always sure that every device is documented in the wiki's device table and it makes reviews a lot easier (therefore more time for coding :wink:).

Success output:
https://travis-ci.org/postmarketOS/pmbootstrap/builds/359092960#L643

### How to test
Preparation:
```
$ yes "" | pmbootstrap aportgen device-not-documented
$ yes "" | pmbootstrap aportgen device-apple-ipt1g
$ test/check_devices_in_wiki.py -h
usage: check_devices_in_wiki.py [-h] [--booting] [--path PATH]

optional arguments:
  --booting    devices must be in the upper table, being in the 'not booting'
               table below is not enough (all devices in pmbootstrap master
               should be in the upper table)
  --path PATH  instead of downloading the devices page from the wiki, use a
               local HTML file
```

Test if the devices are in the wiki at all:
```
$ test/check_devices_in_wiki.py
not-documented: not in the wiki yet.
*** Wiki check failed!
Thank you for porting postmarketOS to a new device! \o/

Now it's time to add some documentation:
1) Create a device specific wiki page as described here:
   <https://wiki.postmarketos.org/wiki/Help:Device_Page>
2) Add your device to the overview matrix:
   <https://wiki.postmarketos.org/wiki/Devices>
3) Run these tests again with an empty commit in your PR:
   $ git commit --allow-empty -m 'run tests again'

Please take the time to do these steps. It will make your
precious porting efforts visible for others, and allow them
not only to use what you have created, but also to build upon
it more easily. Many times one person did a port with basic
functionallity, and then someone else jumped in and
contributed major new features.
```

Test if they are in the booting section:
```
$ test/check_devices_in_wiki.py --booting
apple-ipt1g: wrong section of the wiki, this should be in booting already
not-documented: not in the wiki yet.
*** Wiki check failed!
...
```


test/check_devices_in_wiki.py
```diff
@@ -0,0 +1,121 @@
...
+    pmb_src = os.path.realpath(os.path.join(os.path.dirname(__file__) + "/.."))
+    for path in glob.glob(pmb_src + "/aports/device/device-*/"):
+        device = os.path.dirname(path).split("device-", 1)[1]
+        ret.append(device)
+    return sorted(ret)
+
+
+def get_wiki_devices_html(path):
+    """:param path: to a local file with the saved content of the devices wiki
+                    page or None to download a fresh copy"""
```
z3ntu: return value documentation maybe?

z3ntu:
Interesting idea :) Left one comment.

ollieparanoid:
Thanks for reviewing @z3ntu! Added the missing documentation, do you think it's ready to go?

PR #1372, opened 2018-03-28
change log output order 'flash_system' before 'flash_kernel'

lduboeuf:


ollieparanoid:
Thanks for the PR. Why did you change the order?

lduboeuf:
if i am a noob and just follow on screen logs, i will do a flash_kernel first  and ( at least on my device ) force me to boot again in flash mode

ollieparanoid:
That's a good reason! Alright, could you fix up the PEP8 complaints found by Travis? (you can run the `test/static_code_analysis.sh` script locally as well after installing `flake8` and `shellcheck`).

PureTryOut:
Sorry, I'm not exactly sure why it matters what order you flash it. In my experience, it makes no difference at all, and the system will boot properly doing it either order. How does flashing the kernel first force you to boot your device into flash mode again?

lduboeuf:
Yes, if i flash_kernel first , i can't do anything anymore in flash boot. I need to reboot and then flash_system

PureTryOut:
Interesting. In that case, yeah the change sounds good.

PR #1375, opened 2018-03-29
xiaomi-armani: load wlan module on default runlevel

alive4ever:
Previous post install action results in loading wlan module during
sysinit. This is not possible since wlan device is only exposed after
wcnss-wlan service. This also delays ssh startup for approximately 3
minutes.

Loading wlan module should only be done on default runlevel, after
wcnss-wlan service has been started.


alive4ever:
fb-mode is obtained when weston is running, so it should be correct.

directfb applications still don't work, will cause framebuffer driver crash and forced reboot.

PR #1377, opened 2018-03-29
re-enable rdp for weston

lduboeuf:
re enable weston rdp,
minimal test:
#on device
weston --backend=rdp-backend.so  --rdp-tls-cert=my.crt --rdp-tls-key=my.key --socket=wayland-1

#on computer
xfreerdp /v:172.16.42.1 /sec:tls /size:50%


aports/main/postmarketos-ui-weston/APKBUILD
```diff
@@ -1,11 +1,11 @@
 pkgname=postmarketos-ui-weston
 pkgver=4
-pkgrel=3
+pkgrel=4
 pkgdesc="(Wayland) Reference compositor (demo, not a phone interface)"
 url="https://github.com/postmarketOS"
 arch="noarch"
 license="GPL3+"
-depends="weston-clients weston weston-shell-desktop weston-backend-fbdev
+depends="weston-clients weston weston-shell-desktop weston-backend-fbdev weston-backend-rdp
```
ollieparanoid: What do you think about not enabling it by default?
People need to execute additional commands to start the RDP server anyway, so I would simply include this line in the instructions instead of always installing the subpackage.
```
$ apk add weston-backend-rdp
```

lduboeuf: correct, should be an option,

lduboeuf:
added a mini howto: https://wiki.postmarketos.org/wiki/Weston

lduboeuf:
you're welcome :)

PR #1388, opened 2018-04-02
"pkgrel_bump --auto": Handle subpackages properly

### LINKED ISSUE #1379: pkgrel_bump --auto doesn't work for qt5-qtbase?
As @PureTryOut discovered, our `qt5-qtbase` binary package was broken in the repository because of a soname bump. However, `pmbootstrap pkgrel_bump --auto` did not detect this for some reason.

```
./pmbootstrap.py chroot --add=qt5-qtbase-dev
[00:33:38] (native) install alpine-base
[00:33:48] NOTE: Run 'pmbootstrap pkgrel_bump --auto' to mark packages with outdated dependencies for rebuild. This will most likely fix this issue (soname bump?).
[00:33:48] ERROR: Could not find package 'so:libmysqlclient.so.18' in any aports folder or APKINDEX.
[00:33:48] Run 'pmbootstrap log' for details.
[00:33:48] See also: <https://postmarketos.org/troubleshooting>
```

ollieparanoid:
`pmbootstrap pkgrel_bump --auto` automatically increases the pkgrel for
packages linking against libraries, which don't exist anymore (because
the soname has been changed). The feature is explained in detail in

The previous implementation did not detect soname breakage, when a
subpackage linked against a certain library, but the main package
did not (e.g. `qt5-qtbase-mysql` and `qt5-qtbase`). This was, because
we iterated over the aports/* to find the packages to be checked.

To fix this, we are iterating over the packages found in the APKINDEX
files instead (of both the locally compiled packages and the downloaded
index from the pmOS mirror).

Details:
* `pmb/helpers/pkgrel_bump.py`:
  * Rewrite `auto_apkindex_package()` to act upon a given parsed
    `aport` and `apk` (from the index) instead of finding the `apk`
    dict by itself (we need it earlier anyway).
  * Rewrite `auto()` to iterate over APKINDEX files instead of aports
    * Skip packages already found, so the `pkgrel` does not get
      increased multiple times when the same package was found in
      multipe index files.
* Put the package name at the beginning of the log messages to make
  them more readable
* testdata: Create a new `testsubpkg` aport, where only the subpackage
  links against `testlib`
* Adjust testing code to test everything with `testsubpkg` as well.

NOTE: This makes the command a bit slower, but we could improve
performance again by smart caching of `pmb.parse.apkindex.depends()`.
This could come in a future PR, the important part here is that the
command is bug-free again with this fix.

Fixes #1379.

### DIFF
--- a/pmb/helpers/pkgrel_bump.py
+++ b/pmb/helpers/pkgrel_bump.py
@@ -16,7 +16,6 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with pmbootstrap.  If not, see <http://www.gnu.org/licenses/>.
 """
-import glob
 import logging
 import os
 
@@ -66,7 +65,7 @@ def auto_apkindex_files(args):
     Get the paths to the APKINDEX files, that need to be analyzed, sorted by
     arch. Relevant are the local pmbootstrap generated APKINDEX as well as the
     APKINDEX from the pmOS binary repo.
-.
+
     :returns: {"armhf": "...../APKINDEX.tar.gz", ...}
     """
     pmb.helpers.repo.update(args)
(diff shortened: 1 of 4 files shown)


ollieparanoid:
Thanks for the review \o/

Issue #1383, opened 2018-04-02
[Feature Request] Automatically add an alias that navigates to top of repo


Alexandroid99:
When building Android, the envsetup.sh script adds a few nice commands for navigating the build environment, like croot, for quickly jumping to the root of the repository. It'd be great if pmbootstrap had similar goodies. Just a minor request and not vital, but I think quite a few people could appreciate.

pinoaffe:
I'd rather pmbootstrap leaves my environment untouched

Alexandroid99:
Sure, optional, but what's an extra variable $TOP in your current shell session? Wouldn't hurt that much, or are we suddenly back to the C64 BASIC days where every character in memory has to be considered carefully? :yum: 
Anyway, I'm not good with shell scripting, though maybe I'll try to come up with something. If anyone wants to proceed me, go right ahead, you're probably more versed in it than I am.

ollieparanoid:
Well, `pmbootstrap` is a Python program, which does not get sourced like a shell script, but directly executed instead. So it can't really modify the current shell's environment variables even if we wanted to.

pinoaffe:
One "trick" i could think of to go to the parent dir would be `while [ ! -d .git ]; do cd ..; done`

BlitzKraft:
I think it would be more efficient if there is a variable storing the root directory from where the script will be sourced. The while loop will work, but just "feels" wrong to me.

BlitzKraft:
```bash
$PMB_PATH=$(pwd)
alias pmroot=cd "$PMB_PATH"
```
This should do the trick, when sourced from the pmbootstrap directory.

PR #1390, opened 2018-04-03
Add some aliases.

BlitzKraft:
Currently has `pmbroot` to navigate to the root of this repo. And
`pmbootstrap` to run. Addresses #1383 .


Alexandroid99:
Should be a neat little thing. Personally, I wouldn't mind it sitting in the root folder. What I'd do ontop is add a question to `pmbootstrap init` if the user wants to have the command added automatically, as it's probably the first command any new user would do. Otherwise people might miss it.

BlitzKraft:
@ollieparanoid I'll move it to the helpers folder. This is the output from shellcheck: 

```bash
Line 4:
alias pmbroot="cd \"$PMB_PATH/..\""
                    ^-- SC2139: This expands when defined, not when used. Consider escaping.
 
Line 5:
alias pmbootstrap="env python3 $PMB_PATH/pmbootstrap.py"
                               ^-- SC2139: This expands when defined, not when used. Consider escaping.
```
The warnings are exactly expected behavior. Disabling them inline. And it passes shellcheck after that. 

Question: Should I expect this to be sourced from the helpers directory or the root directory? Depending on that, the alias may need to include a `..` in the path. I am unsure how to proceed.

envsetup.sh
```diff
@@ -0,0 +1,4 @@
+# This needs to be sourced from the root of the repo
+PMB_PATH=$(pwd)
```
z3ntu: This could be improved by using the location of the script instead of the current directory the shell is in. Sourcing it from another directory would work then.

BlitzKraft: That's a better idea than what I thought. I'll do that.

helpers/envsetup.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
+# Explanation here: https://gist.github.com/TheMengzor/968e5ea87e99d9c41782
+PMB_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
```
drebrez: @BlitzKraft I didn't tested it, but what about a solution that works also for other shells and not only bash?
something like:
```
PMB_PATH=$(dirname "$(readlink -f "$0")")
```
and change the shebang to `#!/bin/sh`?

BlitzKraft: I have tried this and a few more variations. All of them end up returning the current working directory (where the source was invoked from) instead of the pmbootstrap path. 

I ran the command as is, but readlink threw an error about an invalid option "b". I escaped the inner most quotes. That was enough for readlink, but it outputs the CWD. This behavior doesn't change independent of "." or "source" is used.

z3ntu: Hmm weird. When executing that command normally it works, but when using it with "source" it breaks...

drebrez: @BlitzKraft I just had the chance to try it and you are right. After some research on internet it seems that there isn't a common way to solve this, each shell require it's own solution.
https://stackoverflow.com/questions/2720092/making-a-shell-source-file-use-relative-paths/2751448#2751448

BlitzKraft: Yup. May be we can address bash and zsh here. Include darwin variants if necessary. That should address majority of the use cases. I don't think we can include all shells. The aliases will work in most shells. So, may be prompt the user to set it if they source this without setting the variable?

ollieparanoid: Alpine's default shell is the `ash` implementation from `busybox`. If we make it dynamically figure out the path, then it would be good to support that shell as well.

> So, may be prompt the user to set it if they source this without setting the variable?

That sounds a bit over engineered TBH. In fact, the whole figuring out the current directory from a sourced script is very complex if you consider that what we are really doing at the end is use two lines to set aliases.

How about something like the following instead?
```shell
#!/bin/sh -e
if [ -e "pmbootstrap.py "]; then
	alias ...
else
	echo "ERROR: Please source this script from the pmbootstrap folder."
fi
```

ollieparanoid:
> What I'd do ontop is add a question to `pmbootstrap init` if the user wants to have the command added automatically, as init is probably the first command any new user would do.

This is not possible, because we don't source the Python script. That means it can't modify the environment variables of the current session (or call another script that modifies them).

> Otherwise people might miss it.

How about we mention it in [here](https://wiki.postmarketos.org/wiki/Porting_to_a_new_device#Shortcut) and in the channel whenever someone asks for the feature? As this does not seem to be a feature that everybody needs, that should give it enough visibility I think.

@BlitzKraft: thanks for the update, answered your question in the inline comments above.

helpers/envsetup.sh
BlitzKraft: Sounds good. This solves all the problems we discussed here.

helpers/envsetup.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/sh -e
+if [ -e "pmbootstrap.py" ]; then
+	PMB_PATH=$(pwd)
+	# shellcheck disable=SC2139
+	alias pmbroot="cd \"$PMB_PATH\""
+	# shellcheck disable=SC2139
+	alias pmbootstrap="python3 $PMB_PATH/pmbootstrap.py"
```
ollieparanoid: I think the `python3` here is redundant, as `pmbootstrap.py` has a `#!` as first line:
```python
#!/usr/bin/env python3
```

BlitzKraft: Done.

PR #1394, opened 2018-04-05
Ask to copy SSH public keys during init

### LINKED ISSUE #1353: Ability to provide password and "copy your ssh key" via command line switch
This has been bothering me for a while now.

It would be great if you could specify in the `pmbootstrap install` command something like `--password=pw123` and `--copy-ssh-key=no` to eliminate the two prompts that exist in the `pmbootstrap install` command. The ssh key switch should be doable fairly easy because it's fully implemented in Python but the `passwd` command unfortunately doesn't accept a password as argument.

drebrez:
This moves the question about copying the SSH keys from the `pmbootstrap install` command to the `pmbootstrap init`.

The only remaining question during the install is for the passwords (of the user and the encryption) and allows something like this:
`yes "mysecretpassword" | pmbootstrap -y install`

Close #1353


ollieparanoid:
> If we do it like you suggest, only ask if any is present, when there aren't it should change the setting to False automatically?

Yes, that's how I would recommend it.

> In case the key is removed later, or just keep in the config the last choice?

I'd keep the last choice in the config in that case. We do something similar with the firmware question for example: When there are no firmware subpackages for the chosen device, it keeps the last choice.

> Is there a reason why we copy only "RSA" and "Ed25519" keys? I checked the history and at the begin it was only RSA, then someone added Ed25519, [...]

Looks like the only reason is that we did not think about using a glob there.

> with the glob you suggest to install all kind of keys right?

Yeah.

Usual disclaimer: This is my opinion and not necessarily the best one. You wrote that you thought about it as well and wrote a reasonable argument, so if you think your choice is better, then we could also do it your way.

drebrez:
@ollieparanoid For me it's fine both solutions. About the glob I agree to include all kind of keys

PR #1395, opened 2018-04-06
osk-sdl: Update to 0.52

ollieparanoid:
The code got refactored. Updating the version used in pmbootstrap, so
it does not diverge too much.

Changes:
https://github.com/postmarketOS/osk-sdl/releases/tag/0.52

It worked on my `samsung-i9100`. But it would be good if someone else reviewed the small change and tested it on their device as well.

### DIFF
--- a/aports/main/osk-sdl/APKBUILD
+++ b/aports/main/osk-sdl/APKBUILD
@@ -1,5 +1,5 @@
 pkgname=osk-sdl
-pkgver=0.51
+pkgver=0.52
 pkgrel=0
 pkgdesc="Onscreen keyboard for unlocking LUKS devices"
 url="https://github.com/postmarketOS/osk-sdl"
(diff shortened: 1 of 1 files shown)


MartijnBraam:
I've tried the PR on hammerhead and osk-sdl doesn't show up, then I checked and it's also broken on master :(

ollieparanoid:
@MartijnBraam: oh, that sucks :-\ maybe we can set some CI for osk-sdl, so we notice breakage like that early. Could you open an issue in the osk-sdl repository with debug logs etc. when you have time for it?

montvid:
@ollieparanoid Just ried out the new version from this pull - it works for flo. There is no visual difference to the old osk-sdl so far I have seen.

montvid:
Okay, there is another thing that osk-sdl or full disk encryption messes with - asus flo wifi firmware load - it does not load anymore. If I install with --no-fde the wifi firmware loads everytime!

craftyguy:
> asus flo wifi firmware load - it does not load anymore.

well that is certainly interesting.. you should file a new issue here and provide the kernel log for both cases (fde and no fde).

ollieparanoid:
@montvid: did I understand you right, that the firmware loading problem is not related to this version upgrade?

montvid:
no, it is not related, i tested just now to be sure. i think it is ok to merge.

PR #1398, opened 2018-04-06
Htc flounder testing branch

Alexandroid99:
Either it doesn't let me suggest pulling to a new branch or I haven't found the option yet. Anyway, the intention is to pull this experimental stuff into a new branch for stuff like the /vendor hook if it starts working and so on. Anything that technically works but isn't the cleanest solution, basically.


Alexandroid99:
Wait, I've got write permissions to this repo? I'm used to being locked out of the original repos I do work on, so I haven't even checked if I can push directly xP

ollieparanoid:
You can push to your own fork of the repository :smile: 
In a checked out git repository, you can add additional "remotes". So you would have both the official `pmbootstrap` repository in there, where you don't have write access, and your own fork.

The commands from here might make it clearer:
https://webapps.stackexchange.com/a/90795

Alexandroid99:
Well I'm aware I can push to my own forked repo, that's no big news. You made it sound like I could push directly to this repo now.

PR #1399, opened 2018-04-06
update oneplus2 kernel

lduboeuf:
not really sure if i used the right workflow to know if kernel update went fine:
I did   zap, init, checksum & build kernel  and then install

Did not catch any issues on booting, network and UI


lduboeuf:
humm i think i made a mistake, sorry.
I took the first commit from the commit history here : https://github.com/LineageOS/android_kernel_oneplus_msm8994/commits/lineage-15.1

I'm confused by the way it is sorted, commits are not sorted by date, but gathered by i don't know what date , need to dig more. 
 
image

PureTryOut:
That just means the commit itself is done on the 21th of March 2017, but the commit is imported in this particular tree at April 4th 2018. So either that commit has always been on a developers own pc all that time, or it has been imported from a different repository. Github (and git itself when using `git log`) sorts commits on the date they are *imported* in the tree, not when they are made.

lduboeuf:
@PureTryOut thanks for the explanation. Humm so by targeting this commit id in APKBUILD, all next commits that occured since won't be taken ?

lduboeuf:
I don't want to pull 3587256 objects on my poor old computer :) so what is the best way to know which commit is effectively the last one onGithub?

lduboeuf:
is travis build issue related to my commit ?

PureTryOut:
The last one listed by Github is the one you want, no matter what date the commit itself was at. But your PR is merged already, why are you still going on about it?

drebrez:
@lduboeuf the travis build on master branch after merging this failed by coincidence with a soname bump.

>RuntimeError: One or more packages need to be rebuilt, because a library they link against had an incompatible upgrade (soname bump). Run 'pmbootstrap pkgrel_bump --auto' to automatically increase the pkgrel in order to trigger a rebuild. If this test case failed during a pull request, the issue needs to be fixed on the 'master' branch first, then rebase your PR on 'master' afterwards.

PR #1400, opened 2018-04-06
Bump pkgrel due to libressl soname upgrade

PureTryOut:
LibreSSL was upgraded upstream. I guess we should wait with merging this till all the arches we use are upgraded though, as the Alpine build server is busy upgrading it still. https://pkgs.alpinelinux.org/packages?name=libressl&branch=edge


PureTryOut:
Good question: at what point do we call pmOS "more stable" and do we decide to switch over to Alpine's stable versions?

ollieparanoid:
In order to do that, all packages that we currently use from testing must also be in the stable folder. And it will be harder to make changes to these packages (they will probably only accept fixes for the releases). So realistically it makes more sense to default to a stable version at some point, but still allow using edge (the Alpine version can already be switched with `pmbootstrap --alpine-version`).

I think it makes sense to look into this when everything else falls into place that is needed for having postmarketOS running as daily driver. But if someone wants to work towards this now, the first step would be checking which packages are still in testing/edge only, that we would need to wait for / try to get into Alpine's next stable release.

z3ntu:
Only s390x is missing now. But as we don't use that architecture I think we can merge this?

PR #1409, opened 2018-04-11
initfs hook_ls: make clear what's not installed

### LINKED ISSUE #1405: pmbootstrap initfs hook_del broken?
@vurpo wrote in #postmarketOS:
> Trying to remove a hook using pmbootstrap initfs hook_del debug-shell, it tells me no such hook is installed

ollieparanoid:
### Example output
```
$ pmbootstrap initfs hook_ls
[19:51:14] * maximum-attention (not installed)
[19:51:14] * debug-shell (installed)
[19:51:14] Done
```

Fixes #1405


z3ntu:
If we would be funny we could write `¬ installed` :stuck_out_tongue_winking_eye: (I recently found out that the british keyboard has that character on the key under the escape key but nobody uses it so it's kinda useless :D)

Issue #1408, opened 2018-04-11
nokia-n900: pmbootstrap install fails for libcrypto.so.42


duncanguthrie:
I'm attempting to run the installation program for the Nokia N900.

I first started by selecting the xfce4 interface. However, the problem is there for no interface at all:

`(002617) [15:31:03] (rootfs_nokia-n900) calculate depends of postmarketos-base, ttf-droid, device-nokia-n900 (pmbootstrap -v for details)`
`(002617) [15:31:04] NOTE: Run 'pmbootstrap pkgrel_bump --auto' to mark packages with outdated dependencies for rebuild. This will most likely fix this issue (soname bump?).`
`(002617) [15:31:04] ERROR: Could not find package 'so:libcrypto.so.42' in any aports folder or APKINDEX.`

I'm doubtful this is an issue due to the recent libressl library bump: I've tried running `pmbootstrap update && pmbootstrap index`; and I've tried running `pmbootstrap pkrel_bump --auto`, too, to no avail. I can build fine for qemu-vexpress. 

Unfortunately, pmbootstrap isn't particularly clear about what is causing the error. If this is a packaging issue, I'd expect it to show what depends on libcrypto.so.42, for instance.

sicelo:
Thanks for opening the issue! I have been struggling with it for some time,
and had come to the conclusion that it  affected only me, as everyone else
on IRC seemed to not have issues.

ollieparanoid:
I guess one of your locally built packages was built against `so:libcrypto.so.42` (which is one of the old libressl's sonames), and it has the same version (pkgver and pkgrel) as the upstream packages. This means, that `pmbootstrap` will your package instead of the one in the postmarketOS binary repository.

This should be fixed with:
* `pmbootstrap zap -p`
* rebase your branch on the official `master` branch

If it still does not work, can you try to reproduce this with the `master` branch?
And what do you get when specifying `pmbootstrap -v` for verbose mode? Then it should tell us where the `libcrypto.so.42` dependency comes from.

Sorry for the trouble, and thanks for opening this bug report! :+1:

sicelo:
On Wed, Apr 11, 2018 at 06:38:12PM +0000, Oliver Smith wrote:
[...]

PR #1410, opened 2018-04-12
Display wiki link on dependency error

ollieparanoid:
Turns out the reasons why a dependency could not be found are not that simple, and the message we had was not that helpful (e.g. in #1408, but there are more examples). Here's a new wiki page, that explains the issue in depth and how to fix it for each scenario: https://wiki.postmarketos.org/wiki/Troubleshooting:dependencies

Current:
> `NOTE: Run 'pmbootstrap pkgrel_bump --auto' to mark packages with outdated dependencies for rebuild. This will most likely fix this issue (soname bump?)`
> `ERROR: Could not find package 'so:libcrypto.so.42' in any aports folder or APKINDEX.`

New:
> `ERROR: Could not find dependency 'so:libcrypto.so.42' in any aports folder or APKINDEX. See: <https://postmarketos.org/depends>`

### DIFF
--- a/pmb/parse/depends.py
+++ b/pmb/parse/depends.py
@@ -151,12 +151,9 @@ def recurse(args, pkgnames, suffix="native"):
 
         # Nothing found
         if not package:
-            logging.info("NOTE: Run 'pmbootstrap pkgrel_bump --auto' to mark"
-                         " packages with outdated dependencies for rebuild."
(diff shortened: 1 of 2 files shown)


ollieparanoid:
Thanks for all the reviews you did today @drebrez, this is really motivating! :grin:

PR #1413, opened 2018-04-14
Update KDE Frameworks to 5.45

PureTryOut:
Everything compiles succesfully.


bhush9:
@ollieparanoid for some reason, the binary repo is missing some kf5 5.45 packages.. is it simply case of "build still going on?"

ollieparanoid:
The `x86_64` build was completed quickly yesterday, because I cached it by building once before the merge. But then it unfortunately got stuck during the `armhf` build (Python code was waiting forever for `abuild` to exit after a successful package build or something, didn't really make sense yet).
I've kicked it off again, and now it's at `modemmanager-qt` for `armhf`, I expect it to have finished building for `armhf` in a few hours and for `aarch64` within 24 hours.

What you can do now:
* `pmbootstrap update` to update the APKINDEX, so `pmbootstrap` knows which binaries were just made available
* compile the missing packages yourself if you need them
* or revert the kf5 upgrade commit locally, so it will use the old packages for now
* or wait until it is completed

Sorry for the inconvenience :\

ollieparanoid:
Packages are compiled for all arches again.

Issue #1385, opened 2018-04-02
pmbootstrap installs unneeded firmware to device


sicelo:
I notice that `pmbootstrap install` installs a lot of firmware that will not get used on the device. My target is Nokia-N900 and in the log I see installation of linux-firmware-i915, linux-firmware-ath9k, linux-firmware-amdgpu, among others, and a whole of other linux-firmware-* packages.

I think this behavior could be changed. I understand that sometimes we may want to use that firmware if, for example, we connect a device via USB host, but in general I doubt this firmware is needed on most devices, at least for sure on N900.

ollieparanoid:
Which firmware files get installed depends on the device. In case of the N900, there's `linux-firmware` mentioned in `aports/device/device-nokia-n900/APKBUILD`. The `linux-firmware` package has been split up recently in Alpine, so it is possible to specify the subpackages you've mentioned instead of the meta-package which depends on all of them.

I agree that we should change all device APKBUILDs where `linux-firmware` is used. Could you make a pull request, that fixes this for the N900?

From reading up [here](https://elinux.org/N900), the only firmware subpackage needed should be:
`linux-firmware-ti-connectivity` (WiFi: wl1251-fw.bin and wl1251-nvs.bin)

(CC @craftyguy and @pavelmachek, do you know about more firmware files used for the modem?)

Thanks for opening this issue @sicelo!

craftyguy:
Wifi definitely needs firmware, I don't recall firmware being loaded for the modem, but it's really difficult to determine if it's required without removing all firmware and trying to load things (unless someone knows of a better way?), since not all drivers will say "I am loading firmware XYZ now" in the log..

ollieparanoid:
@craftyguy: at least from that page, it looks like the bluetooth firmware never made it to the official `linux-firmware`. Regarding cellular modem: are there simple test instructions somewhere, that whoever gives this a shot could try out? Then we would know if behavior changes or not after reducing the installed firmwares.

craftyguy:
> Regarding cellular modem: are there simple test instructions somewhere, that whoever gives this a shot could try out?

Well basically if they can do what was described in [the N900 wiki page for initializing the modem and establishing a data connection](https://wiki.postmarketos.org/wiki/Nokia_N900#GPRS) without any linux-firmware packages installed, then that would confirm the modem does not require loading any extra firmware to operate.

sicelo:
[...]

PR #1414, opened 2018-04-14
Fix Firmware Dependencies for N900

sicelo:
This commit removes dependency on `linux-firmware` in favor of `linux-firmware-ti-connectivity`, which is currently the only needed firmware for the N900, as discussed in issue #1385


sicelo:
Ah yes, fails because upstream uboot-tools has not yet been rebuilt for armhf against the new libressl for yet. See issue #1408 .

PR #1416, opened 2018-04-14
firmwareload: log to system log

scintill:
Is there a reason this wasn't done originally? Maybe we can't count on syslogd being up. Now I'm wondering if this patch is a bad idea for that reason.

### DIFF
--- a/aports/main/postmarketos-base/firmwareload.sh
+++ b/aports/main/postmarketos-base/firmwareload.sh
@@ -2,23 +2,23 @@
 
 FIRMWARE_DIRS="/lib/firmware /lib/firmware/postmarketos"
 
-exec >>/var/log/firmwareload.log 2>&1
+(
+	if [ ! -e "/sys/$DEVPATH/loading" ]; then
+		echo "firmware loader misses sysfs directory"
+		exit 1
+	fi
 
-if [ ! -e "/sys/$DEVPATH/loading" ]; then
-    echo "firmware loader misses sysfs directory"
-    exit 1
-fi
+	for DIR in $FIRMWARE_DIRS; do
+		[ -e "$DIR/$FIRMWARE" ] || continue
+		echo "loading $DIR/$FIRMWARE"
+		echo 1 > "/sys/$DEVPATH/loading"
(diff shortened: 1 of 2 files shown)


aports/main/postmarketos-base/firmwareload.sh
```diff
@@ -2,23 +2,23 @@
 
 FIRMWARE_DIRS="/lib/firmware /lib/firmware/postmarketos"
 
-exec >>/var/log/firmwareload.log 2>&1
+(
+	if [ ! -e "/sys/$DEVPATH/loading" ]; then
+		echo "firmware loader misses sysfs directory"
+		exit 1
```
ollieparanoid: Looks like the `exit 1` won't have the desired effect here:
```shell-session
$ samsung-i9100:~$ (echo testmessage; exit 1) | logger -t testtag && echo "returns 0"
returns 0
```
How about we add a `log()` function on top, that redirects `$1` to `logger -t "firmwareload"`, and call that instead of wrapping the entire code in a subshell? If we do it that way, then the `exit 1` should still work as expected.

scintill: I think it does work; this was my test (dash on Ubuntu):

```
$ (echo test message; exit 42); echo $?
test message
42
$ (echo test message; exit 42) && echo returns 0  
test message
$ 
```

I think your example is checking the return of `logger`, not the subshell. I'm not a shell expert, though... log() might be better for clarity, but I would still want to capture stderr just in case anyway, and subshell is easiest way I know.

scintill:
I'm still not sure about service startup order. inittab contains

```
::sysinit:/sbin/openrc sysinit
::sysinit:/sbin/openrc boot
```

Does this concurrently start syslog and udev, since they are in those two runlevels? Is there a defined order? I'd like the firmware logs to be in syslog, but if only some entries are going to make it in there due to undefined service order, it's too misleading IMO.

ollieparanoid:
I am not sure either to be honest, but if we can make it work it is still a good patch. Some time after the next blog post is out, I could take an in-depth look at it again and do lots of testing to find out if it is feasible or not.

PR #1421, opened 2018-04-16
Introduce the linux-postmarketos-qcom package

bhush9:
- Introduce the newer libdrm version in our aports for now
```
16:24 <robclark> there was briefly a problem with libdrm.. but that has been fixed..
16:25 <robclark> so 2.4.89 had an issue.. but 2.4.90 or newer should be fine
```
- Use https://github.com/postmarketOS/linux branch `bshah/linux-postmarketos-qcom` as base for linux-postmarketos-qcom.
- Switch hammerhead to use mainline kernel.
- Don't use msmfb-refresher

TODO:

- [x] Add adreno 330 firmware
- [x] Add dep to qcom firmware package
- [x] Test Plasma Mobile
- [ ] Test and fix x11 config
- [ ] Add dep to modem package
- [ ] Add some more firmware blobs
- [ ] stuff.


julian-goldsmith:
The Adreno 330 firmware is currently available in the adreno-firmware package.

Will linux-postmarketos-qcom replace that package for mainlined devices?

aports/device/device-lg-hammerhead/APKBUILD
```diff
@@ -1,17 +1,16 @@
 pkgname=device-lg-hammerhead
 pkgver=1
-pkgrel=20
+pkgrel=21
 pkgdesc="Google Nexus 5"
 url="https://github.com/postmarketOS"
 arch="noarch"
 license="MIT"
-depends="postmarketos-base linux-lg-hammerhead mkbootimg msm-fb-refresher mesa-dri-swrast"
+depends="postmarketos-base linux-postmarketos-qcom mkbootimg mesa-dri-freedreno"
```
ollieparanoid: What would you think about packaging the upstream kernel next to the downstream kernel and let the user choose in `pmbootstrap init`? It might be useful to quickly compare how a certain feature works with the downstream kernel vs. upstream.

`pmbootstrap` is [able to do it](https://wiki.postmarketos.org/wiki/Device_specific_package#Multiple_kernels) and I have a PR ready to go that drops the `deviceinfo_msm_refresher` variable from deviceinfo (we would add the `msm-fb-refresher` to the downstream kernel subpackage, and it would automatically get installed as initramfs hook then).

I would be fine with either solution, and I'm curious what you folks think is best.
If we don't use `linux-lg-hammerhead` anymore (i.e. no kernel selection, just use `linux-postmarketos-qcom`), I'd recommend to delete it from `aports`.

aports/firmware/firmware-adreno/APKBUILD
```diff
@@ -1,14 +1,14 @@
...
+pkgrel=6
 pkgdesc="Firmware for adreno gpus"
 url="https://github.com/TheMuppets/proprietary_vendor_sony"
 arch="noarch"
 license="proprietary"
 options="!check"
 
 # Sources
-_files1="a225_pm4.fw a225_pfp.fw a225p5_pm4.fw a300_pfp.fw a300_pm4.fw a330_pfp.fw a330_pm4.fw"
+_files1="a225_pm4.fw a225_pfp.fw a225p5_pm4.fw a330_pfp.fw a330_pm4.fw"
```
ollieparanoid: Why are these files removed? `a300_pfp.fw a300_pm4.fw`

aports/firmware/firmware-adreno/APKBUILD
```diff
@@ -24,15 +24,13 @@ done
 package() {
 	for _i in $_files1 $_files2; do
 		install -Dm644 "$srcdir/$pkgname-$_commit-$_i" \
-			"$pkgdir/lib/firmware/postmarketos/$_i"
+			"$pkgdir/lib/firmware/qcom/$_i"
```
ollieparanoid: So far we put all firmware files packaged in postmarketOS into the `/lib/firmware/postmarketos` folder, so it does not conflict with the files from `linux-firmware`. Why was it changed here, does it not work with the pmOS path?

aports/main/libdrm/APKBUILD
```diff
@@ -0,0 +1,48 @@
+# Maintainer: Natanael Copa <ncopa@alpinelinux.org>
```
ollieparanoid: Please remove the maintainer when copying packages from elsewhere, so people don't send them e-mails for postmarketOS packages. How about replacing it with something like: `# Copied from Alpine to increase $pkgver`

aports/main/linux-postmarketos-qcom/APKBUILD
```diff
@@ -0,0 +1,181 @@
+_flavor=postmarketos-qcom
+_config="config-${_flavor}.${CARCH}"
+
+pkgname=linux-${_flavor}
+
+pkgver=4.16_rc6
+pkgrel=0
+
+arch="armhf"
+pkgdesc="Linux for pmOS supported chipsets (qcom, with extra patches from different devs + linux-next)"
```
ollieparanoid: How about shortening that, e.g.: `Kernel close to linux-next with extra patches for Qualcomm devices`

aports/main/linux-postmarketos-qcom/APKBUILD
```diff
@@ -0,0 +1,181 @@
...
+url="https://kernel.org/"
+depends=""
+makedepends="perl sed installkernel bash gmp-dev bc linux-headers elfutils-dev libressl-dev file bison flex"
+options="!strip !check !tracedeps"
+install=
+source="
+	linux-qcom.tar.gz::https://github.com/bhush9/linux/archive/459e3e6e05ab68f112531209958f4a309de1b1a3.tar.gz
```
ollieparanoid: The hash should go into the archive name, otherwise it will try to use the same tarball even if the download hash changed. Consider using a `$_commit` variable if doing it that way (or provide patch files on top of the official linux-next release, as discussed in IRC).

ollieparanoid:
@julian-goldsmith wrote:
> Will linux-postmarketos-qcom replace that package for mainlined devices?

No, as the `linux-postmarketos-mainline` package is used by the N900 right now. The idea was to make an extra package for qualcomm, so the qcom specific changes don't break non-qcom devices. When some time passed, and the qcom specific changes made their way into upstream's mainline and stable branches, we could [offer these kernels](https://wiki.postmarketos.org/wiki/Device_specific_package#Multiple_kernels) for the hammerhead and others as well.

@bhush9: regarding the TODO list: we don't need to do it all in one PR. When this one is in a good state, even when the features above are not all implemented, I would recommend to merge it as starting point and go from there with new PRs to implement the rest.

aports/device/device-lg-hammerhead/APKBUILD
bhush9: I can try it.. but probably I don't see a reason. I think mainline is more future complete then the downstream :-P Downstream doesn't reallly support anything other then graphics over framebuffer/software compositing, nor will be able to ever unless we use libhybris. I think it's more right thing to do to get rid of old outdated downstream kernel.

I am open to other's suggestions though.

aports/firmware/firmware-adreno/APKBUILD
bhush9: This files are present in linux-firmware-qcom package by upstream, so we don't want to copy them from TheMuppets as long as we are not modifying/binary patching them.

aports/firmware/firmware-adreno/APKBUILD
bhush9: Yeah, upstream kernel expects adreno firmware to be present at qcom/*, That's why it was changed to put it in qcom specific path.

aports/main/libdrm/APKBUILD
bhush9: Will Fix

aports/main/linux-postmarketos-qcom/APKBUILD
bhush9: Will fix description (except s/linux-next/mainline)

aports/main/linux-postmarketos-qcom/APKBUILD
bhush9: Will fix.

aports/device/device-lg-hammerhead/APKBUILD
z3ntu: I would keep both for now.

z3ntu: Not really but choice is good ;)

ollieparanoid: Yes, choice is good, but if the additional choices don't bring us any benefit whatsoever, then it's just additional dead code nobody uses that we still need to carry and maintain.

From that perspective I recommend to remove `linux-lg-hammerhead`. We can link to the old kernel packaging in the hammerhead's wiki page and add it back on demand.

aports/firmware/firmware-adreno/APKBUILD
ollieparanoid: I guess we could patch the kernel / change the kernel config to look in the other path, but then again some adreno blobs are already in the official `linux-firmware` and they do get installed to `/lib/firmware`. So it's better the way you wrote it imho :+1:

ollieparanoid:
@bhush9: Thanks for addressing the comments. I've created it as you suggested as fork of `torvalds/linux` (github import failed yesterday). You and @opendata26 have write access (just let me know if someone else needs it). How about uploading the commits there to a new `linux-postmarketos-qcom` branch (so it's obvious we'll use it as base for this package)?

I think this PR would be good to go as initial version, as soon as
* `linux-lg-hammerhead` is removed (unless we find good reasons to keep it)
* the upstream repo is switched to [postmarketOS/linux](https://github.com/postmarketOS/linux)

(As stated above, I'd recommend to merge it once we have an initial working version, and doing the other TODOs as follow up PRs so it's easier to review and people already have something to play with in the repos.)

bhush9:
@ollieparanoid I've pushed my changes to postmarketOS/linux repo and also I've removed the linux-lg-hammerhead package.

I've also rebased the commits and and squashed some commits, I would prefer if this is rebase merged instead of squash merge, as there are 4 different things..

- New libdrm package
- New linux-postmarketos-qcom package
- Modify the adreno-firmware
- Switch lg-hammerhead device package to linux-postmarketos-qcom.

ollieparanoid:
I've reviewed each commit independently, and they all look good. Merged as single commits as requested. This is fine here in my opinion, because each commit only touches one package, and we know that all packages build and @MartijnBraam tested that they work. (I will build binary packages later this day, they should be available tomorrow. Of course the building process should be optimized long term, so there are shorter delays. I can possibly look into it after the next blog post.)

Thanks again @bhush9 \o/

ollieparanoid:
Maybe someone can update [the wiki page](https://wiki.postmarketos.org/wiki/Google_Nexus_5_(lg-hammerhead)), it has some outdated references to the downstream kernel now.

PR #1423, opened 2018-04-17
Properly style the plasma-mobile

bhush9:
This provides some configuration which provides better plasma mobile
experience in terms of scaling, fonts and theme

### DIFF
--- a/aports/main/postmarketos-ui-plasma-mobile/kdeglobals
+++ b/aports/main/postmarketos-ui-plasma-mobile/kdeglobals
@@ -1,0 +1,21 @@
+[KDE]
+StartDragDist=40
+LookAndFeelPackage=org.kde.plasma.phone
+
+[Icons]
+Theme=breeze
+
+[General]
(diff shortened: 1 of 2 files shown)


aports/main/postmarketos-ui-plasma-mobile/kdeglobals
```diff
@@ -0,0 +1,23 @@
+[KDE]
```
ollieparanoid: Are the two `[KDE]` sections on purpose? (also there's a space at the end of the line)

PureTryOut:
@bhush9 on Matrix:

> Also mind that this would switch to numerical lockscreen only, so if possible just set numerical password

This might be a problem, as right now we allow everything (I guess full unicode?), making it impossible to unlock your screen if you use anything other than numbers in your password. What should we do about this?

ollieparanoid:
Merged on command line in https://github.com/postmarketOS/pmbootstrap/commit/74a6a22fec421b87538f7a4fdbfc7083e038f548.

PR #1424, opened 2018-04-17
Add envkernel.sh: instant mainlining environment

### LINKED ISSUE #1356: kernelenv.sh for mainlining
We have a [mainlining guide](https://wiki.postmarketos.org/wiki/Mainlining_Guide) in the wiki now, that explains how to set up the development environment. Compiling the kernel is done outside of the `pmbootstrap` chroots, because that makes the workflow much better when working on mainline (e.g. you run `make` once, and it only compiles what changed instead of the whole package).

I figured it would be nice if we had a script `pmbootstrap/helpers/kernelenv.sh` that does the following when we source it, to make the setup easier:
* set up the native chroot with a GCC cross compiler, with the appropriate arch as read from the selected device's deviceinfo
* write wrappers that **make it possible with some black magic to use the cross compilers from *outside* of the chroots**, even if the host system uses glibc
* export right `ARCH` and `CROSSCOMPILE` (these are picked up by the kernel's `Makefile`) to point to the wrappers

So that way people wouldn't need to figure out the right cross compiler package name (and other packages, such as bc or openssl-dev) for their distro and install it - the helper script does it all. And it is almost working, the only problem I'm facing is that when using the script, the `-isystem include` option passed to GCC doesn't work anymore. This is used by the kernel's Makefile to use the `include` folder bundled with linux for system libc headers, instead of the ones from musl or glibc.

I've been spending too much time on that already (it's a nice to have after all, the current guide for setting everything up does work), but maybe someone can figure that last bit out so I'm leaving it here for reference. The code isn't the prettiest and could be improved once this actually works.

Usage:
* copy to `pmbootstrap/helpers/kernelenv.sh`
* clone Linux source
* in the Linux folder, run: ` source ~/code/pmbootstrap/helpers/kernelenv.sh`
* run `make` to compile the kernel with the cross compiler from the native chroot

```shell
#!/bin/sh -e

# Alias for pmbootstrap
pmbootstrap_dir="$(dirname "$0")/.."
alias pmbootstrap="$pmbootstrap_dir/pmbootstrap.py"

# Deviceinfo path
device="$(pmbootstrap config device)"
deviceinfo="$pmbootstrap_dir/aports/device/device-$device/deviceinfo"

if  [ -e "$deviceinfo" ]; then
	# Source deviceinfo
	. "$deviceinfo"
	echo "Device: $device"
	echo "Arch: $deviceinfo_arch"

	# Install cross-compiler in chroot
	chroot="$(pmbootstrap config work)/chroot_native"
	flag="$chroot/tmp/kernelenv/setup_done"
	if ! [ -e "$flag" ]; then
		echo "Setting up Alpine chroot (details: 'pmbootstrap log')"
		pmbootstrap -q chroot -- apk -q add \
			abuild \
			bc \
			binutils-"$deviceinfo_arch" \
			gcc-"$deviceinfo_arch" \
			gcc \
			binutils \
			linux-headers \
			libressl-dev \
			musl-dev

		mkdir -p "$chroot/tmp/kernelenv"
		touch "$flag"
	fi

	# Cross-compiler prefix
	prefix="$(source "$chroot/usr/share/abuild/functions.sh";
	arch_to_hostspec "$deviceinfo_arch")"

	# Create wrappers for programs that GCC executes
	gcc_wrapper_dir="$chroot/tmp/kernelenv/libexec/gcc/"
	mkdir -p "$gcc_wrapper_dir"
	for i in cc1 lto-wrapper lto1 collect2; do
		wrapper="$gcc_wrapper_dir/$i"
		{
			echo "#!/bin/sh"
			echo "$chroot/lib/ld-musl-x86_64.so.1" \
				"--library-path=$chroot/lib/:$chroot/usr/lib" \
				"$chroot/usr/libexec/gcc/armv6-alpine-linux-muslgnueabihf/6.4.0/$i" \
				'"$@"'
		} > "$wrapper"
		chmod +x "$wrapper"
	done

	# Create wrappers for GCC and friends
	cpath="$chroot/usr/include/"
	cpath="$cpath:$chroot/usr/lib/gcc/$prefix/6.4.0/include"
	echo "CPATH: $cpath"
	mkdir -p "$chroot/tmp/kernelenv"
	for i in $prefix-gcc $prefix-as $prefix-ld $prefix-ar $prefix-nm \
		$prefix-strip $prefix-objcopy $prefix-objdump; do

		wrapper="$chroot/tmp/kernelenv/$i"
		{
			echo "#!/bin/sh"
			echo "export GCC_EXEC_PREFIX=$gcc_wrapper_dir"
			echo "export CPATH=$cpath"
			echo "$chroot/lib/ld-musl-x86_64.so.1" \
				"--library-path=$chroot/lib/:$chroot/usr/lib" \
				"$chroot/usr/bin/$i" '"$@"'
		} > "$wrapper"
		chmod +x "$wrapper"
	done

	# Export arch and cross-compiler path
	case "$deviceinfo_arch" in
		aarch64*) export ARCH="arm64" ;;
		arm*) export ARCH="arm" ;;
	esac
	export CROSS_COMPILE="$chroot/tmp/kernelenv/$prefix-"
else
	echo "ERROR: Please select a valid device in 'pmbootstrap init' first."
fi
```

ollieparanoid:
Thanks to @Decatf for the idea of aliasing `make`!
CC: @bhush9, @julian-goldsmith, @pavelmachek, @filippz, @flacks, @fourkbomb, @opendata26

### Usage example
```
$ cd ~/code/linux
$ source ~/code/pmbootstrap/helpers/envkernel.sh
Initializing Alpine chroot (details: 'pmbootstrap log')
pmbootstrap envkernel.sh activated successfully.
 * kernel source:  /home/user/code/linux
 * output folder:  /home/user/code/linux/.output
 * architecture:   arm (samsung-i9100 is armhf)
 * aliases: make, kernelroot, pmbootstrap, pmbroot (see 'type make' etc.)
```
`$ type make`
> `make is an alias for echo '*** pmbootstrap envkernel.sh active for /home/user/code/linux! ***'; pmbootstrap -q chroot -- ARCH=arm CROSS_COMPILE=/usr/bin/armv6-alpine-linux-muslgnueabihf- make -C /mnt/linux O=/mnt/linux/.output`

### Notes
* `make menuconfig` works as well
* Sourcing was tested with `zsh` and `bash`, it should be easy to
  extend for other shells

### Changes
* new action `pmbootstrap work_migrate`: does the interactive work
  folder migration if necessary, otherwise it doesn't output anything
  * when calling this first, we can safely use all other commands
    non-interactively without showing the output
* `helpers/envkernel.sh`:
  * installs everything needed for kernel compilation in the native
    chroot
  * mounts the kernel source to `/mnt/linux` inside the chroot
  * creates `/mnt/linux/.output` and chowns it to the `pmos` user, that
    folder will be used for the kernel build output
  * sets up aliases for `make`, `pmbootstrap`, `pmbroot`, `kernelroot`

### Benefits
* Fast setup (especially for people who are new to kernel
  compilation)
  * No need to figure out distribution specific package names
    (cross compilers!)
  * No need to do a test build just to verify that the right
    packages are installed
* Less error prone
  * The right dependencies are always installed
  * `ARCH` and `CROSS_COMPILE` variables always get set automatically
    and based on `deviceinfo_arch`
  * If the build environment is broken for some reason, just zap and
    start over
* Easy to reproduce problems

### How to test
* Source the script like shown above, then compile a full kernel

@MartijnBraam: I've used the SPDX license header here as you recommended in #postmarketOS-offtopic the other day. +1 for deploying that for the entire code base.

Closes #1356


z3ntu:
When I first source envkernel.sh:
```
[luca@g550jk linux]$ source ../../postmarketOS/pmbootstrap/helpers/envkernel.sh
bash: pmbootstrap: command not found
See also: <https://postmarketos.org/troubleshooting>
[luca@g550jk linux]$ type pmbootstrap
pmbootstrap is aliased to `../../postmarketOS/pmbootstrap/helpers/../pmbootstrap.py'
[luca@g550jk linux]$ source ../../postmarketOS/pmbootstrap/helpers/envkernel.sh
pmbootstrap envkernel.sh activated successfully.
 * kernel source:  /mnt/hdd/fp2-mainline-kernel/linux
 * output folder:  /mnt/hdd/fp2-mainline-kernel/linux/.output
 * architecture:   arm (fairphone-fp2 is armhf)
 * aliases: make, kernelroot, pmbootstrap, pmbroot (see 'type make' etc.)
```
Weird behaviour... with `set -x` (and a 'type' call to see if pmbootstrap is set):
```
[luca@g550jk linux]$ set -x
[luca@g550jk linux]$ source ../../postmarketOS/pmbootstrap/helpers/envkernel.sh
+ source ../../postmarketOS/pmbootstrap/helpers/envkernel.sh
++ check_kernel_folder
++ '[' -e Kbuild ']'
++ return
++ export_pmbootstrap_dir bash ''
++ '[' -n ../../postmarketOS/pmbootstrap/helpers/envkernel.sh ']'
+++ dirname ../../postmarketOS/pmbootstrap/helpers/envkernel.sh
++ script_dir=../../postmarketOS/pmbootstrap/helpers
++ export pmbootstrap_dir=../../postmarketOS/pmbootstrap/helpers/..
++ pmbootstrap_dir=../../postmarketOS/pmbootstrap/helpers/..
++ '[' -e ../../postmarketOS/pmbootstrap/helpers/../pmbootstrap.py ']'
++ set_alias_pmbootstrap
++ alias pmbootstrap=../../postmarketOS/pmbootstrap/helpers/../pmbootstrap.py
++ type pmbootstrap
pmbootstrap is aliased to `../../postmarketOS/pmbootstrap/helpers/../pmbootstrap.py'
++ '[' -e /home/luca/.config/pmbootstrap.cfg ']'
++ pmbootstrap work_migrate
bash: pmbootstrap: command not found
++ echo 'See also: <https://postmarketos.org/troubleshooting>'
See also: <https://postmarketos.org/troubleshooting>
```
But actually I just wanted to test if the `make` behaviour was as weird as in my fish-shell port (will push that to this branch then):
```
[luca@g550jk linux]$ make
*** pmbootstrap envkernel.sh active for /mnt/hdd/fp2-mainline-kernel/linux! ***
make: Entering directory '/mnt/linux'
make[1]: Entering directory '/mnt/linux/.output'
getconf: LFS_CFLAGS: unknown variable
getconf: LFS_LDFLAGS: unknown variable
getconf: LFS_LIBS: unknown variable
getconf: LFS_CFLAGS: unknown variable
getconf: LFS_LDFLAGS: unknown variable
getconf: LFS_LIBS: unknown variable
  GEN     ./Makefile
  HOSTCC  scripts/kconfig/zconf.tab.o
scripts/kconfig/zconf.tab.c:2466:23: fatal error: zconf.lex.c: No such file or directory
 #include "zconf.lex.c"
                       ^
compilation terminated.
make[3]: *** [scripts/Makefile.host:107: scripts/kconfig/zconf.tab.o] Error 1
make[2]: *** [../Makefile:524: silentoldconfig] Error 2
make[1]: *** No rule to make target 'include/config/auto.conf', needed by 'include/config/kernel.release'.  Stop.
make[1]: Leaving directory '/mnt/linux/.output'
make: *** [Makefile:146: sub-make] Error 2
make: Leaving directory '/mnt/linux'
```
Using `set -gx ARCH arm; set -gx CROSS_COMPILE arm-none-eabi-` (equivalent of `export`ing the variables) and `make` without the aliases works as expected..

z3ntu:
Oh yeah also the pmbootstrap alias should use an absolute path to its folder and not a relative (as it does now) - fixed that.

helpers/envkernel.fish
```diff
@@ -0,0 +1,196 @@
...
+	case ppc64
+		echo "powerpc64-alpine-linux-musl"
+	case ppc64le
+		echo "powerpc64le-alpine-linux-musl"
+	case s390x
+		echo "s390x-alpine-linux-musl"
+	case x86
+		echo "i586-alpine-linux-musl"
+	case x86_64
+		echo "x86_64-alpine-linux-musl"
```
z3ntu: we could probably remove all arches here except arm stuff as further down we only check for aarch64* and arm*

ollieparanoid:
Thanks for the fish port and for fixing it to use the absolute path! \o/
I thought about how we could get rid of the duplicated code though, and came up with the following:
* `envkernel.fish` calls `envkernel.sh --fish` internally
* `envkernel.sh` sets up the chroot as usually, but prints everything to stderr now
* at the end it prints `alias` lines in fish syntax to stdout
* `envkernel.fish` catches these and executes them
* (NOTE: I've also tried to output the `alias` lines to a custom file descriptor 3, but `envkernel.sh` would complain that `fish` did not open FD3.)

As I just tested it, it works now with `zsh`, `bash` and `fish`. Could you review/test it again?

> Maybe we can use the group of the host user and give the group write folders so we can have both?

This would work for the `envkernel.sh` case, but it would also mean that the `pmos` user gets write access to data owned by the host user in general, which it should rather not have in my opinion.

helpers/envkernel.sh
```diff
@@ -137,29 +139,44 @@ set_alias_pmbroot_kernelroot() {
...
-	echo " * architecture:   $arch ($device is $deviceinfo_arch)"
-	echo " * aliases: make, kernelroot, pmbootstrap, pmbroot" \
-		"(see 'type make' etc.)"
-else
-	# Failure
-	echo "See also: <https://postmarketos.org/troubleshooting>"
-fi
+# Run main() with all output redirected to stderr
+# Afterwards print fish compatible syntax to stdout
+main "$0" >&2 && fish_compat "$1"
```
z3ntu: I would rename the fish_compat function as we might expand to other shells, like `shell_compat`?

ollieparanoid: I'm not sure how many more POSIX incompatible shells there are, that people would use this script with. How about we rename it on demand?

ollieparanoid:
Thanks for fixing that and for all the great reviewing you did once again!

I agree with sudo access being required is annoying here. But I don't think there's much we can do about it - and this script will probably still be a big time saver for people who never set up a kernel cross-compiling environment before.

Do you think it is ready to go?

z3ntu:
Looks good.

PR #1428, opened 2018-04-19
Device samsung i8190 (Galaxy SIII mini)

CUB3D:
**Working**
- Display
- Touch
- USB networking
- FDE

**Not working**
- Everything else


mirh:
I don't want to create drama (and arguably, these are kinda the most updated "actual" sources)
But you should try to see if you can get golden to work with https://github.com/ChronoMonochrome/Chrono_Kernel-1, which is a whole neater world of newness and polishedness than maclaw random trees. 

The thing being a Samsung, should make adapting pretty straightforward... I wouldn't know how much similar U8420 and U8500 are though, but you can get some help [here](https://forum.xda-developers.com/galaxy-ace/ace-2-develop/kernel-u8500-linux-kernel-upgrading-t3603810)

PR #1438, opened 2018-04-22
msm-fb-refresher: use initfs hook

### LINKED ISSUE #1396: msm-fb-refresher: use *.files method to simplify code
Since it's possible to specify hook-specific binaries (see the `debug-shell` hook) that will be added to the initramfs, we can do the following:

* Install a hook with `msm-fb-refresher`, that will add the binary to the initramfs
* Remove the extra code from `mkinitfs.sh` that used `deviceinfo_msm_refersher` to do the same thing
* Remove the `deviceinfo_msm_refresher` variable

ollieparanoid:
The `msm-fb-refresher` updates the screen for msm based devices. It is not needed for all devices, so we had some extra code in the initramfs, that would only add it when the `deviceinfo_msm_refresher` variable was set. However, we are able now to add files to initramfs hooks, so this hack can be removed and simplify everything.

Changes:
* Remove `deviceinfo_msm_refresher` from all deviceinfos
* Add sanity check for it
* Move all `deviceinfo` sanity checks to an extra function
* `postmarketos-mkinitfs`: remove code for msm refresher
* `msm-fb-refresher`: add initramfs hook

### Testing
I have tested, that `pmbootstrap initfs ls` shows the same files before and after the change. But this should be tested on a real device that uses the msm-fb-refresher (I don't have one) before we merge it.

Closes #1396.

---


MartijnBraam:
Can we automatically add initfs hooks for devices or do you need to add the hook manually after every install?

ollieparanoid:
The initramfs hook gets automatically added if the device package depends on `msm-fb-refresher` :+1:

drebrez:
@ollieparanoid nice job with this PR, in the next days I will test on my devices if it works correctly.
I notice some multiple empty lines in the deviceinfo files but IIRC there is already an issue to clean and unify them.
Code looks good, maybe we can add a simple test case that runs the `sanity_check` agains all the existing `deviceinfo`?

ollieparanoid:
> @ollieparanoid nice job with this PR, in the next days I will test on my devices if it works correctly.

Awesome! :tada: 

> I notice some multiple empty lines in the deviceinfo files but IIRC there is already an issue to clean and unify them.

Yes, at least in #688 the next open TODO item there is "Make all deviceinfo files like the template".

> Code looks good, maybe we can add a simple test case that runs the `sanity_check` agains all the existing `deviceinfo`?

Yeah, actually we have that already:

https://github.com/postmarketOS/pmbootstrap/blob/d6391791f7043ee4832012894ac8325510c9725e/test/test_aports.py#L42-L49

ollieparanoid:
Thanks for testing \o/

PR #1440, opened 2018-04-23
Samsung-i8190 wifi now working

CUB3D:
Guide for connecting to wlan is on the wiki page for the device, wifi chip is BCM4334.
Wlan interfaces are wlan0 and p2p0.
Connection seems stable, was able to install some packages.
---


ollieparanoid:
Thanks for making the update! This will be merged soon, right now we're adjusting to the QT upgrade Alpine had yesterday (#1458),

PR #1445, opened 2018-04-27
Only use gtk+2.0-maemo for Hildon

### LINKED ISSUE #1380: File managers having trouble launching
Neither Thunar nor PCManFM have managed to start properly for me. Neither dies completely, pushing out a variety of Gtk, GLib-GObject, and exo warnings/critical errors before stalling totally. Nautilus, I've gotten to launch, and some sort of work, after the installation of gsettings-desktop-schemas, but the title bar is still pretty messed up, with broken icons, etc.

ollieparanoid:
Fixes #1380.

---


michitux:
I did not try this, but from what I can see this changes nothing for i3wm. However, when using i3wm, still software that requires gtk2 could be installed (I have that situation and running `apk add gtk+2.0` did not change anything as gtk+2.0-maemo provides that). Is there any way to change gtk+2.0-maemo to not to be installed unless Hildon is installed?

PureTryOut:
>  Is there any way to change gtk+2.0-maemo to not to be installed unless Hildon is installed?

Yes, by stop making gtk+2.0-maemo "provide" and "replace" gtk+2.0.

https://github.com/postmarketOS/pmbootstrap/blob/master/aports/maemo/gtk%2B2.0-maemo/APKBUILD#L15
https://github.com/postmarketOS/pmbootstrap/blob/master/aports/maemo/gtk%2B2.0-maemo/APKBUILD#L16

Why do we do that anyway? We should probably just make the package that depend on gtk+2.0-maemo specify it in their depends.

Decatf:
> Why do we do that anyway? We should probably just make the package that depend on gtk+2.0-maemo specify it in their depends.

By using the provides variable the package manager will know that there is already alternate gtk+2.0 package installed.

For example: Suppose the provides variable isn't used. If Hildon desktop already has gtk+2.0-maemo installed and something tries to resolve a dependency on gtk+2.0 then the latter will be marked for install. There would a conflict between the two gtk+2.0 packages.

PureTryOut:
Ah I guess they do provide the same files largely, completely forgot.

ollieparanoid:
> So, the proper way to do this then, would be porting either Hildon to GTK3.0 (upstream version, not a fork) or get the Maemo specific GTK2.0 patches upstream.

There [is a thread](https://talk.maemo.org/showthread.php?t=96800) about porting Maemo to GTK3, but it has not had any activity since 2017-09-25. Meanwhile [Maemo Leste](https://github.com/maemo-leste/) came out and they are actively maintaining Maemo/Hildon, and we're using their sources as upstream code now. I think the most feasible way for us as a downstream distribution from their point of view, is to stay close to what they do. And when I asked ~2 months back in the Maemo Leste IRC, it seemed like they don't have plans to use a GTK3 port in the near future. I doubt that they are working on upstreaming the Maemo changes to GTK2 either. If someone wants to work towards upstreaming the Maemo specific bits to GTK2/3, I recommend to get in touch with them.

> [...] from what I can see this changes nothing for i3wm. However, when using i3wm, still software that requires gtk2 could be installed [...]

This is an excellent point, we should fix that!

How about we change the `gtk+2.0-maemo` package to be not conflicting with `gtk+2.0` anymore? We could either change the [sonames](https://en.wikipedia.org/wiki/Soname), or install everything into another prefix (e.g. `/lib/gtk+2.0-maemo/`). And then adjust the Maemo packages to work with that changed version.

In the meantime I would suggest to merge this PR anyway, since it already improves the situation for XFCE4, MATE and matchbox users. What do you think?

Decatf:
I think maybe this should be merged before the one year blog post is up. New comers will be trying pmOS and having issues like the file manager not working.

It seems for now there isn't an easy way to lower the priority of replacement packages to prevent superseding the original package. So I think merge this until a better solution exists. And from now on be very careful about merging such packages which replace others.

PR #1450, opened 2018-04-29
Fix/1448 n900 i3status battery

### LINKED ISSUE #1448: n900 i3status "no battery"
i3 status wont display the battery percentage. By default it will say "no battery"

in i3status.conf, I set battery path
path = "/sys/class/power_supply/bq27200-0/uevent/"

and I get "no '%d' in battery path"

fxkrait:
Sorry its messy. I'm still figuring out how to do this.
Fixes #1448

### DIFF
--- a/aports/device/device-nokia-n900/i3wm/i3status.conf
+++ b/aports/device/device-nokia-n900/i3wm/i3status.conf
@@ -16,7 +16,7 @@ general {
 
 order += "disk /"
 order += "wireless _first_"
-order += "battery all"
+order += "battery 0"
 order += "load"
 order += "tztime local"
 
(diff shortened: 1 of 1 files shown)


sicelo:
Ok, I have a workaround/solution that works for me, based on instructions at https://i3wm.org/i3status/manpage.html#_external_scripts_programs_with_i3status and sample code at https://github.com/i3/i3status/blob/master/contrib/net-speed.sh

The resulting script displays battery status as follows: `83% of 1378mAh (3961mV at 264mA)` and of course one can edit the script to remove/add values he is interested in or not. Will submit a PR in my pmbootstrap fork for your review shortly.

When I find time I will run the i3status test suites and see what to report to upstream i3status, but this solution should suffice for the time being.

ollieparanoid:
Closing in favor of #1470.

PR #1452, opened 2018-04-29
[ui-mate] Set postmarketOS artwork as default wallpaper

drebrez:
Set postmarketOS artwork as default wallpaper.
Tested on qemu and on real device

---


drebrez:
@ollieparanoid I've read somewhere that those schema files have to be compiled with `glib-compile-schemas` first

ollieparanoid:
You're right, [in this howto](http://www.techytalk.info/customize-default-desktop-environment-settings-gnome-centric-linux-distributions/) it is used as well. How about we use `glib-compile-schemas` in the `build()` step then, to convert a readable config file to the binary file that is needed for packaging?

drebrez:
@ollieparanoid it was even simpler, just install the override file you suggested in the schemas folder and rebuild the binary (`glib-compile-schemas`) in the package post-install 👍

PR #1455, opened 2018-04-30
Add support for a separate aports directory & repository.

### LINKED ISSUE #383: Split pmbootstrap and aports in two repositories

marmistrz:
Closes #383. Contains changes from #1454 (to avoid later rebasing)

NB, If Python 3.6 were supported, a much cleaner PEP-498 could be used:

```
        logging.fatal(f"    git clone {APORTS_GIT} {args.aports}")
```
---

### DIFF
--- a/pmb/config/__init__.py
+++ b/pmb/config/__init__.py
@@ -55,7 +55,7 @@ config_keys = ["ccache_size", "device", "extra_packages", "hostname", "jobs",
 # overriden on the commandline)
 defaults = {
     "alpine_version": "edge",  # alternatively: latest-stable
-    "aports": os.path.normpath(pmb_src + "/aports"),
+    "aports": os.path.expanduser("~") + "/.cache/pmbootstrap/aports",
     "ccache_size": "5G",
     # aes-xts-plain64 would be better, but this is not supported on LineageOS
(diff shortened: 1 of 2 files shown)


marmistrz:
I'm wondering why the tests on 0a1f92c fail in the place they do. `pmbootstrap` should actually fail at the very beginning, not able to find the `aports`.

PureTryOut:
If we're going to do this, we should probably move the aports to it's own repository as well.

PR #1460, opened 2018-05-03
Martijnbraam/hammerhead cleanup

MartijnBraam:
It isn't very useful to launch weston demo's on every possible installed DE, even when the weston demo's aren't installed.

The weston demo's still work after this change

This also changes the kernel cmdline to use ttyMSM0 instead of ttyHSL0 since ttyHSL0 is from the downstream kernel.
---


aports/device/device-lg-hammerhead/APKBUILD
```diff
@@ -11,7 +11,6 @@ install=""
 subpackages="$pkgname-weston $pkgname-x11 $pkgname-nonfree-firmware:nonfree_firmware"
```
drebrez: @MartijnBraam please increase the pkg version

PureTryOut:
Note that weston.ini can now be removed entirely, as the settings in there at the moment are the default settings afaik. They only need specifying when they have to be different (say the framebuffer driver rather than the DRM one).

drebrez:
@PureTryOut it's the opposite, the default one use the fbdev backend, see https://github.com/postmarketOS/pmbootstrap/blob/master/aports/main/postmarketos-ui-weston/weston.ini.default

PureTryOut:
Ah, that's our own default. Completely forgot, I was just thinking about Weston upstream defaults. Thanks for the reminder!

PR #1462, opened 2018-05-03
linux-postmarketos-qcom: Remove composite usb gadget for networking

MartijnBraam:
This removes the precomposed usb gadget from the kernel config and enables the configfs usb gadget modules. This makes the usb networking on the hammerhead come up reliably

---


drebrez:
@MartijnBraam also the sony sirius uses the `linux-postmarketos-qcom` kernel, maybe someone have to check if it still works also there?

MartijnBraam:
Yeah this definitely needs some multiple device testing

MartijnBraam:
Hmm it looks like that's a change from running menuconfig on 4.17-rc4 while the old config was for 4.16

MartijnBraam:
This is the least amount of changes I can get it to do with menuconfig, I'm not sure how to make the QMI_HELPER a module again, it changes it on its own when you run menuconfig without touching anything.

screenshot from 2018-05-05 13-44-47

bhush9:
Thinking about please disable, SAMPLE_QMI_CLIENT and effectively QMI_HELPER... as those I had enabled for testing something locally and then those slipped into config for aports but those are not needed

bhush9:
There is some super weirdness is going on, but I don't think is related to this PR, selecting weston makes it not start device at all for some reason which I believe is kernel panic/hang or something.. But I don't think it's related to this PR at all since this PR doesn't touch the graphics part at all and is good to go IMO.

MartijnBraam:
I've bumped the pkgrel, now waiting on testting on the sirius @opendata26

PR #1464, opened 2018-05-04
Use postmarketOS artwork on plasma-mobile

bhush9:


bhush9:
(Please test as I can't test, for some reason qemu is broken in my system)

Issue #1458, opened 2018-05-01
Binary repository problems with today's GCC and QT upgrade


ollieparanoid:
### GCC
**TL;DR:** run `pmbootstrap update` to fix `pmbootstrap` trying to compile `gcc-armhf` or `gcc-aarch64`.

GCC has been updated in upstream in Alpine with a bunch of new patches on top, so our cross compilers need to be updated as well (thanks @MartijnBraam for doing that in #1456).

Two problems:
* The official mirror of the sources of GCC is down, hence users can't compile the packages themselves right now (someone could probably find a mirror to the 6.4.0 patches tarball and use that instead by replacing the `source` in the `APKBUILD`)
* The package does *not* build in strict mode, it tries to install `build-base-armhf` when trying to build `gcc-armhf`. However with the normal `pmbootstrap -y zap; pmbootstrap build gcc-armhf` (where pmbootstrap does the dependency resolution, not apk), it appears to be compiling.

The binary package building PC has the sources in the cache, and I'm building the packages there right now to make them available for all postmarketOS users ASAP. As strict mode can't be used right now, I'm not using the regular repository build script, but the workaround described above.

### QT

Alpine has also updated their QT version from 5.9 to 5.10. This means, we need to update our aports to use the new QT version as well, and then recompile them. Plasma Mobile can probably not be installed right now because of that. I'm on it.

Note: also some packages have been renamed in Alpine for consistency, e.g.
> community/{qt5-websockets => qt5-qtwebsockets}

### TODOs

- [x] compile and upload binary packages for gcc cross compilers
> run `pmbootstrap update` to get the new packages if pmbootstrap tries to build `gcc-armhf` or `gcc-aarch64`
- [x] update our QT packages in the `pmbootstrap` repository to 5.10: https://github.com/postmarketOS/pmbootstrap/commit/6c6d0d3e28270afe699692fbf579fdcf798a314d
- [x] bump pkgrels of KDE packages of private API users: https://github.com/postmarketOS/pmbootstrap/commit/676d22fd8ad4fe3f9d738addae94e6f4ae9121e8
- [x] build binary packages (build is running right now!)
  * Almost all QT packages have been rebuilt already. `qt5-qtwebengine` is failing, which is needed for LuneOS UI (not for Plasma Mobile)
  * KDE packages still need to be rebuilt as of writing
  * See which packages have been built here: http://postmarketos.brixit.nl/
[...]

PR #1465, opened 2018-05-05
Disable locale/translation generation in plasma-workspace [skip ci]

bhush9:
This is workaround to get binary repository working, while I can't
reproduce the ki18n bug at all, and is not fixed upstream, let's
workaround it by disabling po file generation in plasma-workspace for
now.

See issue #1458

TODO: revert when ki18n is fixed upstream.
---


ollieparanoid:
Thank you very much @bhush9! Turns out this did not trigger the bug after all, see https://github.com/postmarketOS/pmbootstrap/issues/1458#issuecomment-386795086

PR #1468, opened 2018-05-05
gcc-*: fix compiling in strict mode [skip ci]

### LINKED ISSUE #1458: Binary repository problems with today's GCC and QT upgrade

ollieparanoid:
Due to changes in abuild, our `gcc-armhf` etc. packages did not build
when using strict mode (i.e. `pmbootstrap build --strict gcc-armhf`)
anymore.

### Changes
* Set `CBUILDDIR=/`, so apk can read a valid package index from there
* Directly set `_cross_configure`, so it does not use CBUILDDIR anymore
* Set `BOOTSTRAP="nobuildbase"` to prevent apk from installing
  `build-base-armhf` etc. (these don't exist in pmOS)
* Remove legacy code for lazy reproducible builds that wrapped
  `package()`
* Fix download links by replacing ftp:// with http:// (also made a PR
  to fix this upstream: https://github.com/alpinelinux/aports/pull/4199)

### How to test
```
$ pmbootstrap build gcc-armhf --force
$ pmbootstrap build hello-world --force --arch=armhf
```

Fixes #1458.

NOTE: Travis fails, because building the packages takes too long for Travis.

---

### DIFF
--- a/aports/cross/gcc-aarch64/APKBUILD
+++ b/aports/cross/gcc-aarch64/APKBUILD
@@ -4,7 +4,6 @@
 
 CTARGET_ARCH=aarch64
 CTARGET="$(arch_to_hostspec ${CTARGET_ARCH})"
-CBUILDROOT="/usr/$CTARGET"
 LANG_OBJC=false
 LANG_JAVA=false
 LANG_GO=false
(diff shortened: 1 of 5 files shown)


ollieparanoid:
Thanks @steamp0rt!

PR #1474, opened 2018-05-07
Install the "device-aaa-bbb-kernel-XXX" package instead of the "linux-postmarketos-XXX" to also install the related depends

drebrez:
This PR will install the proper device subpackage, depending on the kernel selection during init, instead of installing directly the "linux-postmarketos-XXX" using the subpackage name.
In both the current devices with the kernel selection, they have the `depends` variable with the correct kernel package to install, and possibly also other packages.
In theory we can even remove also the lines:
```
     if args.kernel == "downstream":
         return ["linux-" + device]
```
and just install the `device-aaa-bbb-kernel-downstream` package which will have in the `depends` the `linux-aaa-bbb` package.
Or it is by design to not install the device kernel subpackage?

Ps: I noticed that they weren't installed because I had extra stuff in the device kernel mainline subpackage and it was missing.

---


pmb/install/_install.py
```diff
@@ -113,7 +113,7 @@ def get_kernel_package(args, device):
     # Return the pkgname
     if args.kernel == "downstream":
```
drebrez: also remove this?

ollieparanoid:
Thanks for making the change, doing it that way makes much more sense indeed! I'll try to get a review done tomorrow :smiley:

pmb/install/_install.py
ollieparanoid: Yeah, that makes sense!

PR #1475, opened 2018-05-08
matchbox: miscellaneous fixes

duncanguthrie:
The changes may be summarised as:
* Build matchbox-panel against libmatchbox with libxsettings (automatically detect if libmatchbox was built with xsettings, then build with xsettings);
* Build libmatchbox and matchbox-panel with debug symbols;
* Upgrade matchbox-window-manager and matchbox-desktop to version 2 (these now don't use xsettings, but matchbox-window-manager is built with GTK+2 support, and can have certain font configuration applied through that and e.g. its support for pango);
* Change source of matchbox-panel to Debian's source, which had miscellaneous fixes not upstream;
* Amend the terminal-launching panel-applet to look for x-terminal-emulator (e.g. if we use update- alternatives as provided in Alpine's dpkg package, we can change our terminal preference globally);
* Remove unused patches arising from changing package versions and sources.

This fixes the strange issues with Qemu and some issues I had patching the original matchbox-window-manager and matchbox-desktop.

That being said, I haven't enabled PNG themes in matchbox-window-manager, because the program would not build properly otherwise, and they are not essential at the moment.

I tried to get version 2 of matchbox-panel working, but it did not work at all. It is a shame, because this is a rewrite of the original panel with modern features like GTK+3 support. That being said, the components of matchbox now seem to be at least functional and mostly stable, and are usable on real devices.


aports/matchbox/libmatchbox/APKBUILD
```diff
@@ -27,7 +27,10 @@ build() {
 		--enable-pango \
 		--enable-doxygen-docs \
 		--enable-jpeg \
-		--enable-unit-tests 
+		--enable-unit-tests \
+		--enable-debug \
+		LIBS="-lXsettings-client -lX11"
+		LDFLAGS=$LIBS
 	make
```
ollieparanoid: It looks like you wanted to pass `LIBS` and `LDFLAGS` to `make`. I don't think it works that way, you would need to write the `LDFLAGS` variable directly in front of `make`. Example:

```shell
#!/bin/sh

echo "first:"
LDFLAGS=test
env | grep LDFLAGS

echo "second:"
LDFLAGS=test env | grep LDFLAGS
```

output:
```
first:
second:
LDFLAGS=test
```

More thoughts:
* if the `LIBS` variable is not used anywhere else, how about directly setting `LDFLAGS="-lXsettings-client -lX11"`?
* it looks like `LIBS/LDFLAGS` was not set at all, but it still worked (automatic detection of what is installed in configure I guess?) - so how about we remove these variables if they aren't needed anyway?

aports/matchbox/libxsettings-client/APKBUILD
```diff
@@ -14,13 +14,11 @@ source="http://http.debian.net/debian/pool/main/libx/${pkgname}/${pkgname}_${pkg
 builddir="${srcdir}/${pkgname}-${pkgver}"
 options="!check"
 subpackages="$pkgname-doc $pkgname-dev"
-
 prepare() {
 	default_prepare
 	update_config_sub
 	update_config_guess
 }
-
```
ollieparanoid: Is there a reason why you removed the empty lines between `prepare()` and `build()` (not just in this `APKBUILD`)? The example `APKBUILD`s generated by `newapkbuild` (e.g. `pmbootstrap newapkbuild testpkg`) has empty lines there:
```shell
pkgname=testpkg
pkgver=
pkgrel=0
pkgdesc=""
url=""
arch="all"
license=""
depends=""
makedepends=""
install=""
subpackages="$pkgname-dev $pkgname-doc"
source=""
builddir="$srcdir/"

build() {
    cd "$builddir"
}

package() {
    cd "$builddir"
}
```

aports/matchbox/matchbox-desktop/APKBUILD
```diff
@@ -25,7 +22,9 @@ build() {
 		--prefix=/usr \
 		--sysconfdir=/etc \
 		--mandir=/usr/share/man \
-		--localstatedir=/var
+		--localstatedir=/var \
+		LIBS="-lX11" \
+		LDFLAGS=$LIBS
```
ollieparanoid: see "is the backslash intended" and LIBS/LDFLAGS comments above

aports/matchbox/matchbox-window-manager/APKBUILD
```diff
@@ -1,48 +1,39 @@
...
-_pkgrel=osso21
+_pkgname=$pkgname-2
+pkgver=2.0_git20171119
 pkgrel=0
-_pkgver=$pkgver-$_pkgrel
-pkgdesc="Window manager intended for non-desktop embedded platforms such as handhelds and set-top boxes"
-url="http://matchbox-project.org"
+_commit=5f44595c66be118b1a74f489de7f8492035cbdf0 
+pkgdesc="Window manager intended for non-desktop embedded platforms such as handhelds and set-top boxes" 
+url="http://matchbox-project.org"
```
ollieparanoid: (spaces at the end of the `url`, `pkgdesc` and `_commit` lines)

aports/matchbox/matchbox-window-manager/APKBUILD
```diff
@@ -1,48 +1,39 @@
...
 		--host=$CHOST \
 		--prefix=/usr \
 		--sysconfdir=/etc \
 		--mandir=/usr/share/ \
-		--enable-startup-notification \
-		--enable-expat \
+		--with-gtk \
+		--with-pango \
+		LIBS="-lexpat" \
+		LDFLAGS=$LIBS
```
ollieparanoid: see backslash/libs/ldflags comments above

ollieparanoid:
> This fixes the strange issues with Qemu and some issues I had patching the original matchbox-window-manager and matchbox-desktop.

Awesome, this should make testing much easier!

> I tried to get version 2 of matchbox-panel working, but it did not work at all. It is a shame, because this is a rewrite of the original panel with modern features like GTK+3 support. That being said, the components of matchbox now seem to be at least functional and mostly stable, and are usable on real devices.

Maybe that could work in a future PR. This is a great improvement already!

aports/matchbox/libxsettings-client/APKBUILD
duncanguthrie: No, it was style. I don't mind either way, but I suppose it's more readable so I'll add whitespace where appropriate.

aports/matchbox/libmatchbox/APKBUILD
duncanguthrie: The configure script passes the LIBS/LDFLAGS (and CFLAGS) specified when one runs it to make, which is why I put it there. Another alternative is to patch the pkg-config scripts in the source, but this wasn't always very straightforward.

As for saying LIBS=$LDFLAGS, that's not something I intended to leave there. Usually it's going to be sufficient to put only one there. I'll amend the files accordingly.

aports/matchbox/libmatchbox/APKBUILD
```diff
@@ -27,7 +27,10 @@ build() {
 		--enable-pango \
 		--enable-doxygen-docs \
 		--enable-jpeg \
-		--enable-unit-tests 
+		--enable-unit-tests \
+		--enable-debug \
```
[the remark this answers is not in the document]
duncanguthrie: No, it was unintentional - I'll amend the file accordingly.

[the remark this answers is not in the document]
duncanguthrie: In fact, that's not right - see my comment below about passing LIBS into configure being as intended

duncanguthrie:
Could you possibly clarify how changing from ./autogen.sh to autoreconf --install will break aarch64? 

In fact, the contents of autogen.sh in these cases was actually  just autoreconf -i -v (or similar) and then having it run ./configure. I needed to change configure, so I thought it was more readable and more straightforward just to run autoreconf and then configure in the APKBUILD. 

Is that a good approach? I don't think it's necessarily a bad idea to change it.

duncanguthrie:
I managed to get the panel version 2 to run, so I thought it was appropriate to push it to this PR.

duncanguthrie:
Additionally, it makes sense to build matchbox-keyboard as a GTK2 and GTK3 input method, as that can be toggled in window managers which support this. Furthermore, building it with Pango support allows one to customise its font scaling and other properties.

ollieparanoid:
> Could you possibly clarify how changing from ./autogen.sh to autoreconf --install will break aarch64?
>
> In fact, the contents of autogen.sh in these cases was actually just autoreconf -i -v (or similar) and then having it run ./configure. I needed to change configure, so I thought it was more readable and more straightforward just to run autoreconf and then configure in the APKBUILD.
>
> Is that a good approach? I don't think it's necessarily a bad idea to change it.

Sounds like a good idea to me! Changing the autoreconf stuff rang an alarm bell for me regarding aarch64, but as you explained it, your change shouldn't affect it.

> I managed to get the panel version 2 to run, so I thought it was appropriate to push it to this PR.

Awesome! \o/ :tada: 

> Additionally, it makes sense to build matchbox-keyboard as a GTK2 and GTK3 input method, as that can be toggled in window managers which support this. Furthermore, building it with Pango support allows one to customise its font scaling and other properties.

This sounds be a great improvement! I would recommend to make that change in a new pull request though.

duncanguthrie:
I hope my rebase didn't break things...

ollieparanoid:
> I hope my rebase didn't break things...

Oh, I did not realize that you've rebased it. I thought I caused the high number of commits myself, so I just reset it to the last commit before merging `master` in. (I wanted to merge master into this branch to make Travis CI go through, but this doesn't even work on master right now because weston doesn't compile). I'm sorry if that made it all complicated.

The code changes look good and it worked in qemu-amd64.

I think this can be merged as soon as #1477 is solved (I'll look into that tomorrow).> I hope my rebase didn't break things...

Oh, I did not realize that you've rebased it. I thought I caused the high number of commits myself, so I just reset it to the last commit before merging `master` in. (I wanted to merge master into this branch to make Travis CI go through, but this doesn't even work on master right now because weston doesn't compile). I'm sorry if that made it all complicated.

The code changes look good and it worked in qemu-amd64.

I think this can be merged as soon as #1477 is solved (I'll look into that tomorrow).

duncanguthrie:
Looks like Travis is happy. Thanks for merging the fix!

Issue #1368, opened 2018-03-27
Support different values in deviceinfo for upstream and downstream kernels


ollieparanoid:
PR #1363 will make it possible to choose between upstream (official [kernel.org](https://kernel.org) ones, possibly slightly patched) and downstream kernels (the Android kernel forks) for devices.

As @drebrez and @z3ntu pointed out, we will need different values for the [`deviceinfo`](https://wiki.postmarketos.org/wiki/Deviceinfo_reference) variables for upstream kernels.

@drebrez [proposed](https://github.com/postmarketOS/pmbootstrap/pull/1363#issuecomment-376327553) the following solution which I quite like:
> [...] having both lines in `deviceinfo`, e.g.:
```
deviceinfo_kernel_cmdline_mainline="cma=64m msm.vram=16m"
deviceinfo_kernel_cmdline_downstream="console=ttyHSL0,115200,n8"
```
> but in the subpackage installation just remove all the other lines and rename the variable to `deviceinfo_kernel_cmdline`? 
> That way we can have a common logic between all the APKBUILDs

I would recommend to name the suffix `upstream` instead of `mainline` though, because we have `linux-postmarketos-mainline`, `linux-postmarketos-stable` etc. We could implement the logic to create the kernel-specific `deviceinfo` file in the [`devicepkg-dev`](https://wiki.postmarketos.org/wiki/Device_specific_package#devicepkg-dev) package we already have and use.

z3ntu:
So as I understand it, the devicepkg-dev script should go through all deviceinfo lines, remove the suffix from the correct kernel config variables and remove (or leave - doesn't really matter) the others?

drebrez:
@z3ntu right, that was the idea, and do it for every `kernel_*` subpackage to remove the suffix of that subpackage

drebrez:
@ollieparanoid while adding the downstream/mainline choice for the samsung klte kernel I noticed that the downstream uses `deviceinfo_bootimg_qcdt="true"` but the mainline kernel works with the dtb appended to the kernel itself (`zImage-dtb`).
Should we just consider that every mainline kernel doesn't need the qcdt and just remove that line? (or set to false)

drebrez:
draft:
```
kernel="downstream"      # or upstream
if grep -q "deviceinfo_kernel_cmdline_$kernel" "$srcdir/deviceinfo"; then
[...]

PR #1479, opened 2018-05-09
Add samsung-klte mainline kernel

drebrez:
Changes:
- `device-samsung-klte` with both downstream and mainline kernel subpackages
- `linux-postmarketos-qcom` updated to last commit

What works in mainline kernel:
- internal SD card
- volume and home key buttons
- usb network

---

### DIFF
--- a/aports/device/device-samsung-klte/APKBUILD
+++ b/aports/device/device-samsung-klte/APKBUILD
@@ -1,13 +1,17 @@
 pkgname="device-samsung-klte"
 pkgdesc="Samsung Galaxy S5"
-pkgver=0.1
-pkgrel=3
+pkgver=0.2
+pkgrel=0
 url="https://postmarketos.org"
 license="MIT"
 arch="noarch"
 options="!check"
-depends="postmarketos-base linux-samsung-klte mkbootimg msm-fb-refresher"
+depends="postmarketos-base mkbootimg msm-fb-refresher"
 makedepends="devicepkg-dev"
+subpackages="
+	$pkgname-kernel-downstream:kernel_downstream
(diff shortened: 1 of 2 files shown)


drebrez:
Just tested and I can confirm that it doesn't work.
Currently the boot.img is created with QCDT (dtb inside the boot.img) and also the kernel with the appended dtb.
I tried with QCDT and the kernel without the appended dtb but it doesn't work either.
In just shows the samsung download mode with an error message.

With `deviceinfo_bootimg_qcdt="false"` and the dtb appended to the kernel it boots fine.

FYI: the current version of linux-postmarketos-qcom doesn't have a usb node for the samsung-klte, I will add it very soon.

ollieparanoid:
Thanks for the PR!
* Regarding split of the deviceinfo package, see my answer in https://github.com/postmarketOS/pmbootstrap/issues/1368#issuecomment-387881831
* Maybe it makes sense to ship the deviceinfo package split as its own PR, so we could merge it even if running mainline that way does not work with your device yet?

drebrez:
I thought about splitting the PR but it was useful to have the other files
to test it. I will split the PR later.
Tomorrow I will look at the your version

opendata26:
As for qcdt, decompile a downstream DTS and add the properties at the top,(msm-id and a few others) to your upstream dts,

opendata26:
@drebrez yeah

drebrez:
@ollieparanoid I've rebased this PR to only include the mainline kernel for the samsung-klte, It works with the same `deviceinfo` as the downstream kernel since I've enabled the QCDT support and also the usb network.
I will create a new PR for the `devicepkg-dev` with the same version I've added here before and we can continue the discussion there.

PR #1480, opened 2018-05-10
weston: Add wayland-dev make dependency, add patch for compilation with freerdp 2.0.0-rc2

### LINKED ISSUE #1477: Weston doesn't build
Dependency `libfreerdp` has been updated, so we need to rebuild `weston`. But we can't rebuild weston right now, as `libwayland` was updated today, but `mesa-dev` isn't built against it yet in Alpine.

Upstream fix:
https://github.com/alpinelinux/aports/pull/4229

Workaround:
```
$ cd ~/code/pmbootstrap
$ rm -r aports/main/weston
```

This will make `pmbootstrap` use the latest binary package from the postmarketOS repositories (because it doesn't see the current version as outdated anymore). That probably works except for RDP being broken.

z3ntu:
Fixes #1477 

Reason for the patch is https://github.com/FreeRDP/FreeRDP/commit/1f7d33a2f22a372c124ce985a394619e186e06b9
Should probably send that patch upstream.

---


ollieparanoid:
Thanks for the patch, I've missed it while working on this in parallel. Your patch is better, because it does not disable the RDP backend. I'll rebase and then use it, thank you very much @z3ntu!

PR #1509, opened 2018-05-22
Add gconfig/xconfig support

### LINKED ISSUE #891: Add the ability to do "pmbootstrap gconfig" or "xconfig"?
I kinda wanna use the GUI configs, rather than the ncurses-based ones.

steamp0rt:
Hooray, I got gconfig to work! Also, kconfig_check and menuconfig have been merged into one command.
Examples:
```
pmbootstrap kconfig check
pmbootstrap kconfig check --arch=armhf linux-postmarketos-mainline
pmbootstrap kconfig edit amazon-thor
pmbootstrap kconfig edit --frontend=g amazon-thor
pmbootstrap kconfig edit --frontend=ncurses amazon-thor
pmbootstrap kconfig edit --arch=armhf --frontend=gconfig postmarketos-qcom
```
capture

closes #891


steamp0rt:
**NOTE:** This requires kernel APKBUILD changes!
`make ARCH="$_carch" menuconfig` needs to become `make ARCH="$_carch" $kopt`!

aports/main/postmarketos-splash/make-splash.py
```diff
@@ -5,14 +5,18 @@
 import sys
 
 
-def make_splash(width, height, filename, landscape=False, text="", config_file=None, center=False):
-    print("Creating ({}x{}) splashscreen {}".format(width, height, os.path.basename(filename)))
+def make_splash(width, height, filename, landscape=False,
```
PureTryOut: I'm guessing you had some IDE autoformat the whole project, but that isn't related to this PR at all. Could you undo it all?

steamp0rt: I ran autopep8

steamp0rt: To be exact, `autopep8 -ria ~/pmbootstrap`

aports/device/linux-amazon-thor/APKBUILD
```diff
@@ -80,7 +80,8 @@ prepare() {
 # this is so we can do: 'abuild menuconfig' to reconfigure kernel
 menuconfig() {
 	cd "$srcdir"/build
-	make ARCH="$_carch" menuconfig
+	export DISPLAY=:0
```
MartijnBraam: I guess the export line doesn't really have to be in this file and can be set in python instead

helpers/pmbootstrap-autocompletion.zsh
```diff
@@ -31,7 +31,7 @@ _pmbootstrap_targets()
 			find $PMBOOTSTRAP_DIR/aports/ -mindepth 2 -maxdepth 2 -type d \
 				-printf '%f\n' | sed "s|$PMBOOTSTRAP_DIR/aports/||g"
 			;;
-		kconfig_check|menuconfig)
+		kconfig)
 			ls -1 $PMBOOTSTRAP_DIR/aports/*/ | grep linux- \
```
MartijnBraam: This isn't right, this makes it autocomplete `pmbootstrap kconfig [package]`, not `pmbootstrap kconfig check [package]`

aports/main/postmarketos-splash/make-splash.py
MartijnBraam: This makes it quite hard to read the diff.  you can probably reverse it with `git revert 78e2b74`

drebrez: @steamp0rt sometimes I use autopep8 with the `--range line line` to only format the lines I've changed, instead of the whole folder/file

aports/device/linux-amazon-thor/APKBUILD
steamp0rt: yeah this was just during testing

helpers/pmbootstrap-autocompletion.zsh
steamp0rt: I don't understand ZSH autocompletion files, i dunno what to do

MartijnBraam: Me neither :D

.travis/travis_script.sh
```diff
@@ -28,9 +28,9 @@ yes "" | ./pmbootstrap.py init
 fold_end "init"
 
 # pmbootstrap kconfig_check
-fold_start "kconfig_check" "pmbootstrap kconfig_check"
-./pmbootstrap.py kconfig_check
-fold_end "kconfig_check"
+fold_start "kconfig check" "pmbootstrap kconfig scheck"
```
MartijnBraam: typo scheck -> check

pmb/parse/arguments.py
```diff
@@ -185,6 +185,18 @@ def arguments_newapkbuild(subparser):
...
+def arguments_kconfig(subparser):
+    arch_native = pmb.parse.arch.alpine_native()
+    arch_choices = set(pmb.config.build_device_architectures + [arch_native])
+    frontend_choices = ["xconfig", "x", "gconfig", "g", "ncurses", "n"]
+
+    ret = subparser.add_parser("kconfig", help="manage kernel aport options")
+
+    ret.add_argument("--arch", choices=arch_choices, dest="arch")
+    ret.add_argument("--frontend", choices=frontend_choices, default="ncurses", dest="frontend", help="The frontend that should be used for menuconfig.")
+    ret.add_argument("thingy", help="Either 'check' or the package name.")
```
MartijnBraam: thingy is not a good name, maybe it's neater to make this "package" with nargs="?" so it's optional and add a `--check` argument for the non-menuconfig mode and check on execution if  only one of those is supplied.

steamp0rt: Done

ollieparanoid:
> **NOTE:** This requires kernel APKBUILD changes!
`make ARCH="$_carch" menuconfig` needs to become `make ARCH="$_carch" $kopt`!

`pmbootstrap menuconfig` calls `abuild menuconfig` internally, which calls the `menuconfig()` function in the `APKBUILD`. That behavior was copied from `linux-vanilla` from Alpine. However, Alpine removed the `menuconfig()` function from their kernels, so it doesn't seem to be the recommended way to do it anymore (awilfox also commented at one point that it's a hack, and "menuconfig" is not documented in `abuild -h`).

So before we modify the `menuconfig()` function from all kernels, I would suggest the following instead:
* replace the `abuild menuconfig` call with `abuild prepare` and executing `make menuconfig` (or similar command) directly with `pmb.helpers.run.user()`.
* remove the `menuconfig()` function from the template for new kernels (`pmb/aportgen/linux.py`)

Opinions on this?

Thanks for making this @steamp0rt! :tada:

steamp0rt:
@ollieparanoid yes, but how would we handle things like `$srcdir/android_kernel_amazon_hdx-common-${_hash}` in pmbootstrap?

MartijnBraam:
we can parse the value of $builddir from the APKBUILD file

ollieparanoid:
That would work (if we run a shell that sources the APKBUILD and prints `$builddir`). However, 23 of our kernel packages do not have the `builddir` variable set, so it won't work for them.

Here's another method. How about we add a function to `menuconfig.py` like the following (untested)?

```python
import glob


# Find the builddir
def get_builddir(args):
	"""
	Find the builddir by looking for a file that is always present in the
	kernel source.

	:returns: builddir path inside the chroot, e.g. 
			  "/home/pmos/build/src/linux-4.16"
	"""
	# Search pattern
	chroot = args.work + "/chroot_native/"
	srcdir = "/home/pmos/build/src"
	srcdir_outside = chroot + srcdir
	pattern = srcdir_outside + "/*/scripts/kconfig/menu.c"

	# Require one result
	matches = glob.glob(pattern)
	if not matches:
		raise RuntimeError("Could not find builddir, no match for: " + pattern)

	# Get the first folder after srcdir_outside and prepend srcdir
	folder = matches[0][len(srcdir_outside):].split("/")[0]
	return srcdir + folder
```

The file this checks for is in the Linux source tree at least [since 2005](https://github.com/torvalds/linux/commits/master?after=bee797529d7c1ea4e2803fda067d20edbc00bc3d+69&path%5B%5D=scripts&path%5B%5D=kconfig&path%5B%5D=menu.c).

steamp0rt:
@ollieparanoid Thanks. The function needed slight modification, but it worked!

README.md
```diff
@@ -81,12 +81,12 @@ $ pmbootstrap bootimg_analyze ~/Downloads/twrp-3.2.1-0-fp2.img
 
 Check kernel configs:
 ```
-$ pmbootstrap kconfig_check
+$ pmbootstrap kconfig check
```
drebrez: @steamp0rt I think this should be `pmbootstrap kconfig --check`

pmb/build/menuconfig.py
```diff
@@ -82,9 +125,9 @@ def menuconfig(args, pkgname):
                     "/home/pmos/build", log=False, env={"CARCH": arch})
 
     # Run abuild menuconfig
```
drebrez: @steamp0rt update comment, it doesn't use abuild anymore

pmb/build/menuconfig.py
```diff
@@ -67,12 +92,30 @@ def menuconfig(args, pkgname):
     aport = pmb.build.find_aport(args, pkgname)
     apkbuild = pmb.parse.apkbuild(args, aport + "/APKBUILD")
     arch = get_arch(args, apkbuild)
+    kopt = "menuconfig"
 
     # Set up build tools and makedepends
     pmb.build.init(args)
-    depends = apkbuild["makedepends"] + ["ncurses-dev"]
+    if args.frontend == "n" or args.frontend == "ncurses":
```
ollieparanoid: How about refactoring that part to:
```py
depends = apkbuild["makedepents"]
kopt = "menuconfig"
if args.frontend in ["n", "ncurses"]:
	depends += ["ncurses-dev"]
elif args.frontend in ["g", "gconfig"]:
	depends += ["gtk+2.0-dev", "glib-dev", "libglade-dev", "font-noto"]
	kopt = "gconfig"
...
```

pmb/build/menuconfig.py
```diff
@@ -67,12 +92,30 @@ def menuconfig(args, pkgname):
...
 
+    # Symlink host's .xauthority into native
+    if args.frontend != "n" and args.frontend != "ncurses":
+        logging.info("(native) symlink host Xauthority")
+        file = os.environ.get('XAUTHORITY')
+        if not file:
+            file = os.path.join(os.environ['HOME'], '.Xauthority')
+        link = args.work + "/chroot_native/home/pmos/.Xauthority"
+        pmb.helpers.file.symlink(args, file, link, True)
+        pmb.helpers.run.user(args, ["xhost", "+"])
```
ollieparanoid: As written in the chat, this does not work as intended. `xhost +` will allow *any* connection, which completely disables the security with `XAUTHORITY` that is intended to only allow certain users to access the X server.

Also I doubt that the symlink is working since the pmos user inside the chroot (which is the one who will run the "make menuconfig" or similar command) has (intentionally) another user id as the user of the host system.

But you could copy and chown the `.Xauthority` file, then it should work without `xhost +x`.

As @MartijnBraam mentioned, having a function that properly sets up the chroot so X11 applications can run on the host's X server would be useful for qemu as well!

What do you think about creating a new function `pmb.chroot.other.copy_xauthority()` (in `pmb/chroot/other.py`) that checks if the xauthority file exists in the chroot, and if it does not, copies it over and prints out a message to the log?

pmb/build/menuconfig.py
```diff
@@ -81,10 +124,10 @@ def menuconfig(args, pkgname):
     pmb.chroot.user(args, ["abuild", "prepare"], "native",
                     "/home/pmos/build", log=False, env={"CARCH": arch})
 
-    # Run abuild menuconfig
-    logging.info("(native) run menuconfig")
-    pmb.chroot.user(args, ["abuild", "-d", "menuconfig"], "native",
-                    "/home/pmos/build", log=False, env={"CARCH": arch})
+    # Run make menuconfig
+    logging.info("(native) make menuconfig: " + get_builddir(args))
```
ollieparanoid: I'd recommend to put `builddir = get_builddir(args)` here and use that instead of calling `get_builddir()` twice, so it doesn't have to find the path twice.

pmb/helpers/file.py
```diff
@@ -70,7 +70,7 @@ def is_older_than(path, seconds):
     return lastmod + seconds < time.time()
 
 
-def symlink(args, file, link):
+def symlink(args, file, link, su=False):
```
ollieparanoid: As written above, I don't think this works with the symlink. So if I'm not wrong with my assumption, then all modifications to `symlink()` should be reverted.

steamp0rt: Without this, symlink had permission denied

steamp0rt:
I honestly don't like it

.gitignore
```diff
@@ -110,3 +110,6 @@ ENV/
 
 # Pytest
 .pytest_cache
+
+# PyCharm
```
z3ntu: I would call the comment `Jetbrains IDEs` or similar

README.md
ollieparanoid: (This comment is outdated, as it is `kconfig check` everywhere now.)

pmb/build/menuconfig.py
```diff
@@ -67,12 +92,30 @@ def menuconfig(args, pkgname):
...
     pmb.chroot.apk.install(args, depends)
 
+    # Symlink host's .xauthority into native
+    if args.frontend != "n" and args.frontend != "ncurses":
+        logging.info("(native) copy host Xauthority")
+        file = os.environ.get('XAUTHORITY')
+        if not file:
+            file = os.path.join(os.environ['HOME'], '.Xauthority')
+        link = args.work + "/chroot_native/home/pmos/.Xauthority"
+        pmb.helpers.run.root(args, ["rm", "-f", link])
```
ollieparanoid: This will throw an error if the file does not exist, so we need to check first:
```py
if os.path.exists(link):
	pmb.helpers.run.root(args, ["rm", link])
```
This also removes the `-f` flag, which isn't needed, right?

helpers/pmbootstrap-autocompletion.zsh
ollieparanoid: Let's put `echo "edit check"` here for now, so it matches what `pmbootstrap flasher` is completing. This file could use a general update to support more completions (or even autogenerate it: #1517), but this is out of scope for this PR.

pmb/build/menuconfig.py
```diff
@@ -67,12 +92,30 @@ def menuconfig(args, pkgname):
...
+    if args.frontend in ["n", "ncurses"]:
+        depends += ["ncurses-dev"]
+    elif args.frontend in ["g", "gconfig"]:
+        depends += ["gtk+2.0-dev", "glib-dev", "libglade-dev", "font-noto"]
+        kopt = "gconfig"
     pmb.chroot.apk.install(args, depends)
 
+    # Symlink host's .xauthority into native
+    if args.frontend != "n" and args.frontend != "ncurses":
+        logging.info("(native) copy host Xauthority")
```
ollieparanoid: I think it would be nice if we moved the whole "copy Xauthority" block here into its own function, so we can re-use it with Qemu later. See https://github.com/postmarketOS/pmbootstrap/pull/1509#discussion_r190992485:

> What do you think about creating a new function `pmb.chroot.other.copy_xauthority()` (in `pmb/chroot/other.py`) that checks if the xauthority file exists in the chroot, and if it does not, copies it over and prints out a message to the log?

pmb/parse/arguments.py
```diff
@@ -185,6 +185,25 @@ def arguments_newapkbuild(subparser):
...
+    ret = subparser.add_parser("kconfig", help="manage kernel aport options")
+
+    sub = ret.add_subparsers(dest="action_kconfig")
+
+    check = sub.add_parser("check", help="check kernel aport config")
+    check.add_argument("--arch", choices=arch_choices, dest="arch")
+    check.add_argument("package", default="", nargs='?')
+
+    edit = sub.add_parser("edit", help="edit kernel aport config")
+    edit.add_argument("--frontend", choices=frontend_choices, default="ncurses", dest="frontend", help="The frontend that should be used for menuconfig.")
```
ollieparanoid: In my [previous review](https://github.com/postmarketOS/pmbootstrap/pull/1509#pullrequestreview-123492125) I've proposed to replace `--frontend` with `-x` (xconfig) and `-g` (gconfig).
Thoughts on that?

pmb/build/menuconfig.py
```diff
@@ -81,10 +124,10 @@ def menuconfig(args, pkgname):
...
                     "/home/pmos/build", log=False, env={"CARCH": arch})
 
-    # Run abuild menuconfig
-    logging.info("(native) run menuconfig")
-    pmb.chroot.user(args, ["abuild", "-d", "menuconfig"], "native",
-                    "/home/pmos/build", log=False, env={"CARCH": arch})
+    # Run make menuconfig
+    logging.info("(native) make menuconfig")
+    pmb.chroot.user(args, ["make", kopt], "native",
+                    get_builddir(args), log=False, env={"ARCH": pmb.parse.arch.alpine_to_kernel(arch), "DISPLAY": ":0", "XAUTHORITY": "/home/pmos/.Xauthority"})
```
ollieparanoid: This line is pretty long, how about inserting some line breaks?

pmb/parse/kconfig.py
```diff
@@ -90,7 +90,7 @@ def check(args, pkgname, details=False):
                     else:
                         logging.warning("WARNING: " + path + " isn't configured"
                                         " properly for postmarketOS, run"
-                                        " 'pmbootstrap kconfig_check' for"
+                                        " 'pmbootstrap kconfig --check' for"
```
ollieparanoid: That would be `kconfig check` then.

pmb/parse/arguments.py
steamp0rt: It's more consistent as --frontend, though? We don't have `-arm` or `-arm64` or `-x86`

ollieparanoid: Well, we use both styles. We have this:
* `pmbootstrap build  [--arch {x86_64,armhf,aarch64,x86}]`

But also the mutually exclusive groups:
* `pmbootstrap qemu [--display | --spice]`
* `pmbootstrap chroot [-r | -b ... | -s ... ]`
* `pmbootstrap newapkbuild [-a | -C | -m | -p | -y] ...`

So from the consistency perspective I think both would be okay.

I've suggested `-x` and `-g` because they are much shorter (and that way we don't need to add the shorthand values to `--frontend`, which we don't do anywhere else either).

pmb/build/menuconfig.py
```diff
@@ -81,10 +124,11 @@ def menuconfig(args, pkgname):
     pmb.chroot.user(args, ["abuild", "prepare"], "native",
                     "/home/pmos/build", log=False, env={"CARCH": arch})
 
-    # Run abuild menuconfig
-    logging.info("(native) run menuconfig")
-    pmb.chroot.user(args, ["abuild", "-d", "menuconfig"], "native",
-                    "/home/pmos/build", log=False, env={"CARCH": arch})
+    # Run make menuconfig
+    logging.info("(native) make menuconfig")
```
ollieparanoid: How about: `logging.info("(native) make " + kopt)`

pmb/helpers/frontend.py
```diff
@@ -233,23 +233,31 @@ def newapkbuild(args):
...
+        # At least one failure
+        if error:
+            raise RuntimeError("kconfig check failed!")
+        else:
+            logging.info("kconfig check succeded!")
 
-    # At least one failure
-    if error:
-        raise RuntimeError("kconfig_check failed!")
+    elif args.package and args.action_kconfig != "check":
```
ollieparanoid: I would change that to:
`elif args.action_kconfig == "edit":`

The package should be required with `edit` anyway (argparse can check for that).

pmb/parse/arguments.py
```diff
@@ -185,6 +185,25 @@ def arguments_newapkbuild(subparser):
                      " download link to the source archive")
 
 
+def arguments_kconfig(subparser):
+    arch_native = pmb.parse.arch.alpine_native()
+    arch_choices = set(pmb.config.build_device_architectures + [arch_native])
+
+    ret = subparser.add_parser("kconfig", help="manage kernel aport options")
```
ollieparanoid: I'm not sure if that makes clear what `pmbootstrap kconfig` can do. What do you think about `help="change or edit kernel configs"` instead?

pmb/parse/arguments.py
```diff
@@ -185,6 +185,25 @@ def arguments_newapkbuild(subparser):
...
+    sub = ret.add_subparsers(dest="action_kconfig")
+
+    check = sub.add_parser("check", help="check kernel aport config")
+    check.add_argument("--arch", choices=arch_choices, dest="arch")
+    check.add_argument("package", default="", nargs='?')
+
+    edit = sub.add_parser("edit", help="edit kernel aport config")
+    edit.add_argument("--arch", choices=arch_choices, dest="arch")
+    edit.add_argument("-x", dest="x", action="store_true", help="use xconfig rather than ncurses for kernel configuration")
+    edit.add_argument("-g", dest="g", action="store_true", help="use gconfig rather than ncurses for kernel configuration")
```
ollieparanoid: Can we use `dest="gconfig"` and `dest="xconfig"` instead?

pmb/parse/arguments.py
```diff
@@ -185,6 +185,25 @@ def arguments_newapkbuild(subparser):
...
 
 
+def arguments_kconfig(subparser):
+    arch_native = pmb.parse.arch.alpine_native()
+    arch_choices = set(pmb.config.build_device_architectures + [arch_native])
+
+    ret = subparser.add_parser("kconfig", help="manage kernel aport options")
+
+    sub = ret.add_subparsers(dest="action_kconfig")
+
```
ollieparanoid: When we add `sub.required = True` here, it should require a kconfig_action to be set (see comment above).

pmb/parse/arguments.py
```diff
@@ -185,6 +185,25 @@ def arguments_newapkbuild(subparser):
...
+    check = sub.add_parser("check", help="check kernel aport config")
+    check.add_argument("--arch", choices=arch_choices, dest="arch")
+    check.add_argument("package", default="", nargs='?')
+
+    edit = sub.add_parser("edit", help="edit kernel aport config")
+    edit.add_argument("--arch", choices=arch_choices, dest="arch")
+    edit.add_argument("-x", dest="x", action="store_true", help="use xconfig rather than ncurses for kernel configuration")
+    edit.add_argument("-g", dest="g", action="store_true", help="use gconfig rather than ncurses for kernel configuration")
+    edit.add_argument("package", default="", nargs='?')
```
ollieparanoid: `nargs='?'` means: specify zero or more packages here. But when editing the kernel config, we want exactly one package. That would be:
```py
edit.add_argument("package")
```

ollieparanoid:
Getting close! What do you think of the rest of my previous review?

ollieparanoid:
Now it properly *saves* the config. However, it does not load the right config file yet. After saving, "kconfig check" reports warnings for options that weren't changed and the log says:

> using defaults found in arch/x86/configs/x86_64_defconfig

Besides that, as written above:
> Can you add a compatibility wrapper at least for "pmbootstrap menuconfig"?

steamp0rt:
Looks good to me, @ollieparanoid

aports/device/linux-amazon-thor/APKBUILD
```diff
@@ -75,15 +75,6 @@ prepare() {
 		silentoldconfig
 }
 
-
-
-# this is so we can do: 'abuild menuconfig' to reconfigure kernel
-menuconfig() {
```
drebrez: @steamp0rt @ollieparanoid with this change the `menuconfig` function is not necessary anymore? removing it from all the kernel packages is too much?

pmb/helpers/frontend.py
```diff
@@ -188,6 +188,9 @@ def export(args):
 
 
 def menuconfig(args):
+    logging.warning("WARNING: 'pmbootstrap menuconfig' is deprecated and will"
```
drebrez: @steamp0rt @ollieparanoid this warning says that `pmbootstrap menuconfig` will be removed soon, but it's already removed from `pmb/parse/arguments.py`

steamp0rt: I believe he set up an alias?

aports/device/linux-amazon-thor/APKBUILD
drebrez: @steamp0rt @ollieparanoid with this change the `menuconfig` function is not necessary anymore? removing it from all the kernel packages is too much?
maybe without changing the `pkgrel` like in this case, so it's not necessary to rebuild them

pmb/helpers/frontend.py
drebrez: @steamp0rt @ollieparanoid this warning says that `pmbootstrap menuconfig` will be removed soon, but it's already removed from `pmb/parse/arguments.py`

drebrez: Ignore this, I've just noticed the `legacy_menuconfig ` now

pmb/parse/kconfig.py
```diff
@@ -90,7 +90,7 @@ def check(args, pkgname, details=False):
                     else:
                         logging.warning("WARNING: " + path + " isn't configured"
                                         " properly for postmarketOS, run"
-                                        " 'pmbootstrap kconfig_check' for"
+                                        " 'pmbootstrap kconfig  check' for"
```
drebrez: @steamp0rt there are 2 spaces between `kconfig` and `check`

ollieparanoid:
Thanks a lot for the review @drebrez! Do you think it's ready now?

drebrez:
@steamp0rt @ollieparanoid code looks good. I didn't had a chance to test it manually, but I think you both already tested it.

ollieparanoid:
Thanks @drebrez!
Regarding SSH, this won't work out of the box, because you would need the `.xauthority` from the PC where you SSH from, and you would need to forward the X11 port.

ollieparanoid:
I've changed `pmbootstrap menuconfig` to `pmbootstrap kconfig edit` in most places, and `pmbootstrap kconfig_check` to ...`kconfig check` everywhere.
@steamp0rt: do you want to add a section [here in the wiki](https://wiki.postmarketos.org/wiki/Kernel_configuration) that describes how to use `xconfig` or `gconfig`, possibly with screen shots?

PR #1510, opened 2018-05-24
started port

ragon000:
[Wiki Link](https://wiki.postmarketos.org/wiki/Samsung_Wave_2_(samsung-s8530))


PureTryOut:
Next time please mention what device you're porting in the title of the PR. Also, since you removed the default text in the PR, I assume you don't want this merged via the default Github way?

The wiki page says you haven't flashed it yet, do you need help with anything?

mirh:
https://github.com/AMD-Grifon/android_kernel_samsung_aries/tree/cma
This tree is possible better. 

Also, I wonder if there are finally methods that aren't raw flashing of emmc blocks for Wave 3?

PR #1512, opened 2018-05-26
Introduce calls on hammerhead [skip ci]

bhush9:
postmarketOS birthday edition! :tada: 

This can in theory get voice calls working with Plasma Mobile with some changes to plasma-mobile package, but I can't get to test it due to broken SIM atm.

This will also make it possible to call through dbus without any changes.

CC: @scintill please check that ofono patches doesn't mess up on your device :)

---
[ ] Merge on GitHub (see <https://postmarketos.org/merge>)


MartijnBraam:
I've got a phone call working from the hammerhead with the first commit in this PR using the python script from the ofono repository (https://git.kernel.org/pub/scm/network/ofono/ofono.git/tree/test/dial-number?id=7fb48999707c778212707c634a84ef87852ec86e)
call

aports/main/ofono-qt/APKBUILD
```diff
@@ -0,0 +1,52 @@
...
+	08-add-delivery-report.patch
+	09-createMultiparty-synchronous.patch
+	10-use_right_modem_objpath.patch
+	11_use_libdir.patch
+"
+builddir="$srcdir/ofono-qt-1.5+git20120419+bcf0c04"
+
+build() {
+	cd "$builddir"
+    qmake-qt5
```
ollieparanoid: spaces used instead of tabs (this line and the line below)

aports/main/libphonenumber/APKBUILD
```diff
@@ -0,0 +1,46 @@
...
+arch="all"
+license="Apache"
+depends_dev="boost-dev protobuf-dev icu-dev"
+makedepends="$depends_dev cmake gtest-dev"
+subpackages="$pkgname-dev"
+source="$pkgname-$pkgver.tar.gz::https://github.com/googlei18n/${pkgname}/archive/v${pkgver}.tar.gz
+		googletest-1.8.0.tar.gz::https://github.com/google/googletest/archive/release-1.8.0.tar.gz"
+builddir="$srcdir/$pkgname-$pkgver"
+
+# this package doesn't play nicely when built with ccache and distcc, clear ccache
```
ollieparanoid: What happens exactly when building this with ccache multiple times (log output)?

aports/main/libphonenumber/APKBUILD
```diff
@@ -0,0 +1,46 @@
+# Contributor: Bhushan Shah <bshah@kde.org>
+# Maintainer: Bhushan Shah <bshah@kde.org>
+pkgname=libphonenumber
+pkgver=8.9.6
+pkgrel=0
+pkgdesc="Google's common Java, C++ and JavaScript library for parsing, formatting, and validating international phone numbers."
```
ollieparanoid: How about shortening that to:
`pkgdesc="international phone number parsing/formatting/validating library"`

aports/main/ofono-qt/APKBUILD
bhush9: Fixed.

aports/main/libphonenumber/APKBUILD
bhush9: Fixed.

aports/main/libphonenumber/APKBUILD
bhush9: There are multiple weird errors everytime, mainly it comes from the fact that this library is built using boost templates, so the generated code is not completely reproducible and compatible with ccache.

ollieparanoid: Okay, how about we disable ccache in `build()` then with `export CCACHE_DISABLE=1` and adjust the comment?

bhush9:
On Mon, May 28, 2018 at 02:52:26PM -0700, Oliver Smith wrote:
> Okay, how about we disable ccache in `build()` then with `export CCACHE_DISABLE=1` and adjust the comment?

okay I will change that.

ollieparanoid:
Merged.

PR #1515, opened 2018-05-26
Update linux-samsung-p4wifi config to be based on mainline tegra_defconfig

Decatf:
Previously the kernel config was based on the config from a mainline-ish kernel for Android. I have dropped the remaining Android specific patches. It turned out most weren't needed. I generated a new config based on `tegra_defconfig` which is in mainline. The code is now mainline with out-of-tree drivers and corresponding hardware specific patches/fixes/hacks. No more Android stuff.

There's no functionality change. The working / not working status remains the same.

@ollieparanoid:
You asked before about a having a common branch such as `linux-postmarketos-tegra`. There are quite a few device specific quirks which I am guarding with checks in probe/init functions. Example:
```
if (of_machine_is_compatible("samsung,p4wifi")) {
     // handle device quirks
}
```
A pretty low impact example here: https://github.com/Decatf/linux/commit/8f7ebdc547eebad405104a31723c8ab65e019353. There are several such things throughout. This method is taken from mainline but they use it very sporadically. I think there's only a handful of uses.

At the moment there is one ugly hack that I use in order to increase the system partition size. The default is really small (450MB). It's kind of ugly and prevents the kernel from being used in a generic way. I did not guard this with the `of_machine_is_compatible` check. It could be done there too if necessary. Otherwise this would remain as a device specific kernel.

Tegra already has sort of decent mainline support for basic SOC functionality anyways. I don't know what the approach is for `linux-postmarketos-qcom` but there's nothing critical for general functionality of the Tegra 2 SOC that is out-of-tree. What is out-of-tree are only the device specific quirks and ugly hacks for such things.

So yeah that's the state of this device. It doesn't matter to me if it remains as device specific kernel but it's not out of the realm of possibility for this device to exist as part of a common tegra kernel.


Decatf:
> I understand that. However, if you should have time and interest in improving the code more, you could consider reaching out to the tegra kernel developers on the LKML and ask them for advice on how to get rid of the hacks that are still necessary (my guess with my limited understanding of the code in question is, that it could be done by refactoring the existing code to cover the p4wifi use cases as well).

I've got one tiny patch that is currently in linux-next. I don't know when it would actually be in mainline though. Maybe 4.18? Anyways the reason I haven't pressed on everything else is because this tablet has some partition layout that didn't follow convention. So that kind of a big blocker. They (Google/Samsung engineers) tried to mainline it back in the day (2011-ish) which revealed this oddity. The patch wasn't accepted. That's why I don't have much hope that this particular device will be out of the box mainline but a mainline-ish device.

Other Tegra 2 tablets appear to be much closer to reference design. It's evidenced by a couple of Tegra 2 tablets which people got running the stock nvidia kernel with much less complications. I was kinda hoping maybe this would spur some interest in reviving these. There was a whole line Tegra based Asus tablets from this generation that have a keyboard + touchpad dock which would be a good demo.

ollieparanoid:
I haven't seen the discussion there, but assuming you would really want to get this in mainline I think you could ask the subsystem maintainers how the patch should look like in order to get it accepted (even if there are such quirks required). Or am I missing something?

PR #1522, opened 2018-05-30
Update libhybris and add more related packages

NotKit:
* update libhybris
    * change _rev to latest git commit (there was a mistake before)
    * update custom patch for EGL integration with Xorg (fixes QML apps and Blink rendering)
    * install EGL libs into /usr/lib/libhybris-egl as was discussed before to avoid conflicts with MESA
* package lxc-android configs from Halium - mounts Android filesystems and allows to start minimalistic Android system in LXC container
* package xf86-video-hwcomposer and dependencies - allows Xorg to render via Android HWComposer API and utilize EGL via libhybris


aports/hybris/drihybris/APKBUILD
```diff
@@ -0,0 +1,33 @@
...
+pkgver=1.0_git20171130
+pkgrel=0
+_commit=fbaf21e7cb343e0b7cd29f256ec28c46c01c07b8
+pkgdesc="custom DRI3-based Xorg extension for use with libhybris"
+url="https://github.com/NotKit/drihybris"
+arch="all"
+license="MIT"
+subpackages="$pkgname-doc $pkgname-dev"
+depends=
+makedepends="autoconf automake util-macros
```
ollieparanoid: There are still dependencies missing, [Travis](https://travis-ci.org/postmarketOS/pmbootstrap/builds/385719965#L1038) says:

```
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal 
autoreconf: configure.ac: tracing
autoreconf: configure.ac: not using Libtool
autoreconf: running: /usr/bin/autoconf
configure.ac:43: error: possibly undefined macro: AC_DISABLE_STATIC
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
configure.ac:44: error: possibly undefined macro: AC_PROG_LIBTOOL```
```
I think it needs to depend on libtool at least. You can check locally if you have all dependencies with: `pmbootstrap build --strict drihybris`.

aports/hybris/libhybris/libhybris-egl.post-install
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+ARCH=`uname -m`
```
ollieparanoid: [SC2006](https://github.com/koalaman/shellcheck/wiki/SC2006): Please use $(..) instead of legacy backticks.

aports/hybris/libhybris/libhybris-egl.post-install
```diff
@@ -0,0 +1,6 @@
...
+
+if [ ! -f /etc/ld-musl-$ARCH.path ]; then
+    echo "/usr/lib/libhybris-egl:/lib:/usr/local/lib:/usr/lib" > /etc/ld-musl-$ARCH.path
```
ollieparanoid: We use tabs in APKBUILDs, so I'd recommend to use them here as well.

aports/hybris/lxc-android/pre-start.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+lxc_config_path="$(dirname $LXC_CONFIG_FILE)"
+
+# Allow pre-start.sh to be provided by device package, otherwise use the default one
+if [ -f "$lxc_config_path/pre-start.custom.sh" ]; then
+   exec $lxc_config_path/pre-start.custom.sh
```
ollieparanoid: use tabs instead of spaces?

aports/hybris/xf86-video-hwcomposer/APKBUILD
```diff
@@ -0,0 +1,74 @@
...
+		make DESTDIR="${_tmppkgdir}/$pkgname-$_ver" install
+	done
+}
+
+package() {
+	cd "$builddir"
+	install -Dm644 COPYING "$pkgdir"/usr/share/licenses/$pkgname/COPYING
+	install -Dm644 "$srcdir"/30-modules-hybris.conf \
+		"$pkgdir"/etc/X11/xorg.conf.d/30-modules-hybris.conf
+
```
ollieparanoid: unneeded empty line

aports/hybris/lxc-android/lxc-android.post-install
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+
+if [ "`id -u user`" == "1000" ]; then
+	# Change default user uid from 1000 to 100000 to avoid conflicts.
+	# Beware that it ./pmbootstrap install creates user after installing
+	# packages, so it will work only when package is installed on live system
+
+	# usermod -u 100000 user doesn't work due to "user" user is being used
```
ollieparanoid: * This is commented out... so right now the user needs to manually change the UID of `user`?
* How about we simply set the default UID of the user created with the Python code to `100000`?
* We could also change when the user is created, so this always works in `pmbootstrap install`.

NotKit: Currently it does this hard way by editing /etc/passwd with sed. I considered changing default UID, but then we would need some logic to check if 1000 can be used or not depending on installed packages. It doesn't have to be 100000 though, anything higher than 9999 is a safe bet (check https://android.googlesource.com/platform/system/core/+/master/libcutils/include/private/android_filesystem_config.h for reference).

ollieparanoid: > check if 1000 can be used or not depending on installed packages.

What do you mean with that? If we use 100000 (or anything other over 9999) by default, then we can always assume that this is the user's UID without checking?

NotKit: If you're willing to change default UID for all cases (not only when hybris is installed), then yes. I think 12345 is used for pmbootstrap chroot right now?

ollieparanoid: Yes, I don't see any problem with that. And you are right with `12345`, it is set in `pmb.config.chroot_uid_user`.

ollieparanoid:
(Travis still complains that this doesn't build, let us know if you need help with that.)

PR #1523, opened 2018-05-30
Package nexmon.

### LINKED ISSUE #592: Package nexmon -- Broadcom firmware patches to fix broadpwn and enable monitor mode

steamp0rt:
Packages nexmon. WIP.

Closes #592


steamp0rt:
I don't exactly know how to subpackage this.

aports/firmware/nexmon/APKBUILD
```diff
@@ -0,0 +1,33 @@
+pkgname=nexmon
+pkgver=2.2.2_git20180530
```
ollieparanoid: (serious question) Why using the latest git revision instead of the latest stable 2.2.2 version?
https://github.com/seemoo-lab/nexmon/archive/2.2.2.tar.gz

By using stable releases it gets easier to keep the packages updated to the upstream version, and the releases should be more stable and tested than random git commits. However, there may be valid reasons to use the latest git revision (e.g. because the stable release has not been updated in a too long time and it's missing a critical fix or feature).

aports/firmware/nexmon/APKBUILD
```diff
@@ -0,0 +1,33 @@
...
+01_ldflags.patch
+"
+
+builddir="$srcdir/$pkgname-$_hash"
+options="!check"
+
+build() {
+	cd "$builddir"
+	touch DISABLE_STATISTICS
+	source setup_env.sh
```
ollieparanoid: This is not good, they are [shipping their own binary compiler and compiler plugin](https://github.com/seemoo-lab/nexmon/blob/a45ec745484022bae3c64899d87e4513487eff9d/setup_env.sh#L22-L23), and are using that. It would be better to compile them from source.

This is not trivial to be honest. We could try to make the [gcc-nexmon-plugin](https://github.com/seemoo-lab/nexmon/tree/master/buildtools/gcc-nexmon-plugin) work with the GCC packaged in Alpine, and then use Alpine's GCC for everything but this will probably not work out of the box.

aports/firmware/nexmon/APKBUILD
```diff
@@ -0,0 +1,33 @@
...
+	make firmwares
+}
+
```
ollieparanoid: You're missing the `package()` function, that's why the build fails right now.
Regarding your question on how subpackages work, check out:
* https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#subpackages
* aports using subpackages, such as `device-nokia-n900`
* a more sophisticated example of automatically generated subpackages from the folder names is `linux-firmware` from Alpine's aports

aports/firmware/nexmon/APKBUILD
ollieparanoid: Then change it to the stable version please :smile:

aports/firmware/nexmon/APKBUILD
ollieparanoid: Seems to be here:
`/usr/lib/gcc/armv6-alpine-linux-musl/6.4.0/plugin/include/lra-int.h`

See [package contents search](https://pkgs.alpinelinux.org/contents?path=%2A%2Finclude%2A&repo=main&page=1&arch=armhf&branch=edge&name=gcc).

(But as written earlier I advise against looking into this before the rest is resolved.)

PR #1525, opened 2018-05-31
kde/falkon: package 3.0.1

george-hopkins:
This pull-request adds a receipt to package [Falkon](https://www.falkon.org/), a cross-plattform browser based on QtWebEngine (previously known as QupZilla). It is targeted towards KDE users but can be used on other platforms, too.

Tested with QEMU (amd64):

Screenshot

---

### DIFF
--- a/aports/kde/falkon/APKBUILD
+++ b/aports/kde/falkon/APKBUILD
@@ -6,7 +6,8 @@ url="https://www.falkon.org/"
 arch="all"
 license="GPL-3.0-or-later"
 depends="qt5-qtbase-sqlite"
-makedepends="cmake extra-cmake-modules qt5-qtdeclarative-dev qt5-qtwebengine-dev qt5-qtx11extras-dev qt5-qttools-dev xcb-util-dev ki18n-dev"
+makedepends="cmake extra-cmake-modules qt5-qtdeclarative-dev qt5-qtwebengine-dev qt5-qtx11extras-dev qt5-qttools-dev xcb-util-dev ki18n-dev ttf-freefont"
+#checkdepends="ttf-freefont"
 source="https://download.kde.org/stable/falkon/$pkgver/falkon-$pkgver.tar.xz
 	disable-ld-fatal-warnings.patch
 	disable-backtraces.patch
@@ -21,6 +22,12 @@ build () {
 		-DCMAKE_INSTALL_PREFIX=/usr \
 		-DCMAKE_INSTALL_LIBDIR=lib
 	make
+}
+
+check() {
+	cd "$builddir/build"
+	# GUI tests fail (SEGV_MAPERR)
+	CTEST_OUTPUT_ON_FAILURE=TRUE QT_QPA_PLATFORM=offscreen ctest || true
 }
 
 package() {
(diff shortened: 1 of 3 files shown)


aports/kde/falkon/APKBUILD
```diff
@@ -0,0 +1,33 @@
+pkgname=falkon
+pkgver=3.0.1
+pkgrel=0
+pkgdesc="A KDE-based web browser using QtWebEngine rendering engine"
```
bhush9: Might I suggest little description change? "Cross-platform Qt Web Browser using QtWebEngine rendering engine"

(mostly because it works completely fine in !KDE environment :smile: )

george-hopkins: :+1: Sure!

drebrez: @george-hopkins If it's not KDE related, then I think it can be moved outside of the `aports/kde` folder, probably is better in `aports/main`

bhush9: It is kde related in sense upstream is KDE, so yes aports/kde is fine :)

michitux:
I can confirm that this builds natively under armhf and it works nicely, i.e., Falkon is quite usable with touch and so far the best browser I have tried. I have just one question: I have seen that there are some tests built. Is there a reason why you have not included a `check` function to execute the tests?

PureTryOut:
Huh interesting, I had a ready aport laying around for a while, but it wouldn't compile due to LibreSSL not being fully OpenSSL compatible. I tried again after the latest LibreSSL update but it still wouldn't compile. Glad you kept trying though!

aports/kde/falkon/APKBUILD
```diff
@@ -0,0 +1,33 @@
+pkgname=falkon
+pkgver=3.0.1
+pkgrel=0
+pkgdesc="Cross-platform web browser using QtWebEngine rendering engine"
+url="https://www.falkon.org/"
+arch="all"
+license="GPL-3.0-or-later"
```
PureTryOut: "GPL-3.0+"

MartijnBraam: GPL-3.0-or-later seems correct since the APKBUILD reference says to use the SPDX identifier if possible.

> If the license is on the SPDX License List or SPDX License Exceptions, use the identifier specified by SPDX.

george-hopkins:
@michitux Unfortunately, I haven't been able to run tests during the build. I appended the receipt such that the failures are logged and can be examined. On my machine, all GUI tests fail with `SEGV_MAPERR`.

michitux:
@george-hopkins Okay, that doesn't seem trivial to fix. Do you know if Falkon has any kind of CI setup? I couldn't find anything on a quick search. I personally wouldn't mind skipping the tests for now as long as it is documented why.

george-hopkins:
Yes, there seem to be four CI jobs running by KDE. However, I'm not sure whether these are used by the project maintainers. Judging from the `stable-kf5-qt5 ` tests it should be possible to run all tests successfully. Unfortunately, I haven't been able to find out yet how to configure a similar environment.

bhush9:
Most (if not some) tests would require running GUI session (Xvfb) and also dbus which I believe is not possible with pmbootstrap/aports(?)

PR #1526, opened 2018-05-31
[Trivial] Add charging mode detection for kylevess

MayeulC:
As documented here:
https://wiki.postmarketos.org/wiki/Samsung_Galaxy_Trend_lite_(s7390g)

This might be set on other devices as well.

This is my first time committing using GitHub's web interface, it isn't that bad!

---


drebrez:
@MayeulC you forgot to recompute the checksum: `pmbootstrap checksum postmarketos-mkinitfs`

MayeulC:
Right, well, I guess GitHub's web interface won't allow me to do this easily, after all :)

PR #1530, opened 2018-06-01
main/qt5-qtwebchannel: remove dependeny on QML

george-hopkins:
In case an application doesn't use QML, the declared dependencies pull in unneeded packages (especially a lot of *-dev packages). If applications like to use QML, they should depend on qt5-qtdeclarative/qt5-quickcontrols2 themselves. This change might break applications which currently assume QML will be available. This change is in line with the [script from upstream](https://git.alpinelinux.org/cgit/aports/tree/community/qt5-qtwebchannel/APKBUILD).

---


NotKit:
@ollieparanoid the proposed change doesn't affect build time dependencies, so it will still build with QML support this way. I'm not sure why qt5-qtquickcontrols2-dev is currently in runtime deps anyway, looks like a typo to me?

george-hopkins:
Yes, as @NotKit mentioned, QML is still supported, it just not enforced when the package is installed. 
However, I just saw that the script from upstream supports QML too (introduced [in February](https://github.com/alpinelinux/aports/commit/930b767c25fcd95507114f74e20cca11bad49a9e)). I think we should be fine if we just remove this fork. There is no special treatment for `qt5-quickcontrols2-dev`, by including `qt5-declarative-dev` we already get support for QML.

ollieparanoid:
You guys are right, it should have been in depends instead of makedepends, which is exactly what the upstream aport has now. Our fork of the package was deleted with the commit above.

Thanks again!

PR #1532, opened 2018-06-01
kde/falkon: disable checks

ollieparanoid:
Disable checks for good.
* half of them would not run anyway because they require X11
* falkon-cookietest hangs for more than 30 minutes with 100% CPU usage
  while cross compiling for armhf

I'm creating a PR for this, so Travis runs on it.

---


ollieparanoid:
Travis ran through!

PR #1537, opened 2018-06-05
update upstream linux-firmware to latest

yangxuan8282:
update upstream linux-firmware to latest, to include raspberry pi 3B+ Wi-Fi firmware 
---


PureTryOut:
You have some commits in here that have nothing to do with the PR.
That said, is it not better to update the upstream (read: Alpine's) linux-firmware rather than add it to our own repo?

yangxuan8282:
@PureTryOut 
yes, the first 5 commit is not related to this PR, it was created when I was trying to sync master branch with pmos upstream, I don't know how to clean it, maybe I should delete previous forked repos and fork it again

if this changes can be accepted by alpine upstream will be better, but since the alpine maintainer not have any interest merge my `linux-rpi` PR, I guess they won't accept this one either 

this PR only add three files related to rpi 3B+ wifi, and won't harm other user

yangxuan8282:
@ollieparanoid 
hi, I just found this package need to add `!tracedeps` to `options` sections to pass pmbootstrap build.

If all package in temp folder need to compile on local machine? I thought it was build by CI, this package have many subpackages, if the user network is slow that would take some times to download. 

Since this commit is just for add three rpi 3b+ wifi firmware files, how about add a separete package for it? 

It will look like this packages: https://github.com/yangxuan8282/rpi-aports/blob/master/firmware-raspberrypi/APKBUILD , but without brcmfmac43430 related files, since the upstream `linux-firmware` have already include it.

ollieparanoid:
> hi, I just found this package need to add `!tracedeps` to `options` sections to pass pmbootstrap build.

Thanks, I've missed that. Fixed in b2410d5df01896aba7a0211c85586da6402c1875, see the commit message for details.

> If all package in temp folder need to compile on local machine? I thought it was build by CI, this package have many subpackages, if the user network is slow that would take some times to download.

We have the postmarketOS binary package repo with pre-built packages for everything except for aports/firmware (for now, we have a solution for that as well). The repo gets filled up when I manually start the package building script on my dedicated package building PC. I haven't done this since this PR got merged, which is why the binary package was not available (but now it should be up fairly soon, just started the process).

I know this is not the best workflow, it would be better to have a CI as you've said that does this automatically. That's where I want to take it in the near future.

> how about add a separete package for it?

I think this would be a logical change to do in Alpine's aports. Not too long ago, there was only one `linux-firwmare` package without subpackages, which had the Raspberry Pi firmware added from external sources. But now it has subpackages, so it would make sense to put the Raspberry Pi firmware into its own package. Personally, I would advise against making this change in pmOS only and not in Alpine, because then we unnecessarily diverge from upstream (the other changes can be upstreamed in one big raspberry pi update PR, after that is through we could look into the dedicated raspberry pi firmware package in upstream).

PR #1543, opened 2018-06-08
temp/linux-rpi: update to latest rpi kernel

yangxuan8282:
update linux-rpi to 4.14.44
with vc4 enabled, so we can use drm backend on raspberry pi platform 
---


yangxuan8282:
@ollieparanoid 

I'll undo the `DEVTMPFS_MOUNT` changes after #1548 merged 

and there are some info maybe not related to this PR, but related to pmos on rpi, I guess need some tweaks on kernel config

whether use current upstream alpine linux-rpi 4.9 or my forked 4.14 kernel not make big difference ( the most part of kernel config file should same ) 

- sometimes first partition won't auto mount 

I used to mention this in https://github.com/postmarketOS/pmbootstrap/pull/1501#issuecomment-390371805 , and I got this issues again later. After compare dmesg output, 

when it successful auto mount first partition:

```
[   20.916425] EXT4-fs (mmcblk0p2): mounted filesystem with ordered data mode. Opts: (null)
...
[   22.994128] EXT4-fs (mmcblk0p2): re-mounted. Opts: data=ordered
```

when it not mount, it's something like:

(this is not what exactly is, but basically same)

```
EXT4-fs (dm-1): mounted filesystem with ordered data mode. Opts: (null)
...
EXT4-fs (dm-1): re-mounted. Opts: data=ordered
```

I'm not quite sure about how it mount, but seems when the device mounted by device mapper the first partition can't be mount proprely.

- take long time to boot

I have tried several sd card on different rpi model (zero w, 3B, 3B+), it all take long time to boot, it take around 40-50 seconds to waiting desktop appear, the sd card speed is not so fast ( write speed is about 14MB/s, read speed is close ) , but with same device on raspbian it only take about 20 seconds or less to boot. and seems overclock the sd card with `dtparam=sd_overclock=100` in `/boot/usercfg.txt` not really help either ( this overclocking is not some dangers things like overclock cpu, it's just take advantage of your sd card, instead of use default 50, which is just for compatible with more sd card)

with `vfat` added in `deviceinfo_modules_initfs` section of `deviceinfo` can speed up about 10 seconds, so the boot time is 35 seconds, but I think it should be better

yangxuan8282:
seems without `CONFIG_DEVTMPFS_MOUNT` the rpi zero w and 3b can boot normally, but I'm not sure if there are potential danger, I have checked other rpi kernel config file, both raspbian ( [rpi config](https://github.com/raspberrypi/linux/blob/rpi-4.14.y/arch/arm/configs/bcmrpi_defconfig#L390), [rpi2 config](https://github.com/raspberrypi/linux/blob/rpi-4.14.y/arch/arm/configs/bcm2709_defconfig#L395) ) and archlinux arm ( [rpi + rpi2](https://github.com/archlinuxarm/PKGBUILDs/blob/master/core/linux-raspberrypi/config#L1590) ) have included it

ollieparanoid:
> I'll undo the `DEVTMPFS_MOUNT` changes after #1548 merged 

Thanks, it's merged now!

> sometimes first partition won't auto mount

We have code in place, that should wait for the SD card to appear here:

https://github.com/postmarketOS/pmbootstrap/blob/cd181c541a69d0db0980f2f0ce55813e4452c0d3/aports/main/postmarketos-mkinitfs/init.sh.in#L35-L38

https://github.com/postmarketOS/pmbootstrap/blob/cd181c541a69d0db0980f2f0ce55813e4452c0d3/aports/main/postmarketos-mkinitfs/init_functions.sh#L122-L130

But after that, the initramfs-extra has been introduced, so we try to mount the /boot partition before waiting for the SD card. This should work as expected again when moving `wait_root_partition` before `mount_boot_partition`. Would you like to test this, and make a new PR for that change?

> - take long time to boot

OpenRC has sequential boot, and on first boot it generates SSH keys for example (so it blocks until those are generated). We also have chrony running, which may hang (I've noted that in Qemu but not looked into it much more yet).
You could try to enable the console logging during boot (to tty0), so you could see where it hangs.
If you want to optimize harder, you could try to run bootchart while it is booting (probably requires modification of the initramfs code). I agree that this needs to be improved, it is very notable on the phones as well, but no one looked much into that yet.

> with `vfat` added in `deviceinfo_modules_initfs` section of `deviceinfo` can speed up about 10 seconds, so the boot time is 35 seconds, but I think it should be better

Cool, we should do that optimization then. But are you sure that it's the vfat change, and not the SSH keys that were generated on fist boot?

yangxuan8282:
@ollieparanoid 
for boot time, since I have reboot many times, so it's not related to first boot generate ssh keys, I think add `vfat` can help improve boot time for rpi

and after move `wait_root_partition` before `mount_boot_partition`, it take little longer time to boot (it's 40s now), did not got first partition won't auto mount issues for now, but need reboot more times to test

**/usr/share/postmarketos-mkinitfs/init.sh.in**

```
...
wait_root_partition
mount_boot_partition
extract_initramfs_extra @INITRAMFS_EXTRA@
start_charging_mode
#wait_root_partition
...
```

yangxuan8282:
@ollieparanoid 
the build seems failed because of timeout, would you mind increase the build time limit?

ollieparanoid:
Travis CI has a limit of 50 minutes which can not be increased. But I can test it manually on my PC and report back (tomorrow that is, right now I'm building the KDE frameworks update).

> UPDATE: seems move wait_root_partition before mount_boot_partition not fix the mount first partition issues

That is weird! How about we add a similar loop around the `mount_boot_partition` step instead (try it once, on failure show error splash and retry every second), would you like to try that?

yangxuan8282:
@ollieparanoid 
yes, I think it's ready to be merged.

yangxuan8282:
seems Alpine upstream just bump linux-rpi to 4.14.44, based on raspberry pi downstream config file, and add aarch64 support, maybe we should keep use the upstream package?

https://github.com/alpinelinux/aports/commit/ff478c74b2ec904371253c4977a16fe0643ec4d8

daveloyall:
Perfect is the enemy of good, yangxuan8282. :)

ollieparanoid:
Oh... in that case it makes more sense to work with the upstream version. Sorry for the mess and thanks for the work you have put into this. At least they did respond to your upstream PR, so hopefully more attention will be paid for future raspberry pi improvement PRs.

daveloyall:
Oh, I get it now.  I think... 

But, how does [this](https://github.com/alpinelinux/aports/commit/bcb6feb8283f565da1b62082d9ebe8466f7e5e94) make it into` pmbootstrap`?

Is it pulled in at runtime?

PureTryOut:
You're asking how pmbootstrap knows how to install that package? The package is available pre-compiled in the Alpine repos, so pmbootstrap just pulls it from there. On a running system apk will see it's version is newer than the one you already have installed, and will update it for you.

yangxuan8282:
@ollieparanoid 
I found the slow boot maybe related to chrony, I did a alpine install on rpi with this [script](https://github.com/yangxuan8282/gen-rpi_os/blob/master/gen-alpine_rpi.sh), and add same openrc service as pmos(except hwclock, I delete it because rpi not have hardware clock). because alpine display openrc start progress, so I found every time it stuck at chronyd for over 10 seconds even I have replace default ntp server. after replace chrony with busybox ntp (/etc/init.d/ntpd) the boot progress is much faster than before, the boot time(from rpi color splash screen to see desktop) is basically same as raspbian.
on pmos when I replace chrony with busybox ntp, I found it didn't work

```
logread | grep ntp
Jun 13 04:26:12 raspberrypi daemon.err /etc/init.d/ntpd[1035]: start-stop-daemon: failed to start `/usr/sbin/ntpd'
Jun 13 04:26:12 raspberrypi daemon.err /etc/init.d/ntpd[1015]: ERROR: ntpd failed to start
```

I'm not sure why this happen

PR #1548, opened 2018-06-09
Allow DEVTMPFS_MOUNT in kernel config

ollieparanoid:
It is important, that we have DEVTMPFS enabled in the kernel config. But it does not hurt to have DEVTMPFS_MOUNT enabled as well, and some of Alpine's kernel configs have that by default.

This commit removes the check that forbids the option in kernel configs, so we can fork the raspberry pi kernel package from Alpine without changing unrelated options.

---


z3ntu:
What was the reason we disallowed it in the first place?

z3ntu:
@Wouter92 You added that option in the wiki page: https://wiki.postmarketos.org/index.php?title=Kernel_configuration&type=revision&diff=569&oldid=568

MartijnBraam:
@z3ntu I thought it was because busybox mdev doesn't start when /dev was already mounted

ollieparanoid:
But we use udev (eudev) instead of mdev, so that should not matter even if it was the case. What do you think folks, shall we merge it?

MartijnBraam:
@ollieparanoid we do actually use mdev in the initfs: https://github.com/postmarketOS/pmbootstrap/blob/master/aports/main/postmarketos-mkinitfs/init_functions.sh#L36

ollieparanoid:
Oh you're right, I forgot about that. Thanks for testing it, with that result I don't see any reason why this should not be merged.

PR #1555, opened 2018-06-12
Update Plasma to 5.13.0

PureTryOut:
This update also means we can update `plasma-phone-components` again, so I did just that. Everything builds successfully on x86_64.


ollieparanoid:
Built successfully for `x86_64`, `armhf` and `aarch64`. Shipping! :ship:

Issue #1551, opened 2018-06-11
porting to new device: pmbootstrap kconfig edit Error


Moocow9m:
Command ran:  pmbootstrap kconfig edit samsung-i927

Error: make menuconfig
         /bin/sh: cd: line 1: can't cd to /home/pmos/build/src/build: No such file or directory
        make: *** No rule to make target 'menuconfig'.  Stop.

LOG: https://pastebin.com/dtSDFgLb

MayeulC:
Mmm... Sounds like there is a mismatch between the folder you are trying to cd into and the folders created when extracting your kernel... It would be interesting to see your apkbuild, especially the prepare() part.

You can also `pmbootstrap chroot` in after it failed, and manually look around /home/pmos/build, then change the path in the APKBUILD to the correct one.

Moocow9m:
@MayeulC APKBUILD: https://pastebin.com/YWJeCrYB

ollieparanoid:
Welcome to postmarketOS @Moocow9m and thanks for opening this issue!

This is possibly a regression from #1509. I will look into this tomorrow.

In the meantime, you could revert that PR (`git revert 30df072`) and add this `menuconfig()` function to your kernel's `APKBUILD`. In case this really was caused by a regression, then the menuconfig should appear with  `pmbootstrap menuconfig samsung-i927`.

```shell
menuconfig() {
	cd "$srcdir"/build
	make ARCH="$_carch" menuconfig
	cp .config "$startdir"/$_config
}
```

In case it does not help, then there's something different with your kernel repository compared to the other kernel repos (consider trying another kernel).

Sorry for the inconvenience, we'll figure this out :+1:

Moocow9m:
@ollieparanoid Same error. I am undoing the revert now.

ollieparanoid:
Thanks for trying that out, this made it easier to figure out the problem! I've reproduced your error now, and found a solution. When changing the last part of `prepare()` to the following, it works as expected:


```shell
    # Prepare kernel config ('yes ""' for kernels lacking olddefconfig)
    mkdir -p "$srcdir"/build
    cp "$srcdir"/$_config "$builddir"/.config
    yes "" | make O="$srcdir"/build ARCH="$_carch" HOSTCC="$HOSTCC" oldconfig
}
```

As I understand it, your kernel has a different default value for `KBUILD_OUTPUT` defined somewhere (although I could not find where exactly).

PR #1556 will add this by default to the newly generated templates.

[...]

PR #1556, opened 2018-06-12
aportgen linux: always override KBUILD_OUTPUT

ollieparanoid:
Some kernels have a different `KBUILD_OUTPUT` path (e.g. #1551). When
the output path is different from `$srcdir/build`, then
`pmbootstrap kconfig edit` will not work (same with the previous
`pmbootstrap menuconfig` implementation).

This commit forces the output path to be `$srcdir/build` in the template
for new kernel aports, so we won't have that issue with future ports.

---


z3ntu:
Fixes the issue for my FP1 aport.

PR #1557, opened 2018-06-12
Fix conflict with mkbootimg and android-tools

ollieparanoid:
### Motivation
Alpine ships `mkbootimg` with the `android-tools` package now. This conflicts with the `mkbootimg` fork from osm0sis (see #441).

### Changes
* Rename `mkbootimg` to `mkbootimg-osm0sis` (aport and binary name)
* `mkbootimg-osm0sis`: provides `mkbootimg` now (so we don't need to change all the device aports), update version to 2018.05.10
* Adjust our `mkinitfs` script to call `mkbootimg-osm0sis`
* Better aport description

### Failing test case can be ignored here

`test_soname_bump` [does not pass](https://travis-ci.org/postmarketOS/pmbootstrap/builds/391009778#L834):
> RuntimeError: One or more packages need to be rebuilt, because a library they link against had an incompatible upgrade (soname bump). Run 'pmbootstrap pkgrel_bump --auto' to automatically increase the pkgrel in order to trigger a rebuild. If this test case failed during a pull request, the issue needs to be fixed on the 'master' branch first, then rebase your PR on 'master' afterwards.

But this is not an issue in this case, here's why:
* Many device packages depend on `mkbootimg`
* We have a binary `mkbootimg` package in the repositores, which satifies this requirement (the old package before this PR)
* `pkgrel_bump --auto` finds missing dependencies by looking at which packages provide dependencies.
* after this commit, `mkbootimg-osm0sis` provides `mkbootimg`, so the code complains that `mkbootimg-osm0sis` does not have a binary package
* it can not recognize that it's fine in this case, because `mkbootimg` has binary packages and provides the dependency as well

I've tested locally that the test case passes as soon as binary packages of `mkbootimg` are built for `armhf` and `aarch64`, which is what I will do right after the merge.

### How to test

* Run `pmbootstrap bootimg_analyze`
* Run `pmbootstrap install` for a device that needs the `bootimg` format (e.g. `lg-mako`)

(Without this PR, this crashes due to the conflict with `android-tools` and the old `mkbootimg`, and with this PR it works again.)

---


MayeulC:
Question: Does this mean we don't install `android-tools` anymore, like we were supposed to in https://wiki.postmarketos.org/wiki/Mainlining_Guide#Patch_boot.img ?

If so, I'll adjust the instructions. But I didn't really get how this would fix the conflict, if `mkbootimg` is still part of that package. The `mkbootimg` binaries from `android-tools` and `mkbootimg-osm0sis` would still conflict, right?

MayeulC:
Sorry for the delay, I will review this PR in a few hours.

ilpianista:
I got the conflict error on `flasher boot`, then I merged this and now I get this error on `install`:
```
ERROR: 'deviceinfo_generate_bootimg' is set, but the package 'mkbootimg' was not
installed! Please add 'mkbootimg' to the depends= line of your device's
```

aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -232,7 +232,7 @@ create_uboot_files()
 create_bootimg()
 {
 	[ "${deviceinfo_generate_bootimg}" == "true" ] || return
-	require_package "mkbootimg" "mkbootimg" "generate_bootimg"
+	require_package "mkbootimg-osm0sis" "mkbootimg" "generate_bootimg"
```
MayeulC: I don't really understand why `"mkbootimg"` was required twice here. Moreover, since it is now provided by `mkbootimg-osmosis`, do we really need to explicitly require `mkbootimg-osmosis` ?
Of course, it would work that way, and I didn't try with just `mkbootimg`.

aports/main/postmarketos-mkinitfs/mkinitfs.sh
```diff
@@ -256,7 +256,7 @@ create_bootimg()
 			exit 1
 		fi
 	fi
-	mkbootimg \
+	mkbootimg-osm0sis \
```
MayeulC: Same here: do you prefer `mkbootimg-osm0sis` because it is more explicit? We could also leave `mkbootimg` in place as it is more generic.

aports/main/postmarketos-mkinitfs/mkinitfs.sh
z3ntu:
```
# Required command check with useful error message                                                                                     
# $1: command (e.g. "mkimage")                                                                                                         
# $2: package (e.g. "uboot-tools")                                                                                                     
# $3: related deviceinfo variable (e.g. "generate_bootimg")
```

PR #1554, opened 2018-06-11
device-amazon-thor: Modernize APKBUILD

steamp0rt:
I also added a new fb.modes that seems to work

[x] Merge on GitHub (see <https://postmarketos.org/merge>)

PR #1558, opened 2018-06-12
linux-amazon-thor: Change MDSS TX timeout to 1500.

steamp0rt:
osk-sdl.... might work? I don't know if this is related, but it worked while messing around with this and fb.modes (which I included in #1554)

---


ollieparanoid:
Sorry, but "might work?" is not good enough to get this (or the other PR) merged.

Could you do a fresh installation with:
* #1558 applied only
* #1554 applied only
* both PRs applied
...and then check which of these (or the combination of both?) really fixes the issue?

Between the tests, you can manually delete the amazon-thor related packages like this:
```shell
$ sudo rm -v ~/.local/var/pmbootstrap/packages/armhf/*amazon-thor*
$ pmbootstrap index
```

PR #1566, opened 2018-06-14
New device: Fairphone 1

z3ntu:
```
 ~  telnet 172.16.42.1                                                                                                                                                               Do 14 Jun 2018 21:44:14 CEST
Trying 172.16.42.1...
Connected to 172.16.42.1.
Escape character is '^]'.

Type 'pmos_continue_boot' to continue booting:
/ # uname -a
Linux (none) 3.4.5 #1-postmarketOS SMP PREEMPT Thu Jun 14 19:28:51 UTC 2018 armv7l Linux
/ # cat /proc/version 
Linux version 3.4.5 (pmos@g550jk) (gcc version 6.4.0 (Alpine 6.4.0) ) #1-postmarketOS SMP PREEMPT Thu Jun 14 19:28:51 UTC 2018
```

Tested until now:
debug-shell hook works (usb networking works), maximum-attention hook works

A new aport for "mtk-mkimage" was added, which adds a 512-byte header to both initramfs and zImage. A new deviceinfo flag was added to enable it on the initramfs, the kernel is currently just done in the kernel aport (maybe the flag should do both in the mkinitfs script?)

---


aports/main/mtk-mkimage/APKBUILD
```diff
@@ -0,0 +1,34 @@
+# Contributor:
+# Maintainer:
+pkgname=mtk-mkimage
+pkgver=1
+pkgrel=0
+pkgdesc="Mediatek tool to add a custom header to an image"
+url="http://example.com"
+arch="all"
+license="???"
```
ollieparanoid: While we can't do much about firmware and their licenses, we should at least make sure that software we can compile from source has a proper license in my opinion.

The program is pretty simple so here's a libre implementation as shell script that looks nothing like the original one:

https://gist.github.com/ollieparanoid/6eaaba3d520cf938cee10c72eb31f0a8

(I'd include that directly in the aport, not download it from the gist site.)

I have tested on x86_64 that this outputs exactly the same as the program from MediaTek with various input files.

aports/main/mtk-mkimage/APKBUILD
```diff
@@ -0,0 +1,34 @@
+# Contributor:
+# Maintainer:
```
ollieparanoid: Please either add yourself there or remove the maintainer and contributor lines.

aports/main/mtk-mkimage/APKBUILD
```diff
@@ -0,0 +1,34 @@
...
+pkgname=mtk-mkimage
+pkgver=1
+pkgrel=0
+pkgdesc="Mediatek tool to add a custom header to an image"
+url="http://example.com"
```
ollieparanoid: Is there a documentation site of the header format? That would make a good "url". Otherwise you could put postmarketos.org here.

aports/main/mtk-mkimage/APKBUILD
z3ntu: I also saw a Python implementation for the same thing, but I'll stay with this one for now as it's less work :wink:

aports/main/mtk-mkimage/APKBUILD
z3ntu: http://rex-shen.net/customize-flash-mt6589-rom/ documents it pretty well under the header `The MTK Specified File Header`

z3ntu:
> I can't really see a benefit in either way, so I would keep it like you have it now.

Okay

aports/main/mtk-mkimage/APKBUILD
z3ntu: Also, the WiFi tool (`6620_launcher`) is normally a binary but there are source versions available that I will probably need to use so I guess it wouldn't be the only thing with a questionable license ^^

z3ntu: > Maybe we can find a libre implementation for that as well, or write our own one.

I still don't really exactly know what that tool does (except making a bunch of ioctl calls) and I don't think any alternative for that exists currently. I found a modification from MediaDeb which used that binary (outside of Android - in Debian) successfully but I don't think any other person tried doing that ;) Maybe some Chinese or Russian people but I won't find anything there.^^

>      

Would you mind using the one linked in the gist with the free license? it works exactly like the one you have now, except that the call syntax is slightly different:

PR #1568, opened 2018-06-15
New device: Raspberry Pi 3B/3B+(aarch64)

yangxuan8282:
add Raspberry Pi 3B/3B+(aarch64) support

tested desktop environment:  xfce4, weston ( drm backend, need load vc4 )

device functions: wifi (2.4G), ethernet

---


PureTryOut:
It seems ethernet does not work for me on this build (or it just doesn't completely boot, however the ethernet led is lit), it doesn't show up on my router. Can anyone reproduce this?

yangxuan8282:
@PureTryOut 
can you give more specific information?

like the out of 

`dmesg | grep smsc95xx`

if you are using the rpi as headless server( no monitor, no keyboard ), you can try debug with serial port, or chroot into the system on other linux computer and add some auto connect service ( maybe something like [this](https://gist.github.com/yangxuan8282/d40d590779b93d07effba3d556570652/raw/ea6fdac798f9d5c69aeca717aaf1ff9142e1a432/wifi-helper.md) ) to connect wifi

aports/device/device-raspberry-pi3/usercfg.txt
```diff
@@ -0,0 +1,7 @@
+enable_uart=1
+
+# uncomment to take advantage of your high speed sd card
+#dtparam=sd_overclock=100
+
+# uncomment to use drm backend ( for weston and wayland )
+#dtoverlay=vc4-kms-v3d
```
ollieparanoid: Why don't we enable these by default?

yangxuan8282: if we enable vc4 by default we can’t see pmos splash screen, long time blackscreen may make user think there is something wrong happen, in raspberry pi downstream kernel config, vc4 was config as mouldes( instead of y), maybe need to add it into deviceinfo

yangxuan8282:
@ollieparanoid 
hi, sorry I'm not really be familiar with git workflow, I'm trying to resolve the conflict with rebase:

```
git checkout device-raspberry-pi3
git rebase master
```

but seems nothing happen

PureTryOut:
```
git checkout master
git pull
git checkout device-raspberry-pi3
git rebase master
```

*or* (assuming origin points to this repository, not your own fork)

```
git checkout device-raspberry-pi3
git fetch origin
git rebase origin/master
```

yangxuan8282:
seems I mess up every thing, some commit not related this PR was include in, maybe I should delete the forked repos and close the PR, and re fork again

yangxuan8282:
@PureTryOut 
I tried build this again on 3B, seems sometimes did got same issues about ethernet, it got wrong ip ( 172.16.42.1 ), after reboot the issues gone. I remember used to got same issues with package device-raspberry-pi once

the same ip appear in https://github.com/postmarketOS/pmbootstrap/blob/master/aports/main/postmarketos-mkinitfs/init_functions.sh#L3 , although I'm not sure if it's related, @ollieparanoid do yout think if there are any related service may cause this problem?

on 3B+ the wired connection work normally

for wifi I tried over 5 times, not got any issues

MartijnBraam:
I guess we need to add something to make it account of having an actual ethernet interface on the device instead of it being an usb-ethernet thing. I think having an ethernet port on the PI triggers the qemu code that assigns it a 172.16/16 address.

Does anyone know why? I think that we only need to set the ip address on the `rndis0` and `usb0` interfaces and just run a dhcp client on `eth0` interfaces (which requires a workaround for qemu)

yangxuan8282:
@MartijnBraam 

here is some info about rpi 3B ethernet:

```
 raspberrypi:~$ dmesg | grep smsc95xx
[    0.441627] usbcore: registered new interface driver smsc95xx
[    1.811489] smsc95xx v1.0.6
[    1.857565] smsc95xx 1-1.1:1.0 eth0: register 'smsc95xx' at usb-3f980000.usb-1.1, smsc95xx USB 2.0 Ethernet, b8:27:eb:38:ed:f3
[   20.128652] smsc95xx 1-1.1:1.0 eth0: hardware isn't capable of remote wakeup
[   21.741489] smsc95xx 1-1.1:1.0 eth0: link up, 100Mbps, full-duplex, lpa 0xCDE1
```

the ethernet port was also mount to the usb interfaces , so it's also an usb-ethernet , just sometime it won't trigger pmos usb0 interfaces settings

yangxuan8282:
@PureTryOut 
hi, do you still have the eth0 issues? can you try modified `boot_delay=0` to `boot_delay=3` or other number and see the result

aports/device/device-raspberry-pi3/usercfg.txt
ollieparanoid: Yeah, adding it to `deviceinfo_modules_initfs` should fix this. Could you try that out?

ollieparanoid:
> Does anyone know why? I think that we only need to set the ip address on the rndis0 and usb0 interfaces and just run a dhcp client on eth0 interfaces (which requires a workaround for qemu)

We set the IP [here](https://github.com/postmarketOS/pmbootstrap/blob/9311f9c2f24de8728c11aff3fed9590ce57d4e0c/aports/main/postmarketos-mkinitfs/init_functions.sh#L251). But does it make sense to start the DHCP server in a LAN by default, where usually a router is in place with a DHCPCD?


I would recommend that we add a new deviceinfo option:
`deviceinfo_disable_dhcpcd`

And when that is set, we only print a message in `start_udhcpd()` and return. That way, the IP does not get set, and OpenRC can take care of the network settings. Thoughts?

lawl:
IMO (if possible) it would make more sense to only start the dhcpd if pmos knows it's a usb adapter and then *only* bind to that.

Multiple dhcp servers in one network tend to screw the network up pretty bad, so imo pmos should try to not do that.

Thus I think if you explicitly want the opposite i.e. run dhcpd on a non usb interface it should be opt-in and not opt-out.

MayeulC:
We could try to query an existing DHCP server, and start our own if there is none. We could miss DHCP probes from other devices, though, as those will likely start sending probes as soon as the link comes up.

Or maybe only start it briefly after the link comes up, as the other device will likely send probes at that time (and kill the dhcp server, then probe for an existing DHCP server; an option is to do this periodically).

but I think that @lawl's solution is quite reasonable as well.

MartijnBraam:
@lawl's is good but I think it should be the inverse since most devices do need dhcp. I can't think of a good name  to call the variable. Basically you want dhcpd on point-to-point links and dhcpc on point-to-multipoint links.

ollieparanoid:
@MayeulC: that sounds pretty complex and error prone. As @lawl and @MartijnBraam mentioned, raspberry like devices are the minority. So I would advise against implementing a check for dhcp servers on the network.

#1580 implements `deviceinfo_disable_dhcpcd`. Reviews and tests highly apprecated!

yangxuan8282:
@ollieparanoid 
I have tested #1580 , desktop use weston, xfce4, and plasma mobile, use drm backend, 
with `deviceinfo_disable_dhcpcd="true"` in deviceinfo, the eth0 seems not got usb network issues again for now (but I'm not sure if it will happen again)
as for pmos splash screen, although add `deviceinfo_modules_initfs="vc4"` in deviceinfo, but it's still blackscreen, use `dtoverlay=vc4-fkms-v3d` in /boot/usercfg.txt, sometimes it can display pmos splash screen

on plasma mobile mouse and keyboard not work (I remember it used to work) 

by the way, seems alpine upstream somehow upgrade rpi related packages pretty often those days, they add rpi bluetooth and 3B+ wifi firmware in https://github.com/alpinelinux/aports/commit/507d8ef3e038dd654b9953e3ffbcbd5116019df5 , so as a result, there are file conflict with aports/firmware/firmware-pi-bluetooth, we may need delete firmware-pi-bluetooth and related section in device-raspberry-pi, since they update linux-firmware, we may should delete aports/temp/linux-firmware

PureTryOut:
No it seems at least the mouse isn't working in PlaMo. I couldn't test the keyboard as I can't launch an application to test it in. I'm guessing this is the same issue as #1585.

yangxuan8282:
@ollieparanoid 
seems use `dtoverlay=vc4-fkms-v3d` and add `deviceinfo_modules_initfs="vc4"` in deviceinfo, the pmos splash screen can display now, I guess it's related to this commit https://github.com/alpinelinux/aports/commit/fa980e8f2ed77d48a51e1df9c94e8501f5a1e89a from @clandmeter

yangxuan8282:
the new PR is here: https://github.com/postmarketOS/pmbootstrap/pull/1600

PR #1570, opened 2018-06-15
Feature/devicepkg dev kernel functions

drebrez:
Changes:

- Update the `devicepkg-dev` with a new `kernelpkg_functions.sh` script which contains the `downstreamkernel_prepare` function discussed in https://github.com/postmarketOS/pmbootstrap/issues/688#issuecomment-384747760

- Revert the changes in #1556 and use the `downstreamkernel_prepare` in the `prepare` function

- Fix `pmbootstrap kconfig edit` by reading the value of the `builddir` and executing `make menuconfig` directly inside the builddir

- Update some old APKBUILD kernel packages (to be continued)

Close: #1551

---

### DIFF
--- a/aports/device/linux-amazon-thor/APKBUILD
+++ b/aports/device/linux-amazon-thor/APKBUILD
@@ -10,7 +10,7 @@ _flavor="amazon-thor"
 url="https://kernel.org"
 license="GPL2"
 options="!strip !check !tracedeps"
-makedepends="perl sed installkernel bash gmp-dev bc linux-headers elfutils-dev xz dtbtool"
+makedepends="perl sed installkernel bash gmp-dev bc linux-headers elfutils-dev xz dtbtool devicepkg-dev"
 HOSTCC="${CC:-gcc}"
 HOSTCC="${HOSTCC#${CROSS_COMPILE}}"
 
@@ -34,23 +34,8 @@ source="
 builddir="$srcdir/${_repository}-${_commit}"
 
 prepare() {
-	default_prepare
-
-	# gcc6 support
-	cp -v "$srcdir/compiler-gcc6.h" "$builddir/include/linux/"
-
-	# Remove -Werror from all makefiles
-	local i
-	local makefiles="$(find . -type f -name Makefile)
-		$(find . -type f -name Kbuild)"
-	for i in $makefiles; do
-		sed -i 's/-Werror-/-W/g' "$i"
-		sed -i 's/-Werror//g' "$i"
-	done
-
-	# Prepare kernel config ('yes ""' for kernels lacking olddefconfig)
-	cp "$srcdir"/$_config "$builddir"/.config
-	yes "" | make ARCH="$_carch" HOSTCC="$HOSTCC" oldconfig
+	. /usr/share/pmbootstrap/kernelpkg_functions.sh
+	downstreamkernel_prepare
 }
 
 build() {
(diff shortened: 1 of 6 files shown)


aports/device/linux-amazon-thor/APKBUILD
```diff
@@ -85,7 +70,7 @@ package() {
 }
 
 sha512sums="7d9f9001c6fc11954922070e3ed18f7e2843d5cd45dbab0f86ab332d6c45ee0279dd2c9bf498d91c017322f8b56f7f7417f132201bf2104ea5826212d09224f7  linux-amazon-thor-f73e4eec8717c6104eca954cf9c259d9d37c38a5.tar.gz
-d682b04c30b7829d6e90b4ad3eb40201ff253dfdb1717742b6b6574b8d3c86abc95e6f7ee8e8ab097520b41e0240b06bdd75fc8be727ac481e656c4d23f8dd30  config-amazon-thor.armhf
+903a8bac7a63278d56a9bea412b700b6d833cddc19065f2a53afbaee1fccce66b6df170413bd4667620f2739662b2032ed8d0cc926856229595afd41a27e8f1c  config-amazon-thor.armhf
```
steamp0rt: ?

ollieparanoid:
@drebrez: Thank you very much for all the work you have put into this!
I need to understand the problem better before I can review this, see https://github.com/postmarketOS/pmbootstrap/issues/1551#issuecomment-397789336.

ollieparanoid:
In my opinion, this does too many things at once. Let's work on a fix for #1551 first, that fixes it for all kernels, before we add new features like `downstreamkernel_prepare`. With that being said, I think `downstreamkernel_prepare` is a really good addition for a future PR and as I wrote in https://github.com/postmarketOS/pmbootstrap/issues/1551#issuecomment-397886525 I will work on a fix now, that uses some of the code in this PR to fix up `kconfig edit`. We could rebase this PR on it afterwards, if you like.

Thanks again!

aports/device/linux-amazon-thor/APKBUILD
drebrez: @steamp0rt probably a mistake, It's when I was trying to fix the `kconfig edit`. I will fix it. Thanks

ollieparanoid:
Thanks for the update!

What do you think about introducing the `downstreamkernel_prepare` function and using it in the template for new aports, but not modifying existing ones in this PR? (That way we would have less conflicts)

> I don't really like sourcing the `kernelpkg_functions.sh` file directly in the prepare function, but I don't know how to do it differently.

Me neither, I was wondering why you did that in the first place, but of course that's the only way to access `default_prepare` from `abuild`.

> Maybe we can do like in the device package and pass all the variables but we will not be able to call `default_prepare`, but we can copy that code directly in the script.

Yes, the `default_prepare` is [pretty short](https://github.com/alpinelinux/abuild/blob/07b6609d745dfffd009f76fc93599e8960d54fbb/abuild.in#L620-L635), so we could copy its contents. I would recommend against naming the function the same way, because that may cause confusion.

`abuild` [is under the GPL-2](https://github.com/alpinelinux/abuild/blob/07b6609d745dfffd009f76fc93599e8960d54fbb/abuild.in#L7), so the easiest way would be to change the license in the APKBUILD from MIT to GPL-2 if you are fine with that ([totally legit](https://softwareengineering.stackexchange.com/a/105926)).

PR #1572, opened 2018-06-17
device-samsung-manta: Enable touchscreen input

BakaOsaka:
Adds touchscreen support to device-samsung-manta (Google Nexus 10) with assistance from ```/dev/ponies``` on Riot.
Tested successfully on the Weston demo .
---

### DIFF
--- a/aports/device/device-samsung-manta/90-android-touch-dev.rules
+++ b/aports/device/device-samsung-manta/90-android-touch-dev.rules
@@ -1,3 +1,0 @@
-# Touchscreen
-SUBSYSTEM=="input", ATTRS{name}=="Atmel maXTouch Touchscreen", \
-ENV{ID_INPUT}="1", ENV{ID_INPUT_TOUCHSCREEN}="1"
--- a/aports/device/device-samsung-manta/90-android-touch-dev.rules
+++ b/aports/device/device-samsung-manta/90-android-touch-dev.rules
@@ -1,0 +1,3 @@
+# Touchscreen
+SUBSYSTEM=="input", ATTRS{name}=="Atmel maXTouch Touchscreen", \
+ENV{ID_INPUT}="1", ENV{ID_INPUT_TOUCHSCREEN}="1"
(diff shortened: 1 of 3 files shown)


aports/device/device-samsung-manta/90-android-touch-dev.rules
```diff
@@ -0,0 +1,3 @@
+# Touchscreen
+SUBSYSTEM=="input", ATTRS{name}=="Atmel maXTouch Touchscreen", \
+ENV{ID_INPUT}="1", ENV{ID_INPUT_TOUCHSCREEN}="1"
```
MayeulC: Is this file really necessary? It should be automatically generated with the device package (see https://github.com/postmarketOS/pmbootstrap/pull/1554#discussion_r195552363 right, @ollieparanoid ?)

Otherwise, that's nitpicking, but you [miss a newline](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline) at the end of file

aports/device/device-samsung-manta/APKBUILD
```diff
@@ -12,11 +12,11 @@ makedepends="devicepkg-dev"
 source="deviceinfo"
 
 build() {
-	devicepkg_build $startdir $pkgname
+       devicepkg_build $startdir $pkgname
```
PureTryOut: We use tabs for indentation here, please revert this change (same for the one below)

aports/device/device-samsung-manta/APKBUILD
```diff
@@ -1,7 +1,7 @@
 # Reference: <https://postmarketos.org/devicepkg>
 pkgname="device-samsung-manta"
 pkgdesc="Google Nexus 10"
-pkgver=0.1
+pkgver=0.3
```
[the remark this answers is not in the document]
BakaOsaka: My mistake - I incremented this twice, once for each commit. I realize since this will get squashed there's no need to do so. I've set it back to 0.2.

aports/device/device-samsung-manta/APKBUILD
BakaOsaka: Fixed in latest commit.

PR #1573, opened 2018-06-17
pmbootstrap kconfig edit: support $builddir and $srcdir/build

### LINKED ISSUE #1551: porting to new device: pmbootstrap kconfig edit Error

ollieparanoid:
The linux APKBUILDs write the kernel config either to `$builddir` (default from the template) or `$srcdir/build` (legacy, and I reverted to that in #1556, which was not the proper fix for this regression).

With this commit, `pmbootstrap kconfig edit` is able to edit both versions, and prints a note when the APKBUILD is still using the old style.

Fixes #1551.

---


pmb/build/menuconfig.py
```diff
@@ -55,6 +55,38 @@ def get_arch(args, apkbuild):
...
+        logging.warning("*****")
+        logging.warning("NOTE: The code in this linux APKBUILD is pretty old."
+                        " Consider making a backup and migrating to a modern"
+                        " version with: pmbootstrap aportgen " + pkgname)
+        logging.warning("*****")
+
+        return ret
+
+    # New style ($builddir)
+    cmd = "srcdir=$PWD/src source APKBUILD; echo $builddir"
```
drebrez: @ollieparanoid what's the purpose of `srcdir=$PWD/src`? I guess is for the `--src` argument?

ollieparanoid: This is, because `$srcdir` is not defined when sourcing the APKBUILD, and part of `$builddir` is always `$srcdir`. It could also be written as:

```python
cmd = "srcdir=/home/pmos/build/src source APKBUILD; echo $builddir"
```

...which is probably less confusing. How about we switch it to that?

drebrez: @ollieparanoid ah ok, now it's clear and yes written like that is less confusing.

ollieparanoid: Thanks, I'll update it. besides that, do you think the PR is ready to go?

drebrez: @ollieparanoid Yes I've tested it with a lot of linux packages, old and new, and works fine 👍

ollieparanoid:
@drebrez: thanks again for the testing!

PR #1579, opened 2018-06-19
kde/kwin: workaround the regression introduced in upstream (#1579)

### LINKED ISSUE #1574: Plasma mobile regression
I'm a bit busy these days, so I'm only scrolling over the chat backlog currently, not reading every message anymore. But @PureTryOut and @bhush9 mentioned, that plasma mobile does not work since the last update.

Regarding the process:
* How about @PureTryOut and I check if PlaMo is still working at least in Qemu before making the PR/merging the PR (currently we are both only testing if it compiles, right)?
* It would be very nice if someone could create an issue in the bug tracker about such problems, even if there aren't logs available yet. Then the issue is more visible than somewhere in the chatlog.

---

@PureTryOut, @bhush9: do you have any details/logs about the issue, or any way I could hep out with this?

bhush9:
I have better fix planned but unfortunately I won't get it to work
before Plasma 5.13.2 release (Plasma 5.13.1 is due today).

Fixes #1574

---


bhush9:
@MartijnBraam can you check it on hammerhead? I  am unable to test at moment.

PR #1580, opened 2018-06-19
Don't run dhcpcd on the raspberry pi

### LINKED ISSUE #1547: Wifi on the RPi3 is terribly slow
Making an SSH connection over wifi to a RPi3 running (as of right now the latest commit) postmarketOS is unbelievably slow. I'm not sure what we can do about this and if I'm the only one or not, but right now it's unbearable. Response to typing in a character can be either instant or take a second or two.

```
▶ ping  192.168.1.180
PING 192.168.1.180 (192.168.1.180) 56(84) bytes of data.
64 bytes from 192.168.1.180: icmp_seq=1 ttl=64 time=5.49 ms
64 bytes from 192.168.1.180: icmp_seq=2 ttl=64 time=2.75 ms
64 bytes from 192.168.1.180: icmp_seq=6 ttl=64 time=5.86 ms
64 bytes from 192.168.1.180: icmp_seq=7 ttl=64 time=4.19 ms
64 bytes from 192.168.1.180: icmp_seq=11 ttl=64 time=8.67 ms
64 bytes from 192.168.1.180: icmp_seq=12 ttl=64 time=4.18 ms
^C
--- 192.168.1.180 ping statistics ---
12 packets transmitted, 6 received, 50% packet loss, time 11164ms
rtt min/avg/max/mdev = 2.755/5.195/8.676/1.854 ms
```

As you can see there is quite a bit of packet loss as well.

ollieparanoid:
Adds `deviceinfo_disable_dhcpcd` and sets it to true for the pi.

Related: #1547.

@PureTryOut: could you please review the few lines of code changes and test if the network problems go away with this PR?

---


aports/main/postmarketos-mkinitfs/init_functions.sh
```diff
@@ -241,6 +241,14 @@ start_udhcpd() {
 	# Only run once
 	[ -e /etc/udhcpd.conf ] && return
 
+	# Skip if disabled
+	# shellcheck disable=SC2154
+	if [ "$deviceinfo_disable_dhcpcd" = "true" ]; then
```
drebrez: @ollieparanoid why `dhcpcd` ? in the function is called `start_udhcpd`, the config file `udhcpd.conf` and the executable `udhcpd`. 
Why not calling the variable `deviceinfo_disable_udhcpd`?

drebrez: From https://wiki.archlinux.org/index.php/dhcpcd:

`dhcpcd` => DHCP ***client*** daemon
`dhcpd` => DHCP ***(server)*** daemon
`udhcpd` => micro DHCP ***(server)*** daemon

with the variable we actually do 2 things:
- do not set the static ip to the interface
- do not start the DHCP ***(server)*** daemon

Don't know a better name for the variable but like this it might be confused with the DHCP client deamon which has nothing to do here.

ollieparanoid: Oh, the muscle memory of typing manually `dhcpcd ...` so often.

Thanks for catching that, I will change it to `deviceinfo_disable_dhcpd`, and describe the *two* things it does in the wiki description!

My reasoning for `dhcpd` instead of `udhcpd` is, because `udhcpd` is the implementation name of the busybox program. But the implementation name does not matter here, we might as well switch that out with another dhcp server if the udhcpd one doesn't behave like we want at some point (as it happened with busybox tar, which did unexpected things).

ollieparanoid:
@PureTryOut: thanks for testing!
@drebrez: thanks for the review, what do you think of the new version?

PR #1583, opened 2018-06-21
device-raspberry-pi: use linux-firmware-brcm

ollieparanoid:
Alpine's firmware packages have been updated to include the latest raspberry pi firmware as noted by @yangxuan8282 in https://github.com/postmarketOS/pmbootstrap/pull/1568#issuecomment-398620666. This commit adjusts our aports.

### Details:
* remove obsolete `aports/temp/linux-firmware`
* `firmware/firmware-pi-bluetooth`: rename to `device/bluetooth-raspberry-pi`
  and remove firmware files (they are part of Alpine's `linux-firmware` now)
* `device-raspberry-pi`: depend on `bluetooth-raspberry-pi`
* `device-raspberry-pi`: not that Alpine's kernels depend non free firmware

@drebrez, @MartijnBraam, @PureTryOut, @yangxuan8282:
Could you please review and test this?

---


aports/device/bluetooth-raspberry-pi/APKBUILD
```diff
@@ -0,0 +1,22 @@
+# Maintainer: He Yangxuan <yangxuan8282@gmail.com>
+
+pkgname=bluetooth-raspberry-pi
+pkgver=2
+pkgrel=0
+pkgdesc="Service and udev rule for built-in bluetooth on Raspberry Pi zero w, 3B, 3B plus"
+url="https://github.com/RPi-Distro/bluez-firmware"
+arch="armhf"
```
yangxuan8282: since this package only include openrc init scripts and udev rules, I guess we can change the arch to `noarch` or `all`, and this will compatible with rpi3 aarch64

ollieparanoid:
@drebrez wrote:
> `linux-firmware-brcm` is for the bluetooth or for the wifi? Maybe the description can be updated to explain that is also for the wifi

Yes, `linux-firmware-brcm` has both the Wi-Fi and bluetooth firmware. As it is right now, Alpine's kernels depend on `linux-firmware-brcm` already, so what saying yes/no to the non free firmware question does is change whether `bluetooth-raspberry-pi` gets installed or not.

I think that's a problem, even if we say there's also Wi-Fi firmware being installed, answering the non free firmware question does not do what it is supposed to do. People would say "yes" for the raspberry pi 1 and get the bluetooth service installed, although they don't need it.

How about we split up the `device-raspberry-pi` into multiple packages, so we can add the `bluetooth-raspberry-pi` package to the `nonfree_firmware()` function only for the bluetooth devices? That would give us the following packages:
* `device-raspberry-pi` (Raspberry Pi 1, Zero; no bluetooth; `linux-rpi` armhf)
* `device-raspberry-pi-zero-w` (Raspberry Pi Zero W; with bluetooth; `linux-rpi` armhf)
* `device-raspberry-pi-2` (Raspberry Pi 2; with bluetooth; `linux-rpi2` or `linux-rpi` aarch64?)
* `device-raspberry-pi-3` (Raspberry Pi 3; with bluetooth; `linux-rpi` aarch64)

yangxuan8282:
@ollieparanoid 
split to multi package seems make things complex, the `nonfree_firmware` actually is refer to firmware file in `linux-firmware-brcm`, since currently the upstream `linux-rpi` depends on `linux-firmware-brcm`, we can not really provide the option to select it or not
the `bluetooth-raspberry-pi` in this PR only include udev rules file and openrc init scripts, which is not nonfree, we can put it repos, and tell user this package exist(in wiki or somewhere else), they can install it if they need

ollieparanoid:
Now that's a much better idea, thank you very much!

TODO:
- [ ] update the PR to remove the nonfree firmware subpackage
- [ ] merge PR after review
- [ ] mention in the wiki that `bluetooth-raspberry-pi` should be added as necessary by the user in the extra packages question in init


PS: I've written to clandmeter about removing the firmware dependency from the kernel, maybe that will happen in the future.

ollieparanoid:
@yangxuan8282, @drebrez: does one of you have time to take another look at this again?

yangxuan8282:
looks fine, I guess we can merge it

Issue #1567, opened 2018-06-14
Testing the new Raspberry Pi kernel update


ollieparanoid:
We are directly using the Raspberry Pi kernels from Alpine in postmarketOS, and they have just been updated. Could we check if the new kernels work properly? @clandmeter from Alpine asked if we could test this, so potential problems could be fixed before the upcoming 3.8 release of Alpine. And besides helping out our Alpine friends, it would be good to know whether it works for pmOS uses as well!

(based on the list of who has which device that @drebrez created in #1501):
- [x] Raspberry Pi Zero W (@steamp0rt, @yangxuan8282)
- [ ] Raspberry Pi 1 Model B (@duncanguthrie, @yangxuan8282?)
- [x] Raspberry Pi 2 Model B (@drebrez, @yangxuan8282?)
- [x] Raspberry Pi 3 (@MartijnBraam, @PureTryOut)
- [x] Raspberry Pi 3 B+ (@yangxuan8282)

I've already mentioned the networking problem #1547 to @clandmeter. CC: @ncopa

yangxuan8282:
@ollieparanoid 
it happen yesterday I wrote a small [script](https://gist.github.com/yangxuan8282/d40d590779b93d07effba3d556570652) to install alpine on rpi, after test the new kernel, I can confirm it works on zero w, rpi 3b, 3b+(5G wifi not tested, no idea how to set region on alpine), seems that scripts not work for aarch64(maybe need uboot), it stuck at color splash screen, it indicate that can't found kernel, and I'm not familiar with arm64, so not sure if the aarch64 linux-rpi kernel works
will test on pmos later

PureTryOut:
Well I can confirm it definitely works on the RPi3, but it didn't resolve the network issue I'm having. Everything else appears to work fine just as before though.

drebrez:
@ollieparanoid I can confirm that works on the RPI 2 Model B, I've upgraded the packages, rebooted and everything was working as before.
@PureTryOut FYI: I don't have any issues with the ethernet network, constant delay of 15-20ms (depending what I ping) and 0% packet loss

ollieparanoid:
@ncopa, @clandmeter: Assuming that the Raspberry Pi 1 should work when the Zero works (because they use the same kernel), all versions have been tested :+1:

Slightly off topic pi + Alpine related questions to you:
[...]

PR #1597, opened 2018-06-25
temp/raspberrypi-bootloader: remove (to use Alpine's package again)

ollieparanoid:
This has been updated to the latest official release in Alpine, so we shouldn't need our version anymore.

@yangxuan8282: would you like to review this?

---


yangxuan8282:
@ollieparanoid 
sure, since upstream one is up to date now, we don't really need keep it anymore

yangxuan8282:
@ollieparanoid 
after search again I found the info I provide about fixup_x.dat, start_x.elf maybe incorrect, and I have edited the comment after found it: https://github.com/postmarketOS/pmbootstrap/issues/1567#issuecomment-399749996

from raspberry pi [documents](https://www.raspberrypi.org/documentation/configuration/config-txt/boot.md): 

    start_x=1 implies start_file=start_x.elf fixup_file=fixup_x.dat
    start_x=1 should be specified when using the camera module.

so it should not related to testing version, it includes camera and software codecs

and from that pages, it can be confirmed that start_db.elf, fixup_db.dat is for debug use

although not sure if the upstream subpackage name `raspberrypi-bootloader-experimental` is related to the info I provided, but seems use `raspberrypi-bootloader-x` should better, do you think if we should inform the alpine upstream?

ollieparanoid:
> do you think if we should inform the alpine upstream?

Sure, if you notice something like that, I think the best way is to contact Alpine devs (clandmeter in this case) directly via IRC (or e-mail, or in that GitHub issue we're using for Raspberry chatting with upstream right now). Asking via IRC also is a good method to make sure PRs does not get overlooked :+1:

yangxuan8282:
@clandmeter would you mind check this https://github.com/postmarketOS/pmbootstrap/issues/1567#issuecomment-399749996 and this https://github.com/postmarketOS/pmbootstrap/pull/1597#issuecomment-400339404 

 the info I provide about fixup_x.dat, start_x.elf seems incorrect, maybe we should use the name `raspberrypi-bootloader-x` instead of `raspberrypi-bootloader-experimental`

clandmeter:
I renamed it as upstream suggested it holds eXperimental features.
ref: https://github.com/raspberrypi/firmware/commit/b4ba992ff772c759cf603e00b359f9a0b4cc888e

If you found a better reference for the X naming i can make a change.

yangxuan8282:
then seems info on elinux is correct

Issue #1158, opened 2018-01-23
Use Alpine's QEMU for "pmbootstrap qemu"


ollieparanoid:
We want to use Alpine's Qemu for launching the VM, so we can avoid dealing with various QEMU versions/configurations shipped with the host distributions.

TODO:
- [ ] Fork Alpine's QEMU package, and add virgl (#543) and SDL2 support
- Run Alpine's QEMU as proof of concept from the host OS like shown [here](https://github.com/postmarketOS/pmbootstrap/issues/453#issuecomment-326358124) 
  - [ ] Headless version (for `--spice`)
  - [ ] SDL2 version
  - [ ] GTK version
- [ ] Adjust `pmbootstrap qemu`:
  - [ ] install the right QEMU package in the native chroot (depends on the VM architecture)
  - [ ] Run `--spice` with Alpine's QEMU
  - [ ] Run GTK and SDL2 with Alpine's QEMU
- [ ] Proper upstreaming
  - [ ] Upstream changes to Alpine's QEMU package
  - [ ] Remove our fork of the package

ollieparanoid:
If I understood @PureTryOut and @bhush9 right in #postmarketOS, using the following with Qemu 2.10 or above is required to get the terminal (or was it something else?) in Plasma Mobile working:
> -device (virtio|qxl)-vga,xres=XRES,yres=YRES

That version is not in all the Linux distributions yet, which are used by people running `pmbootstrap`. If we used Alpine's QEMU, we could safely add this parameter.

PureTryOut:
> or was it something else?

It was something else 😉 With Qemu 2.10 or above it's possible to give custom dimensions to Virtio, so you can tell the VM to use a portrait mode resolution and such. So instead of the default 800x600, you could do 600x800, or 200x300, or whatever you prefer.

PureTryOut:
It would be nice if someone with some extra time could take a look at this. It'd make pmbootstrap way more host independent, and it would make other arch emulation way easier for the users as well.

ollieparanoid:
When #1509 is merged, we have a nice `pmb.chroot.other.copy_xauthority()` function, that makes it possible to start X11 applications from within the chroot and display them on the host's X11 server.

This should make it a lot easier already :)

Decatf:
Interestingly it's possible to run a spice remote client (`virt-viewer` package) that is installed in chroot as well.

**Headless version (for `--spice`)**

Server QEMU / spice:
```shell
[...]

PR #1598, opened 2018-06-26
Use Alpine's qemu "pmbootstrap qemu"

Decatf:
Use Alpine's QEMU rather than the host system QEMU (as described in issue #1158). The QEMU GTK/SDL version and Spice version are working as previously.

This also partially addresses #534. It forks Alpine's QEMU with two changes:
 - Compile with `virglrenderer`
 - Add patch to use `QEMU_MODDIR` env variable as a search path for QEMU modules which is required for GTK and SDL display.

There are two new `pmbootstrap` `--device` and `--vga` arguments which are passed to qemu's equivalent args. The virgl renderer is a buggy from my testing. These two parameters are the relevant ones used for testing virgl. Run it with:

```shell
pmbootstrap qemu --display sdl,gl=on --device (virtio-vga|virtio-gpu-pci),virgl=on --vga virtio
```
Sometimes it's useful to omit one or the other argument depending on the graphical environment I've had different results depending on which.

---

### DIFF
--- a/aports/main/qemu/0001-elfload-load-PIE-executables-to-right-address.patch
+++ b/aports/main/qemu/0001-elfload-load-PIE-executables-to-right-address.patch
@@ -1,0 +1,89 @@
+From 6818f32f74981d9bccec8afbab37c42b50ab58be Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Timo=20Ter=C3=A4s?= <timo.teras@iki.fi>
+Date: Thu, 4 Jul 2013 15:50:36 +0300
+Subject: [RFC PATCH] elfload: load PIE executables to right address
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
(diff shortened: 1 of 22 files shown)


Decatf:
> I made a glaring omission in the initial PR and didn't account for when guest arch != host arch it should continue using the host qemu.

That's not correct either.

Sorry. I'm going to stew on this whole thing for a bit and then try again.

ollieparanoid:
Thanks for the PR @Decatf and no problem.

Regarding the `pkgver` of the forked Qemu package: we need to change this to `pkgver=9999`, otherwise our `qemu` package would get replaced with Alpine's package once they update it. (Also because of the pkgver change: `pkgrel=0`)

Besides that, I've opened a discussion about placing all forked aports from Alpine in `aports/temp`, see #1603.

NOTE: [We're moving to gitlab on Saturday](https://postmarketos.org/blog/2018/06/27/moving-to-gitlab/), so we will have a bit of a downtime during the import, where no pull requests / merge requests can be submitted.

PR #1599, opened 2018-06-26
Add kdeconnect aport

PureTryOut:
So I don't have a device to test it on, but it runs in Qemu.

KDE Connect, mainly famous for it's Android app, has been working on making a mobile app for Plasma Mobile. Of course we want that too! The CMake line `-DEXPERIMENTALAPP_ENABLED=true` enables this app, and as the line itself says, it's still experimental. Things will be broken and we'll have to continuously update it, but that's part of the fun isn't it? Found bugs can be reported on (preferably) bugs.kde.org or on [Telegram](https://t.me/joinchat/AOS6gA37orb2dZCLhqbZjg).

It didn't compile at first due to a musl issue, but I [committed a fix](https://github.com/KDE/kdeconnect-kde/commit/0e90fe0bd06e687b2028d269db560a2bde7e79c8) for that.

### DIFF
--- a/aports/kde/kdeconnect/APKBUILD
+++ b/aports/kde/kdeconnect/APKBUILD
@@ -1,0 +1,57 @@
+# Contributor:
+# Maintainer:
+pkgname=kdeconnect
+pkgver=1.3.1_git20180621
+pkgrel=0
+_commit="0057c44a30cc5205b23db8bb48ac86a5a9a368e3"
+pkgdesc="Adds communication between KDE and your smartphone"
+url="https://community.kde.org/KDEConnect"
+arch="all"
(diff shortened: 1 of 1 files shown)


aports/kde/kdeconnect/APKBUILD
```diff
@@ -0,0 +1,57 @@
+# Contributor:
+# Maintainer:
```
ollieparanoid: You've added yourself here in the other kde packages. Would you like to maintain this one as well?

PureTryOut: Completely forgot :see_no_evil: 
Added myself as maintainer

Issue #1577, opened 2018-06-18
Make device packages consistently use devicepkg-dev


ollieparanoid:
Using `devicepkg-dev` is the modern way of creating `device-` aports. It can automatically generate udev rules for the touch screen (which works for most devices, if it does not, we can still add an extra udev rule on top of that, or ideally improve devicepkg-dev, that it also works for those devices).

It would be nice if all device packages used `devicepkg-dev`, so we have them consistent. To make sure that we don't have regressions, we would need to test each device after changing it. So if you own one of the devices on the list, you could help out with resolving this with the following steps:
* copy your current device- aport to a temporary folder
* run `pmbootstrap aportgen device-CODENAME`
* adjust the newly generated aport, so it has the same features as the old-style one
* test the new aport
* when the functionallity is the same, please create a pull request and mention that you have tested that it works on the device.

These devices were built with an old template which does not have the `# Reference: <https://postmarketos.org/devicepkg>` line (and in most cases that means, they are not using `devicepkg-dev` either):

 - [ ] amazon-thor
 - [ ] asus-flo
 - [ ] asus-grouper
 - [ ] asus-t00f
 - [ ] fairphone-fp2
 - [ ] google-glass
 - [ ] htc-ace
 - [ ] htc-bravo
 - [ ] huawei-angler
 - [ ] huawei-y530
 - [ ] lenovo-karate
 - [ ] lg-d285
 - [ ] lg-d855
 - [ ] lg-dory
 - [ ] lg-hammerhead
 - [ ] lg-lenok
 - [ ] lg-mako
 - [ ] motorola-maserati
 - [ ] motorola-osprey
 - [ ] motorola-titan
 - [ ] nokia-n9
 - [ ] nokia-n900
 - [ ] oneplus-bacon
 - [ ] oneplus-oneplus2
 - [ ] oneplus-onyx
 - [ ] qemu-aarch64
 - [ ] qemu-amd64
 - [ ] qemu-vexpress
 - [ ] samsung-i747m
 - [ ] samsung-i9003
 - [ ] samsung-i9100
 - [ ] samsung-i9305
 - [ ] samsung-i9505
 - [ ] samsung-klte
 - [ ] samsung-n7100
 - [ ] sony-amami
 - [ ] sony-aries
 - [ ] sony-castor-windy
 - [ ] sony-honami
 - [ ] sony-sirius
 - [ ] sony-yuga
 - [ ] t2m-flame
[...]

PR #1602, opened 2018-06-28
device-huawei-y530: Modernize APKBUILD

drebrez:
Modernize the APKBUILD of the Huawei Ascend Y530

Tested with weston and everything still works, other UI have problems starting Xorg, I didn't investigated the reason.

also added the missing devicepkg reference to the Samsung Galaxy S5 APKBUILD

Ref: #1577

---


drebrez:
Probably yes, I don't think the problem is with the small change in the udev rule. Kernel isn't changed. Tomorrow I can try without the PR

PR #1604, opened 2018-06-29
New device - Sony Xperia M (sony-nicki)

ewentual:
Initial bringup of postmarketOS for Sony Xperia M. That includes everything that has been documented on [wiki](https://wiki.postmarketos.org/wiki/Sony_Xperia_M_(sony-nicki))
---


PureTryOut:
You have Plasma documented as "works but for 15 seconds only". I would list it as not working instead. Do note that it's not because of a device specific issue, but due to the device being armhf and not having hardware acceleration. Most armhf devices have this problem, so there is no reason to debug it for this one.

ewentual:
I don't understand why has it failed xfce4 check, it works fine on my device, as can be seen there:
photo5886744908646165770

aports/device/device-sony-nicki/APKBUILD
```diff
@@ -0,0 +1,34 @@
...
+package() {
+	devicepkg_package $startdir $pkgname
+	install -D -m644 "$srcdir"/90-android-touch-dev.rules \
+	"$pkgdir"/etc/udev/rules.d/90-android-touch-dev.rules
+	install -D -m644 "$srcdir"/screen-workaround.sh \
+	"$pkgdir"/etc/profile.d/screen-workaround.sh
+	install -D -m644 "$srcdir"/fb.modes \
+	"$pkgdir"/etc/fb.modes
+	install -D -m644 "$srcdir"/pointercal \
+	"$pkgdir"/etc/pointercal
```
ollieparanoid: Please indent the continued lines with one more tab character:

```shell
	install -D -m644 "$srcdir"/90-android-touch-dev.rules \
		"$pkgdir"/etc/udev/rules.d/90-android-touch-dev.rules
(...)
```

aports/device/device-sony-nicki/90-android-touch-dev.rules
```diff
@@ -0,0 +1,8 @@
+# udev rules file
```
ollieparanoid: You have probably tried this, but just in case: did you try this without the udev rule? The modern `devicepkg_build` function you're using below makes the udev rule obsolete in most (but not all) cases.

aports/device/device-sony-nicki/screen-workaround.sh
```diff
@@ -0,0 +1,11 @@
+main() {
```
ollieparanoid: You are installing this to `/etc/profile.d`, which means it gets executed as soon as the user logs in on the TTY, right? Currently we have autologin on the TTY configured for the user interfaces. However, the plan is to use a display manager in the future, in which case this will not work anymore as I understand it.

In either case, I would recommend to make this an initramfs hook instead, so the screen fix is active as early in the booting process as possible. If we can't make it an initramfs hook, we could also set this script up as daemon.

Could you try out if that works, and if it does, update your PR?

Examples:
* `device-htc-m8` has an initramfs hook
* `device-xiaomi-armani` has its own OpenRC service and enables it

aports/device/device-sony-nicki/90-android-touch-dev.rules
ewentual: Not sure if you mean whole file or line 1, but without file TS won't work

